PR #22, opened 2022-06-05
create a build procedure.  Add additional files needed to make installed wheel work

Dr-Irv:
Idea in build procedure is to copy the files from `typings/pandas` into `pandas-stubs` so that the wheel will install into `pandas-stubs`


Dr-Irv:
Merging but there is a pyright issue https://github.com/microsoft/pyright/issues/3540 that should be fixed before we distribute.

PR #23, opened 2022-06-06
Project Refactoration

BrenoJesusFernandes:
So How to improve the project?💡
    
- My goal here it was to bring poetry to the battle ⚔️ . I need help to manage dependencies, test and deploy the project the easier way possible, the project is simple and should keep that way. So I bring everything inside pyproject.toml, so we do not need to worry anymore to be lost in thousands of config files and the develop dependencies are explicit declared separately. In resume, It becomes really easy to setup, tests and publish. I just adapted everything to poetry, include github actions.

 poetry install -> make magic code 🪄 -> poetry run all_tests  .... _voilàà_ ( do you want to publish? _poetry publish_ 😉)

- Another important step it was to improve the project structure. Since PEP 561 the default structure to a stubs-only package is just "my_package-stubs", so i though better to bring the package to pandas-stubs and the tests to outside the pandas folder. Furthermore, I added a docs folder to documentation and a scripts folder to util scripts.

There is still a lot a work to make, but I hope with this strong base we could improve the project.


.gitignore
```diff
@@ -127,3 +127,4 @@ dmypy.json
 
 # Pyre type checker
 .pyre/
+/poetry.lock
```
twoertwein, line 130: Probably not important for now: It could be helpful to add `poetry.lock` to the repository (`poetry install` will take less time and it improves reproducibility) but every now and then someone needs to update `poetry.lock` (is there a github bot for this?).

pyproject.toml
```diff
@@ -1,3 +1,32 @@
...
+[tool.poetry.dependencies]
+python = ">=3.8,<4.0"
+
+[tool.poetry.dev-dependencies]
+pandas = ">=1.4.2"
+mypy = ">=0.960"
+typing-extensions = ">=4.2.0"
+matplotlib = ">=3.3.2"
+pytest = ">=7.1.2"
+pyright = ">=1.1.251"
```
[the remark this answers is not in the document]
BrenoJesusFernandes: It's not affiliate but it is the same code, he just wraps it with python and runs node in the background. I think there is no problem.

.gitignore
BrenoJesusFernandes: Hmm i see, probably a good ideia let poetry.lock available. But what do you mean with github bot, I didn't get the ideia.

twoertwein:
I will approve this PR so that the CI can run (and then undo my approval - hope it doesn't cancel the CI run)
image

Dr-Irv:
> I will approve this PR so that the CI can run (and then undo my approval - hope it doesn't cancel the CI run) 

I don't think you have to do this yet, because the workflow has been run in his github account. See https://github.com/BrenoJesusFernandes/pandas-stubs/actions

.github/workflows/test.yml
```diff
@@ -2,64 +2,37 @@ name: 'Test'
...
-      #     extra-args: --dependencies
-      #     warn-partial: true
+    - name: Run Pytest
+      run: poetry run pytest
 
-      - name: Run mypy against dist
-        run:
-          mypy tests/pandas
+    - name: Run MyPy
+      run: poetry run mypy pandas-stubs tests
```
Dr-Irv, line 38: One issue that I have run into with respect to using the more traditional method of a build using `setup.py` is that the wheel that is built, when installed, isn't passing the tests.  So if you look at https://github.com/pandas-dev/pandas-stubs/blob/main/.github/workflows/test.yml , you will see this in there:
```
     - name: Build wheel and install and remove typings
        run: |
          python setup.py build bdist_wheel
          find ./dist/*.whl | xargs pip install
          rm -rf typings
          pip install pyright
      
      # - name: run pyright command line
      #   run:
      #     pyright -p pyrightdistconfig.json --dependencies

      # - name: Run pyright against dist
      #   uses: gramster/pyright-action@main
      #   with:
      #     project: pyrightdistconfig.json
      #     extra-args: --dependencies
      #     warn-partial: true

      - name: Run mypy against dist
        run:
          mypy tests/pandas
```

The problem right now is that `pyright` is failing with this, and there is a long discussion at https://github.com/microsoft/pyright/issues/3540 , there seems to be an issue with the presence of the `py.typed` file that is interpreted differently on MacOS versus Windows and Linux.

In any case, adding the tests here to build the wheel, install it, and then test pyright and mypy against the installed wheel is something that should be done.

README.md
```diff
@@ -8,9 +8,14 @@ This is the home for pandas typing stubs supported by the pandas core team.  The
...
-## Build instructions
+### Documentation
 
-Use `python setup.py build bdist_wheel` to build the wheel.  NOTE:  `setuptools 62.3.2` is required!
+- Enter in docs to see:
+    - 1 - How to setup the enviroment
+    - 2 - How to test the project
+    - 3 - How to follow the code style 
+    - 4 - Security stuffs
+    - 5 - How to publish
```
Dr-Irv, line 18: Can you create links from here to the docs you created?

pyproject.toml
```diff
@@ -1,3 +1,32 @@
+[tool.poetry]
+name = "pandas-stubs"
+version = "0.0.1"
```
Dr-Irv, line 3: I'd like to use a version numbering convention of pandas_version.yymmdd for the stubs.  So if these were done today, the version would be 1.4.2.220607 .  Then we mark the releases of the stubs by their release date, and also know which version of pandas the stubs correspond to.

pyproject.toml
```diff
@@ -1,3 +1,32 @@
...
+description = "Stubs for pandas"
+authors = ["Breno Fernandes <brenojesusfernandes@pm.com>"]
+packages = [
+    { "include" = "pandas-stubs"}
+]
+
+[tool.poetry.dependencies]
+python = ">=3.8,<4.0"
```
Dr-Irv, line 11: maybe we should say <=3.10 since we only support pandas up to that version

Dr-Irv (changes requested):
Thanks for this great contribution.  We need to straighten out testing of the built wheel.  See below.  I will need to try out this PR soon so I can understand better how it builds a wheel we can give to pypi.org

.github/workflows/test.yml
BrenoJesusFernandes: I had a lot of trouble figuring out the "py.typed" with "partial\n" string it was causing the problem, at least in my environment. not just in wheel installed case. So for a while, I just delete the file to show the poetry idea. I have the three Local OS here, I'm gonna test and see if I can help to fix the bug in the official pyright repository. To add the test with the wheel is not gonna be hard when the pyright is gonna work properly.

README.md
BrenoJesusFernandes: Ohh yeah, it is easier to see.

pyproject.toml
BrenoJesusFernandes: I did not work with the poetry version yet, because we did not have a convention yet and my future ideia is to automate the version, tag, and publish system, so we do not need to worry with the static version in pyproject.toml file.

pyproject.toml
BrenoJesusFernandes: Nice, I did not even notice. I'm gonna fix this. Noted.

.github/workflows/test.yml
Dr-Irv: So it turns out that if I remove `py.typed`, then everything works locally and with the installed wheel.  That version is now in `main`.  So remove `py.typed` from your version and it should be OK

Dr-Irv:
One thing that is making this PR hard to evaluate is that you applied `black` and `isort` to the PYI files, which means many of them changed, and I think you may have added some PYI files that may not be necessary. 

Can you split the PR into two PRs?  One is just the `poetry` stuff and then another PR to do the `black` and `isort` part.  Could be done in either order.

pyproject.toml
Dr-Irv: > Oh another important point to work is to upgrade the current stub version to work in 1.4.2 version, I guess you already saw it, but it is compatible just with API 1.2.x.

What makes you believe it is not compatible with 1.4?  Are we missing something?  

The stubs that are here came from the Microsoft repo  `https://github.com/microsoft/python-type-stubs` (and they just retired the pandas stubs there).  I've used them successfully in projects using pandas 1.4.2, so I'd like to understand your comment about compatibility with API versions.

Issue #27, opened 2022-06-16
Operator "<=" not supported for types "Timestamp" and "TimestampSeries"


timovp:
Hi all, 
I raised an issue over at pylance: 
https://github.com/microsoft/pylance-release/issues/2933
These has been an update, but the error became TimestampSeries instead Series[Timestamp]
to replicate: 
```python
import pandas as pd

df =  pd.DataFrame(['2020-01-01','2019-01-01'])
dates = pd.to_datetime(df[0], format = '%Y-%m-%d')
date_to_compare = pd.to_datetime('2019-02-01', format = '%Y-%m-%d')
date_mask = date_to_compare <= dates
```
Generated the type error in vs code: 

> Operator "<=" not supported for types "Timestamp" and "TimestampSeries"
  Operator "<=" not supported for types "Timestamp" and "TimestampSeries"
  Pylance[reportGeneralTypeIssues](https://github.com/microsoft/pylance-release/blob/main/DIAGNOSTIC_SEVERITY_RULES.md#diagnostic-severity-rules)

Would like to create a PR to help, but just started to figure out how type stubs work, so might need a time :-)

Dr-Irv:
@timovp To encourage you to do the PR, here is what you'd need to do.  Look in `series.pyi` and the overloads for `__add__`.  You'll need to do something similar for `__lt__`, `__le__`, `__gt__`, `__ge__`, except the results in all cases are `Series[bool]`

timovp:
created a PR #28 for it. thanks again for the encouragement!

Dr-Irv:
closed in #28

PR #28, opened 2022-06-16
Update series.pyi to support comparisons with Timestamp

timovp:
added TimestampSeries comparisons with pd.Timestamp object 
Created to solve https://github.com/pandas-dev/pandas-stubs/issues/27


pandas-stubs/core/series.pyi
```diff
@@ -1094,8 +1094,14 @@ class Series(IndexOpsMixin, NDFrame, Generic[S1]):
     def __div__(self, other: Union[num, _ListLike, Series[S1]]) -> Series[S1]: ...
     def __eq__(self, other: object) -> Series[_bool]: ...  # type: ignore
     def __floordiv__(self, other: Union[num, _ListLike, Series[S1]]) -> Series[int]: ...
+    @overload
     def __ge__(self, other: Union[num, _ListLike, Series[S1]]) -> Series[_bool]: ...
+    @overload
+    def __ge__(self, other: Timestamp) -> Series[_bool]: ...
```
Dr-Irv, line 1100: I think this may be easier to do without using overloads.  E.g., for `__ge__`, use:
```python
def __ge__(self, other: Union[num, Timestamp, _ListLike, Series[S1]]) -> Series[_bool]: ...
```

Dr-Irv (changes requested):
Can you also add tests to `tests/test_timefuncs.py` that test these 4 comparison operators?

Use `assert_type()` to validate the results.

pandas-stubs/core/series.pyi
timovp: Right yes, thanks for the suggestions! Ill have a look tomorrow!

timovp: I added the comparisons the way you suggested, as well as the testing with the use of `assert_type()`

timovp:
> Can you also add tests to `tests/test_timefuncs.py` that test these 4 comparison operators?
> 
> Use `assert_type()` to validate the results.

While writing tests I noticed not for all tests `assert_type` was used, I implemented the tests with `assert_type()`, but please let me know if the test should have been implemented differently.

tests/test_timefuncs.py
```diff
@@ -46,6 +46,17 @@ def test_types_comparison() -> None:
     tsr2: bool = ts > ts2
 
 
+def test_tupes_timestamp_series_comparisons()->None:
```
Dr-Irv, line 49: Can you add a line here that says #GH 27 to point to the original issue?

Dr-Irv (changes requested):
One small change to add a comment, otherwise looks good to go

Dr-Irv:
> While writing tests I noticed not for all tests `assert_type` was used, I implemented the tests with `assert_type()`, but please let me know if the test should have been implemented differently.

That's fine.  We're inconsistent because many of the tests existed before `assert_type` was available.  I'm glad you used it.

tests/test_timefuncs.py
timovp: Done!

Dr-Irv (approved):
Thanks @timovp

PR #39, opened 2022-06-23
Fix to_dict and from_dict type stubs

KianShah:


KianShah:
There seems to be a difference between the from_dict stub in the https://github.com/VirtusLab/pandas-stubs repo vs the stub here. There, they use the Orientation type in https://github.com/VirtusLab/pandas-stubs/blob/master/third_party/3/pandas/_typing.pyi, which obviously needs to be changed. Need some guidance from more senior contributors on where we need to make the change. I know that you guys are in the process of migrating to this repo, so it may not be an easy question to answer

Dr-Irv:
> There seems to be a difference between the `from_dict` stub in the https://github.com/VirtusLab/pandas-stubs repo vs the stub here. There, they use the `Orientation` type in https://github.com/VirtusLab/pandas-stubs/blob/master/third_party/3/pandas/_typing.pyi, which obviously needs to be changed. Need some guidance from more senior contributors on where we need to make the change. I know that you guys are in the process of migrating to this repo, so it may not be an easy question to answer

The stubs here evolved from work done by Microsoft.  These stubs try to match what is in the pandas source (and there is more work to do to make that happen).  The VirtusLab stubs were done independently, but after discussion between them, Microsoft and the pandas team, we went forward with the Microsoft stubs as they were more complete.  In the case of `to_dict` and `from_dict`, in the pandas source, we haven't created a type to represent the possible `orient` arguments, so it's fine to just use the `Literal` designation here.  Also, the `orient` arguments are asymmetric in the two methods, so having a type to represent it doesn't help much.

Having said that, can you add a test for this change in `test_frame.py` ?

Also, the title of your PR says you're changing `to_dict()`, which could also use a type for `orient` as well, but I didn't see that in your proposed change.

KianShah:
I added 'tight' as an acceptable Literal for the to_dict overload. Is that not enough?

Dr-Irv:
We're trying to make it that when we add something to the stubs, we also add a code snippet that tests the stub.  So, for example, if you have a small piece of code that fails type checking without your change and succeeds with your change, that would be sufficient.

Dr-Irv:
@KianShah let us know if you intend to complete this PR by adding a test.  Thanks.

KianShah:
Yes sorry, I just got really busy. Will update either today or tomorrow

KianShah:
I've updated with some tests for `from_dict` and `to_dict`. Not sure how you guys test, but I just put calls to `from_dict` and `to_dict`. I also added parameter types in the `from_dict` method.

Dr-Irv (changes requested):
Now failing due to formatting issues.  Do the following on your local machine, until all tests pass.  Then make a new commit and push again, and I will trigger the CI
```shell
poetry run poe test_all
```

KianShah:
Ok I've run the pre-commit this time and poe test-all returns all green lights

Dr-Irv (approved):
Thanks @KianShah

PR #46, opened 2022-06-28
Use Sequence instead of List for drop

thomkeh:
The advantage is that `Sequence` is covariant which allows passing in `List[str]`, for example.

I also added `self` to `reindex` because mypy was complaining about that.

### DIFF
--- a/pandas-stubs/core/frame.pyi
+++ b/pandas-stubs/core/frame.pyi
@@ -398,15 +398,15 @@ class DataFrame(NDFrame, OpsMixin):
         fill_axis: AxisType = ...,
         broadcast_axis: Optional[AxisType] = ...,
     ) -> DataFrame: ...
-    def reindex(**kwargs) -> DataFrame: ...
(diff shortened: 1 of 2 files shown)


pandas-stubs/core/frame.pyi
```diff
@@ -398,39 +398,39 @@ class DataFrame(NDFrame, OpsMixin):
         fill_axis: AxisType = ...,
         broadcast_axis: Optional[AxisType] = ...,
     ) -> DataFrame: ...
-    def reindex(**kwargs) -> DataFrame: ...
+    def reindex(self, **kwargs) -> DataFrame: ...
```
thomkeh, line 401: This just brings it in line with the definition of `assign()` above.

Dr-Irv:
> I also added `self` to `reindex` because mypy was complaining about that.

I'm really surprised that our tests didn't pick that up.

Dr-Irv:
Thanks @thomkeh

PR #50, opened 2022-06-28
CI/CD Update

BrenoJesusFernandes:
This update aims to complete our pipeline, automating the publication and distribution of the pandas stubs.

New workflow:
- check_style -> stubs-test -> publish github and pypi 

Details:
- Add style check before all tests
- Dynamic version update the project ( pandas version + current date, use “poe build” to refresh the project version and build a new whell )
- Add automatically publish on Github and PyPi after all tests in main branch (They are commented for a while) 
- Add local test “poe check_style” 
- Add local test “poe format_style” 
- All code are now in black and isort style

Please add, PYPI_TOKEN_NAME and PYPI_TOKEN secrets to the publish works correctly.


Dr-Irv (changes requested):
A few comments:
1) Can you split this PR into two?  First do a PR for the changes to the PYI files due to black and isort.   Then do a PR about creating the pipeline.
2) I don't know if we can do automated updates to pypi, because I think the token would have to be stored in the repo, and that means anyone could push to pypi by just copying the token, so it's a security hole.  Unless there is some other way to do it that you know about.
3) I like that you added the `RELEASE.md` file.  Do you know if there is a way we can auto-generate this on a merge from a PR into main, so that we take the title of the PR, create a link to the PR, and that becomes the line in `RELEASE.md` ?  If not, we'd just have to edit it by hand (or maybe ask people doing PR's to do it)

BrenoJesusFernandes:
1. Yeahp no problem

2. Don't worry there is a safe way to store secrets in repo config, just the owners have access. https://octopus.com/blog/githubactions-secrets

BrenoJesusFernandes:
3. I think it should be possible, but we need to be careful because each new PR is gonna publish a new version, and if we have two versions in the same day we will get a conflict.

Dr-Irv:
> 2. Don't worry there is a safe way to store secrets in repo config, just the owners have access. https://octopus.com/blog/githubactions-secrets

That's going to be a little tricky for a public project like this, because even at the organizational level, we don't want our personal tokens to be in there, so I'll need to discuss with the pandas-core team how/if we want to manage this.

Thanks again for the suggestion.

scripts/__init__.py
```diff
@@ -1,12 +1,12 @@
...
-from pathlib import Path
 
-pkg_path = [x for x in sys.path if x.endswith('site-packages')]
+pkg_path = [x for x in sys.path if x.endswith("site-packages")]
 
-if not Path(fr'{pkg_path[0]}/my_paths.pth').exists():
-    with open(fr'{pkg_path[0]}/my_paths.pth', 'w') as file:
+if not Path(rf"{pkg_path[0]}/my_paths.pth").exists():
+    with open(rf"{pkg_path[0]}/my_paths.pth", "w") as file:
         file.write(str(Path.cwd()))
```
twoertwein, line 10: Why do you create this file? For debugging? It is used nowhere except in this file:

```sh
$ grep -R my_paths .
./scripts/__init__.py:if not Path(fr'{pkg_path[0]}/my_paths.pth').exists():
./scripts/__init__.py:    with open(fr'{pkg_path[0]}/my_paths.pth', 'w') as file:
```

BrenoJesusFernandes:
> 
> > 2. Don't worry there is a safe way to store secrets in repo config, just the owners have access. https://octopus.com/blog/githubactions-secrets
> 
> That's going to be a little tricky for a public project like this, because even at the organizational level, we don't want our personal tokens to be in there, so I'll need to discuss with the pandas-core team how/if we want to manage this.
> 
> Thanks again for the suggestion.

when the secret is stored nobody can see the value, so it should not be a problem.

scripts/__init__.py
BrenoJesusFernandes: just a temporary workaround to mypy and pyright works properly. There's already an open issue to remove this.

BrenoJesusFernandes:
> > 3. I think it should be possible, but we need to be careful because each new PR is gonna publish a new version, and if we have two versions in the same day we will get a conflict.
> 
> Another possibility is that a bot is created that runs at 2AM on Sunday mornings and we just publish new versions once a week instead of with each PR.
> 

Great! I am sure there is a way to schedule github actions to run. So we could apply this ideia.

scripts/__init__.py
twoertwein: Would be good to add a small comment which bug this workaround addresses. Looks a bit weird to create a file in the first site-package.

Dr-Irv: The issue is that `mypy` requires a `-stubs` package to be installed.  It won't look at a `-stubs` package anywhere else.  So the file is needed to tell `mypy` to look in the project directory, but when doing the test, we should create it, then remove it before/after running `mypy`

`pyright` does not have this issue.

BrenoJesusFernandes: @twoertwein thanks for the feedback, you are absolutely right, probably a comment would clarify all the situation. @Dr-Irv  thx the idea <3

Dr-Irv:
> @Dr-Irv I checked here and it's really easy to schedule the publish with github actions, we could run every sunday at 2am if we want, but I was thinking we could check every day if there are new PR's and publish this version in the end of the day.

Thanks.  I need to look into creating a bot account for the project which could then be used for the automatic publishing.

Dr-Irv:
So it turns out that I can create a secret with my pypi login that is good only for this project, and store it at the repository level, and it is then available to github actions that can do the publishing.

If you want to move forward with this, I think we should leverage this project:
https://github.com/pypa/gh-action-pypi-publish

So if you want to do a PR for that, and create a placeholder for the token, then I'll create a token.

I would prefer that the action is only run once a day, but only if changes were made to `main` in the past 24 hours.

If you can figure that out, that would be great!

Dr-Irv:
There is also this example that is used by a member of the pandas core team on `geopandas`:
https://github.com/geopandas/geopandas/blob/main/.github/workflows/release_to_pypi.yml

BrenoJesusFernandes:
It does not make sense to me, I know this project and it's wonderful for projects don't use poetry, but what I think is worring you is that in our pipeline there's a line with our token and anyone could see, but if that's the case you should not worry, github hides any secrets for anyone in the repository, even the owners. So my point is that we are going to use another project when we should just use `poetry publish --build --username 'CONTEND-TOTALY-HIDE' --password`  'CONTEND-TOTALY-HIDE2'. Underneath that package do the same thing.

BrenoJesusFernandes:
I see, thank you for the suggestions! When I finished the other two PR I am gonna start this one, it should not be hard, the last puzzle I still do not search about is how to generate the release notes automatically.

twoertwein:
> I see, thank you for the suggestions! When I finished the other two PR I am gonna start this one, it should not be hard, the last puzzle I still do not search about is how to generate the release notes automatically.

Some time-based projects such as dask simply use the git commit messages as a changelog/release notes. Personally, I'm not a fan of it in the case of dask (new cool features and major overhauls get lost in a long list of commits) but for `pandas-stubs` which  has only type annotations, it could be fine to just dump the commit messages.

twoertwein:
Slightly unrelated question: are the tests also run on main, visually it seems that the tests were not run for some commits (no green check mark/red cross):

image

Dr-Irv:
> Slightly unrelated question: are the tests also run on main, visually it seems that the tests were not run for some commits (no green check mark/red cross):

TBH, I'm not sure.  Looking at the past history, when I've done PR's and then merged them in, then the test is run on `main`.  It seems that if someone else does the PR, and I do the merge, it doesn't run on `main`.  

I don't know how to set that up!  Or maybe as part of the test procedure, we take the PR and merge with main, and run the PR test on the merged code?

BrenoJesusFernandes:
> Slightly unrelated question: are the tests also run on main, visually it seems that the tests were not run for some commits (no green check mark/red cross):
> 
> image
> 

I noticied that, please remove the paths config in the header of 'test.yaml'. For some reason it is not working correctly.

BrenoJesusFernandes:
> > I noticied that, please remove the paths config in the header of 'test.yaml'. For some reason it is not working correctly.
> 
> Could probably have something like what pandas has but just for the main branch:
> https://github.com/pandas-dev/pandas/blob/231b9fad17475918a94fc922d5bbd452c5b67a2f/.github/workflows/code-checks.yml#L3
> 
> (I will not work on it, I have very little knowledge about GitHub workflows.)
> 

Sure I do, my purpose here was to run tests only when source code was modified, because if you are working in docs you do not need to run the tests.

Issue #36, opened 2022-06-23
poetry test_dist destroys uncommitted changes


Dr-Irv:
When testing the distribution with `poetry run poe test_dist` if you have not committed changes (including new files), those files get lost, since we remove the directory `pandas-stubs` to check things.

Probably best to not test the dist if there are uncommitted changes (or new files) in the `pandas-stubs` folder, and output an error message asking if the developer wants to commit.

@BrenoJesusFernandes

Dr-Irv:
Closed via #51

PR #51, opened 2022-06-29
Fix Issue #36:  Avoid deleting uncommitted changes

BrenoJesusFernandes:
Basically solves https://github.com/pandas-dev/pandas-stubs/issues/36, when we run local tests now all changes are temporarily committed and restored after the tests. If there are not any new changes it is just gonna ignore.


BrenoJesusFernandes:
> Would it make sense to skip mypy&pyright on the installed pandas-stubs when there are unstaged or uncommited changes (or even new files that are not tracked by git)?

I don't think so, and you @Dr-Irv ?

twoertwein:
The reason for all of the git commands is that you want to make sure that when you run mypy&pyright on the installed pandas-stubs, that it doesn't use the local pandas-stubs, or? Instead of changing any files/git, how about changing into the `tests` folder,  overwrite the mypy_path/stubPath to use the current folder (which doesn't contain any stubs; in the worst case you still have to alter a file, but it is only pyproject.toml), mypy&pyright then cannot use the local stubs anymore, and then run `mypy .` and `pyright .` (which hopefully finds the installed stubs)?

Dr-Irv:
Here's another suggestion. Instead of doing this with `git`, what we really care about is that the local copy of the `pandas-stubs` folder is not used when we test an installed distribution.  So how about we just temporarily rename `pandas-stubs` to `tempfile.TemporaryDirectory` and after the distribution is tested, you just rename it back.

twoertwein:
> Here's another suggestion. Instead of doing this with `git`, what we really care about is that the local copy of the `pandas-stubs` folder is not used when we test an installed distribution. So how about we just temporarily rename `pandas-stubs` to `tempfile.TemporaryDirectory` and after the distribution is tested, you just rename it back.

I think that is by far the easiest method! Would need to be careful to rename the folder back to pandas-stubs (even in case of an error), otherwise the `tempfile.TemporaryDirectory` will get deleted. I would also be happy to just rename `pandas-stubs` to `_pandas-stubs` (and error if the folder already exists).

BrenoJesusFernandes:
That's it, Done. Thx for the support guys!!

scripts/test/procedures.py
```diff
@@ -31,8 +29,9 @@ def install_dist():
     subprocess.run(cmd, check=True)
 
 
-def remove_src():
-    shutil.rmtree(r"pandas-stubs")
+def rename_src():
+    if Path(r"pandas-stubs").exists():
```
Dr-Irv, line 33: can you abort here if for some reason `_pandas-stubs` already exists?

scripts/test/procedures.py
```diff
@@ -44,14 +43,15 @@ def run_pyright_dist():
...
+
 def uninstall_dist():
     cmd = ["pip", "uninstall", "-y", "pandas-stubs"]
     subprocess.run(cmd, check=True)
 
 
 def restore_src():
-    cmd = ["git", "checkout", "HEAD", "pandas-stubs"]
-    subprocess.run(cmd, check=True)
+    if Path(r"_pandas-stubs").exists():
```
Dr-Irv, line 53: Can you abort here if `_pandas-stubs` somehow already exists?

scripts/test/__init__.py
```diff
@@ -38,7 +39,7 @@ def test_dist(clean_cache: bool = False):
     steps.extend([
```
Dr-Irv, line 39: These steps are a subset of what is in `test_all`, so can you create a function for adding these steps that avoids the repeated code in `test_all()` and `test_dist()` ?

scripts/test/__init__.py
```diff
@@ -62,11 +63,11 @@ def test_all(clean_cache: bool = False):
         Step(name="Run Pytest Against Source Code", run=procedures.run_pytest_src),
```
Dr-Irv, line 63: Similar to my remark about the `test_dist()` steps, the 3 steps here are the same as in `test_src()`, so can you set things up to only define each set of steps in one place and then use them accordingly?

Dr-Irv (changes requested):
See comments to make this a bit cleaner and more error-proof

.github/workflows/test.yml
```diff
@@ -52,8 +52,8 @@ jobs:
     - name: Install Distribution
       run: poetry run poe install_dist
 
-    - name: Remove Source Code
-      run: poetry run poe remove_src
+    - name: Rename Source Code
+      run: poetry run poe rename_src
```
twoertwein, line 56: Optional: would be nice to have it as a context manager

with rename_src():
    tests()

BrenoJesusFernandes: I did not get it, could you explain more details?

scripts/test/procedures.py
BrenoJesusFernandes: Done

scripts/test/__init__.py
BrenoJesusFernandes: Done

Dr-Irv (changes requested):
This looks good.  Can you merge with main and resolve conflicts?

BrenoJesusFernandes:
> This looks good. Can you merge with main and resolve conflicts?

Sure

BrenoJesusFernandes:
Please do not merge this PR now, for some reason pyupgrade is not working in my macos setup, I gonna test in other OS's to see what's going on.

Dr-Irv (approved):
I'm good with what is in here now.  Let me know if I can merge once the tests pass

BrenoJesusFernandes:
> I'm good with what is in here now. Let me know if I can merge once the tests pass

Everything is ok now.

Dr-Irv:
Thanks @BrenoJesusFernandes

PR #59, opened 2022-07-01
TYP/CI: enable more pyright checks

twoertwein:
There were some overlapping overloads in interval.pyi and a few default values in multiple files.


Dr-Irv:
Thanks @twoertwein

PR #60, opened 2022-07-01
CI: enable a few more mypy checks

twoertwein:
- require specific error codes (`enable_error_code = "ignore-without-code"`)
- require explicit exports of imports (through redudant `as` or in `__all__`, pyright does the same, `implicit_reexport`)
- when pyright and mypy disagree, we can use pyright-specific ignore comments (`warn_unused_ignores`)

What is the goal of the tests inside `TYPE_CHECKING`? Is that for compatibility with older versions or are these placeholder for future "not-tests"?


Dr-Irv:
> * require specific error codes (`enable_error_code = "ignore-without-code"`)
> * require explicit exports of imports (through redudant `as` or in `__all__`, pyright does the same, `implicit_reexport`)
> * when pyright and mypy disagree, we can use pyright-specific ignore comments (`warn_unused_ignores`)
> 
> What is the goal of the tests inside `TYPE_CHECKING`? Is that for compatibility with older versions or are these placeholder for future "not-tests"?

The goal of those tests are code that would raise an `Exception` if you executed it.  In theory, each of those lines should be picked up by the type checker as being incorrect, so you would need a `# type: ignore` to tell the type checker that you are _expecting_ it to find an error.  The issue here is that between `pyright` and `mypy`, either or both of them will pick up that the code is incorrect.  If both do, or just `mypy` does, we're fine with a `# type: ignore` comment.  If only `pyright` does, then we can use the `pyright` notation you included in this PR.  If none of them do, it means our stubs are too wide.

tests/test_interval.py
```diff
@@ -47,10 +48,10 @@ def test_interval_length() -> None:
 
     assert_type(idres, "pd.Interval[pd.Timestamp]")
     if TYPE_CHECKING:
-        20 in i1  # type: ignore
-        i1 + pd.Timestamp("2000-03-03")  # type: ignore
-        i1 * 3  # type: ignore
-        i1 * pd.Timedelta(seconds=20)  # type: ignore
+        20 in i1
```
Dr-Irv, line 51: Removing this `type: ignore` means we think this code is legal, but it shouldn't be.  You can't check that an `int` is contained inside of an `Interval[Timestamp]`, so the type checker should pick that up.

I'm hesitant to make this change unless we can fix the stub to error out on this line.

I'm pretty sure at one point it did used to pick that up - not sure what changed.

tests/test_interval.py
```diff
@@ -66,8 +67,8 @@ def test_interval_length() -> None:
     assert_type(i2 * 4.2, "pd.Interval[float]")
 
     if TYPE_CHECKING:
-        pd.Timestamp("2001-01-02") in i2  # type: ignore
-        i2 + pd.Timedelta(seconds=20)  # type: ignore
+        pd.Timestamp("2001-01-02") in i2  # pyright: ignore[reportGeneralTypeIssues]
+        i2 + pd.Timedelta(seconds=20)  # pyright: ignore[reportGeneralTypeIssues]
```
Dr-Irv, line 71: This is one of these cases where `pyright` is correctly picking up the error, but `mypy` is not.

Dr-Irv (commented):
As I wrote here: https://github.com/pandas-dev/pandas-stubs/pull/60#issuecomment-1172946284 I'm hesitant to make these changes.  We want to have tests that the type checker should pick up as being invalid code.  Not sure if we change the stubs to make both `mypy` and `pyright` pick them up, or we leave the `warn_unused_ignores` as it was.

Also, see discussion here:  https://github.com/microsoft/pyright/issues/2839

`pyright` did add a feature `reportUnnecessaryTypeIgnoreComment ` as a result of that issue, and maybe you could experiment with that as well.

twoertwein:
> The goal of those tests are code that would raise an Exception if you executed it.

Great, I wasn't aware that there are already negative test cases!

> We want to have tests that the type checker should pick up as being invalid code.

The current way of having the ignore comments but not checking whether they are needed, doesn't test these negative test cases. As you mentioned it would be good to also use pyright to check whether it needs the ignore comments.

I will try to look into why mypy/pyright accept some of these negative tests. If we are not able to make mypy/pyright error on the negative tests, should we simply create an issue to keep track of these negative tests being accepted?

Dr-Irv:
> > The goal of those tests are code that would raise an Exception if you executed it.
> 
> Great, I wasn't aware that there are already negative test cases!
> 
> > We want to have tests that the type checker should pick up as being invalid code.
> 
> The current way of having the ignore comments but not checking whether they are needed, doesn't test these negative test cases. As you mentioned it would be good to also use pyright to check whether it needs the ignore comments.
> 
> I will try to look into why mypy/pyright accept some of these negative tests. If we are not able to make mypy/pyright error on the negative tests, should we simply create an issue to keep track of these negative tests being accepted?

Maybe we should move all of the negative tests to a separate test file, keep the ones (with the correct `# type: ignore` or `#pyright` comments, and then note the ones that are currently "failing", in the sense that we want to type checker to error out on them, and those are the ones that go into an issue.

twoertwein:
I will separate this PR into two PRs: w/wo `warn_unused_ignores`.

pyproject.toml
```diff
@@ -170,18 +170,19 @@ no_implicit_optional = true
...
 # Suppressing errors
 show_none_errors = true
 ignore_errors = false
+enable_error_code = "ignore-without-code"
 # Miscellaneous strictness flags
 allow_untyped_globals = false
 allow_redefinition = false
 local_partial_types = false
-implicit_reexport = true
+implicit_reexport = false
```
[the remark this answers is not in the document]
twoertwein: `implicit_reexport` is still true in pandas

twoertwein:
It will be difficult to have mypy and pyright both reporting which ignores the don't need as mypy has no mypy-specific ignore comments (https://github.com/python/mypy/issues/12358) but pyright does not need a lot of the ignores in the pandas-stubs itself:

```
pandas-stubs/_libs/tslibs/timedeltas.pyi:118:17 - error: Unnecessary "# type: ignore" comment
pandas-stubs/_libs/tslibs/timestamps.pyi:104:67 - error: Unnecessary "# type: ignore" comment
pandas-stubs/_libs/tslibs/timestamps.pyi:136:54 - error: Unnecessary "# type: ignore" comment
pandas-stubs/_libs/tslibs/timestamps.pyi:137:54 - error: Unnecessary "# type: ignore" comment
pandas-stubs/_libs/tslibs/timestamps.pyi:138:54 - error: Unnecessary "# type: ignore" comment
pandas-stubs/_libs/tslibs/timestamps.pyi:139:54 - error: Unnecessary "# type: ignore" comment
pandas-stubs/_libs/tslibs/timestamps.pyi:141:17 - error: Unnecessary "# type: ignore" comment
pandas-stubs/_libs/tslibs/timestamps.pyi:148:17 - error: Unnecessary "# type: ignore" comment
pandas-stubs/core/arraylike.pyi:15:72 - error: Unnecessary "# type: ignore" comment
pandas-stubs/core/arraylike.pyi:16:72 - error: Unnecessary "# type: ignore" comment
pandas-stubs/core/series.pyi:1107:61 - error: Unnecessary "# type: ignore" comment
pandas-stubs/core/series.pyi:1138:61 - error: Unnecessary "# type: ignore" comment
pandas-stubs/core/series.pyi:1751:61 - error: Unnecessary "# type: ignore" comment
pandas-stubs/core/series.pyi:1752:20 - error: Unnecessary "# type: ignore" comment
pandas-stubs/core/arrays/datetimelike.pyi:45:104 - error: Unnecessary "# type: ignore" comment
pandas-stubs/core/arrays/datetimes.pyi:15:64 - error: Unnecessary "# type: ignore" comment
pandas-stubs/core/indexes/base.pyi:208:51 - error: Unnecessary "# type: ignore" comment
```

tests/test_interval.py
twoertwein: I think it will take quite some effort to remove annotations like these
```py
def __add__(self, other: TimestampSeries) -> TimestampSeries: ...
def __add__(self, other: Timestamp) -> TimestampSeries: ..
```
but also allow legit combinations.

Dr-Irv:
> It will be difficult to have mypy and pyright both reporting which ignores the don't need as mypy has no mypy-specific ignore comments ([python/mypy#12358](https://github.com/python/mypy/issues/12358)) but pyright does not need a lot of the ignores in the pandas-stubs itself:

Yes, and that creates the dilemma for us.  We have to put in `#type: ignore` where we know `mypy` is wrong, or where we are expecting the type checker to report an error.  But we can't separate out the cases between `mypy` and `pyright`

I'm presuming that this PR is not ready to be reviewed, given you said you'd split it into two.

twoertwein:
> I'm presuming that this PR is not ready to be reviewed, given you said you'd split it into two.

I think it should be ready to merge: I removed `warn_unused_ignores`.

Personally, I think it would be an improvement to already enable `warn_unused_ignores` (and use pyright-specific codes where needed without requiring that pyright reports its unneeded ignore comments). Especially, after your recent PRs, there are many negative test cases without any ignore comments.

Dr-Irv (approved):

Dr-Irv:
Thanks @twoertwein

Dr-Irv:
> Especially, after your recent PRs, there are many negative test cases without any ignore comments.

Actually, I didn't add any negative tests.  It turned out that using `pytest.skip()` was causing both `mypy` and `pyright` to not check the code that followed.  But that code would not execute with `pytest` (e.g., `df.plot()`).  So I used `if TYPE_CHECKING` to make sure the API was checked

PR #61, opened 2022-07-01
CI: run pyupgrade

twoertwein:
pyupgrade is meant to be used with pre-commit but I squeezed it into the existing black/isort format.

Notes:

- Many files have `from __future__ import annotations`, pyupgrade would re-write all the unions. I disabled that for now (`--keep-runtime-typing`) but I wouldn't mind enabling it.
- I run pyupgrade before black&isort, so that if pyupgrade replaces the Literal import,  isort can sort the imports afterwards.
- `run_pyupgrade_check` still changes a file (but errors immediately after it).


twoertwein:
I think neither black/isort/pyupgrade run on the CI?

BrenoJesusFernandes:
> I think neither black/isort/pyupgrade run on the CI?

Not really, I could antecipated part of CI/CD update and just add this check, what do you think?

Dr-Irv:
> * Many files have `from __future__ import annotations`, pyupgrade would re-write all the unions. I disabled that for now (`--keep-runtime-typing`) but I wouldn't mind enabling it.

I'm not sure we want to rewrite the unions yet in case doing so makes things not work right with python 3.8 or some other odd combination of tools.  We could consider seeing if the stubs work OK without `from __future__ import annotations`

Issue #45, opened 2022-06-28
Pandas and Matplotlib Dependency in the Stubs Package


lovetheguitar:
Could you please remove pandas and matplotlib from the dependencies of the pandas-stubs package.

Why would you have a separate stubs package and then depend on the main package and even matplotlib?

This package was super slim until 6 days ago: https://github.com/VirtusLab/pandas-stubs/blob/master/setup.py#L38-L40

Thank you.

Dr-Irv:
> Most of matplotlib is not annotated. I think the matplotlib type used the most is `Axes` which unfortunately resolves to `Any`.

`matplotlib` *does* have annotations that are shipped with VS Code.  
> 
> ```python
> from matplotlib.axes import Axes
> 
> reveal_type(Axes)  # mypy: any, pyright: Unknown
> ```
> 
In the above example, within VS Code, you will get `Type[Axes]` on the `reveal_type()`

> If the goal of pandas-stubs is to eventually be part of typeshed and be recommeneded by mypy, I think pandas-stubs cannot depend on any actual code (and only contain pyi files).

I'm not sure that the goal is to make `pandas-stubs` part of `typeshed`, although I could see that happening in the future.

twoertwein:
> matplotlib does have annotations that are shipped with VS Code.

Thank you, I didn't know that! If they have a standalone stub package, pandas-stubs (and pandas) should install that at least for the CI. mypy&pyright infer Axes on the pandas CI as Any/Unknown.

Dr-Irv:
> > matplotlib does have annotations that are shipped with VS Code.
> 
> Thank you, I didn't know that! If they have a standalone stub package, pandas-stubs (and pandas) should install that at least for the CI. mypy&pyright infer Axes on the pandas CI as Any/Unknown.

To clarify, the `matplotlib` annotations are maintained by Microsoft in https://github.com/microsoft/python-type-stubs, and then Microsoft merges those stubs into the VS Code distribution.

Dr-Irv:
closed via #66

PR #65, opened 2022-07-01
added type hint to read_csv usecols

sofcalca:
- [ x ] Closes #54
- [ x ] Tests added: Please use `assert_type()` to assert the type of any return value

I could not write a failing test as the previous type hint was Any

Issue #42, opened 2022-06-27
groupby should accept dict, Series, and Index


crusaderky:
mypy 0.961
pandas-stubs 1.4.2.220626
```python
import pandas

df: pandas.DataFrame
df.groupby(df.index)
```
output:
```
error: Argument 1 to "groupby" of "DataFrame" has incompatible type "Index"; expected "Union[List[str], str, None]"  [arg-type]
```

The documentation explicitly caters for Series and dict; it should be expanded to clarify that Index is also OK:
https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.DataFrame.groupby.html

PR #66, opened 2022-07-02
remove installation dependencies on pandas and matplotlib

### LINKED ISSUE #62: Latest release of Pandas (1.4.3) cannot be used with the latest release of pandas-stubs (1.4.2.220626)
Hi,
The latest version of pandas-stubs (1.4.2.220626) depends on Pandas version 1.4.2. This prevents updating to Pandas 1.4.3 if we want to use the latest version of pandas-stubs.

```
ERROR: Cannot install -r requirements.txt (line 22) and pandas==1.4.3 because these package versions have conflicting dependencies.

The conflict is caused by:
    The user requested pandas==1.4.3
    pandas-stubs 1.4.2.220626 depends on pandas==1.4.2

To fix this you could try to:
1. loosen the range of package versions you've specified
2. remove package versions to allow pip attempt to solve the dependency conflict

ERROR: ResolutionImpossible: for help visit https://pip.pypa.io/en/latest/topics/dependency-resolution/#dealing-with-dependency-conflicts
```

Dr-Irv:
- [x] Closes #62 and #45

Moved the dependencies for `pandas`, `matplotlib` and `typing_extensions` to be dev only.

No tests added.  Tested locally to see that you can install the stubs without pandas and matplotlib being installed, and `pyright` and both `mypy` still will check source.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,6 +1,6 @@
 [tool.poetry]
 name = "pandas-stubs"
-version = "1.4.2.220626"
+version = "1.4.3.220702"
 description = "Type annotations for pandas"
 authors = ["The Pandas Development Team <pandas-dev@python.org>"]
 license = "BSD-3-Clause"
@@ -32,9 +32,6 @@ packages = [
 
 [tool.poetry.dependencies]
 python = ">=3.8,<3.11"
-pandas = "1.4.2"
-typing-extensions = ">=4.2.0"
-matplotlib = ">=3.3.2"
 
 [tool.poetry.dev-dependencies]
 mypy = ">=0.960"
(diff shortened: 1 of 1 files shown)


Dr-Irv:
@BrenoJesusFernandes Can you give me your thoughts on this simple change?  I think it will close the two issues #42 and #65 and everything else will still work.

BrenoJesusFernandes:
> @BrenoJesusFernandes Can you give me your thoughts on this simple change? I think it will close the two issues #42 and #65 and everything else will still work.

I read the issues, the arguments make sense to me and it seems not to affect any workflow, so I think everything is ok.

twoertwein:
> Moved the dependencies for `pandas`, `matplotlib` and `typing_extensions` to be dev only.

Depending on how mypy&pyright work, it might be necessary that users have `typing_extensions` installed.

Dr-Irv:
> Depending on how mypy&pyright work, it might be necessary that users have `typing_extensions` installed.

I did try that.  When you install `mypy`, it pulls in `typing_extensions` as a dependency.  If you install `pyright` and not `mypy`, the tests pass, although you get warnings about the fact that `pandas` is not installed.  I'm not worried about that - it seems odd that you'd do type checking of pandas code without pandas being installed.

PR #83, opened 2022-07-04
CI: run style checks on CI

twoertwein:
Run black&isort&pyupgrde on the CI.

There is no need to run it for all python versions/platforms: only running on 3.8@linux for now.


Dr-Irv (approved):
Thanks @twoertwein

twoertwein:
@Dr-Irv Would you prefer using pre-commit for black/isort/pyupgrade? Using pre-commit would also make it easier to add more linters in the future (maybe flake8-pyi).

Dr-Irv:
> @Dr-Irv Would you prefer using pre-commit for black/isort/pyupgrade? Using pre-commit would also make it easier to add more linters in the future (maybe flake8-pyi).

Yes, that would be preferred. Can you set that up?

PR #89, opened 2022-07-05
CI: pre-commit

twoertwein:
I removed black and isort as a project dependency - could also keep it.

@Dr-Irv


twoertwein:
also added flake8 and flake8-pyi but disabled many rules for now

Dr-Irv (approved):

.github/workflows/test.yml
```diff
@@ -49,7 +49,7 @@ jobs:
 
     - if: matrix.python-version == '3.8' && matrix.os == 'ubuntu-latest'
       name: Check Style
-      run: poetry run poe check_style
+      run: poetry run pre-commit run --all-file --verbose
```
twoertwein, line 52: Should probably use `uses: pre-commit/action@v3.0.0` otherwise pre-commit does not cache the tools it has to install. Installing the tools takes longer than running all the checks!

Should probably still keep pre-commit in `pyproject.toml` so that contributors don't need to manually install pre-commit.

twoertwein: switched to github action and also bumped the version of the other github actions.

Issue #91, opened 2022-07-05
CI: Making the CI faster


twoertwein:
Compared to Pandas, the CI is amazingly fast :) but it could be faster. Installing all the dependencies takes the most time compared to any other CI step. Installing dependencies seems especially slow on windows (1min56), it is faster on mac (1min11) and ubuntu (49s).

I think there are at least two approaches:

- Caching: can probably be easily achieves (in combination) with poetry github actions:  [install-poetry](https://github.com/marketplace/actions/install-poetry-action) or [actions-poetry](https://github.com/marketplace/actions/python-poetry-action)
- poetry.lock: If we are fine with all developers using the same version of packages, we could simply commit poetry.lock to the repository. This makes poetry significantly faster as it doesn't need to resolve which packages/versions are compatible with each other. We would need to update poetry.lock every now and then (for example when there is a new mypy version).

PR #92, opened 2022-07-05
CI: testing impact of poetry.lock

twoertwein:
xref #91

Just testing:

- whether it doesn't break anything among the different OS/python versions; and
- how much it speeds up the CI

(If poetry.lock turns out to be helpful, it might be good to commit @Dr-Irv's poetry.lock in case there are some small/annoying differences between different lock files.)

### DIFF
--- a/poetry.lock
+++ b/poetry.lock
@@ -1,0 +1,933 @@
+[[package]]
+name = "atomicwrites"
+version = "1.4.0"
+description = "Atomic file writes."
+category = "dev"
+optional = false
+python-versions = ">=2.7, !=3.0.*, !=3.1.*, !=3.2.*, !=3.3.*"
+
+[[package]]
+name = "attrs"
+version = "21.4.0"
+description = "Classes Without Boilerplate"
+category = "dev"
+optional = false
+python-versions = ">=2.7, !=3.0.*, !=3.1.*, !=3.2.*, !=3.3.*, !=3.4.*"
+
+[package.extras]
+dev = ["coverage[toml] (>=5.0.2)", "hypothesis", "pympler", "pytest (>=4.3.0)", "six", "mypy", "pytest-mypy-plugins", "zope.interface", "furo", "sphinx", "sphinx-notfound-page", "pre-commit", "cloudpickle"]
+docs = ["furo", "sphinx", "zope.interface", "sphinx-notfound-page"]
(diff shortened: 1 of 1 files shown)


twoertwein:
Seems to speed things up: both on python 3.8 (main vs PR)

- ubuntu: 49s vs 27s
- mac: 1min11 vs 27s
- windows: 1min56 vs 1min4

twoertwein:
Instead of commiting `poetry.lock`, it might be possible to have the CI cache poetry.lock for some time (maybe a week).

twoertwein:
On python 3.8
|   | Windows | Mac | Ubuntu |
| - | - | - | - |
| Main  | 1min56  | 1min11 | 49s |
| poetry.lock |  1min4 |27s  | 27s |
| poetry.lock+cache (miss) | 75s  | 38s |  49s |
| poetry.lock+cache (hit) | 27s | 36s | 3s |

For caching, I included the time needed for the cache hook.

[I think there is quite some noise in these measures]

I think the lock file is the more important part (and simple part).

pyproject.toml
```diff
@@ -36,7 +36,7 @@ python = ">=3.8,<3.11"
 [tool.poetry.dev-dependencies]
 mypy = ">=0.960"
 pytest = ">=7.1.2"
-pyright = ">=1.1.251"
+pyright = ">=1.1.255"
```
Dr-Irv, line 39: pyright has released `1.1.257`.  We'll have to figure out a way to keep that updated

twoertwein: Without a lock file, poetry will try to install the latest version anyways. If there is a cached lock file, it might use an older version.

Even if we install an old pyright (from pip, not the "real" pyright), the pip-pyright will first check whether there is a newer version of the "real" pyright. https://pypi.org/project/pyright/

> We also automatically upgrade the pyright npm package to its latest version on every run, see below for how to change this behaviour.

I just bumped the version to check the speed impact when we can use the old venv but have to re-create the lock file.

twoertwein:
> So I'm good to merge this in, if you think it's ready.

I think it is ready! A more conservative alternative could be to just cache poetry.lock: much smaller cache (should especially reduce the time on windows to load/safe the cache) but it doesn't help as much as caching the entire venv.

PR #102, opened 2022-07-06
MAINT: Catch expected warnings

bashtage:
Catch expected warnings to reduce noise in testing


bashtage:
Just did a few to make sure this is of interest before catching the others

Dr-Irv (approved):
Thanks for doing this. Definitely supportive of getting rid of the pytest warnings throughout.

bashtage:
I'll get the others caught soon, then we can enforce no warnings from pandas (but allow some from other tools).

PR #104, opened 2022-07-06
MAINT: Catch all warnings

bashtage:
Catch all warning generated during testing


Dr-Irv (approved):
Thanks @bashtage .  I was thinking of doing this at some point.  Glad you helped out!

PR #110, opened 2022-07-06
a few more flake8-pyi fixes

twoertwein:
Unless `__all__` exists, all variables that do not start with `_` are considered public: need to make TypeVars private.

Based on [flake8-pyi](https://github.com/PyCQA/flake8-pyi#list-of-warnings), only Y037 can be problematic for some type checkers: meaning we could also go for Y022:

> Imports linting: use typing-module aliases to stdlib objects as little as possible (e.g. builtins.list over typing.List, collections.Counter over typing.Counter, etc.).


pandas-stubs/core/reshape/concat.pyi
```diff
@@ -9,13 +9,12 @@ from typing import (
 )
 
 from pandas._typing import (
-    Axis as Axis,
-    DataFrame as DataFrame,
-    Series as Series,
```
twoertwein, line 14: I don't think we want to encourage people doing something like this `from pandas.core.reshape.concat import DataFrame`

Dr-Irv: The `DataFrame as DataFrame` and `Series as Series` appears in LOTS of the files, so at some point those should get cleaned up.

Dr-Irv (approved):
Thanks @twoertwein

PR #112, opened 2022-07-07
BUG: Correct Categorical.codes

### LINKED ISSUE #111: BUG: `Categorical.codes` has wrong type.
**Describe the bug**
Categorical.codes is `list[int]` but should be `ndarray[Any, np.signedinteger]]`

**To Reproduce**
```python
import string
import numpy as np
import pandas as pd

codes: np.ndarray
chars = np.random.choice([c for c in string.ascii_letters],1000*3, replace=True)
codes = pd.Categorical(["".join(c) for c in z.reshape((1000,3))]).codes
print(f"Actual type: {type(codes)}")
```

Output is 

(run)
```
Actual type: <class 'numpy.ndarray'>
```
(typed)
```
error: Incompatible types in assignment (expression has type "List[int]", variable has type "ndarray[Any, Any]")
```

- OS: Windows 11
- OS 22H2
- Python 3.9
- mypy 0.961 (compiled: yes)
- pandas 1.4.2
- pandas-stubs 1.4.3.220704

bashtage:
Set type to correct type

- [X] Closes #111 (Replace xxxx with the Github issue number)

### DIFF
--- a/tests/test_series.py
+++ b/tests/test_series.py
@@ -11,7 +11,6 @@ from typing import (
 )
 
 import numpy as np
-from numpy import typing as npt
 import pandas as pd
 from pandas.api.extensions import ExtensionArray
 from pandas.core.window import ExponentialMovingWindow
@@ -21,6 +20,9 @@ from typing_extensions import assert_type
 from pandas._typing import Scalar
 
 from tests import check
+
+if TYPE_CHECKING:
+    from pandas._typing import np_ndarray_int  # noqa: F401
 
 
 def test_types_init() -> None:
@@ -801,11 +803,5 @@ def test_types_iter() -> None:
 
 def test_categorical_codes():
     # GH-111
-    codes: Union[
-        npt.NDArray[np.int8],
-        npt.NDArray[np.int16],
-        npt.NDArray[np.int32],
-        npt.NDArray[np.int64],
-    ]
     cat = pd.Categorical(["a", "b", "a"])
-    codes = cat.codes
+    assert_type(cat.codes, "np_ndarray_int")


tests/test_series.py
```diff
@@ -781,3 +782,15 @@ def test_iloc_setitem_ndarray() -> None:
...
+def test_categorical_codes():
+    # GH-111
+    codes: Union[
+        npt.NDArray[np.int8],
+        npt.NDArray[np.int16],
+        npt.NDArray[np.int32],
+        npt.NDArray[np.int64],
+    ]
+    cat = pd.Categorical(["a", "b", "a"])
+    codes = cat.codes
```
Dr-Irv, line 796: Can you do this instead?
1.  import `np_ndarray_int` as follows at the top:
```python
if TYPE_CHECKING:
    from pandas._typing import np_ndarray_int
```
2. Then  you can remove the type of `codes` in the test, and just do:
```python
assert_type(cat.codes, "np_ndarray_int")
```

Dr-Irv (changes requested):

tests/test_series.py
```diff
@@ -781,3 +784,10 @@ def test_iloc_setitem_ndarray() -> None:
     values_s.iloc[indices_u16] = -1
     values_s.iloc[indices_u32] = -1
     values_s.iloc[indices_u64] = -1
+
+
+def test_categorical_codes():
+    # GH-111
+    cat = pd.Categorical(["a", "b", "a"])
+    codes = cat.codes
```
Dr-Irv, line 792: Needs the `assert_type`

Dr-Irv (changes requested):

Dr-Irv:
@bashtage can you add the `assert_type()` as discussed above?  Then we can get this one done.

tests/test_series.py
bashtage: Added

bashtage:
I had to noqa the import to get it to pass flake8 since the import is only used in the assert type.

Dr-Irv (approved):
Thanks @bashtage

twoertwein:
> I had to noqa the import to get it to pass flake8 since the import is only used in the assert type.

The second argument of `assert_type` can not only be of type `str` but also the type itself.

bashtage:
@twoertwein When I remove the quotes it doesn't work in two different ways:

1. If I leave the if TYPE_CHCKING on the import, pytest fails `NameError: name 'np_ndarray_int' is not defined`
2. If I take away the if TYPECHECKING on the import, I get ` ImportError: cannot import name 'np_ndarray_int' from 'pandas._typing'`

Issue #97, opened 2022-07-05
BUG: Overloads for accessing columns of a DataFrame are too restrictive


bashtage:
**Describe the bug**
Current overloads for accessing columns of a DataFrame are too restrictive

**To Reproduce**
```
from __future__ import annotations
import pandas as pd

df = pd.DataFrame({"a":[1,2,3],1:[3,4,5]})
key: list[int | str]
key = [1]
df[key]
```
returns

```
Invalid index type "List[Union[int, str]]" for "DataFrame"; expected type "Union[Tuple[Any, ...], Series[bool], DataFrame, List[str], List[Union[str, bytes, date, datetime, timedelta, bool, int, float, complex, Timestamp, Timedelta]], Index, ndarray[Any, dtype[str_]], ndarray[Any, dtype[bool_]], Sequence[Tuple[Union[str, bytes, date, datetime, timedelta, bool, int, float, complex, Timestamp, Timedelta], ...]]]"
``

I suspect this is a covariance issue (or invariance of `list`)

**Please complete the following information:**
- Windows 11
- Python 3.9
```
mypy                              0.961
mypy-extensions                   0.4.3
pandas                            1.4.2                              
pandas-stubs                      1.4.3.220704
```

**Comment by Dr-Irv** [comment]
> Ignoring non-list sequences: I think it might be sufficient to replace `List[str]` with `list[HashableT]` where `HashableT = TypeVar("HashableT", bound=Hashable)`.

`list[HashableT]` would be too wide.  Would accept any hashable type, and that would include things that are not acceptable to pandas (e.g., a list of numpy arrays)

I'm a bit surprised that a type of `List[Union[int, str]]` didn't match `List[Scalar]` that is defined for `DataFrame.__getitem__()`.  This may require further investigation.  Could be a mypy bug.  Should check with pyright.

**Comment by bashtage** [comment]
I don't think NumPy arrays are washable. 
```
In [1]: from typing import Hashable

In [2]: import numpy as np

In [3]: np.array([1,2])
Out[3]: array([1, 2])

In [4]: isinstance(np.array([1,2]),Hashable)
Out[4]: False

In [7]: hash(np.array([1,2]))
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
Input In [7], in <cell line: 1>()
----> 1 hash(np.array([1,2]))

TypeError: unhashable type: 'numpy.ndarray'
```

[...]

PR #113, opened 2022-07-07
TYP: Series/DataFrame/Index are not Hashable

twoertwein:
- [x] relate to #97


twoertwein:
I cannot reproduce the dist failure of pyright:

```sh
poetry run poe install_dist
poetry run poe rename_src
poetry run poe pyright_dist  # no errors
```

tests/__init__.py
```diff
@@ -0,0 +1,27 @@
...
+    elif expected is None:
+        error = actual_object is not None
+    else:
+        error = not isinstance(actual_object, expected)
+
+    if error:
+        raise TypeError(f"Expected '{expected}' got '{actual}'")
+
+
+typing_extensions.assert_type = assert_type
```
twoertwein, line 27: A bit ugly and only a rough check (if the actual object is of a sub-class, the caller might need to pass the type and not the name of the class) but it might make the test more concise.

Dr-Irv:
> I cannot reproduce the dist failure of pyright:

I looked in your log and saw this in the `install_dist` step:
```
pandas-stubs is already installed with the same version as the provided wheel. Use --force-reinstall to force an installation of the wheel.
```

So it may be that the caching you introduced is keeping around a version of `pandas-stubs` from a previous test??

tests/__init__.py
Dr-Irv: I don't understand why you're overriding the default implementation

Dr-Irv:
> Yes, that must be the case! Thank you for finding this! Let's just cache poetry.lock.

MIght be a good idea to change `build_dist` to do a `--force-reinstall`

tests/__init__.py
twoertwein: Ideally, we should have test like this:
```py
assert_type(pd.DataFrame.__hash__, None)
assert pd.DataFrame.__hash__ is None
```
Replacing `assert_type` at runtime with a rudimentary check, helps us to keep the tests simpler.

twoertwein: Happy to remove it, just thought it might be helpful (even though it is an ugly hack)

Dr-Irv: > Happy to remove it, just thought it might be helpful (even though it is an ugly hack)

I just don't see the benefit and I'd rather not be hacking things in libraries.

The goal of the tests is two-fold:
1. Test with `mypy` and `pyright` to see that "user" code would pass the type checkers
2. Test with `pytest` to see that it is valid pandas code that is runnable.

`assert_type()` at runtime just returns the first argument.  Based on my reading of the docs, I think you could do something like:
```python
assert_type(pd.DataFrame.__hash__, None) is None
```

Dr-Irv:
ping when you want me to review the whole thing (and it's all green)

twoertwein:
Green and ready for review @Dr-Irv

Caching just `poetry.lock` seems to make the installation twice as fast while caching `poetry.lock` and the venv makes it three times as fast. I'm fine just caching `poetry.lock`.

pandas-stubs/core/frame.pyi
```diff
@@ -440,7 +443,7 @@ class DataFrame(NDFrame, OpsMixin):
         *,
         axis: Axis = ...,
         index: Hashable | Sequence[Hashable] = ...,
-        columns: Hashable | Sequence[Hashable] = ...,
+        columns: Hashable | Sequence[Hashable] | Index = ...,
```
Dr-Irv, line 446: same as above.  If `Index` is now `Hashable`, do you need to change this?

pandas-stubs/core/frame.pyi
```diff
@@ -452,7 +455,7 @@ class DataFrame(NDFrame, OpsMixin):
...
```
Dr-Irv, line 458: same as above.  If `Index` is now `Hashable`, do you need to change this?

Dr-Irv (changes requested):

pandas-stubs/core/frame.pyi
```diff
@@ -428,7 +431,7 @@ class DataFrame(NDFrame, OpsMixin):
...
```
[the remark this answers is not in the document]
twoertwein: `Index` was declared to be `Hashable` but it isn't.

pandas-stubs/core/indexes/base.pyi
```diff
@@ -169,7 +170,7 @@ class Index(IndexOpsMixin, PandasObject):
     def where(self, cond, other=...): ...
     def is_type_compatible(self, kind) -> bool: ...
     def __contains__(self, key) -> bool: ...
-    def __hash__(self) -> int: ...
+    __hash__: ClassVar[None]  # type: ignore[assignment]
```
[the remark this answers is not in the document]
twoertwein: done

pandas-stubs/core/frame.pyi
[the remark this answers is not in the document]
twoertwein: `Index` probably needs to be added to many more annotations. Just added it here because a test was falling without it.

Dr-Irv:
You could add a negative test to verify that these types are not `Hashable`, e.g.,
```python
def test_not_hashable():
    def test_func(h: Hashable):
        pass
    test_func(pd.Series()) # type: ignore
```

twoertwein:
> You could add a negative test to verify that these types are not `Hashable`, e.g.,

Added and green

pandas-stubs/core/generic.pyi
```diff
@@ -89,7 +90,7 @@ class NDFrame(PandasObject, indexing.IndexingMixin):
     def bool(self) -> _bool: ...
     def __abs__(self) -> NDFrame: ...
     def __round__(self, decimals: int = ...) -> NDFrame: ...
-    def __hash__(self): ...
+    __hash__: ClassVar[None]  # type: ignore[assignment]
```
Dr-Irv, line 93: I think we should move this one up to the beginning of the class too.  Sorry for not catching it earlier.

Dr-Irv (changes requested):

pandas-stubs/core/generic.pyi
twoertwein: done and green again :)

Dr-Irv (approved):
Thanks @twoertwein

PR #114, opened 2022-07-08
assert types at runtime

twoertwein:
replace `assert_type(...)` with `assert isinstance(assert_type(...), ...)` or `assert assert_type(...) is None`


Dr-Irv (commented):
So I like this change, but I'm wondering if we should create some standard functions in `tests/__init__.py` that help shorten up the `assert` statements.  For example, we could have
```python
def check_series(s: pd.Series, dtype: Optional[DTypeArg] = None):
    assert isinstance(s, pd.Series)
    if dtype is not None:
        assert s.dtype is dtype
```
then instead of
```python
assert isinstance(
        assert_type(series, "pd.Series[bool]"), pd.Series
    ) and series.dtype is np.dtype(bool)
assert isinstance(assert_type(s, pd.Series), pd.Series)
```
could do this:
```python
    check_series(assert_type(series, "pd.Series[bool]"), np.dtype(bool))
    check_series(assert_type(s, pd.Series))
```
Then we have a standard way of checking series with types.  Could do same with the generic `Interval` as well.

Before `assert_type()` was introduced, I did have something like that in the MS stubs.  See
https://github.com/microsoft/python-type-stubs/blob/6b800063bde687cd1846122431e2a729a9de625a/tests/pandas/__init__.py

But now that we know that `assert_type()` returns the value, we could combine that with the ideas there.

I could let this PR go as it is, and we make my suggested change in a later PR.  Open to your opinion.

tests/__init__.py
```diff
@@ -0,0 +1,20 @@
...
+
+    if not isinstance(actual, klass):
+        raise RuntimeError(f"Expected type '{klass}' but got '{type(actual)}'")
+    if dtype is None:
+        return None
+
+    if hasattr(actual, "__iter__"):
+        value = next(iter(actual))  # type: ignore[call-overload]
+    else:
+        value = actual.left  # type: ignore[attr-defined]
```
twoertwein, line 16: checking the actual value is imho better than checking `.dtype`:

- users care about the actual values and
- it simplified the type checking (otherwise we have sometimes a np.dtype object, sometimes an actual type, and sometimes a function)

Dr-Irv, line 16: The `.left` attribute is arbitrary.  So could add parameter to `check` which is `attr: Optional[str] = None` and then do:
```python
if attr is None:
    value = next(iter(actual)) # type: ignore[call-overload]
else:
    value = actual.__getattr__(attr)
```
Then the caller can decide which attribute to check.

Dr-Irv: Maybe we should check value _and_ `.dtype`.  For things like `Series[bool]`, we want to make sure that the `dtype` correspods to the generic attribute.

tests/__init__.py
```diff
@@ -0,0 +1,20 @@
+from __future__ import annotations
+
+from typing import Callable
+
+
+def check(actual: object, klass: type, dtype: type | None = None) -> None:
```
Dr-Irv, line 6: One advantage of having separate `check` functions for each type (as was done prior to the introduction of `assert_type()`) is that we get extra type checking.  So `check_dataframe_result(result: pd.DataFrame)` is then checked by the type checker if we do `check_dataframe_result(pd.DataFrame({"x": [1,2,3]})`

Dr-Irv (changes requested):
Suggested improvements to `check()`

tests/__init__.py
twoertwein: I will try it but I think that dtype is often a numpy dtype while the values (when retrieving them) are builtin types

twoertwein: I prefer to revisit that maybe in the future: dtype is  often a numpy dtype which is not compatible with the type of the value (str vs. numpy.dtype[object_], ...).

twoertwein: Added the `attr` argument. I gave it the default value `"left"` to keep the `check(...)` calls short.

tests/__init__.py
twoertwein: If the `check` functions becomes too large (maybe when checking the dtype), it would definitely be good to have dedicated check functions. At the moment I slightly prefer one function with a short name to keep the tests somewhat more skimmable (avoiding more line break).

Dr-Irv (approved):
Thanks @twoertwein

PR #115, opened 2022-07-08
ENH: Overload to_datetime when errors="coerce"

### LINKED ISSUE #88: BUG: `to_datetime` can return `NaTType`
```python

import pandas as pd
out = pd.to_datetime("dajsldaljd", errors="coerce")
reveal_type(out)
```

wrong type returns

```
note: Revealed type is "pandas._libs.tslibs.timestamps.Timestamp"
```

Also, `mypy` on

```python
import pandas as pd
out = pd.to_datetime("dajsldaljd", errors="coerce")
out is pd.NaT
````

returns

```
error: Non-overlapping identity check (left operand type: "Timestamp", right operand type: "NaTType")
```

### PR DESCRIPTION
**Pull request by bashtage**
Have to_datetime returning NaTType when errors are coerced and scalar input

- [X] Closes #88 (Replace xxxx with the Github issue number)

### DIFF
--- a/tests/test_timefuncs.py
+++ b/tests/test_timefuncs.py
@@ -215,8 +215,26 @@ def test_comparisons_datetimeindex() -> None:
 
 
 def test_to_datetime_nat() -> None:
-    assert_type(pd.to_datetime("2021-03-01", errors="ignore"), "pd.Timestamp")
-    assert_type(pd.to_datetime("2021-03-01", errors="raise"), "pd.Timestamp")
-    assert_type(
-        pd.to_datetime("2021-03-01", errors="coerce"), "Union[pd.Timestamp, NaTType]"
+    # GH 88
+    assert isinstance(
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: tests/test_timefuncs.py [hunk 1]
```diff
@@ -209,3 +212,28 @@ def test_comparisons_datetimeindex() -> None:
     assert_type((dti <= ts), np_ndarray_bool)
     assert_type((dti == ts), np_ndarray_bool)
     assert_type((dti != ts), np_ndarray_bool)
+
+
+def test_to_datetime_nat() -> None:
```
Dr-Irv, line 217: Can you add the GH issue here?

Dr-Irv (changes requested):
One small change, otherwise looks good

tests/test_timefuncs.py
bashtage: Added now

Dr-Irv (approved):
Thanks @bashtage

PR #118, opened 2022-07-08
Remove unintended redundant imports

twoertwein:
There might be many more cases. I grep'ed for redundant imports of Index/Series/DataFrame and then adjusted those files (excluding api, and init files).


bashtage:
LGTM.  Good to have a single authoritative entry point for types.

bashtage (approved):
Good idea.  Probably can have follow ups looking at other `as` imports.

Dr-Irv (approved):
Thanks @twoertwein

PR #120, opened 2022-07-08
is the mypy_pkg_file needed?

### LINKED ISSUE #68: poetry mypy_src step needs to create the pth file
`poetry mypy_src` fails in a clean environment because that step doesn't include the `create_mypy_pkg_file` step

@BrenoJesusFernandes 

Workaround now is to do `poetry test_src`

twoertwein:
closes #68

The dist test run locally fine for me without having to create this file.

### DIFF
--- a/scripts/test/__init__.py
+++ b/scripts/test/__init__.py
@@ -17,7 +17,7 @@ def test_src(profile: str, clean_cache: bool = False):
     elif profile == "pytest":
         steps.extend([_step.pytest_src])
     elif profile == "full":
-        steps.extend([mypy_src, _step.pyright_src, _step.pytest_src])
+        steps.extend([_step.pyright_src, _step.pytest_src])
     else:
         raise ModuleNotFoundError("Profile not found!")
 
@@ -61,7 +61,6 @@ def test_all(clean_cache: bool = False):
 
     steps.extend(
         [
-            step.mypy_src,
             _step.pyright_src,
             _step.pytest_src,
             _step.build_dist,
--- a/scripts/test/__init__.py
+++ b/scripts/test/__init__.py
@@ -17,7 +17,7 @@ def test_src(profile: str, clean_cache: bool = False):
     elif profile == "pytest":
         steps.extend([_step.pytest_src])
     elif profile == "full":
-        steps.extend([_step.pyright_src, _step.pytest_src])
+        steps.extend([_step.mypy_src, _step.pyright_src, _step.pytest_src])
     else:
         raise ModuleNotFoundError("Profile not found!")
 
@@ -61,6 +61,7 @@ def test_all(clean_cache: bool = False):
 
     steps.extend(
         [
+            _step.mypy_src,
             _step.pyright_src,
             _step.pytest_src,
             _step.build_dist,
--- a/scripts/test/__init__.py
+++ b/scripts/test/__init__.py
@@ -13,26 +13,11 @@ def test_src(profile: str, clean_cache: bool = False):
         )
 
     if profile in (None, "", "default"):
-        steps.extend(
-            [
-                _step.create_mypy_pkg_file,
-                _step.mypy_src,
-                _step.destroy_mypy_pkg_file,
-                _step.pyright_src,
-            ]
-        )
+        steps.extend([_step.mypy_src, _step.pyright_src])
     elif profile == "pytest":
         steps.extend([_step.pytest_src])
     elif profile == "full":
-        steps.extend(
-            [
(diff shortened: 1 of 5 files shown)


Dr-Irv:
> The dist test run locally fine for me without having to create this file.

The file is created for the source test.  Put a typing bug in a test, and you'll see it is not found because the local copy of `pandas-stubs` is not used.

Here are the possible ways for the local copy to be tested:
1) change name of directory to `typings/pandas` and modify `pyproject.toml` to tell `mypy` to look there.
2) Install the local copy (but that is what `dist` does)
3) Create a `.pth` file so it knows to look in the current directory.

BrenoJesusFernandes:
tests are passing maybe because the cache from previous builds.

twoertwein:
> The file is created for the source test. Put a typing bug in a test, and you'll see it is not found because the local copy of `pandas-stubs` is not used.

It sounds as this is only needed for local development, or? I installed the distribution, and then added a bug in the local copy, ran mypy, and it found the issue. Do you mind posting the steps which create the issue that you observe?

twoertwein:
> tests are passing maybe because the cache from previous builds.

no longer caching the venv, just the poetry.lock file

twoertwein:
> ```shell
> mypy tests/test_frame.py
> ```

I can confirm that I get errors then, but I usually never run `mypy <specific file>`, I always run `mypy tests/ pandas-stubs/` which works fine even in that case.

twoertwein:
How about we add `files = ["pandas-stubs", "tests"]` to pyproject.toml, and run just `mypy` (no file arguments are needed)?

I don't think it is worth having this workaround on the CI just to make `mypy tests/test_frame.py` work, mypy is very fast on pandas-stubs and the tests.

BrenoJesusFernandes:
I am feeling we are missing something ... when I arrive at my machine I am gonna test again to see if the behaviour is the same.

twoertwein:
Do you have a refrence to an existing mypy issue about this? I think it could be a mypy bug, if we add `mypy_path="."` it should always find pandas-stubs, but it doesn't!

Dr-Irv:
See this:
https://mypy.readthedocs.io/en/stable/installed_packages.html#installed-packages

> Starting in mypy 0.900, most third-party package stubs must be installed explicitly.

And this:
https://mypy.readthedocs.io/en/stable/stubs.html

> Note that [PEP 561](https://peps.python.org/pep-0561/) stub-only packages must be installed, and may not be pointed at through the MYPYPATH

Also, `mypy tests` doesn't work either.

twoertwein:
Thank you @Dr-Irv that explains why it isn't working. Then how about:

- only testing the `pandas-stubs` (locally) with mypy and pyright
- install `pandas-stubs`
- and then test `tests` with mypy and pyright using the installed stubs?

Dr-Irv:
No. Because in development, it is really useful (and it saves time) to run the tests without building a dist and installing them.

I have had times where the tests pass `pyright`, but not `mypy`, so it is useful to just run `mypy` by itself.

twoertwein:
You can always run `mypy pandas-stubs/ tests/` without having to install anything (for some reason that works).

scripts/test/__init__.py
```diff
@@ -76,9 +61,6 @@ def test_all(clean_cache: bool = False):
 
     steps.extend(
         [
-            _step.create_mypy_pkg_file,
-            _step.mypy_src,
```
Dr-Irv, line 80: can you keep this step in there?

scripts/test/__init__.py
```diff
@@ -13,26 +13,11 @@ def test_src(profile: str, clean_cache: bool = False):
...
-            ]
-        )
+        steps.extend([_step.mypy_src, _step.pyright_src])
     elif profile == "pytest":
         steps.extend([_step.pytest_src])
     elif profile == "full":
-        steps.extend(
-            [
-                _step.create_mypy_pkg_file,
-                _step.mypy_src,
```
Dr-Irv, line 30: can you keep the mypy step in there?

Dr-Irv (changes requested):
So since `mypy tests pandas-stubs` works on the local source, let's keep that in there so you can just test locally.

scripts/test/__init__.py
twoertwein: Thank you, I removed that line accidentally.

Dr-Irv (approved):
thanks @twoertwein

BrenoJesusFernandes:
Indeed is working without the file in my setup, but I'm still curious why? haha

Issue #121, opened 2022-07-08
CI failing on Windows - maybe due to poetry.lock


Dr-Irv:
See https://github.com/pandas-dev/pandas-stubs/runs/7256706305?check_suite_focus=true which happened after I merged a PR to `main`.  Might be due to the `poetry.lock` stuff??

Dr-Irv:
> It could be a instabability in pypi serve, did you try to re-run the failed job again?

Yes, I did.  It failed the second time as well.  Looking at the log, it is failing when doing an install.

BrenoJesusFernandes:
maybe replacing in CI file `poetry install -vvv --no-root` with `poetry update -vvv` could fix it.

twoertwein:
if it is caused by poetry.lock, a simple workaround is to make any change to pyproject.toml.

I think poetry update, would also ignore the poetry.lock

Dr-Irv:
I updated `pyproject.toml` and now `main` is passing. 

@BrenoJesusFernandes or @twoertwein - can one of you create a PR that does what was suggested here? https://github.com/pandas-dev/pandas-stubs/issues/121#issuecomment-1179278018

I'm concerned that if I merge in the future, then I'll have the same problem where a PR looks good, but I merge, and then the action on `main` fails.

twoertwein:
In theory, peotry.lock should exactly be the same across different python and OS version. I will create a PR to cache poetry-lock separately for each OS/python version.  If it still fails then, then it could be a poetry/network issue.

BrenoJesusFernandes:
> maybe replacing in CI file `poetry install -vvv --no-root` with `poetry update -vvv` could fix it.

@twoertwein I tested here this idea and it looks like it works

twoertwein:
Doesn't update completely ignore poetry.lock? Meaning we get no speedup by caching poetry.lock? Could go back to caching the entire venv. Personally, I find caching poetry.lock very elegant, one small file (almost no overhead on a miss) and quite some speedup.

BrenoJesusFernandes:
hmm I see the point, I totally forgot about the cache system.

twoertwein:
unrelated: do we need to update pip on the CI? It seems that it is already the newest version and in [some cases](https://github.com/pandas-dev/pandas-stubs/runs/7259308979?check_suite_focus=true) checking for updates takes 50 seconds.

BrenoJesusFernandes:
Personally, I don't think is necessary.

[...]

PR #122, opened 2022-07-08
cache poetry.lock for each OS/python version/poetry version

twoertwein:
xref #121

There was also a [poetry update today](https://github.com/python-poetry/poetry/releases/tag/1.1.14), maybe that caused some weird issues.


Dr-Irv (approved):
Thanks @twoertwein

PR #123, opened 2022-07-08
flake8: py/pyi specific ruleset

twoertwein:


pandas-stubs/api/types/__init__.pyi
```diff
@@ -1,6 +1,48 @@
 from pandas._libs.lib import infer_dtype as infer_dtype
 
-from pandas.core.dtypes.api import *
+from pandas.core.dtypes.api import (
```
twoertwein, line 3: We should consider using `__all__` instead of using the redundant exports (I think typeshed is using `__all__`). I believe that mypy/pyright allow simple operations like  `__all__ = [...] + other_module.__all__`

Dr-Irv (approved):
Thanks @twoertwein

PR #124, opened 2022-07-10
Allow subtypes of List[Scalar] by introducing ScalarArg

### LINKED ISSUE #97: BUG: Overloads for accessing columns of a DataFrame are too restrictive

Dr-Irv:
- [x] Closes #97 
- [x] Tests added:
    - `test_frame.py:test_columns_mixlist()`


pandas-stubs/_typing.pyi
```diff
@@ -122,6 +122,7 @@ Scalar = Union[
     Timestamp,
     Timedelta,
 ]
+ScalarArg = TypeVar("ScalarArg", bound=Scalar)
```
twoertwein, line 125: Maybe `ScalarT` to emphasize that it is a `TypeVar`.

Even though `_typing.py` is private, it might be good to make all variables that are not actually in pandas explicitly private. Users might want to import some of the type variables from pandas to annotate their own code.

Dr-Irv: > Maybe `ScalarT` to emphasize that it is a `TypeVar`.
> 
Good idea.  Done in next commit.

> Even though `_typing.py` is private, it might be good to make all variables that are not actually in pandas explicitly private. Users might want to import some of the type variables from pandas to annotate their own code.

That could be something we do in the future.  One thing I would like to do is reconcile the things in `pandas._typing.py` and `pandas-stubs._typing.pyi`, because there are some inconsistencies.   Created #128

Dr-Irv:
@twoertwein once this is approved and merged, I'm going to do a new release.

twoertwein (approved):

twoertwein:
Thank you @Dr-Irv !

PR #129, opened 2022-07-10
CI: run local type tests only one time

twoertwein:
The local tests are (mostly) covered by the installed tests. The only use case I can see is to test that the mypy workaround to avoid installing the stubs works.

This seems to save around 40s-50s on windows.


twoertwein:
Closing as the dist tests only test using the stubs for the tests but don't check the stubs themselves.

PR #136, opened 2022-07-12
enable reportUnusedImport

twoertwein:
flake8 also has a check for unused imports but it seems that it isn't very useful in pyi files (seems to report all imports as being unused?).


Dr-Irv (approved):
Thanks @twoertwein 

Is it a goal to see if we can get more of the `pyright` strict tests to pass?  Or will that be pretty hard to do?

twoertwein:
> Is it a goal to see if we can get more of the `pyright` strict tests to pass? Or will that be pretty hard to do?

Some will be pretty hard.

Some of them would need to be fixed in pandas: reportInconsistentConstructor and reportPrivateUsage.

Some need to be fixed in other libraries: reportMissingTypeStubs and reportUntypedBaseClass.

Some can be fixed here: reportIncompleteStub and reportUnusedVariable (all variables should be used, even if it is just in `assert_type`)

PR #137, opened 2022-07-13
Rename poe tasks

twoertwein:
This should make `poe --help` easier to skim and clearer.

image

Happy to also rename `Steps` (in `scripts/test/_step.`py) and the github workflow steps, if the changes in pyproject.toml are okay.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -80,9 +80,13 @@ script = "scripts.test:test(clean_cache, src=True, dist=True)"
     options = ["-c", "--clean_cache"]
     default = false
 
-[tool.poe.tasks.pytest_src]
+[tool.poe.tasks.pytest]
 help = "Run pytest"
-script = "scripts.test.run:pytest_src"
+script = "scripts.test.run:pytest"
+
+[tool.poe.tasks.style]
+help = "Run pre-commit"
+script = "scripts.test.run:style"
 
 [tool.poe.tasks.mypy_src]
 help = "Run mypy on 'tests' (using the local stubs) and on the local stubs"
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -50,6 +50,16 @@ isort = ">=5.10.1"
 requires = ["poetry-core>=1.0.0"]
 build-backend = "poetry.core.masonry.api"
 
+[tool.poe.tasks.test_all]
+help = "Run all tests"
+script = "scripts.test:test(clean_cache, src=True, dist=True)"
+
+    [[tool.poe.tasks.test_all.args]]
+    help = "remove cache folders (mypy and pytest)"
+    name = "clean-cache"
+    options = ["-c", "--clean_cache"]
+    default = false
+
 [tool.poe.tasks.test_src]
 help = "Run local tests (includes 'mypy_src', 'pyright_src', 'pytest', and 'style')"
 script = "scripts.test:test(clean_cache, src=True)"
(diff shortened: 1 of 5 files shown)


pyproject.toml
```diff
@@ -51,75 +51,72 @@ requires = ["poetry-core>=1.0.0"]
...
+help = "Run mypy on 'tests' using the installed stubs"
+script = "scripts.test.run:mypy_dist"
+
 [tool.poe.tasks.pyright_src]
-help = "CI Test | Run pyright against source code"
+help = "Run pyright on 'tests' (using the local stubs) and on the local stubs"
 script = "scripts.test.run:pyright_src"
 
-[tool.poe.tasks.pytest_src]
-help = "CI Test | Run pytest against source code"
```
Dr-Irv, line 98: Can we keep this one in there?  It's useful sometimes to just run `pytest` and not `mypy` and `pyright` ?

Dr-Irv (changes requested):

pyproject.toml
twoertwein: test_src has some redundancy with the other tests. Pytest can still be run with `poe test_src --profile=pytest`.

I think there are two options:

1. add more options to test_src (for example only mypy and only pyright) and remove pytest_src, mypy_src, and pyright_src or
2. keep pytest_src but use test_src only for running all local tests (remove mypy+pyright, only pytest, only pre-commit).

I slightly prefer option 2.

twoertwein: Added `pytest_src` back and removed the profiles from `test_src`

Dr-Irv: How about this:

- have `test_src` have no profile, do `mypy`, `pyright`, `pytest` and `style`
- add back `pytest_src`
- add `style`

then we have the following options:

- local tests
  -  `test_src` (does all of the next 4)
  - `mypy_src`
  - `pyright_src`
  - `pytest_src`
  - `style`
- dist tests (each of the following builds the dist, installs the dist, renames the source, does the test(s), renames the source back, uninstalls the dist)
  - `test_dist` (does all of the next 3)
  - `mypy_dist`
  - `pyright_dist`
  - `pytest_dist` (should add that)
  
Don't think we need `install_dist` and `rename_src` as top level options.

twoertwein: >     * add `style`

Yeah, that is missing right now.

>     * dist tests (each of the following builds the dist, installs the dist, renames the source, does the test(s), renames the source back, uninstalls the dist)
>       
>       * `test_dist` (does all of the next 3)
>       * `mypy_dist`
>       * `pyright_dist`

I like that as it makes testing the installation easier while developing but it will make the CI slower (unless the CI just calls test_dist)

>       * `pytest_dist` (should add that)

tests/ is not part of the wheel (and I think it is good that it isn't part of the wheel)

Dr-Irv: > I like that as it makes testing the installation easier while developing but it will make the CI slower (unless the CI just calls test_dist)

I'm fine if the CI just calls `test_dist` or just calls the 3 individual pieces (which I think it does now)

> tests/ is not part of the wheel (and I think it is good that it isn't part of the wheel)

If you look at the `poe` run, both `mypy_dist` and `pyright_dist` test the `tests` folder against the installed distribution, so we're fine.

Dr-Irv: looks good

Dr-Irv (approved):
Thanks @twoertwein .  Nice improvement.

Issue #133, opened 2022-07-12
BUG: Invalid index type "Tuple[slice, int]" for "_LocIndexerFrame";


bashtage:
**Describe the bug**
Standard loc operations fails

**To Reproduce**
```python

import pandas as pd

df = pd.DataFrane({1000: [1,2,3], 2000:[3,4,5]})
df.loc[:, 2000]
````

Produces

```
 error: Invalid index type "Tuple[slice, int]" for "_LocIndexerFrame"; expected type "Union[Union[slice, ndarray[Any, dtype[signedinteger[_64Bit]]], Index, List[int], Series[int]], Union[Series[bool], ndarray[Any, dtype[bool_]], List[bool]],
 List[Union[str, str_]], Tuple[Union[Union[slice, ndarray[Any, dtype[signedinteger[_64Bit]]], Index, List[int], Series[int]], Union[Series[bool], ndarray[Any, dtype[bool_]], List[bool]], slice, List[Union[str, str_]], Tuple[Union[str, int, slice], ...]], Union[List[Union[str, str_]], slice, Series[bool], Callable[..., Any]]]]"
```

**Please complete the following information:**
 - Windows
 - 11
 - 3.10
 - mypy 0.961
 - pandas-stubs                      1.4.3.220710

**Comment by bashtage** [comment]
Loosely speaking I think `loc` should take some combinations like

```
_sliceT = Union[slice, pd.IndexSlice]
Label = Hashable | None

tuple[Sequence[Label], Sequence[Label]]
tuple[Sequence[Label]]
tuple[_sliceT]
tuple[Sequence[Label], _sliceT ]
tuple[_sliceT , Sequence[Label]]
tuple[_sliceT , Sequence[Label]]
```

bashtage:
closed in #138

PR #138, opened 2022-07-13
ENH: Improve DataFrame loc indexing

bashtage:
Allow any scalar types in .loc to return a Series

xref #133

- [X] Closes N/A but makes forward progress

### DIFF
--- a/pandas-stubs/core/frame.pyi
+++ b/pandas-stubs/core/frame.pyi
@@ -138,17 +138,10 @@ class _LocIndexerFrame(_LocIndexer):
     @overload
     def __getitem__(
         self,
-        idx: Union[int, StrLike],
-    ) -> Series: ...
-    @overload
-    def __getitem__(
-        self,
-        idx: Tuple[Union[IndexType, MaskType], StrLike],
-    ) -> Series: ...
-    @overload
-    def __getitem__(
-        self,
-        idx: Tuple[Tuple[slice, ...], StrLike],
+        idx: Union[
+            Union[ScalarT, None],
+            Tuple[Union[IndexType, MaskType, Tuple[slice, ...]], Union[ScalarT, None]],
+        ],
     ) -> Series: ...
     @overload
     def __setitem__(
--- a/tests/test_frame.py
+++ b/tests/test_frame.py
@@ -1,5 +1,5 @@
 # flake8: noqa: F841
-from datetime import date
+import datetime
 import io
 from pathlib import Path
 import tempfile
@@ -943,7 +943,7 @@ def test_types_regressions() -> None:
     ss2: pd.Series = pd.concat([s1, s2])
 
     # https://github.com/microsoft/python-type-stubs/issues/110
-    d: date = pd.Timestamp("2021-01-01")
+    d: datetime.date = pd.Timestamp("2021-01-01")
     tslist: List[pd.Timestamp] = list(pd.to_datetime(["2022-01-01", "2022-01-02"]))
     sseries: pd.Series = pd.Series(tslist)
     sseries_plus1: pd.Series = sseries + pd.Timedelta(1, "d")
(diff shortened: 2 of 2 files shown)


pandas-stubs/core/frame.pyi
```diff
@@ -138,17 +138,17 @@ class _LocIndexerFrame(_LocIndexer):
...
     def __getitem__(
         self,
-        idx: Tuple[Union[IndexType, MaskType], StrLike],
+        idx: Tuple[Union[IndexType, MaskType], Union[ScalarT, None]],
     ) -> Series: ...
     @overload
     def __getitem__(
         self,
-        idx: Tuple[Tuple[slice, ...], StrLike],
+        idx: Tuple[Tuple[slice, ...], Union[ScalarT, None]],
```
Dr-Irv, line 151: Now that I look at this, I think we could combine the 3 overloads here into one with the argument `Union[ScalarT, None, Tuple[Union[IndexType, MaskType], Union[ScalarT, None]], Tuple[Tuple[slice, ...], Union[ScalarT, None]]]`

I think this combination was just missed in the past.

Other issue in using `ScalarT` is whether it might be too wide.  Are we sure that all the types that are in `Scalar` can be used as a column name in a `DataFrame` ? (e.g., `Timestamp`, `Timedelta`, `datetime.date`, etc.)

Dr-Irv (changes requested):
1. I suggested combining the 3 overloads, but if that doesn't work with the current tests, then you can leave it as is.
2. Can you add tests for all the types in `Scalar` to see if they are valid column names?

pandas-stubs/core/frame.pyi
bashtage: I managed to simplify it even more since the two `Tuple` types could be further compressed into a single `Tuple`

bashtage:
All scalar types work as expected.

tests/test_frame.py
```diff
@@ -1180,3 +1180,72 @@ def test_columns_mixlist() -> None:
...
+        timestamp,
+        pd_timedelta,
+        none,
+    ]
+    values = np.arange(len(idx))[:, None] + np.arange(len(idx))
+    df = pd.DataFrame(values, columns=idx, index=idx)
+
+    # Note: bool_ cannot be tested since the index is object and pandas does not
+    # support boolean access using loc except when the index is boolean
+    assert_type(df.loc[str_], pd.Series)
```
Dr-Irv, line 1222: Can you add the `check` function to all of the tests you added, so this test would look like this:
```python
check(assert_type(df.loc[str_], pd.Series), pd.Series)
```

Dr-Irv (changes requested):

tests/test_frame.py
```diff
@@ -1219,33 +1219,34 @@ def test_frame_scalars_slice() -> None:
...
-    assert_type(df.loc[:, complex_], pd.Series)
-    assert_type(df.loc[:, timestamp], pd.Series)
-    assert_type(df.loc[:, pd_timedelta], pd.Series)
-    assert_type(df.loc[:, none], pd.Series)
+    check(assert_type(df.loc[str_], pd.Series), pd.Series)
+    check(assert_type(df.loc[bytes_], pd.Series), pd.Series)
+    check(assert_type(df.loc[date], pd.Series), pd.Series)
+    check(assert_type(df.loc[datetime_], pd.Series), pd.Series)
+    check(assert_type(df.loc[timedelta], pd.Series), pd.Series)
+    print(df.loc[int_])
```
Dr-Irv, line 1227: can you remove the `print`

Dr-Irv (changes requested):
one small change

tests/test_frame.py
bashtage: I added these.  They highlighted that these type defs are fundamentally wrong since scalar (only) selection using loc produces `Union[Series, DataFrame]` always.  Here is an example:

```python

import pandas as pd
df = pd.DataFrame({1:[1,2,3,4],b"1":[1,2,3,4],True:[1,2,3,4],1.0:[1,2,3,4]}, index=[1,b"1",1.0,True])
df.loc[1]
```

returns
```
      1  b'1'
1     1     1
1.0   3     3
True  4     4
```

The equality test must not be exact.  Possibly a bug in pandas.  

This said, I think the pattern here covers the common cast:

Another "gotcha" is 

```python
import numpy as np

df = pd.DataFrame(np.arange(9).reshape((3,-1)))
df.columns = [1,2,1]
df.index=[1,2,1]
df.loc[1]
```

which returns

```
   1  2  1
1  0  1  2
1  6  7  8
```

and 

```python
df[:, 1]
```

which returns

```
   1  1
1  0  2
2  3  5
1  6  8
```

tests/test_frame.py
bashtage: Removed, sorry.

tests/test_frame.py
Dr-Irv: In your first example, the values `[1, 1.0, True]` are equal to 1 in python:
```python
>>> [(1 == a) for a in [1, 1.0, True]]
[True, True, True]
```
So no bug in pandas.

In your second example, the issue here is that if you have duplicate values in an axis (either the index or the columns), then that changes the type of the result.   

The goal of these stubs is to handle the most common use cases (no duplicates!).  We're doing static typing, but when the results are dynamic based on the underlying values of the `DataFrame` object, we take a best guess as to the "usual" result.

Dr-Irv (approved):
Thanks @bashtage

tests/test_frame.py
bashtage: I was under the impression that `DataFrame`s were more dict-like in that 1, 1.0, and True are all distinct keys.

Dr-Irv: That's not true for dicts:
```python
>>> d={1: 10, 1.0: 20, True: 30}
>>> d
{1: 30}
```

bashtage: I thought it is more like `is` when it came to key distinction.  Interesting, and makes things internally consistent then. Thanks.

PR #141, opened 2022-07-14
try setup-python's cache

twoertwein:
setup-python has a built-in cache option that even supports poetry! It doesn't help us to cache poetry.lock, but it will cache the virtual environment that poetry creates.

I changed the CI based on the example provided by setup-python: https://github.com/actions/setup-python#caching-packages-dependencies


twoertwein:
While it often improves performance, it can take a very long time to retrieve the cache (mac+py39 it took over 2 min!).

PR #144, opened 2022-07-16
make -c/--clean_cache a flag

twoertwein:
c/clean_cache is a flag to control whether to clear the mypy/pytest cache but `poe test_src -c` did not work (it expected an argument after c/clean_cache).

After making this keyword of type boolean, no argument is required: `poe test_src -c` works :)


bashtage (approved):
Simple and LGTM.

Dr-Irv (approved):
Thanks @twoertwein

PR #146, opened 2022-07-17
Fix Series.cat accessors

### LINKED ISSUE #139: [Bug] `rename_categories` is no member of `CategoricalAccessor`
**Describe the bug**
`CategoricalAccessor.rename_categories()` is unknown.

**To Reproduce**
```py
# pyright: strict, reportUnknownMemberType=false
"""Example."""
import pandas as pd

pd.Series().astype('category').cat.rename_categories({})  # Cannot access member "rename_categories" for type "CategoricalAccessor" Member "rename_categories" is unknown
```
grafik

 - OS:  Linux
 - OS Version 20.04
 - python version: 3,8,10
 - version of type checker: pylance v2022.7.30 (Not in v2022.7.20 as `.cat` is Unknown there)
 - version of installed `pandas-stubs`: pylance v2022.7.30 bundeld ones.

**Additional context**

Dr-Irv:
- [x] Closes #139
  - Added additional tests to `test_series:test_cat_accessor()`


tests/test_series.py
```diff
@@ -732,6 +732,19 @@ def test_cat_accessor() -> None:
     # GH 43
     s = pd.Series(pd.Categorical(["a", "b", "a"], categories=["a", "b"]))
     check(assert_type(s.cat.codes, "pd.Series[int]"), pd.Series, int)
+    # GH 139
+    ser = pd.Series([1, 2, 3], name="A").astype("category")
+    check(assert_type(ser.cat.set_categories([1, 2, 3]), pd.Series), pd.Series)
```
twoertwein, line 737: Not familiar with how generic Series is. Do you still have access to the type of the elements here?

Dr-Irv: > Not familiar with how generic Series is. Do you still have access to the type of the elements here?

Yes, it will be the underlying type of the categories.  We can't do `Series[category]`, because categories could be strings, ints, etc.

Nevertheless, I changed things to test the type of the value inside the `Series` in the new commit.

twoertwein: Okay! In general, I would be fine with only adding runtime checks to ensure the type annotations. In this case, I wouldn't mind removing  the checks for `int` as `int` isn't part of the type annotation.

twoertwein:
Thanks @Dr-Irv !

PR #147, opened 2022-07-17
fix args to DataFrame.set_index

### LINKED ISSUE #140: [Bug] `DataFrame.set_index` does not accept `pd.Index` anymore.
**Describe the bug**
`DataFrame.set_index` does not accept `pd.Index` anymore.

**To Reproduce**
```py
# pyright: strict, reportUnknownMemberType=false
"""Example."""
import pandas as pd

df = pd.DataFrame()
expected = df.set_index(pd.Index([])) # Argument of type "Index" cannot be assigned to parameter "keys" of type "Label | Sequence[Unknown]" in function "set_index"
```
grafik

 - OS: Linux
 - OS Version: 20.04
 - python version: 3.8.10
 - version of type checker: pylance v2022.7.31 (not in v2022.7.30)
 - version of installed `pandas-stubs`: i belive `_stub_version = Literal["1.4.3.220710"]` (pylane bundled)

**Additional context**

Dr-Irv:
- [x] Closes #140
- [x] Tests added: 
  - `test_frame.py:test_types_set_index()` from original issue - added one test

### DIFF
--- a/pandas-stubs/core/frame.pyi
+++ b/pandas-stubs/core/frame.pyi
@@ -577,7 +577,9 @@ class DataFrame(NDFrame, OpsMixin):
     @overload
     def set_index(
         self,
-        keys: Union[Label, Sequence],
+        keys: Union[
+            Label, Series, Index, np.ndarray, Iterator[Hashable], List[Hashable]
+        ],
         drop: _bool = ...,
         append: _bool = ...,
         verify_integrity: _bool = ...,
@@ -587,7 +589,9 @@ class DataFrame(NDFrame, OpsMixin):
     @overload
     def set_index(
         self,
-        keys: Union[Label, Sequence],
+        keys: Union[
+            Label, Series, Index, np.ndarray, Iterator[Hashable], List[Hashable]
+        ],
         drop: _bool = ...,
         append: _bool = ...,
         verify_integrity: _bool = ...,
@@ -597,7 +601,9 @@ class DataFrame(NDFrame, OpsMixin):
     @overload
     def set_index(
         self,
-        keys: Union[Label, Sequence],
+        keys: Union[
+            Label, Series, Index, np.ndarray, Iterator[Hashable], List[Hashable]
+        ],
         drop: _bool = ...,
         append: _bool = ...,
         *,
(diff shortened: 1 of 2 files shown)


pandas-stubs/core/frame.pyi
```diff
@@ -577,7 +577,9 @@ class DataFrame(NDFrame, OpsMixin):
     @overload
     def set_index(
         self,
-        keys: Union[Label, Sequence],
+        keys: Union[
+            Label, Series, Index, np.ndarray, Iterator[Hashable], List[Hashable]
```
twoertwein, line 581: I think `Iterator[Hashable]` includes `List[Hashable]`, but I didn't test that.

Dr-Irv: > I think `Iterator[Hashable]` includes `List[Hashable]`, but I didn't test that.

No, it's not.  Here's an example where `pyright` shows the incompatibility:
```python
from typing import Iterator


def fun(a: Iterator[int]):
    k = next(a, None)
    while k is not None:
        print(k)


li = [1, 2, 3]
fun(li)
```
On the last line, `pyright` reports:
```text
iterlist.py:11:5 - error: Argument of type "list[int]" cannot be assigned to parameter "a" of type "Iterator[int]" in function "fun"
    "list[int]" is incompatible with protocol "Iterator[int]"
      "__next__" is not present (reportGeneralTypeIssues)
```

twoertwein: `Iterable[int]` wouldn't require `__next__` but I'm not sure whether `set_index` works with any iterable.

twoertwein: > `Iterable[int]` wouldn't require `__next__` but I'm not sure whether `set_index` works with any iterable.

pandas has no type annotations for `keys`. Let's keep it as-is.

twoertwein:
Thanks @Dr-Irv !

pandas-stubs/core/frame.pyi
bashtage: If you believe the docs, then the types that could be used are

```
Label
Series
Index
np.ndarray
Iterator
List[Union[Label, Series, Index, np.ndarray, Iterator]]
```

Clearly it also supports `Sequence` everywhere `Series` is used.

PR #148, opened 2022-07-18
groupby.__iter__() fix types

### LINKED ISSUE #142: Unpacking group keys is refused in multi-columns groups
**Describe the bug**
Unpacking group keys is refused in multi-columns groups, group key is wrongly assumed to be a str

**To Reproduce**
```
import pandas as pd
df = pd.DataFrame({"a": [0, 1, 2], "b": [4, 5, 6], "c": [7, 8, 9]})
for (k1, k2), g in df.groupby(["a", "b"]):
    print(k1, k2)
```

Script output
```
0 4
1 5
2 6
```

mypy output
```
foo.py:5: error: Unpacking a string is disallowed
foo.py:6: error: Cannot determine type of "k1"
foo.py:6: error: Cannot determine type of "k2"
Found 3 errors in 1 file (checked 1 source file)
```

 - OS: Ubuntu 22.04 
 - python=3.9.13
 - pandas=1.4.3
 - pandas-stubs=1.4.3.220710
 - mypy=0.961

Dr-Irv:
- [x] Closes #142
- [x] Tests added: 
  - `test_frame.py:test_groupby_result()`

The `groupby` internal class hierarchies are not what is currently in pandas.  They were changed in v1.2, but the type stubs were initially created on top of pandas 1.1.   Will create an issue about that.


twoertwein:
> The `groupby` internal class hierarchies are not what is currently in pandas. They were changed in v1.2, but the type stubs were initially created on top of pandas 1.1. Will create an issue about that.

stubtest will be really useful to (slowly) align the type stubs with the implementation structure.

tests/test_frame.py
```diff
@@ -1251,3 +1252,13 @@ def test_boolean_loc() -> None:
     df = pd.DataFrame([[0, 1], [1, 0]], columns=[True, False], index=[True, False])
     check(assert_type(df.loc[True], pd.Series), pd.Series)
     check(assert_type(df.loc[:, False], pd.Series), pd.Series)
+
+
+def test_groupby_result() -> None:
+    # GH 142
+    df = pd.DataFrame({"a": [0, 1, 2], "b": [4, 5, 6], "c": [7, 8, 9]})
+    lresult = [(cast(Tuple[int, int], k), g) for k, g in df.groupby(["a", "b"])]
```
twoertwein, line 1260: I would be very cautious adding `cast`s to the type tests, if we add them we might in end up testing (only) the runtime API of pandas
```py
index, value = assert_type(df.groupby(["a", "b"]), tuple[Hashable, DataFrame]
assert isinstance(index.__hash__(), int)  # if possible isinstance(index, Hashable) 
assert isinstance(value, DataFrame)
```

Dr-Irv: I did something like this in the next commit.  `cast` has been removed.  It's a bit tricky to test `__iter__()`

Dr-Irv: Changed.

Also did a scheme to address the original issue.

tests/test_frame.py
```diff
@@ -1251,3 +1253,26 @@ def test_boolean_loc() -> None:
     df = pd.DataFrame([[0, 1], [1, 0]], columns=[True, False], index=[True, False])
     check(assert_type(df.loc[True], pd.Series), pd.Series)
     check(assert_type(df.loc[:, False], pd.Series), pd.Series)
+
+
+def test_groupby_result() -> None:
+    # GH 142
+    df = pd.DataFrame({"a": [0, 1, 2], "b": [4, 5, 6], "c": [7, 8, 9]})
+    index, value = next(df.groupby(["a", "b"]).__iter__())
```
twoertwein, line 1261: One small change: a two-element list can also be unpacked into two variables.

I tested the following small change on your branch:

```diff
     Iterable,
+    Iterator,
     List,

...
-    index, value = next(df.groupby(["a", "b"]).__iter__())
+    iterator = df.groupby(["a", "b"]).__iter__()
+    assert_type(iterator, Iterator[Tuple[Tuple, pd.DataFrame]])
+    index, value = next(iterator)
```

pandas-stubs/core/groupby/generic.pyi
```diff
@@ -46,6 +48,12 @@ def pin_whitelisted_properties(
     klass: Type[FrameOrSeries], whitelist: FrozenSet[str]
 ): ...
 
+class SeriesGroupByScalar(SeriesGroupBy):
+    def __iter__(self) -> Iterator[Tuple[Scalar, Series]]: ...
+
+class SeriesGroupByNonScalar(SeriesGroupBy):
+    def __iter__(self) -> Iterator[Tuple[Tuple, Series]]: ...
```
twoertwein, line 55: No strong opinion: I see the point from a user-perspective, but I think deviating too much from the pandas implementation will make it challenging to use stubtest later (and maybe also maintaining the stubs).

twoertwein: should probably make these classes that are only in the stubs but not in pandas private

twoertwein: > No strong opinion

I think I develop a slightly stronger opinion. I would prefer not to introduce too many non-pandas classes: I believe that this will keep pandas-stubs more maintainable in the long-run.

Personally, I prefer the previous state of the PR where you returned `tuple[Hashable, NDFrameT]` (or making SeriesGroupBy generic, might be more invasive).

tests/test_frame.py
Dr-Irv: > One small change: a two-element list can also be unpacked into two variables.
> 
> I tested the following small change on your branch:
> 
> ```diff
>      Iterable,
> +    Iterator,
>      List,
> 
> ...
> -    index, value = next(df.groupby(["a", "b"]).__iter__())
> +    iterator = df.groupby(["a", "b"]).__iter__()
> +    assert_type(iterator, Iterator[Tuple[Tuple, pd.DataFrame]])
> +    index, value = next(iterator)
> ```

OK, that is in the next commit.

pandas-stubs/core/groupby/generic.pyi
Dr-Irv: > should probably make these classes that are only in the stubs but not in pandas private

Did that in the next commit

twoertwein:
Thanks @Dr-Irv !

pandas-stubs/core/groupby/generic.pyi
twoertwein: While I prefer pandas and pandas-stubs to be aligned (I would like if pandas and pandas-stub could in the future converge and only be different in challenging cases: `__getitem__, Timestamp, ...` - might be a long list) I'm fine with introducing classes/variables that are not in pandas as long as they are private (or some other prefix).

PR #151, opened 2022-07-18
TYP: future annotations

### LINKED ISSUE #150: Use PEP 604 style annotations
We might consider using the new(er) type annotation system (replace `Union` with `|`, use `list` and so on):

- Pandas uses them universally since https://github.com/pandas-dev/pandas/pull/47769
- and based on [flake8-pyi](https://github.com/PyCQA/flake8-pyi) those annotations are supported in almost all cases (as far as I understand we shouldn't be using them in `pandas/_typing.pyi` where we define most/all the type aliases):

> Y037 (enforcing PEP 604 syntax everywhere) is not yet fully compatible with the mypy type checker, which has [a few bugs](https://github.com/python/mypy/issues?q=is%3Aopen+is%3Aissue+label%3Atopic-pep-604+label%3Atopic-type-alias) regarding PEP 604 type aliases.

twoertwein:
- [x] Closes #150


twoertwein:
I will try to separate it in two commits: automatic (sed+pre-commit) and then manual changes to make it easier to review

twoertwein:
- first commit: `sed -i '1s;^;from __future__ import annotations\n;'` + `pre-commit` + `isort` + `black`
- second commit: removing unused imports and gettign mypy/pyright to pass
- thrid commit: manually chaning offening (flake8-pyi Y022) Union/List/Dict/Tuple/Set cases
- fourth commit: remove unused flake8 ignore comments

pyi files do not need the future import but without it pre-commit while not re-write the annotations.

pandas-stubs/core/series.pyi
```diff
@@ -994,102 +975,98 @@ class Series(IndexOpsMixin, NDFrame, Generic[S1]):
...
         **kwargs,
     ) -> Series[S1]: ...
     def asfreq(
         self,
         freq,
-        method: Optional[
-            Union[_str, Literal["backfill", "bfill", "pad", "ffill"]]
-        ] = ...,
-        how: Optional[Union[_str, Literal["start", "end"]]] = ...,
+        method: None | (_str | Literal["backfill", "bfill", "pad", "ffill"]) = ...,
```
Dr-Irv, line 1005: put `None` at end, remove parens

pandas-stubs/io/json/_json.pyi
```diff
@@ -72,75 +72,69 @@ class JSONTableWriter(FrameWriter):
...
-    encoding: Optional[str] = ...,
+    date_unit: str | None = ...,
+    encoding: str | None = ...,
     lines: bool = ...,
-    chunksize: Optional[int] = ...,
-    compression: Optional[
-        Union[str, Literal["infer", "gzip", "bz2", "zip", "xz"]]
-    ] = ...,
+    chunksize: int | None = ...,
+    compression: None | (str | Literal["infer", "gzip", "bz2", "zip", "xz"]) = ...,
```
Dr-Irv, line 99: put `None` at end, remove parens

pandas-stubs/core/arrays/string_.pyi
```diff
@@ -1,18 +1,18 @@
...
+_type_StringArray = type[StringArray]
+
 class StringDtype(ExtensionDtype):
     name: str = ...
     na_value = ...
     @property
     def type(self) -> type: ...
     @classmethod
-    def construct_array_type(cls) -> type[StringArray]: ...
+    def construct_array_type(cls) -> _type_StringArray: ...
```
bashtage, line 15: Is this one a bug or something?

pandas-stubs/io/json/_json.pyi
```diff
@@ -72,75 +72,69 @@ class JSONTableWriter(FrameWriter):
...
```
Dr-Irv, line 117: put `None` at end, remove parens

pandas-stubs/core/arrays/boolean.pyi
```diff
@@ -19,7 +19,7 @@ class BooleanDtype(ExtensionDtype):
     @property
     def kind(self) -> str: ...
     @classmethod
-    def construct_array_type(cls) -> type[BooleanArray]: ...
+    def construct_array_type(cls) -> _type_BooleanArray: ...
```
bashtage, line 22: Same

bashtage (approved):
2, 3, and 4 all look good.  1 is harder to check, but probably can be trusted.

pandas-stubs/core/arrays/integer.pyi
```diff
@@ -7,7 +7,7 @@ from .masked import BaseMaskedArray as BaseMaskedArray
 class _IntegerDtype(ExtensionDtype):
     name: str
     base = ...
-    type: Type
+    type: type
```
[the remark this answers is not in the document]
twoertwein: it didn't trigger a mypy error, but I didn't check with `reveal_type` whether `type` has the correct annotation. I will look into that and your other comments later today.

pandas-stubs/core/arrays/string_.pyi
Dr-Irv: > That is a mypy bug (pyright doesn't need this workaround): when a class has a method called str/bool/type, mypy gets confused when we use these annotations inside the class.

If we switch to `type_t` as I suggested, it will make this issue go away.  Then you could do `type_t[StringArray]`

twoertwein: `type_t[StringArray]` does not work for mypy (even when explicitly delcared as a `typing_extensions.TypeAlias`)

> error: "type" expects no type arguments, but 1 given  [type-arg]

twoertwein: (should maybe retry after installing mypy 0.971)

twoertwein: I can try that again, I was using `type_t  = type`

twoertwein: importing typing.Type isn't compatible with Y022:

> Y022 Use "type[MyClass]" instead of "typing.Type[MyClass]" (PEP 585 syntax)

but more importantly, even when `implicit_reexport = true`

> pandas-stubs/core/dtypes/base.pyi:5: error: Module "pandas._typing" has no attribute "type_t"  [attr-defined]
> pandas-stubs/core/arrays/string_.pyi:5: error: Module "pandas._typing" has no attribute "type_t"  [attr-defined]
> pandas-stubs/core/arrays/integer.pyi:3: error: Module "pandas._typing" has no attribute "type_t"  [attr-defined]
> pandas-stubs/core/arrays/boolean.pyi:5: error: Module "pandas._typing" has no attribute "type_t"  [attr-defined]

I honestly don't understand that (still pushed it).

twoertwein:
But neither Dict nor List is at those lines? (locally `poe test_all` works for me - on 3.10)
> pandas-stubs/core/frame.pyi:194: error: Name "Dict" is not defined  [name-defined]
> pandas-stubs/core/frame.pyi:194: note: Did you forget to import it from "typing"? (Suggestion: "from typing import Dict")
> pandas-stubs/core/frame.pyi:197: error: Name "List" is not defined  [name-defined]
> pandas-stubs/core/frame.pyi:197: note: Did you forget to import it from "typing"? (Suggestion: "from typing import List")

Dr-Irv:
You can retrigger the CI from github.  I had a failure due to network connectivity, and just went it and told it to rerun the jobs.

twoertwein:
> You can retrigger the CI from github. I had a failure due to network connectivity, and just went it and told it to rerun the jobs.

Thanks! It seems that github requires that the branch has no conflict with main (that might also explain why Dict and List appeared out of nowhere):
https://github.community/t/run-actions-on-pull-requests-with-merge-conflicts/17104/26

I resolved all conflicts.

Dr-Irv:
Let me know when I should look at this again.

pandas-stubs/core/arrays/string_.pyi
twoertwein: > Let me know when I should look at this again.

The CI seems to be working now and I believe I addressed all comments except for `type` (I use `_type` for now).

I'm happy to address the _type in a follow up (I might open a pandas PR to change Type -> type there).

twoertwein: I would like that but that doesn't work for mypy. I'll open an issue at mypy (we cannot `_type[something]` if we import it from another file)

twoertwein: when you think it is ready to merge, let me know and I will do a final rebase. There might be non-trivial conflicts (PRs adding `Dict` to an annotation but this PR removes the `Dict` import).

Dr-Irv: So I should review now?

twoertwein: (the first four PRs did not change)

twoertwein: there is already an issue about importing Type or type (as TypeAliases) from another file: https://github.com/python/mypy/issues/10068

The reason why `from typing import Type as type_t` works in pandas is that it isn't a type alias. Using this in pandas-stubs does for some reason not work as mypy doesn't seem to find `type_t` when it is being imported from `_typing` (even when implicit reexports are enabled). There seems to be some difference in how mypy treats pyi and py files.

Dr-Irv: Based on that `mypy` issue, this seems to work in `_typing.pyi` and we can use `type_t` to correspond to pandas:
```python
from builtins import type as type_t

__all__ = ["type_t"]
```
flake8 and mypy then seem to be happy.

Dr-Irv (changes requested):
So I think there is a better workaround for the `type` issue, where we define `type_t` once, export it from `_typing.pyi` and then import `type_t` instead of always creating `_type`

Otherwise, a few small things.

pandas-stubs/core/arrays/string_.pyi
twoertwein: nice!

pandas-stubs/_typing.pyi
```diff
@@ -208,3 +209,5 @@ GroupByObjectNonScalar = Union[
     tuple, list[Label], Function, Series, np.ndarray, Mapping[Label, Any], Index
 ]
 GroupByObject = Union[Scalar, GroupByObjectNonScalar]
+
+__all__ = ["npt", "type_t"]
```
Dr-Irv, line 213: can we go back to ` import numpy.typing as npt` ?

Dr-Irv (approved):
Thanks @twoertwein .  I now have to get used to not using `Union`, `Dict`, `List`, etc. !

PR #152, opened 2022-07-19
Add Tuple[slice] in iLocIndexerFrame. Fix #143

### LINKED ISSUE #143: [BUG] Argument of type "tuple[slice]" cannot be assigned to parameter "idx"
**Describe the bug**
False positive error warning when using `iloc` function.

**To Reproduce**
Using pyright on
```
import pandas as pd

df = pd.DataFrame({'Char': ['A', 'B', 'C'], 'Number': [1, 2, 3]})
df = df.iloc[0:2,]
```
produces
```
Argument of type "tuple[slice]" cannot be assigned to parameter "idx" of type "IndexType | MaskType | Tuple[IndexType | MaskType, IndexType | MaskType]" in function "__getitem__"    Type "tuple[slice]" cannot be assigned to type "IndexType | MaskType | Tuple[IndexType | MaskType, IndexType | MaskType]"      "tuple[slice]" is incompatible
```

 - OS: ArchLinux
 - python 3.10.5
 - pyright 1.1.260
 - pandas-stub 1.4.3.220704

AchmadFathoni:
- [x] Closes #143


bashtage:
Is that an actual supported mode for iloc? Or just an accident that works?

https://pandas.pydata.org/docs/reference/api/pandas.DataFrame.iloc.html

AchmadFathoni:
I don't know, I'm out of deep in this topic. I just following @Dr-Irv suggestion.

Dr-Irv:
> Is that an actual supported mode for iloc? Or just an accident that works because tuples are unpacked in `__s/getitem__`?
> 
> https://pandas.pydata.org/docs/reference/api/pandas.DataFrame.iloc.html

The example in the issue is `df.iloc[0:2,]`.  Here the arguments to `__getitem__()` is `tuple(slice,)`.  Agreed it's an odd usage, but valid.

Dr-Irv (changes requested):
Can you also add a test in `test_frame.py` that fails without this change, and works with it?  Copying the code from the original issue will suffice.  Make sure to include a comment that refers to the issue.

bashtage:
> The example in the issue is `df.iloc[0:2,]`. Here the arguments to `__getitem__()` is `tuple(slice,)`. Agreed it's an odd usage, but valid.

I agree it works.  I am just querying if pandas-stubs should be adding types that aren't part of the docstring. Shouldn't there be a change upstream to confirm that this is a supported format?

Dr-Irv (changes requested):
CI failing due to `black`.  Do `poe style` to make sure everything is OK.

pandas-stubs/core/frame.pyi
```diff
@@ -89,7 +89,10 @@ class _iLocIndexerFrame(_iLocIndexer):
     @overload
     def __getitem__(
         self,
-        idx: IndexType | MaskType | tuple[IndexType | MaskType, IndexType | MaskType] | tuple[slice],
+        idx: IndexType
+        | MaskType
+        | tuple[IndexType | MaskType, IndexType | MaskType]
+        | tuple[slice],
```
Dr-Irv, line 95: Sorry - I should have picked up on this before.  Use 
```python
tuple[IndexType | MaskType, IndexType | MaskType | slice]
```

Dr-Irv (changes requested):

pandas-stubs/core/frame.pyi
AchmadFathoni: That change fails the test I made
```
Type "tuple[slice]" cannot be assigned to type "IndexType | MaskType | tuple[IndexType | MaskType, IndexType | MaskType]"
```

Dr-Irv: My mistake.  I missed the comma

Dr-Irv (approved):
Thanks @AchmadFathoni

PR #154, opened 2022-07-20
BUG: Use HashableT rather than Hashable to improve compat

### LINKED ISSUE #153: BUG: set_index rejects `List[str]`
**Describe the bug**

**To Reproduce**
```python
import pandas as pd

df = pd.DataFrame({"a":[1,2,3]})
_l: list[str] = ["a","b","c"]
df.set_index(_l)
```

Error with mypy 0.971 is

```
exp.py:5: error: Argument 1 to "set_index" of "DataFrame" has incompatible type "List[str]"; expected "Union[Optional[Hashable], Series[Any], Index, ndarray[Any, Any], Iterator[Hashable], List[Hashable]]"
exp.py:5: note: "List" is invariant -- see https://mypy.readthedocs.io/en/stable/common_issues.html#variance
exp.py:5: note: Consider using "Sequence" instead, which is covariant
exp.py:5: note: "list" is missing following "Iterator" protocol member:
exp.py:5: note:     __next__
```

 - OS: Windows
 - OS Version 11
 - python version: 3.10
 - version of type checker: mypy
 - version of installed `pandas-stubs`:   1.4.3.220718

**Additional context**
Looks like a regression in #147

bashtage:
Use HashableT so that any hashable object can be use in a list or iterator

- [X] Closes #153(Replace xxxx with the Github issue number)

### DIFF
--- a/pandas-stubs/core/frame.pyi
+++ b/pandas-stubs/core/frame.pyi
@@ -58,6 +58,7 @@ from pandas._typing import (
     FilePathOrBuffer,
     FilePathOrBytesBuffer,
     GroupByObjectNonScalar,
+    HashableT,
     IgnoreRaise,
     IndexingInt,
     IndexLabel,
@@ -587,7 +588,7 @@ class DataFrame(NDFrame, OpsMixin):
     def set_index(
         self,
         keys: Union[
-            Label, Series, Index, np.ndarray, Iterator[Hashable], List[Hashable]
+            Label, Series, Index, np.ndarray, Iterator[HashableT], List[HashableT]
         ],
         drop: _bool = ...,
         append: _bool = ...,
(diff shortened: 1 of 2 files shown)


Dr-Irv (approved):
Thannks @bashtage

pandas-stubs/core/frame.pyi
```diff
@@ -611,7 +612,7 @@ class DataFrame(NDFrame, OpsMixin):
     def set_index(
         self,
         keys: Union[
-            Label, Series, Index, np.ndarray, Iterator[Hashable], List[Hashable]
+            Label, Series, Index, np.ndarray, Iterator[HashableT], List[HashableT]
```
twoertwein, line 615: I think the Iterator wouldn't have needed the TypeVar:

```py
from typing import Iterator, Hashable

def test(x: Iterator[Hashable]):
    ...

def iter() -> Iterator[str]:
    ...

test(iter())
```

PR #156, opened 2022-07-20
ENH: Hashable -> hashablet

bashtage:
* Improves the use of list[Hashable] using list[HashableT]
* Replaces `Hashable | List[Hashable]` with `Hashable | Sequence[Hashable]` since the issue with strings isn't an issue here
* Fixes the npt import issue


pandas-stubs/_typing.pyi
```diff
@@ -42,6 +42,8 @@ from pandas._libs.tslibs import (
 
 from pandas.core.dtypes.dtypes import ExtensionDtype
 
+npt = numpy.typing
```
Dr-Irv, line 45: why make this change.  We're following the convention in pandas of doing `from numpy import typing as npt`

Dr-Irv (changes requested):
I really think we should stop using `Sequence`, because if you have `Sequence[str]`, `Sequence[Hashable]` or `Sequence[HashableT]`, as a parameter, then a string by itself will match.
```python
from ast import FunctionDef
from typing import Sequence, TypeVar, Hashable

HashableT = TypeVar("HashableT", bound="Hashable")


def fun1(a: Sequence[str]):
    pass


def fun2(b: Sequence[Hashable]):
    pass


def fun3(c: Sequence[HashableT]):
    pass


fun1("abc")

fun2("def")

fun3("ghi")
```
The above code passes `mypy` and `pyright`, but we really don't want it to if our intent is that `Sequence[Hashable]` does not match a single string.

Maybe make this change, but change `Sequence` to `List`

pandas-stubs/_typing.pyi
twoertwein: that is needed for mypy 0.971. npt needs to be re-exported

twoertwein:
> I really think we should stop using `Sequence`, because if you have `Sequence[str]`, `Sequence[Hashable]` or `Sequence[HashableT]`, as a parameter, then a string by itself will match.

In the following cases, str is already accepted

> Replaces Hashable | List[Hashable] with Hashable | Sequence[Hashable] since the issue with strings isn't an issue here

pandas-stubs/core/generic.pyi
```diff
@@ -250,10 +251,10 @@ class NDFrame(PandasObject, indexing.IndexingMixin):
         sep: _str = ...,
         na_rep: _str = ...,
         float_format: Optional[_str] = ...,
-        columns: Optional[Sequence[Hashable]] = ...,
+        columns: Optional[Sequence[HashableT]] = ...,
```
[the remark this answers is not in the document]
bashtage: Fixed.

bashtage:
I'm not introducing and new `Sequence` except in the case I noted above where `Hashable | List[HashableT]` becomes `Hashable | Sequence[Hashable]` which is more general while not encountering the issue that `str` is `Sequence[Hashable]` since `str` is `Hashable`.

pandas-stubs/_typing.pyi
bashtage: Let me revert and see if it passes CI.

PR #158, opened 2022-07-20
remove cache options

twoertwein:
mypy is already called with `--no-incremental` which disables the cache. I simply added the corresponding flag for pytest, to remove this flag.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -145,7 +145,7 @@ enable_error_code = "ignore-without-code" # same as in pandas
 allow_untyped_globals = false
 allow_redefinition = false
 local_partial_types = false
-implicit_reexport = false # pyright behaves the same
+implicit_reexport = true
 strict_equality = true
 # Configuring error messages
 show_error_context = false
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -35,7 +35,7 @@ python = ">=3.8,<3.11"
 types-pytz = ">= 2022.1.1"
 
 [tool.poetry.dev-dependencies]
-mypy = ">=0.961"
+mypy = ">=0.971"
 pytest = ">=7.1.2"
 pyright = ">=1.1.255"
 poethepoet = ">=0.13.1"
@@ -145,7 +145,7 @@ enable_error_code = "ignore-without-code" # same as in pandas
 allow_untyped_globals = false
 allow_redefinition = false
 local_partial_types = false
-implicit_reexport = true
+implicit_reexport = false # pyright behaves the same
 strict_equality = true
 # Configuring error messages
 show_error_context = false
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -53,36 +53,15 @@ build-backend = "poetry.core.masonry.api"
 
 [tool.poe.tasks.test_all]
 help = "Run all tests"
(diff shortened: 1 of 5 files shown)


Dr-Irv:
> mypy is already called with `--no-incremental` which disables the cache. I simply added the corresponding flag for pytest, to remove this flag.

I don't think it is called with `--no-incremental` when you run locally.  So I think this option should be kept.

twoertwein:
`mypy --help` says:

> --no-incremental          Disable module cache (inverse: --incremental)

So I assume that is effectively the same as deleting the cache (but it it less code on our side to achieve it)

Dr-Irv:
> `mypy --help` says:
> 
> > --no-incremental          Disable module cache (inverse: --incremental)
> 
> So I assume that is effectively the same as deleting the cache (but it it less code on our side to achieve it)

Yes, but I don't see that we are running with `--no-incremental` by default??

Dr-Irv:
> That seems to be the case on main:
> 
> https://github.com/pandas-dev/pandas-stubs/blob/9072b1596e5364268648899a7adaf021a8d7e8c2/scripts/test/run.py#L7

Ah, good.  So you're right that we don't need the option!

pyproject.toml
```diff
@@ -166,7 +145,7 @@ enable_error_code = "ignore-without-code" # same as in pandas
 allow_untyped_globals = false
 allow_redefinition = false
 local_partial_types = false
-implicit_reexport = false # pyright behaves the same
+implicit_reexport = true
```
twoertwein, line 148: Could re-enable this option again as mypy 0.971 seems to need `npt = numpy.typing` in either case

Dr-Irv (commented):
Given that you had to change the `npt` reference because of `mypy` 0.971, can you bump the version in `pyproject.toml` ?

scripts/test/run.py
```diff
@@ -14,7 +13,7 @@ def pyright_src():
 
 
 def pytest():
-    cmd = ["pytest"]
+    cmd = ["pytest", "--cache-clear"]
```
Dr-Irv, line 16: double checking that not clearing the cache will always work.

scripts/test/run.py
```diff
@@ -61,13 +60,3 @@ def restore_src():
...
-
-
-def clean_mypy_cache():
-    if Path(".mypy_cache").exists():
-        shutil.rmtree(".mypy_cache")
-
-
-def clean_pytest_cache():
-    if Path(".mypy_cache").exists():
-        shutil.rmtree(".pytest_cache")
```
Dr-Irv, line 73: the test was wrong here, so the `-c` never worked!

twoertwein: hahaha :) I didn't notice that when I removed it :)

scripts/test/run.py
twoertwein: Locally, I do not need to clear the cache. Happy to remove it. The CI will anyways have a clean plate every time it runs.

Dr-Irv: So I see.  Now your proposal is to *always* clear the cache.  I think that's a good idea.  Equivalent to what will happen with `mypy` where we *always* use `--no-incremental`

Dr-Irv (approved):
Thanks @twoertwein

PR #159, opened 2022-07-20
ENH: Improve io/stata.py

bashtage:
Pulls over typing information from pandas including argumnts added since 1.1


Dr-Irv (approved):
While I usually prefer having tests, this could be hard to test, so I'm going to approve it.

Thanks @bashtage

PR #162, opened 2022-07-22
update docs

twoertwein:
updated tests.md, style.md is already covered in tests.md, and removed html formatting tags.


docs/tests.md
```diff
@@ -1,16 +1,15 @@
...
-    - Pytest: Runs only pytest </br> `poe test_src --profile=pytest` </br> 
-    - Pre-commit: Runs all style checks in pre-commit </br> `poe test_src --profile=style` </br> 
-    - Full: Run all the above tests </br> `poe test_src --profile=full` </br> </br>
+[Poe](https://github.com/nat-n/poethepoet) is used to run all tests.
 
-- Run local tests against distribution: </br> `poe test_dist` </br> </br>
+Here are the most important options. Fore more details, please use `poe --help`.
 
-- Run all local tests: </br> `poe test_all` </br> </br>
+- Run all tests (includes installing the stubs): `poe test_all`
```
Dr-Irv, line 7: I think this should say "includes testing the stubs when installed"

Dr-Irv (changes requested):
small change suggested

docs/tests.md
twoertwein: `test_all` installes the stubs, then tests them, and then uninstalled them

Dr-Irv: Sure, but the remark seems to indicate that it is _just_ installing the stubs.  So that's why I suggested the change.

How about "includes testing stubs as installed"

twoertwein: I merged main and slightly modified the existing description

Dr-Irv (approved):
Thanks @twoertwein

PR #163, opened 2022-07-22
ENH: Add default mypy and pyright tasks

bashtage:
Add simpler task names to simplify running tools

### DIFF
--- a/docs/tests.md
+++ b/docs/tests.md
@@ -1,15 +1,14 @@
 ## Test
 
-- Run local tests against source code. </br> `poe test` </br> </br> 
+- Run all local tests against source code. </br> `poe test` </br> </br> 
   - Profiles: </br>
     - Default: Runs only mypy and pyright tests </br> `poe test --profile=default` </br>
     - Pytest: Runs only pytest </br> `poe test --profile=pytest` </br> 
     - Pre-commit: Runs all style checks in pre-commit </br> `poe test --profile=style` </br> 
-    - Full: Run all the above tests </br> `poe test --profile=full` </br> </br>
 
-- Run local tests against distribution: </br> `poe test_dist` </br> </br>
+- Run local tests against an installed distribution: </br> `poe test_dist` </br> </br>
 
-- Run all local tests: </br> `poe test_all` </br> </br>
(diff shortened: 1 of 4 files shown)


bashtage:
@twoertwein Not sure how you feel about these, but I like the idea of just running `poe mypy` like I run `poe pytest`.  This only runs one of the mypy tasks, but that it usually enough to find any errors IME.

twoertwein:
> @twoertwein Not sure how you feel about these, but I like the idea of just running `poe mypy` like I run `poe pytest`. This only runs one of the mypy tasks, but that it usually enough to find any errors IME.

Cannot you already do that with `poe pyright_src` and `poe mypy_src` (or did I accidentally remove that)?

twoertwein:
Happy to shorten `mypy_src` to `mypy` and `pyright_src` to `pyright` - but we shoud not have both `mypy` and `mypy_src`

bashtage:
> Happy to shorten `mypy_src` to `mypy` and `pyright_src` to `pyright` - but we shoud not have both `mypy` and `mypy_src`

That is my idea.  To have a "default" mypy which would just be called mypy.  Same for pyright.  I'm used to using `mypy <project>`, so something like `poe mypy` feels intuitive (like `poe pytest` since `pytest <project>` is more common).

bashtage:
I have removed the three comands that end in `src` so that their "base"s are now the preferred names

```
poe mypy
poe pyright
poe test
```

twoertwein (approved):
Thank you, looks good to me!

@Dr-Irv

docs/tests.md
```diff
@@ -1,11 +1,11 @@
 ## Test
 
-- Run local tests against source code. </br> `poe test_src` </br> </br> 
+- Run local tests against source code. </br> `poe test` </br> </br>
```
Dr-Irv, line 3: We should say which is the default profile

Dr-Irv (changes requested):
one small change in the docs

docs/tests.md
twoertwein: There are no more profiles, that's why I opened #162

Dr-Irv (approved):
Thanks @bashtage

PR #165, opened 2022-07-23
CLEAN: remove pandas/core/internals

Dr-Irv:
Remove references to `pandas/core/internals`


pandas-stubs/core/generic.pyi
```diff
@@ -38,15 +37,6 @@ _str = str
 class NDFrame(PandasObject, indexing.IndexingMixin):
     __hash__: ClassVar[None]  # type: ignore[assignment]
 
-    def __new__(
-        cls,
-        data: BlockManager,
```
twoertwein, line 43: I don't know whether `NDFrame.__new__` is public/private. If it is public, it might be best to keep it but simply drop the annotation for `data`.

Dr-Irv: `NDFrame` is not public.  But it is needed because it defines methods common to `Series` and `DataFrame`.  So by removing `__new__()` from the stubs, we are saying that you should not instantiate it in code.

twoertwein:
Thank you @Dr-Irv

PR #166, opened 2022-07-23
CLEAN: Align Groupby types

### LINKED ISSUE #149: CLEAN: Fix up Groupby classes
In pandas 1.1., the classes were `_GroupBy` and `Groupby`.  Since that time, `_Groupby` was replaced by `BaseGroupBy`, and generics were introduced.  

The original version of these stubs were generated on top of pandas 1.1.

This should get reconciled at some point so the class names are the same, including the various subclassing that is going on.

In pandas, `BaseGroupBy` inherits from `SelectionMixin[NDFrameT]`, but if that is used in the stubs, mypy complains because the definitions of `aggregate()` are different in `SelectionMixin` and `GroupBy`

So there is some work to do here.
 
See #142 and #148

Dr-Irv:
- [X] Closes #149 
  - None

pandas has `BaseGroupBy` (which should be private), whereas the stubs had `_GroupBy`, but I wanted to get things aligned.  Probably `BaseGroupBy` should be private in pandas.

In addition, `SelectionMixin` is really an abstract class, so removing `aggregate` makes sense, as the pandas implementation raises an exception anyway.

### DIFF
--- a/pandas-stubs/core/base.pyi
+++ b/pandas-stubs/core/base.pyi
@@ -1,24 +1,18 @@
 from __future__ import annotations
 
 from typing import (
-    Callable,
     Generic,
     Literal,
 )
 
 import numpy as np
-from pandas import (
-    DataFrame,
-    Index,
-    Series,
-)
+from pandas import Index
 from pandas.core.accessor import DirNamesMixin
 from pandas.core.arrays import ExtensionArray
 from pandas.core.arrays.categorical import Categorical
 
 from pandas._typing import (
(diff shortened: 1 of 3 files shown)


twoertwein (approved):
> In addition, SelectionMixin is really an abstract class, so removing aggregate makes sense, as the pandas implementation raises an exception anyway.

Could let it return `NoReturn` but that is not worth it.

twoertwein:
Thank you @Dr-Irv

PR #168, opened 2022-07-25
Fix groupby.apply() and sum(axis=1)

### LINKED ISSUE #167: [BUG] DataFrameGroupBy.apply can return Series
**Describe the bug**
Calling `apply` on a grouped DataFrame - ie a DataFrameGroupBy object - usually returns a DataFrame, and this is correctly typed in the stubs. However, if the callable being applied returns a scalar rather than a DataFrame or Series, then `apply` will in turn return a Series. This is documented in example 3 in the [docs for `GroupBy.apply`](https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.core.groupby.GroupBy.apply.html).

**To Reproduce**
Toy example: 
```
df = pd.DataFrame({"col1": [1, 2, 3], "col2": [4, 5, 6]})
result: pd.Series = df.groupby('col1').apply(lambda x: x.sum().mean())
assert(isinstance(result, pd.Series))
```
This code runs correctly and the assertion succeeds, but mypy gives "Incompatible types in assignment (expression has type "DataFrame", variable has type "Series[Any]")".

 - OS: MacOS
 - OS Version: 12.5
 - python version: 3.9.9
 - mypy version: 0.971
 - version of installed `pandas-stubs`: 1.4.3.220718

**Additional context**
It's not obvious how this could be fixed. I tried using `overload` to give different return types depending on the type of Callable passed:
```
    @overload
    def apply(self, func: Callable[[DataFrame], Union[DataFrame, Series]], *args, **kwargs) -> DataFrame: ...
    @overload
    def apply(self, func: Callable[[DataFrame], Union[Scalar, Sequence]], *args, **kwargs) -> Series: ...
```
and this correctly lints the example above, but fails in at least one particular circumstance: Pandas special-cases the built-in functions `sum`, `max` and `min` when passed to `apply`, replacing them with their numpy equivalents, so the above code causes the test in https://github.com/pandas-dev/pandas-stubs/blob/main/tests/test_frame.py#L558 to fail.

Dr-Irv:
- [x] Closes #167
- [x] Tests added:
  - `test_frame.py:test_groupby_apply()`

This revealed an issue with `DataFrame.sum(axis=1)` and we probably need to fix things like `DataFrame.mean(axis=1)`, etc., but I think we should just create an issue for that.


pandas-stubs/core/frame.pyi
```diff
@@ -1902,14 +1902,24 @@ class DataFrame(NDFrame, OpsMixin):
         fill_value: float | None = ...,
     ) -> DataFrame: ...
     @overload
+    def sum(
+        self,
+        axis: Literal[1, "columns"],
```
twoertwein, line 1907: Isn't the 3rd overload the same as this one but has a broader `axis` annotation? Is it possible to simply move the third overload up?

danielroseman:
Thanks for looking into this so quickly. Unfortunately I don't think this is quite right.

`df.sum()` with no parameters returns a Series. The `summean` function here returns a float, not a series - I'm not sure why this isn't showing a type error in your test. (In fact using latest head, `reveal_type(df.sum())` gives Any, which is a bit strange.)

But returning a scalar is the only situation in which `apply` would return a Series. In all other cases, including using the built-in `sum` (which maps to `np.sum` internally), `apply` returns a DataFrame:

```
In [2]: df = pd.DataFrame({"col1": [1, 2, 3], "col2": [4, 5, 6]})

In [3]: df.groupby('col1').apply(sum)
Out[3]:
      col1  col2
col1
1        1     4
2        2     5
3        3     6


In [4]: df.sum().mean()   # returns float, not Series
Out[4]: 10.5

In [5]: df.gropuby('col1').apply(lambda x: x.sum().mean())
Out[5]:
col1
1    2.5
2    3.5
3    4.5
dtype: float64
```

So I think there are three cases here: 

1. `func` returns a DataFrame or a Series: `apply(func)` returns a DataFrame
2. `func` is the built-in `sum`, `min` or `max` [ie the keys of `pandas.core.common._builtin_table`]: `apply(func)` returns a DataFrame
3. `func` returns a scalar (actually, anything that can go into a single DF element, including eg a string, list, tuple or dict): `apply(func)` returns a Series

pandas-stubs/core/frame.pyi
Dr-Irv: Yes, that's right.  Changed in next commit

Dr-Irv:
> 
> So I think there are three cases here:
> 
> 1. `func` returns a DataFrame or a Series: `apply(func)` returns a DataFrame
> 2. `func` is the built-in `sum`, `min` or `max` [ie the keys of `pandas.core.common._builtin_table`]: `apply(func)` returns a DataFrame
> 3. `func` returns a scalar (actually, anything that can go into a single DF element, including eg a string, list, tuple or dict): `apply(func)` returns a Series

I'm not sure I got all the cases right.  Would need better tests.  Current version passes the example you initially supplied.

One thing to be aware of is that using a `lambda` gives you a `Callable` that is not typed.  So I changed the test to type the callable.  It's inconvenient from a coding standpoint, but there is no way for a type checker to know the arguments and return type of a `lambda`.

pandas-stubs/core/groupby/generic.pyi
```diff
@@ -111,7 +111,13 @@ class _DataFrameGroupByNonScalar(DataFrameGroupBy):
...
     def any(self, skipna: bool = ...) -> DataFrame: ...
     def all(self, skipna: bool = ...) -> DataFrame: ...
-    def apply(self, func, *args, **kwargs) -> DataFrame: ...
+    # mypy sees the two overloads as overlapping
+    @overload
+    def apply(  # type: ignore[misc]
+        self, func: Callable[[DataFrame], Series | Scalar], *args, **kwargs
+    ) -> Series: ...
+    @overload
+    def apply(self, func: Callable, *args, **kwargs) -> DataFrame: ...
```
twoertwein, line 120: They are overlapping `Callable[[DataFrame], Series | Scalar]` is also `Callable` but their return type is different.

Not sure whether it is possible, you can add more specialized functions for the DataFrame case (so that the first and second overload do not overlap anymore). And at the end have a "catch-all" overload `def apply(self, func: Callable, *args, **kwargs) -> DataFrame | Series: ...`

Dr-Irv: `pyright` doesn't complain about this - just `mypy` does.

Just did a commit where I added another `overlay` and made the catchall return `DataFrame | Series` as you suggested.

pandas-stubs/core/frame.pyi
```diff
@@ -1906,22 +1906,22 @@ class DataFrame(NDFrame, OpsMixin):
         self,
         axis: AxisType | None = ...,
         skipna: _bool | None = ...,
+        level: None = ...,
         numeric_only: _bool | None = ...,
         min_count: int = ...,
-        *,
-        level: Level,
```
twoertwein, line 1912: Is that true for a MultiIndex? If not, it might be good to mention that the stubs focus on getting the most common usage patterns rights (as pandas can not completely be typed because of the current type system/pandas API), maybe in the README (I think numpy has a "limitation"/"difference from implementation" section on their type page).

pandas-stubs/core/groupby/generic.pyi
```diff
@@ -111,7 +112,17 @@ class _DataFrameGroupByNonScalar(DataFrameGroupBy):
...
     def any(self, skipna: bool = ...) -> DataFrame: ...
     def all(self, skipna: bool = ...) -> DataFrame: ...
-    def apply(self, func, *args, **kwargs) -> DataFrame: ...
+    # mypy sees the two overloads as overlapping
+    @overload
+    def apply(  # type: ignore[misc]
+        self, func: Callable[[DataFrame], Series | Scalar], *args, **kwargs
+    ) -> Series: ...
+    @overload
+    def apply(  # type: ignore[misc]
```
twoertwein, line 121: I replaced my local version of generic.pyi with this one and it seems that neither pyright nor mypy need these two ignores (I would have thought both of them need ignores, a DataFrame is Iterable).

pandas-stubs/core/frame.pyi
Dr-Irv: If you specify a `level`, then you get a different result than when you don't specify it.

I did write this here: https://github.com/pandas-dev/pandas-stubs/blob/main/docs/philosophy.md

> The goal of the pandas-stubs project is to provide type stubs for the public API that represent the recommended ways of using pandas.

pandas-stubs/core/groupby/generic.pyi
Dr-Irv: If I remove both of these ignores, I get errors from both `pyright` and `mypy`

twoertwein: you are right, not sure what I messed up. Unless you add a test covering this second overload, I would be tempted to remove it.

Dr-Irv: > you are right, not sure what I messed up. Unless you add a test covering this second overload, I would be tempted to remove it.

There is a test already.   That's why I needed to add the second overload.

https://github.com/pandas-dev/pandas-stubs/blob/77b357f1b64d3ee39057cedf099f167d02503696/tests/test_frame.py#L558

pandas-stubs/core/frame.pyi
twoertwein: When running this:
```py
type(pd.DataFrame({"a": [1], "b": [2], "c": [3]}).set_index(["a", "b"]).sum(level="a"))  # <class 'pandas.core.frame.DataFrame'>
FutureWarning: Using the level keyword in DataFrame and Series aggregations is deprecated and will be removed in a future version. Use groupby instead. df.sum(level=1) should use df.groupby(level=1).sum().
```

I wouldn't mind omitting the level argument in sum.

Dr-Irv: Can you create an issue and we'll do a separate PR for that?

pandas-stubs/core/groupby/generic.pyi
twoertwein: Okay, I'm not a fan of this overload but it seems to work :)

twoertwein:
Thanks @Dr-Irv

PR #169, opened 2022-07-25
CLN: update flake8-pyi

twoertwein:
The big changes are:

> Y041 detects redundant numeric unions. For example, PEP 484 specifies that type checkers should treat int as an implicit subtype of float, so int is redundant in the union int | float. In the same way, int is redundant in the union int | complex, and float is redundant in the union float | complex.
> Y044 from __future__ import annotations has no effect in stub files, as forward references in stubs are enabled by default.
>


pandas-stubs/_typing.pyi
```diff
@@ -43,16 +41,14 @@ from pandas.core.dtypes.dtypes import ExtensionDtype
 
 ArrayLike = Union[ExtensionArray, np.ndarray]
 AnyArrayLike = Union[Index, Series, np.ndarray]
-PythonScalar = Union[str, int, float, bool, complex]
+PythonScalar = Union[str, bool, complex]
```
twoertwein, line 44: PR to make these changes in pandas https://github.com/pandas-dev/pandas/pull/47850

Dr-Irv (approved):
Thanks @twoertwein

PR #170, opened 2022-07-26
CLEAN: remove nonpublic stubs

Dr-Irv:
- [x] Tests added
  - added `test_pandas.py:test_unique()`

### Removals
- `pd.value_counts()` - while exported, it is not documented
- Everything in `algorithms.pyi` except for `factorize()` and `unique()`
- `pd.api.extensions.take` - not documented
- From `pandas.core`:  `apply.pyi`, `sorting.pyi`, `nanops.pyi` - nothing there is public

### Refinements
- `pd.unique()` - added tests and overloads based on the docs


twoertwein:
> * `pd.value_counts()` - while exported, it is not documented

I guess we follow a "remove first, wait for user issue, create doc issue at pandas, add back"-approach?

Dr-Irv:
> > * `pd.value_counts()` - while exported, it is not documented
> 
> I guess we follow a "remove first, wait for user issue, create doc issue at pandas, add back"-approach?
> 
> edit: might not be the best user-experience but it is probably the best way to expand the pandas docs.

I will raise an issue in the pandas repo to ask if this is a docs oversight, or a coding error.  Either it is not supposed to be public, so the code should get fixed, or the docs need to get fixed.  If the former, then the change here is fine.  If the latter, then most likely no user is going to complain about it until we document it, so removing it now causes no harm.

Dr-Irv:
> I will raise an issue in the pandas repo to ask if this is a docs oversight, or a coding error. Either it is not supposed to be public, so the code should get fixed, or the docs need to get fixed. If the former, then the change here is fine. If the latter, then most likely no user is going to complain about it until we document it, so removing it now causes no harm.

Based on this: https://www.programcreek.com/python/example/101338/pandas.value_counts
it seems that it is meant to be public, and just missing from the docs, given how much it is used.

I will put it back in here.

twoertwein (approved):
Looks good to me

twoertwein:
Thanks @Dr-Irv !

(When in doubt, I wouldn't mind removing slightly more than necessary: in the worst case we need to add it back but it will slowly make the pandas doc more complete.)

PR #173, opened 2022-07-29
Standardised aggregate functions typing

amotzop:
Made all `agg/aggregate/transform` functions use same basic typing aliases `AggFuncType`, `AggFuncTypeBase` and `AggFuncTypeDict`.


pandas-stubs/core/frame.pyi
```diff
@@ -919,28 +922,28 @@ class DataFrame(NDFrame, OpsMixin):
     ) -> DataFrame: ...
     def diff(self, periods: int = ..., axis: AxisType = ...) -> DataFrame: ...
     @overload
-    def agg(self, func: Callable | _str, axis: AxisType = ..., **kwargs) -> Series: ...
+    def agg(self, func: AggFuncTypeBase, axis: AxisType = ..., **kwargs) -> Series: ...
     @overload
     def agg(
         self,
-        func: list[Callable] | dict[_str, Callable] = ...,
+        func: list[AggFuncTypeBase] | AggFuncTypeDict = ...,
```
twoertwein, line 929: This and some of the other changes allow more types. Can you please add tests for these newly accepted types, for example `func: list[str]`, `func: dict[str, list[Callable]]`, and `func: dict[str, list[str]]`?

Dr-Irv (changes requested):
Given that the results of `agg`, `aggregate`, etc. could be of different types, can you modify the tests to have a pattern like this:
```python
check(assert_type(df.agg(min), pd.Series), pd.Series)
```
Adding that will confirm that the typing result and actual result in execution match up.

tests/test_frame.py
```diff
@@ -548,9 +548,8 @@ def test_types_groupby() -> None:
     df.groupby(by="col1", sort=False, as_index=True)
     df.groupby(by=["col1", "col2"])
 
-    df1: pd.DataFrame = df.groupby(by="col1").agg("sum")
```
Dr-Irv, line 551: please don't remove existing tests

tests/test_frame.py
```diff
@@ -560,10 +559,37 @@ def test_types_groupby() -> None:
     df8: pd.DataFrame = df.groupby("col1").transform("sum")
     s1: pd.Series = df.set_index("col1")["col2"]
     s2: pd.Series = s1.groupby("col1").transform("sum")
-    s3: pd.Series = df.groupby("col1")["col3"].agg(min)
-    df9: pd.DataFrame = df.groupby("col1")["col3"].agg([min, max])
-    df10: pd.DataFrame = df.groupby("col1").agg(
-        new_col=pd.NamedAgg(column="col2", aggfunc="max")
```
Dr-Irv, line 566: I think the test for `df10` got lost

tests/test_frame.py
```diff
@@ -630,6 +656,15 @@ def test_types_window() -> None:
...
     df.rolling(2, axis=1, center=True)
 
+    df.rolling(2).agg("max")
+    df.rolling(2).agg(max)
+    df.rolling(2).agg(["max", "min"])
+    df.rolling(2).agg([max, min])
+    df.rolling(2).agg({"col2": "max"})
+    df.rolling(2).agg({"col2": max})
+    df.rolling(2).agg({"col2": ["max", "min"]})
+    df.rolling(2).agg({"col2": [max, min]})
```
Dr-Irv, line 666: Put in the `check` and `assert_type` here as well.

tests/test_series.py
```diff
@@ -458,6 +482,11 @@ def test_types_window() -> None:
     s.rolling(2)
     s.rolling(2, axis=0, center=True)
 
+    s.rolling(2).agg("sum")
+    s.rolling(2).agg(sum)
+    s.rolling(2).agg(["max", "min"])
+    s.rolling(2).agg([max, min])
```
Dr-Irv, line 488: add `check` and `assert_type` here as well

tests/test_series.py
```diff
@@ -496,9 +525,40 @@ def test_types_compare() -> None:
 
 def test_types_agg() -> None:
     s = pd.Series([1, 2, 3], index=["col1", "col2", "col3"])
-    s.agg("min")
-    s.agg(x=max, y="min", z=np.mean)
-    s.agg("mean", axis=0)
+    check(s.agg("min"), np.int64)
+    check(s.agg(min), np.int64)
```
Dr-Irv, line 529: add `assert_type` on the above 2

tests/test_series.py
```diff
@@ -496,9 +525,40 @@ def test_types_compare() -> None:
...
+    check(assert_type(s.agg(["min", "max"]), pd.Series), pd.Series)
+    check(assert_type(s.agg([min, max]), pd.Series), pd.Series)
+    check(assert_type(s.agg({"a": "min"}), pd.Series), pd.Series)
+    check(assert_type(s.agg({0: min}), pd.Series), pd.Series)
+    check(assert_type(s.agg(x=max, y="min", z=np.mean), pd.Series), pd.Series)
+    check(s.agg("mean", axis=0), np.float64)
```
Dr-Irv, line 535: add `assert_type`

Dr-Irv (changes requested):
I know we're asking to have you add `check` and `assert_type` on existing tests that were already there.  You have the option of leaving those tests in there as they were OR, for completeness given the nature of your changes, we'd prefer you add in those checks on existing test functions.  Also have to make sure that no tests were dropped.

tests/test_series.py
amotzop: I tried but had a small problem as the return type is `S1` and I couldn't get both mypy and pyright to pass (mypy wanted `Any` while pyright wanted `Unknown`).

tests/test_frame.py
amotzop: It moved to `test_types_groupby_agg` and has a return type checking now.

tests/test_series.py
amotzop: Maybe the return type in this case should just be `Scalar`?

Dr-Irv: If you use `Scalar`, then it is a `Union` of types, so that isn't friendly.  `S1` is correct.  But when you create a `Series` as you did, it is generic, so tracked as `Series[Any]`.  I think you could do
```python
check(assert_type(s.agg("min"), Any), np.int64)
```
or maybe this:
```python
check(assert_type(s.agg("min"), "S1"), np.int64)
```

tests/test_frame.py
Dr-Irv: OK, now I see that.  Thanks.

tests/test_series.py
twoertwein: or could cast `s` to be of type `Series[int]`?

amotzop: Just thought that `S1` is not entirely correct, for example if you aggregate using `"mean"` you can get floats from ints.
`Any` doesn't pass the pyright checks (expected "Any" but received "Unknown")
"S1" also doesn't seem to work (it's not defined).
Making `s` into a `Series[int]` creates problems because then all the type checks need to be compared against `pd.Series[int]` but then the pytest run fails because the "real" pandas Series type isn't generic.

Dr-Irv: I think you could get `"S1"` to work by doing 
```python
if TYPE_CHECKING:
    from pandas._typing import S1
```

I agree that the result won't always be right, e.g., with "mean", but you could then create an overload that says if "mean" or `mean` are passed, the result is float.

amotzop: It doesn't work because `S1` is an unbound `TypeVar`.

Dr-Irv: This is coming down to different behaviors with `mypy` and `pyright`.  I prefer the `mypy` behavior here.  I've raised an issue with `pyright` to see what they say.    https://github.com/microsoft/pyright/issues/3794

Dr-Irv: It appears that this is a `pyright` bug that will get fixed in the "next release", which will probably happen within a week, so let's wait for that, and then you can do `assert_type()` versus `Any` with both `mypy` and `pyright`

Dr-Irv: It's now fixed in pyright 1.1.266 so you should be able to do
```python
check(assert_type(s.agg("min"), Any), np.int64)
```
You should update `pyproject.yaml` to refer to that version.

Dr-Irv (approved):
thanks @amotzop

PR #176, opened 2022-08-01
Make relational operators work with all scalars

### LINKED ISSUE #175: Unsupported operand types for > ("Series[Any]" and "Timedelta")
`pandas` supports comparison for `Series[Timedelta]` and `Timedelta`, but `pandas-stubs` does not.

```py
import pandas as pd

s = pd.Series([pd.Timedelta(minutes=2)])
result = s > pd.Timedelta(hours=1)
```

Using mypy:

```sh
$ mypy tests.py
tests.py:4: error: Unsupported operand types for > ("Series[Any]" and "Timedelta")
```

Version of installed `pandas-stubs`: 1.4.3.220724

https://stackoverflow.com/questions/73064606/how-can-i-stop-mypy-complaining-when-comparing-a-pd-series-to-single-value

Note: This is similar to #27, which reported for `Timestamp`.

Dr-Irv:
- [x] Closes #175 
- [x] Tests added:
  - `test_series.py:test_relops()`

Made it so any scalar is comparable to a `Series`.


twoertwein (approved):
Happy to merge!

There are places where the annotations for the operators is relatively strict in regard to the type of Series ,e.g., to prevent `__ge__(self: Series[int], other: Series[str])`. Could replace `Scalar` with `S1` but I assume that might be too strict/create too many false negatives?

Dr-Irv:
> There are places where the annotations for the operators is relatively strict in regard to the type of Series ,e.g., to prevent `__ge__(self: Series[int], other: Series[str])`. Could replace `Scalar` with `S1` but I assume that might be too strict/create too many false negatives?

I like the idea of using `S1` here.  If the `Series` has a known type, we are then saying you can only compare to that type.  Most of the time, the `Series` is really `Series[Any]` so allowing any comparison to `S1` will work fine.

This will then prevent doing `Series[Timestamp]` comparing to `Timedelta`, which is something we want to pick up.

Let's put it out there and see if we get any reports.  Tests are passing with using `S1`.

twoertwein:
Thanks @Dr-Irv !

PR #177, opened 2022-08-01
More specific types for GroupBy.apply.

danielroseman:
- [x] Refs #168

Thanks for the previous change (#168) on this issue, but unfortunately it didn't quite cover all the use cases - on my code base it frequently fell through to the union of DataFrame and Series, which would have required a lot of extra narrowing. This change is more specific and adds some extra test cases which should catch all the variants.

Also added a missing `sum` method on SeriesGroupBy.


pandas-stubs/core/groupby/generic.pyi
```diff
@@ -110,18 +111,20 @@ class _DataFrameGroupByNonScalar(DataFrameGroupBy):
 class DataFrameGroupBy(GroupBy):
     def any(self, skipna: bool = ...) -> DataFrame: ...
     def all(self, skipna: bool = ...) -> DataFrame: ...
-    # mypy sees the two overloads as overlapping
+    # mypy and pyright see these overloads as overlapping
     @overload
     def apply(  # type: ignore[misc]
-        self, func: Callable[[DataFrame], Series | Scalar], *args, **kwargs
+        self, func: Callable[[Iterable], float], *args, **kwargs
```
Dr-Irv, line 117: The `Callable[[Iterable]...` should come last, because you want the other ones to match first.  I think that will make `pyright` not see things as overlapping.

Dr-Irv (changes requested):

pandas-stubs/core/groupby/generic.pyi
danielroseman: I thought that originally. But then mypy gives the following error:
```
tests/test_frame.py:559: error: Incompatible types in assignment (expression has type "Series[Any]", variable has type "DataFrame")  [assignment]
```

I can't really explain that behaviour, because according to everything I can find mypy should also match in order.

Dr-Irv: I just tried moving it last and it worked.  Still had `pyright` complain about the last overload never being matched without the `#type: ignore` in there, but it did work.  Can you try again?

Dr-Irv (changes requested):

Dr-Irv:
`pyright` reports an overloading problem due to what I think is a bug with `strict` type checking.  Reported here:  https://github.com/microsoft/pyright/issues/3780

pandas-stubs/core/groupby/generic.pyi
danielroseman: Apologies, looks like I had the previous version of mypy. Running the reordered code against the latest version works correctly.

Dr-Irv (approved):
Thanks @danielroseman

PR #179, opened 2022-08-03
Updated Dataframe.to_excel and Dataframe.to_latex

p5k369:
For both, _header_ accepts boolean or a list of strings.


tests/test_frame.py
```diff
@@ -1080,6 +1080,15 @@ def test_frame_getitem_isin() -> None:
     check(assert_type(df[df.index.isin([1, 3, 5])], pd.DataFrame), pd.DataFrame)
 
 
+def test_to_excel() -> None:
+    if TYPE_CHECKING:  # skip pytest
```
twoertwein, line 1084: Why does this need to be skipped? Is it because one of the excel dependencies is not installed?

p5k369: Otherwise openpyxl would be a dependency only for this case. As far as I know it is not needed elsewhere. Do you think thats worth the dependency?

Dr-Irv: I think we can add it as a `dev-dependencies` in the `pyproject.toml` .  Can you try that?

tests/test_frame.py
```diff
@@ -1080,6 +1080,15 @@ def test_frame_getitem_isin() -> None:
...
+
+        df = pd.DataFrame(data={"col1": [1, 2], "col2": [3, 4]})
+        df.to_excel("foo.xlsx", startrow=1, startcol=1, header=False)
```
Dr-Irv, line 1087: Can you use the pattern that is in `test_pandas.py` with `NamedTemporaryFile` that is used to test `to_csv()` ?

Dr-Irv (changes requested):

Dr-Irv:
Looking into this, it seems that we are not supporting `Series.to_excel()` and `Series.to_latex()`, so we should fix these here as well.

Can you take `to_excel()` and `to_latex()` out of `frame.pyi` and then use just a common stub in `core/generic.pyi` ?

tests/test_frame.py
p5k369: I added openpyxl to the dev-dependencies.

tests/test_frame.py
p5k369: The test is now no longer skipped by pytest.

p5k369:
> Can you take `to_excel()` and `to_latex()` out of `frame.pyi` and then use just a common stub in `core/generic.pyi` ?

Hey @Dr-Irv,
To be honest, I'm not sure what I had to do other than removing `to_excel()` and `to_latex()` from `frame.pyi` because they both were already present in `generic.pyi`. Did I misunderstand something?

Dr-Irv:
> To be honest, I'm not sure what I had to do other than removing `to_excel()` and `to_latex()` from `frame.pyi` because they both were already present in `generic.pyi`. Did I misunderstand something?

No, but I wanted to make sure that the stubs in `generic.pyi` were the same as the ones in `frame.pyi`, and they were, so we're all good on that front.

Dr-Irv (approved):
Thanks @p5k369

PR #182, opened 2022-08-05
some cleanups from stubtest

twoertwein:


pandas-stubs/api/indexers/__init__.pyi
```diff
@@ -1,6 +1,6 @@
-from pandas.core.indexers import check_array_indexer as check_array_indexer
-from pandas.core.window.indexers import (
+from pandas.core.indexers import (
```
twoertwein, line 1: `indexers` is a file in the stubs but a folder in pandas. `BaseIndexer` and so on are technically important from a file in this folder not from `__init__`.

pandas-stubs/io/common.pyi
```diff
@@ -17,15 +17,7 @@ def validate_header_arg(header) -> None: ...
...
 ) -> FilePathOrBuffer[AnyStr]: ...
-def is_s3_url(url) -> bool: ...
-def is_gcs_url(url) -> bool: ...
 def urlopen(*args, **kwargs) -> IO: ...
-def get_filepath_or_buffer(
-    filepath_or_buffer: FilePathOrBuffer,
-    encoding: str | None = ...,
-    compression: str | None = ...,
-    mode: str | None = ...,
-): ...
```
Dr-Irv, line 28: I have a feeling that this entire file can be deleted, because it has nothing that is public in it.

Dr-Irv (changes requested):

pandas-stubs/io/common.pyi
twoertwein: I agree. Happy to delete it - wanted to avoid "controversial" changes.

Dr-Irv (approved):
Thanks @twoertwein

PR #188, opened 2022-08-10
BUG: groupby not accepting list[str] as by argument

### LINKED ISSUE #187: BUG: groupby by argument raises with mypy when argument is annotated as ``list[str]``
**Describe the bug**

Using ``df.groupby(by=cols)`` where cols is annotated as ``list[str]`` raises a mypy error

```
error: Argument "by" to "groupby" of "DataFrame" has incompatible type "List[str]"; expected "Optional[Union[Tuple[Any, ...], List[Optional[Hashable]], Union[ufunc, Callable[..., Any]], Series[Any], ndarray[Any, Any], Mapping[Optional[Hashable], Any], Index]]"  [arg-type]
``` 

**To Reproduce**

```
def _call_groupby(df: pd.DataFrame, cols: list[str]):
    df.groupby(by=cols).count()
```

This also raises when adding:

```
    cols: list[str] = ["col1", "col2"]
    df.groupby(by=cols).sum()
```

to ``test_types_groupby`` in ``test_frame.py``

``list[str]`` should be valid? If passing simply ``["col1", "col2"]`` without an annotation this works

 - OS:  ubuntu
 - OS Version 22.04
 - python version 3.10
 - version of installed `pandas-stubs` newest 20220807

**Additional context**

phofl:
- [x] Closes #187 (Replace xxxx with the Github issue number)

Please let me know, if there is anything else I should add

cc @twoertwein

### DIFF
--- a/tests/test_frame.py
+++ b/tests/test_frame.py
@@ -565,6 +565,7 @@ def test_types_groupby() -> None:
     df10: pd.DataFrame = df.groupby("col1").agg(
         new_col=pd.NamedAgg(column="col2", aggfunc="max")
     )
+    # GH#187
     cols: list[str] = ["col1", "col2"]
     check(assert_type(df.groupby(by=cols).sum(), pd.DataFrame), pd.DataFrame)
 
--- a/tests/test_frame.py
+++ b/tests/test_frame.py
@@ -546,13 +546,6 @@ def test_types_groupby() -> None:
     df.groupby(by="col1")
     df.groupby(level="ind")
     df.groupby(by="col1", sort=False, as_index=True)
-
-    cols: list[str] = ["col1", "col2"]
-    df.groupby(by=cols).sum()
-
-    cols_opt: list[str | None] = ["col1", "col2"]
-    df.groupby(by=cols_opt).sum()
-
     df.groupby(by=["col1", "col2"])
 
     df1: pd.DataFrame = df.groupby(by="col1").agg("sum")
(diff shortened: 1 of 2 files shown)


tests/test_frame.py
```diff
@@ -546,6 +546,13 @@ def test_types_groupby() -> None:
     df.groupby(by="col1")
     df.groupby(level="ind")
     df.groupby(by="col1", sort=False, as_index=True)
+
+    cols: list[str] = ["col1", "col2"]
+    df.groupby(by=cols).sum()
+
+    cols_opt: list[str | None] = ["col1", "col2"]
+    df.groupby(by=cols_opt).sum()
```
twoertwein, line 554: If you want, you can also test the return value

twoertwein (approved):

tests/test_frame.py
phofl: Added

twoertwein: `check(assert_type(df.groupby(by=cols).sum(), pd.DataFrame), pd.DataFrame)`

phofl: Wouldn't this also fail with the assignment? Is the assert_type way the proper way of checking this?

twoertwein: If the return type was `Any`, `df: pd.DataFrame` would not report an error.

phofl: Ah, did not think about that. Thanks

phofl: Independently of this pr: We should check everything like this then, correct?

twoertwein: > Independently of this pr: We should check everything like this then, correct?

Ideally yes, but as a starting point, we like new code to use `assert_type` (for static typing) and also `check` (to test the runtime behavior; and for sequence-like items to check the dtpye of the items).

tests/test_frame.py
```diff
@@ -565,6 +565,11 @@ def test_types_groupby() -> None:
     df10: pd.DataFrame = df.groupby("col1").agg(
         new_col=pd.NamedAgg(column="col2", aggfunc="max")
     )
+    cols: list[str] = ["col1", "col2"]
```
Dr-Irv, line 568: can you add a #GH 187 comment here?

Dr-Irv (changes requested):
One small change, otherwise looks good

tests/test_frame.py
phofl: Done

Dr-Irv (approved):
Thanks @phofl

PR #192, opened 2022-08-13
Allow int and None in DataFrame.loc

### LINKED ISSUE #189: BUG: loc.__getitem__ stricter than it should be with list indexers
**Describe the bug**

``loc.__getitem__`` is stricter than it should be I think

```
error: List item 0 has incompatible type "int"; expected "str"  [list-item]
error: List item 0 has incompatible type "None"; expected "str"  [list-item]

``` 

**To Reproduce**

```
df = pd.DataFrame([[1, 2]], columns=Index([1, None], dtype=object))

df.loc[:, [None]]
df.loc[:, [1]]

```

Alternatively you could define df as 

```
df = pd.DataFrame({"a": [1, 2], None: 5})
```

Both of them should be valid. Especially since

```

df.loc[:, None]
df.loc[:, 1]

```

both work.

 - OS:  ubuntu
 - OS Version 22.04
 - python version 3.10
 - version of installed `pandas-stubs` newest 20220807

**Additional context**

Dr-Irv:
- [x] Closes #189
  - `test_frame.py:test_loclist()`

### DIFF
--- a/pandas-stubs/core/frame.pyi
+++ b/pandas-stubs/core/frame.pyi
@@ -114,20 +114,20 @@ class _LocIndexerFrame(_LocIndexer):
         self,
         idx: IndexType
         | MaskType
-        | list[StrLike]
+        | list[HashableT]
         | tuple[
             IndexType
             | MaskType
             | slice
-            | list[StrLike]
+            | list[HashableT]
             | tuple[str | int | slice, ...],
-            list[StrLike] | slice | Series[bool] | Callable,
+            list[HashableT] | slice | Series[bool] | Callable,
         ],
     ) -> DataFrame: ...
     @overload
     def __getitem__(
         self,
-        idx: tuple[int | StrLike | tuple[StrLike, ...], StrLike],
+        idx: tuple[int | StrLike | tuple[HashableT, ...], int | StrLike],
     ) -> Scalar: ...
     @overload
     def __getitem__(
@@ -141,19 +141,19 @@ class _LocIndexerFrame(_LocIndexer):
         self,
         idx: MaskType
         | StrLike
-        | tuple[MaskType | Index | Sequence[Scalar] | Scalar | slice, ...],
+        | tuple[MaskType | Index | Sequence[ScalarT] | Scalar | slice, ...],
         value: S1 | ArrayLike | Series | DataFrame,
     ) -> None: ...
     @overload
     def __setitem__(
         self,
-        idx: tuple[tuple[StrLike | Scalar | slice, ...], StrLike],
+        idx: tuple[tuple[HashableT | Scalar | slice, ...], HashableT],
         value: S1 | ArrayLike | Series[S1] | list,
     ) -> None: ...
     @overload
     def __setitem__(
         self,
-        idx: tuple[tuple[StrLike | Scalar | slice, ...], StrLike],
+        idx: tuple[tuple[HashableT | Scalar | slice, ...], HashableT],
         value: S1 | ArrayLike | Series[S1] | list,
     ) -> None: ...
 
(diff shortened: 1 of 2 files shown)


pandas-stubs/core/frame.pyi
```diff
@@ -114,20 +114,20 @@ class _LocIndexerFrame(_LocIndexer):
...
             | tuple[str | int | slice, ...],
-            list[StrLike] | slice | Series[bool] | Callable,
+            list[HashableT] | slice | Series[bool] | Callable,
         ],
     ) -> DataFrame: ...
     @overload
     def __getitem__(
         self,
-        idx: tuple[int | StrLike | tuple[StrLike, ...], StrLike],
+        idx: tuple[int | StrLike | tuple[HashableT, ...], int | StrLike],
```
twoertwein, line 130: Wouldn't  a tuple of tuples return a DataFrame?

pandas-stubs/core/frame.pyi
```diff
@@ -141,19 +141,19 @@ class _LocIndexerFrame(_LocIndexer):
...
         | StrLike
-        | tuple[MaskType | Index | Sequence[Scalar] | Scalar | slice, ...],
+        | tuple[MaskType | Index | Sequence[ScalarT] | Scalar | slice, ...],
         value: S1 | ArrayLike | Series | DataFrame,
     ) -> None: ...
     @overload
     def __setitem__(
         self,
-        idx: tuple[tuple[StrLike | Scalar | slice, ...], StrLike],
+        idx: tuple[tuple[HashableT | Scalar | slice, ...], HashableT],
```
twoertwein, line 150: This will bind `HashableT` to the same type: if the second tuple element is of type `str`, the first element needs to be a `tuple[str, ...]`.

pandas-stubs/core/frame.pyi
phofl, line 130: I think the column indexer should be HashableT too?

Dr-Irv:
```python
>>> df = pd.DataFrame.from_records([["a", "x", 1], ["a", "y", 2]], columns=["ab", "xy", "val"]).set_index(["ab", "xy"])
>>> df
       val
ab xy
a  x     1
   y     2
>>> df.loc[("a", "x"), "val"]
1
```
Here the argument is a tuple `(("a", "x"), "val")` and it returns a single value

pandas-stubs/core/frame.pyi
Dr-Irv: Can you give an example that is valid pandas where we matched before, and don't match now?  Or an example that is valid pandas that should match and we're not catching it?

pandas-stubs/core/frame.pyi
Dr-Irv: I tried that, and then `mypy` says the overloads were overlapping (which could be a mypy bug)

twoertwein: I assume the first and second overload would overlap then. The first overload accepts `tuple[ tuple[slice, ...], slice]` and the second overload would accept `tuple[tuple[HashableT, ...], HashableT]`. `slice` is `Hashable` and would match both overloads but return different return values.

pandas-stubs/core/frame.pyi
twoertwein: Maybe it doesn't bind within the same annotation - I couldn't create a working example that is not accepted :)

Are the 2nd and 3rd the same?

pandas-stubs/core/frame.pyi
Dr-Irv: Yes, that must be it.  So I think we have to go with what I put in here

pandas-stubs/core/frame.pyi
Dr-Irv: Good catch. That's been there since we pulled the stubs over from Microsoft!

Next commit removes the duplicate.

twoertwein (approved):

pandas-stubs/core/frame.pyi
twoertwein: Happy to merge after rebasing (conflicts with the other merged PR).

cc @phofl

Dr-Irv: have done the rebasing.  I will do a new release once this is merged.

twoertwein:
Thanks @Dr-Irv !

PR #195, opened 2022-08-16
read_excel accepts PathLike, buffers

tdsmith:
pandas.read_excel is documented to accept

> str, bytes, ExcelFile, xlrd.Book, path object, or file-like object

This expands the annotation to include str, bytes, paths,
and file-likes. Bytes are coerced into a BytesIO inside ExcelReader.

### DIFF
--- a/tests/test_frame.py
+++ b/tests/test_frame.py
@@ -1273,20 +1273,20 @@ def test_read_excel() -> None:
 
 def test_read_excel_io_types() -> None:
     df = pd.DataFrame([[1, 2], [8, 9]], columns=["A", "B"])
-    with tempfile.NamedTemporaryFile(delete=False, suffix=".xlsx") as file:
+    with tempfile.NamedTemporaryFile(suffix=".xlsx") as file:
         as_str: str = file.name
         df.to_excel(file.name)
-        file.close()
 
-    pd.read_excel(as_str)
-    as_path = Path(as_str)
-    pd.read_excel(as_path)
+        check(assert_type(pd.read_excel(as_str), pd.DataFrame), pd.DataFrame)
 
-    with as_path.open("rb") as as_file:
-        pd.read_excel(as_file)
+        as_path = Path(as_str)
+        check(assert_type(pd.read_excel(as_path), pd.DataFrame), pd.DataFrame)
 
-    as_bytes = as_path.read_bytes()
-    pd.read_excel(as_bytes)
+        with as_path.open("rb") as as_file:
+            check(assert_type(pd.read_excel(as_file), pd.DataFrame), pd.DataFrame)
+
+        as_bytes = as_path.read_bytes()
+        check(assert_type(pd.read_excel(as_bytes), pd.DataFrame), pd.DataFrame)
 
 
 def test_join() -> None:
--- a/tests/test_frame.py
+++ b/tests/test_frame.py
@@ -1277,19 +1277,19 @@ def test_read_excel_io_types() -> None:
     with tempfile.NamedTemporaryFile(delete=False, suffix=".xlsx") as file:
         as_str: str = file.name
         df.to_excel(file)
-    try:
-        check(assert_type(pd.read_excel(as_str), pd.DataFrame), pd.DataFrame)
 
-        as_path = Path(as_str)
-        check(assert_type(pd.read_excel(as_path), pd.DataFrame), pd.DataFrame)
+    check(assert_type(pd.read_excel(as_str), pd.DataFrame), pd.DataFrame)
 
-        with as_path.open("rb") as as_file:
-            check(assert_type(pd.read_excel(as_file), pd.DataFrame), pd.DataFrame)
+    as_path = Path(as_str)
+    check(assert_type(pd.read_excel(as_path), pd.DataFrame), pd.DataFrame)
 
-        as_bytes = as_path.read_bytes()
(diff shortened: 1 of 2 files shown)


Dr-Irv (changes requested):
Can you please add tests for this?

tdsmith:
What should a test demonstrate?

Dr-Irv:
> What should a test demonstrate?

You create sample code that doesn't pass the type checker without the changes to the `pyi` files, but does pass it with those changes.

See https://github.com/pandas-dev/pandas-stubs/blob/44e3f17f04a966770ddd8ea546602d6f1bb65f11/tests/test_frame.py#L1258 for an example.

tdsmith:
Got it, thanks! I noticed the feather and arrow tests actually write out a fixture and then read it back in, so I tried to ape that, but let me know if we shouldn't actually do I/O. It looks like openpyxl is in the dev dependency set.

tests/test_frame.py
```diff
@@ -1271,6 +1271,24 @@ def test_read_excel() -> None:
...
+
+    pd.read_excel(as_str)
+    as_path = Path(as_str)
+    pd.read_excel(as_path)
+
+    with as_path.open("rb") as as_file:
+        pd.read_excel(as_file)
+
+    as_bytes = as_path.read_bytes()
+    pd.read_excel(as_bytes)
```
Dr-Irv, line 1289: Can you indent the part that is referring to the temporary file to be within the first `with` block? 

while looking at this, it uncovered an issue with respect to us not deleting the files (which we should do), so I'll create an issue for that, but it will be easier to address if the code above is within the `with` block.

Also, we now have the convention of checking return types with new PRs, so for each `pd.read_excel()` statement, can you do something like:
```python
check(assert_type(pd.read_excel(as_bytes), pd.DataFrame), pd.DataFrame)
```

tests/test_frame.py
```diff
@@ -1271,6 +1271,24 @@ def test_read_excel() -> None:
         df17: dict[int | str, pd.DataFrame] = pd.read_excel("foo", sheet_name=None)
 
 
+def test_read_excel_io_types() -> None:
```
Dr-Irv, line 1274: Can you add a comment here referring to this PR, i.e.,:
```python
    # GH 195
```

Dr-Irv (changes requested):

tests/test_frame.py
tdsmith: Done.

tdsmith:
It looks like Windows needs us to close the file before we can re-open it, so we shouldn't shuffle everything into the context manager, but we can unlink it in a finally: block.

tests/test_frame.py
```diff
@@ -1271,6 +1271,27 @@ def test_read_excel() -> None:
...
+    # GH 195
+    df = pd.DataFrame([[1, 2], [8, 9]], columns=["A", "B"])
+    with tempfile.NamedTemporaryFile(delete=False, suffix=".xlsx") as file:
+        as_str: str = file.name
+        df.to_excel(file)
```
Dr-Irv, line 1279: Just insert `file.close()` here and then you don't need the `try`..`finally` block.   That's what worked elsewhere.

Dr-Irv (changes requested):

tests/test_frame.py
tdsmith: I'm not sure that's right --

- If we pass delete=False to NamedTemporaryFile, then we have to take responsibility for the file or else it will get left behind. Exiting the context block just closes the file and does not delete it.

- If we pass delete=True (the default) to NamedTemporaryFile, calling file.close() will unlink the file -- NamedTemporaryFile returns a [_TemporaryFileWrapper](https://github.com/python/cpython/blob/d4792ce916b94d090b6c7bce8b0f973e840c9e4e/Lib/tempfile.py#L459) which overrides close(). In this case we can't close the file without unlinking it, so the tests will fail because the file will be gone.

So we need to pass delete=False so we can access the file after we close it, which means that we need to take some explicit action to clean up the file.

I think some other tests are calling delete=False and then not cleaning up the file, which I think is what you wanted to avoid here.

Dr-Irv: > I think some other tests are calling delete=False and then not cleaning up the file, which I think is what you wanted to avoid here.

Yes, I agree, and I created #196 so we remember to address that. 

I'm trying to figure out a way to address it in a uniform way.   So we could find a way here in this PR, and then address the other places later.

If you removed `try` and `finally` and just had `Path(as_str).unlink()` at the end, would that work?  Or is the concern that if the test fails when we run `pytest`, the file would still be left behind?

tdsmith: Roger!

> If you removed try and finally and just had Path(as_str).unlink() at the end, would that work? Or is the concern that if the test fails when we run pytest, the file would still be left behind?

Yes, that works fine unless an assertion fails.

Dr-Irv: > > If you removed try and finally and just had Path(as_str).unlink() at the end, would that work? Or is the concern that if the test fails when we run pytest, the file would still be left behind?
> 
> Yes, that works fine unless an assertion fails.

I'm leaning towards leaving out `try` and `finally` because if it does fail, you can at least see the file if you want to debug the issue.

Dr-Irv (approved):
Thanks @tdsmith

tdsmith:
Thanks for your patience!

PR #199, opened 2022-08-16
fix Series.split with expand=True

### LINKED ISSUE #194: Series.str.split with expand=True should return DataFrame
**Describe the bug**
```python
s = pd.Series(["abc-def", "ghi-jkl"])
s.str.split("-", expand=True)
reveal_type(s)
```
Above code returns `DataFrame`, but `reveal_type` returns `Series[Unknown]`

**To Reproduce**
1. See above
2. pyright and mypy
3. See above.

Need to modify `pandas-stubs/core/strings.pyi` so that `split` returns `DataFrame` when `expand=True`.  Create proper overload.

Dr-Irv:
- [x] Closes #194
- [x] Tests added
  - `test_series.py` in `test_string_accessors` - added a test with `expand=True`
  - `test_indexes.py`, added `test_str_split`


pandas-stubs/core/strings.pyi
```diff
@@ -44,9 +52,19 @@ class StringMethods(NoNewAttributesMixin, Generic[T]):
         na_rep: str | None = ...,
         join: Literal["left", "right", "outer", "inner"] = ...,
     ) -> T: ...
+    @overload
+    def split(
+        self, pat: str = ..., n: int = ..., *, expand: Literal[True], regex: bool = ...
+    ) -> _TS: ...
```
twoertwein, line 58: it might be nice to have all the `*` after the same argument and not at different levels across overloads.

pandas-stubs/core/strings.pyi
```diff
@@ -5,17 +5,25 @@ from typing import (
...
+)
 from pandas.core.base import NoNewAttributesMixin
 
 from pandas._typing import T
 
-class StringMethods(NoNewAttributesMixin, Generic[T]):
+# The _TS type is what is used for the result of str.split with expand=True
+_TS = TypeVar("_TS", DataFrame, MultiIndex)
+
+class StringMethods(NoNewAttributesMixin, Generic[T, _TS]):
```
twoertwein, line 26: how is the value of `_TS` determined? It is not in `__init__` and you didn't add it explicitly (return value of `StringMethods[MultiIndex]`)

pandas-stubs/core/strings.pyi
Dr-Irv: done in next commit

pandas-stubs/core/strings.pyi
Dr-Irv: It is passed in as a parameter to the `StringMethods` class so we know what the return type should be of `str.split` when `expand==True`

twoertwein: sorry, I'm blind - I didn't see the annotation of `str()` :)

twoertwein:
thanks @Dr-Irv

PR #201, opened 2022-08-17
ENH: Improve clipboard

bashtage:
Synchronize clipboard with upstream pandas


bashtage:
Pretty hard to add tests for so nothing added so far.

Dr-Irv (changes requested):
Given that the `kwargs` are the same as `read_csv` for `read_clipboard` and the same as `to_csv` for `to_clipboard`, we should copy those arguments over to here, and remove the `kwargs`

For testing, you could do a `to_clipboard` followed by `read_clipboard`

bashtage:
Clipboard testing doesn't work on CI without a lot of extra setup. Maintaining the CI setup is not worth it IMO.

bashtage:
Test added.  It runs on windows but not linux.

tests/test_io.py
```diff
@@ -0,0 +1,18 @@
+from pandas import DataFrame
+import pytest
+from typing_extensions import assert_type
+
+from tests import check
+
+from pandas.io.clipboard import PyperclipException
+from pandas.io.clipboards import read_clipboard
```
twoertwein, line 8: Directly importing it from the pandas namespace would probably be better as it is documented that way https://pandas.pydata.org/docs/reference/api/pandas.read_clipboard.html

If it is intended to be called from `pandas.io`, then maybe also a test for it. Would probably need to escalate that as an issue to pandas.

bashtage: Fixed.

Dr-Irv:
> Given that the `kwargs` are the same as `read_csv` for `read_clipboard` and the same as `to_csv` for `to_clipboard`, we should copy those arguments over to here, and remove the `kwargs`
> 

@bashtage Do you think you could do this as well?  No need to add a test for all the args, but you could copy them over.

Dr-Irv (approved):
thanks @bashtage

Issue #196, opened 2022-08-16
CLEAN:  Delete temporary files in tests


Dr-Irv:
When we use `NamedTemporaryFile` in our tests, the files stay around once the tests are complete.  We should put in code to delete the files after each is created and then used.

PR #202, opened 2022-08-17
ENH: Synchronize io/stata with pandas master

bashtage:
Sychronize and remvoe classes not part of the public API


bashtage:
No tests yet, but can do some here

pandas-stubs/io/stata.pyi
```diff
@@ -32,57 +34,46 @@ def read_stata(
...
+    path: FilePath | ReadBuffer[bytes],
+    convert_dates: bool = ...,
+    convert_categoricals: bool = ...,
+    index_col: str | None = ...,
+    convert_missing: bool = ...,
+    preserve_dtypes: bool = ...,
+    columns: list[HashableT] | None = ...,
+    order_categoricals: bool = ...,
+    chunksize: int | None = ...,
+    *,
```
twoertwein, line 52: so the second and third overload only different in whether the arguments are keyword-only or can also be provided as positional arguments?

If pandas would have deprecated positional arguments, I would remove the positional overloads - but they are not yet deprecated. (Will try to open a PR for that later today at pandas.)

bashtage: Yup. They are needed to handle the cases where `iterator=True` either using a keyword argument or only positional.

bashtage:
Test added.  I took the CM from pandas. It could be made available somewhere else, or we could stub it in even though it is not public.

pandas-stubs/core/frame.pyi
```diff
@@ -236,15 +240,19 @@ class DataFrame(NDFrame, OpsMixin):
...
-        byteorder: _str | Literal["<", ">", "little", "big"] | None = ...,
-        time_stamp=...,
+        byteorder: Literal["<", ">", "little", "big"] | None = ...,
+        time_stamp: _dt.datetime | None = ...,
         data_label: _str | None = ...,
-        variable_labels: dict | None = ...,
-        version: int = ...,
-        convert_strl: list[_str] | None = ...,
+        variable_labels: dict[Hashable, str] | None = ...,
+        version: int | None = ...,
```
Dr-Irv, line 250: reading the code and the docs, only valid versions of `version` are `[114, 117, 118, 119, None]`

pandas-stubs/core/frame.pyi
```diff
@@ -236,15 +240,19 @@ class DataFrame(NDFrame, OpsMixin):
     ) -> np.recarray: ...
     def to_stata(
         self,
-        path: FilePathOrBuffer,
-        convert_dates: dict | None = ...,
+        path: FilePath | WriteBuffer[bytes],
+        convert_dates: dict[Hashable, str] | None = ...,
```
Dr-Irv, line 244: Looking at the code, only values for `str` are `[ 
        "tc",
        "%tc",
        "td",
        "%td",
        "tw",
        "%tw",
        "tm",
        "%tm",
        "tq",
        "%tq",
        "th",
        "%th",
        "ty",
        "%ty",
    ]`

pandas-stubs/core/frame.pyi
```diff
@@ -236,15 +240,19 @@ class DataFrame(NDFrame, OpsMixin):
...
         data_label: _str | None = ...,
-        variable_labels: dict | None = ...,
-        version: int = ...,
-        convert_strl: list[_str] | None = ...,
+        variable_labels: dict[Hashable, str] | None = ...,
+        version: int | None = ...,
+        convert_strl: list[HashableT] | None = ...,
+        compression: CompressionOptions = ...,
+        storage_options: StorageOptions = ...,
+        *,
```
Dr-Irv, line 254: why include the `*` ?

Dr-Irv (changes requested):
I realize that you copied this from pandas source, but for the stubs, we can make things narrower.

tests/test_io.py
```diff
@@ -0,0 +1,93 @@
...
+from pandas.io.stata import (
+    StataReader,
+    read_stata,
+)
+
+DF = DataFrame({"a": [1, 2, 3], "b": [0.0, 0.0, 0.0]})
+
+
+@contextmanager
+def ensure_clean(filename=None, return_filelike: bool = False, **kwargs: Any):
```
Dr-Irv, line 27: Any reason we can't use `NamedTemporaryFile` here as we do in the tests elsewhere?

twoertwein: in pandas we don't use it because some of the to_* methods try to remove the file themselves on failure (I think only one of them does that, don't remember which one). This does not work on windows since NamedTemporaryFile has an open file handle to the file.

Dr-Irv: That relates to why I created #196 .  We're using `NamedTemporaryFile` already in the tests.

So I think we should have a separate PR to address #196 (maybe using `ensure_clean` throughout  - not opposed to that), and then this PR just uses `NamedTemporaryFile` for now.

Or we do the other PR first to introduce `ensure_clean()` first, then do this one for `stata`

twoertwein: going with `NamedTemporaryFile` for pandas-stubs is probably better: it is what people often use. And as long as we don't have testcases that internally error, we will not encounter the same issue as on pandas.

pandas-stubs/io/stata.pyi
twoertwein: I opened https://github.com/pandas-dev/pandas/pull/48128 to make most arguments keyword-only

twoertwein: using positional keywords will now be deprecated in 1.5 (except for the first non-self argument)

pandas-stubs/core/frame.pyi
bashtage: The * is to handle the case where someone does something like `read_stata(path, iterator=True)`. I can't have an elipses there since it would cause an overlap in type checking.

tests/test_io.py
bashtage: IMO a context manager is important when writing tests that do IO.  It could be simpler, but it would probably be easiest to just use `tm.ensure_clean` without bringing it over.  We could just stub this single function from `pandas._testing` and then everything should pass without needing to worry about the code.

Dr-Irv: > IMO a context manager is important when writing tests that do IO. It could be simpler, but it would probably be easiest to just use `tm.ensure_clean` without bringing it over. We could just stub this single function from `pandas._testing` and then everything should pass without needing to worry about the code.

I'm fine with that, but can we do that in a separate PR, importing `tm.ensure_clean()`, adding a stub for it, and changing the use of `NamedTemporaryFile` throughout ?  You could do that PR first, then finish this one, or use `NamedTemporaryFile` here, and then do the PR with `tm.ensure_clean()`

pandas-stubs/io/stata.pyi
```diff
@@ -32,57 +36,31 @@ def read_stata(
...
-    def __init__(self, value: float) -> None: ...
-    def __eq__(self, other: object) -> bool: ...
-    @property
-    def string(self) -> str: ...
-    @property
-    def value(self) -> float: ...
-    @classmethod
-    def get_base_missing_value(cls, dtype): ...
-
 class StataParser:
```
Dr-Irv, line 64: `StataParser` is not public, but `StataReader` inherits from it, but we should just keep stubs for the methods that are documented in `StataReader`

pandas-stubs/io/stata.pyi
```diff
@@ -147,65 +125,52 @@ class StataWriter(StataParser):
         self,
```
Dr-Irv, line 125: StataWriter is not public, but `StataWriter.write_file()` is, so we can just simplify this and only include the `write_file()` method

pandas-stubs/io/stata.pyi
```diff
@@ -147,65 +125,52 @@ class StataWriter(StataParser):
...
-        self,
-        df: DataFrame,
-        columns: Sequence[str],
-        version: int = ...,
-        byteorder: str | None = ...,
-    ) -> None: ...
-    def generate_table(self) -> tuple[dict[str, tuple[int, int]], DataFrame]: ...
-    def generate_blob(self, gso_table: dict[str, tuple[int, int]]) -> bytes: ...
-
 class StataWriter117(StataWriter):
```
Dr-Irv, line 141: `StataWriter117` is not public, so we can delete

pandas-stubs/io/stata.pyi
```diff
@@ -147,65 +125,52 @@ class StataWriter(StataParser):
...
+        variable_labels: dict[HashableT, str] | None = ...,
         convert_strl: Sequence[Hashable] | None = ...,
         compression: CompressionOptions = ...,
         storage_options: StorageOptions = ...,
         *,
-        value_labels: dict[Hashable, dict[float, str]] | None = ...,
+        value_labels: dict[HashableT, dict[float, str]] | None = ...,
     ) -> None: ...
 
 class StataWriterUTF8(StataWriter117):
```
Dr-Irv, line 159: `StataWriterUTF8` is not public, so we can delete

Dr-Irv (changes requested):

Dr-Irv (approved):
thanks @bashtage

PR #204, opened 2022-08-18
ENH: Synchronize ORC with upstream

bashtage:
Synchronize the orc read and write functions with upstream pandas


bashtage:
No tests because it would require pyarrow, which is not installed.

pandas-stubs/io/orc.pyi
```diff
@@ -1,7 +1,26 @@
...
+from pandas._typing import (
+    FilePath,
+    ReadBuffer,
+    WriteBuffer,
+)
 
 def read_orc(
-    path: FilePathOrBuffer, columns: list[str] | None = ..., **kwargs
+    path: FilePath | ReadBuffer[bytes],
+    columns: list[str] | None = ...,
```
Dr-Irv, line 16: Should this be `list[HashableT]` ?

pandas-stubs/io/orc.pyi
```diff
@@ -1,7 +1,26 @@
...
+    **kwargs: Any,
 ) -> DataFrame: ...
+def to_orc(
```
Dr-Irv, line 19: `pandas.io.to_orc()` is not documented, but v1.5 will have `DataFrame.to_orc()`, so can you put the stub there?

Dr-Irv (changes requested):

Dr-Irv:
> No tests because it would require pyarrow, which is not installed.

It would just have to go in the section in `pyproject.toml` under `[tool.poetry.dev-dependencies]` so wouldn't be an installation requirement for the stubs, just for testing, and I think we should do that.

Dr-Irv:
I see that you are skipping the tests if we have pandas < 1.5.  Did you test this with `main` from pandas?

pandas-stubs/io/orc.pyi
bashtage: Fixed.

Dr-Irv:
I'm good with these changes, but can you answer this question: https://github.com/pandas-dev/pandas-stubs/pull/204#issuecomment-1220603356

Also need to resolve conflicts.  (realizing this is a PITA as I keep approving PR's)

bashtage:
> I see that you are skipping the tests if we have pandas < 1.5. Did you test this with `main` from pandas?

Yes, it passes on main.

bashtage:
Ready for merge.

Dr-Irv (approved):
thanks @bashtage

PR #205, opened 2022-08-18
ENH: Synchronize gbq with upstream

bashtage:
Align with actual API

### DIFF
--- a/pandas-stubs/io/gbq.pyi
+++ b/pandas-stubs/io/gbq.pyi
@@ -1,4 +1,7 @@
-from typing import Any
+from typing import (
+    Any,
+    Literal,
+)
 
 from pandas.core.frame import DataFrame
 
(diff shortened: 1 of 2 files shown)


bashtage:
No tests due to optional dependencies and extra CI overhead of trying to test against big query.

pandas-stubs/io/gbq.pyi
```diff
@@ -12,10 +12,10 @@ def read_gbq(
     dialect: str | None = ...,
```
Dr-Irv, line 12: `dialect` is `'standard'` or `'legacy'` according to the docs

pandas-stubs/io/gbq.pyi
```diff
@@ -12,10 +12,10 @@ def read_gbq(
...
     location: str | None = ...,
     configuration: dict[str, Any] | None = ...,
-    credentials=...,
+    # Google type, not available
+    credentials: Any = ...,
     use_bqstorage_api: bool | None = ...,
```
Dr-Irv, line 17: according to docs, `use_bqstorage_api` can't be `None`

pandas-stubs/io/gbq.pyi
```diff
@@ -12,10 +12,10 @@ def read_gbq(
...
-    private_key=...,
-    verbose=...,
+    max_results: int | None = ...,
     progress_bar_type: str | None = ...,
```
Dr-Irv, line 19: `progress_bar_type` can be `Literal["tqdm", "tqdm_notebook", "tqdm_gui"]` according to docs

pandas-stubs/io/gbq.pyi
```diff
@@ -12,10 +12,10 @@ def read_gbq(
...
 ) -> DataFrame: ...
 def to_gbq(
```
Dr-Irv, line 21: `pandas.io.gbq` isn't public, so this should just be in `frame.pyi`

Dr-Irv (changes requested):

Dr-Irv:
> No tests due to optional dependencies and extra CI overhead of trying to test against big query.

Here I agree with you - the overhead of testing against big query isn't worth it.

pandas-stubs/io/gbq.pyi
bashtage: The default value is `None`, so it clearly is accepted.

Dr-Irv: Probably worthwhile to create an issue in the `pandas` repo

Dr-Irv:
@bashtage can you avoid doing force pushes, as it makes it hard to see the previous comments I made within the code changes you made?

Dr-Irv (approved):
Thanks @bashtage

bashtage:
I only did a force because I rebased.  Commits are linear

Dr-Irv:
> I only did a force because I rebased. Commits are linear

I don't think you have to do a force when you rebase.  Whereas the commits are linear, github loses the connection between my previous comments and the forced push version.

PR #209, opened 2022-08-19
Fixed axis and return types for DataFrame.all/any

amotzop:
- Added the option to pass `None` as `axis` in `DataFrame.all\any`.
- Fixed the return type to be `Series[bool]` instead of just `Series`.

### DIFF
--- a/pandas-stubs/core/frame.pyi
+++ b/pandas-stubs/core/frame.pyi
@@ -1160,7 +1160,6 @@ class DataFrame(NDFrame, OpsMixin):
         axis: None,
         bool_only: _bool | None = ...,
         skipna: _bool = ...,
-        level: None = ...,
         **kwargs,
     ) -> _bool: ...
     @overload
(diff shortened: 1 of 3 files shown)


amotzop:
I thought about also adding relevant overloads to `Series.all/any` according to the level argument, but apparently the level argument is deprecated (I got a `FutureWarning`), so I wasn't sure if it was better to add it or keep it as is (and maybe remove the relevant overloads for `DataFrame.all/any`?).

Dr-Irv:
> I thought about also adding relevant overloads to `Series.all/any` according to the level argument, but apparently the level argument is deprecated (I got a `FutureWarning`), so I wasn't sure if it was better to add it or keep it as is (and maybe remove the relevant overloads for `DataFrame.all/any`?).

If `level` is deprecated, we should remove it from these stubs so that people stop using it.  Can you do that?

However, the docs aren't showing that `level` is deprecated, which is an oversight, so can you create an issue in the pandas repo that indicates this documentation error?

Dr-Irv (approved):

Dr-Irv:
Thanks @amotzop

PR #212, opened 2022-08-20
DatetimeIndex accessors and reorganize the datetime accessors for Series

### LINKED ISSUE #193: No types for DatetimeIndex attributes
**Describe the bug**
```python
import pandas

x = pandas.DatetimeIndex()
x.month
```
Fails to typecheck in mypy even though this attribute is documented here https://pandas.pydata.org/docs/reference/api/pandas.DatetimeIndex.month.html

**To Reproduce**
1. Run type checker of your choice on the above snippet with latest version of pandas and pandas-stubs

 - OS: Linux
 - OS Version NixOS unstable
 - python version 3.9.13
 - version of installed `pandas-stubs` 1.4.3.220807

**Additional context**
Also tested with pyright language server and got same error, not sure what version of pyright.

Dr-Irv:
- [x] Closes #193
- [x] Tests added
  - `test_timefuncs.py:test_datetimeindex_accessors`

Reorganized the accessors to correspond to the pattern used in defining them for `DatetimeArray`.  Used a lot of `Generic` types and `TypeVar` to get all the return types right that are different between `Series.dt.xxx` and `DatetimeIndex.xxx` for various accessors `xxx`. 

Removed deprecated methods and associated tests.


twoertwein:
wow, we have to do all of this because of `inherit_names`? I would like to remove `inherit_names` but it is moving slowly since it also requires implementation changes https://github.com/pandas-dev/pandas/pull/48015

pandas-stubs/core/indexes/accessors.pyi
```diff
@@ -192,7 +314,65 @@ class PeriodProperties(DatetimeAndPeriodProperties):
...
+        TimestampSeries,
+        Series[str],
+        PeriodSeries,
+    ],
+    _TimedeltaPropertiesNoRounding,
+    _PeriodProperties,
 ):
     def __new__(cls, data: Series): ...
+
+class TimestampProperties(
```
twoertwein, line 342: Could mark newly introduced classes that are not in the implementation, maybe `_TimestampProperties`? Same for `TimestampProperties` and `PeriodSeries`

Dr-Irv:
> wow, we have to do all of this because of `inherit_names`? I would like to remove `inherit_names` but it is moving slowly since it also requires implementation changes [pandas-dev/pandas#48015](https://github.com/pandas-dev/pandas/pull/48015)

It's really because the way that the properties are hooked in to `Series.dt` and into `DatetimeIndex` is via the properties on `DatetimeArray` , which is using `inherit_names` so that if you call `Series.dt.month` or `DatetimeIndex.month` (for example), it delegates that to `DatetimeArray.month`, but returns a different result (`Series` versus `Index` respectively). 

The other option is that we just document all the properties for `Series.dt` and `DatetimeIndex`, but that then becomes different than the implementation.

pandas-stubs/core/indexes/accessors.pyi
Dr-Irv: > Could mark newly introduced classes that are not in the implementation, maybe `_TimestampProperties`? Same for `TimestampProperties` and `PeriodSeries`

Not sure.  Since `TimestampProperties` is the result of `TimestampSeries.dt`, we are then importing it there.  Do we want to have imports of "private" classes internally within the stubs?

So what I chose to do here is make any class that gets imported by another stub file public (no underscore) for the purpose of the stubs.

twoertwein: okay, sounds reasonable.

twoertwein:
Happy to keep your current approach - it works :) removing `inherit_names` will take quite a while.

Dr-Irv:
> Happy to keep your current approach - it works :) removing `inherit_names` will take quite a while.

Let me know if I should make any other changes.

twoertwein:
Thanks @Dr-Irv hope it becomes easier to type in the future when pandas changes.

PR #219, opened 2022-08-23
TYP: Add typing for HDF and synchronize API

bashtage:


bashtage:
A slightly larger set included than strict documented API. I used my knowledge of interacting with `HDFStore` when deciding what should be in it. Same for `TableIterator`

pandas-stubs/io/pytables.pyi
```diff
@@ -1,76 +1,102 @@
...
+    errors: Literal[
+        "strict",
+        "ignore",
+        "replace",
+        "surrogateescape",
+        "xmlcharrefreplace",
+        "backslashreplace",
+        "namereplace",
+    ]
+    | None = ...,
```
Dr-Irv, line 43: Not sure that `None` is acceptable for `errors` argument.

tests/test_io.py
```diff
@@ -90,3 +96,39 @@ def test_clipboard_iterator():
         assert_type(read_clipboard(iterator=False, chunksize=1), TextFileReader),
         TextFileReader,
     )
+
+
+def test_hdf():
+    with ensure_clean() as path:
+        check(assert_type(DF.to_hdf(path, "df"), None), type(None))
+        check(assert_type(read_hdf(path), Union[DataFrame, Series, Index]), DataFrame)
```
Dr-Irv, line 104: Should we check `to_hdf()` for a `Series`?  Also, I don't think there is an `Index.to_hdf()`, so is `Index` a possible result of `read_hdf()`?

tests/test_io.py
```diff
@@ -90,3 +96,39 @@ def test_clipboard_iterator():
...
+
+
+def test_hdfstore():
+    with ensure_clean() as path:
+        store = HDFStore(path, model="w")
```
twoertwein, line 109: might be good using it as a context manager - tests also whether `__enter__` and `__exit__` are declared.

tests/test_io.py
bashtage: I thought one could store indices but nope. Remvoed.

tests/test_io.py
bashtage: Added a test

pandas-stubs/io/pytables.pyi
bashtage: Fixed.

Dr-Irv (commented):
Can you add a test for writing/reading a `Series` ?

bashtage:
> Can you add a test for writing/reading a `Series` ?

Added one.

Dr-Irv (approved):
thanks @bashtage

PR #221, opened 2022-08-24
ENH: Improve SPSS typing

bashtage:


Dr-Irv (approved):
thanks @bashtage

PR #225, opened 2022-08-24
ENH: Improve parquet io

bashtage:
Synchronize with upstream
Eliminate redundant code


pandas-stubs/core/frame.pyi
```diff
@@ -275,23 +275,23 @@ class DataFrame(NDFrame, OpsMixin):
...
-        compression: Literal["snappy", "gzip", "brotli"] = ...,
-        index: _bool | None = ...,
-        partition_cols: list | None = ...,
+        path: FilePath | WriteBuffer[bytes],
+        engine: ParquetEngine = ...,
+        compression: Literal["snappy", "gzip", "brotli"] | None = ...,
+        index: bool | None = ...,
+        partition_cols: list[HashableT] | None = ...,
+        storage_options: StorageOptions = ...,
         **kwargs,
```
Dr-Irv, line 284: I see that you typed `**kwargs` as `Any` in `read_parquet()`, so why not do it here (or not type it there)?

Dr-Irv (changes requested):
One small question about consistency in typing `**kwargs`, otherwise looks OK

pandas-stubs/core/frame.pyi
bashtage: Added, thanks.

bashtage:
Should be ready once green.

Dr-Irv (approved):
thanks @bashtage

PR #226, opened 2022-08-25
ENH: Clean JSON Api

bashtage:
Cleand and synchronize JSON api


bashtage:
Still needs tests

pandas-stubs/io/json/_json.pyi
```diff
@@ -11,194 +9,89 @@ from pandas.core.series import Series
...
 def read_json(
-    path: FilePathOrBuffer,
-    orient: str | None = ...,
-    dtype=...,
-    convert_axes=...,
-    convert_dates: bool = ...,
+    path_or_buf: FilePath | ReadBuffer[str] | ReadBuffer[bytes],
+    *,
+    orient: Literal["split", "records", "index", "columns", "values", "table"]
+    | None = ...,
```
Dr-Irv, line 24: Don't see that `None` is an allowable argument here in the docs.  Which probably means the docs need to be updated.  Can you create an issue in pandas repo for that?

pandas-stubs/io/json/_json.pyi
```diff
@@ -11,194 +9,89 @@ from pandas.core.series import Series
...
+    typ: Literal["series"],
```
Dr-Irv, line 49: If `typ` is `"series"`, that limits the values of `orient` to `split`, `records` or `index`

pandas-stubs/io/json/_json.pyi
```diff
@@ -11,194 +9,89 @@ from pandas.core.series import Series
...
-    date_unit: str | None = ...,
+    date_unit: Literal["s", "ms", "us", "ns"] | None = ...,
     encoding: str | None = ...,
+    encoding_errors: Literal[
+        "strict", "ignore", "replace", "backslashreplace", "surrogateescape"
+    ]
+    | None = ...,
     lines: bool = ...,
-    chunksize: int | None = ...,
+    chunksize: int,
```
Dr-Irv, line 38: according to docs, `chunksize` can only be passed if `lines==True`, so we might as well force the type of `lines` to be `True` in this overload

Dr-Irv (changes requested):
We have tests for `json_normalize()` already, but can you add tests for `read_json()`?

pandas-stubs/io/json/_json.pyi
bashtage: `None` is special since it handles the different defaults automatically.

pandas-stubs/io/json/_json.pyi
bashtage: Fixed.

pandas-stubs/io/json/_json.pyi
bashtage: Fixed.

pandas-stubs/io/json/_json.pyi
bashtage: Issue opened.

bashtage:
> We have tests for `json_normalize()` already, but can you add tests for `read_json()`?

Added tests for read_json.

tests/test_io.py
```diff
@@ -231,3 +233,31 @@ def test_spss():
...
+    check(assert_type(DF.to_json(), str), str)
+
+
+def test_json_chunk():
+    with ensure_clean() as path:
+        check(assert_type(DF.to_json(path), None), type(None))
+        json_reader = read_json(path, chunksize=1, lines=True)
+        check(assert_type(json_reader, JsonReader), JsonReader)
+        for sub_df in json_reader:
+            check(assert_type(sub_df, Union[DataFrame, Series]), DataFrame)
```
Dr-Irv, line 262: So we can make this a little better.  If you make `JsonReader` a `Generic` that allows either `Series` or `DataFrame`, then you know that `JsonReader.__next__()` will return that type, which means that you can change the two overloads to return `JsonReader[Series]` when `typ=="series"` and `JsonReader[DataFrame]` when `typ=="frame"`

Looking in the current pandas source, they did this, but used strings rather than the types as the arguments to `Generic`

Dr-Irv (changes requested):

tests/test_io.py
bashtage: Made it generic `JsonReader[DataFrame]` or `JsonReader[Series]`.

pandas-stubs/io/json/_json.pyi
```diff
@@ -110,10 +115,10 @@ def read_json(
     storage_options: StorageOptions = ...,
 ) -> DataFrame: ...
 
-class JsonReader(abc.Iterator):
+class JsonReader(abc.Iterator, Generic[FrameSeriesStrT]):
     def read(self) -> DataFrame | Series: ...
     def close(self) -> None: ...
-    def __iter__(self) -> JsonReader: ...
+    def __iter__(self) -> JsonReader[FrameSeriesStrT]: ...
     def __next__(self) -> DataFrame | Series: ...
```
Dr-Irv, line 122: Can't this return `FrameSeriesStrT` as well?

pandas-stubs/io/json/_json.pyi
```diff
@@ -18,6 +21,8 @@ from pandas._typing import (
     StorageOptions,
 )
 
+FrameSeriesStrT = TypeVar("FrameSeriesStrT", bound=Union[Series, DataFrame])
```
Dr-Irv, line 24: Two things to investigate:
1. I think (but not sure) that you can use `NDFrameT` from `_typing.py` here and everything will work as expected, without having to create a new `TypeVar`
2. If you have to use this `TypeVar` and you can't  use `NDFrameT`,  then make it private to the module with an underscore and remove `Str` from the name, so it is `_FrameSeriesT` .

Dr-Irv (changes requested):

pandas-stubs/io/json/_json.pyi
bashtage: NDFrameT works, so simplified.

pandas-stubs/io/json/_json.pyi
bashtage: No longer relevant.

Dr-Irv (approved):
Thanks @bastage.  Use of `NDFrameT` makes it look very clean

PR #228, opened 2022-08-25
ENH: Synchronize to/read_feather

bashtage:
Synchronize and clean API


Dr-Irv (approved):
Thanks @bashtage

PR #229, opened 2022-08-25
ENH: Improve sql io api

bashtage:


bashtage:
Testing this might not be worth it.

bashtage:
I included some classes and their attributes as these are required for some of the modes. These classes are used in the docs as examples.

pandas-stubs/core/generic.pyi
```diff
@@ -163,15 +166,17 @@ class NDFrame(PandasObject, indexing.IndexingMixin):
     def to_sql(
         self,
         name: _str,
-        con,
+        con: Any,
         schema: _str | None = ...,
         if_exists: _str = ...,
```
Dr-Irv, line 171: according to docs, `if_exists` is `Literal["fail", "replace", "append"]`

pandas-stubs/core/generic.pyi
```diff
@@ -163,15 +166,17 @@ class NDFrame(PandasObject, indexing.IndexingMixin):
...
-        index_label: _str | Sequence[_str] | None = ...,
+        index_label: IndexLabel = ...,
         chunksize: int | None = ...,
-        dtype: dict | Scalar | None = ...,
-        method: _str | Callable | None = ...,
-    ) -> None: ...
+        dtype: DtypeArg | None = ...,
+        method: Literal["multi"]
+        | None
+        | Callable[[SQLTable, Any, list[str], Iterable], int | None] = ...,
```
Dr-Irv, line 178: We've been using the convention to put `None` last so can you exchange the order of `Callable` and `None` here?

pandas-stubs/io/sql.pyi
```diff
@@ -1,191 +1,139 @@
...
 
-def execute(sql, con, cur=..., params=...): ...
+@overload
 def read_sql_table(
     table_name: str,
-    con,
-    schema: str | None = ...,
-    index_col: str | Sequence[str] | None = ...,
+    con: Any,
+    schema: str | list[str] = ...,
```
Dr-Irv, line 21: `schema` can't be `list[str]` according to docs.  Also allows `None`

pandas-stubs/io/sql.pyi
```diff
@@ -1,191 +1,139 @@
...
+    columns: list[str] | None = ...,
+    chunksize: None = ...,
 ) -> DataFrame: ...
+@overload
+def read_sql_query(
+    sql: str,
+    con: Any,
+    index_col: str | list[str] | None = ...,
+    coerce_float: bool = ...,
+    params: list[str] | dict[str, str] | None = ...,
```
Dr-Irv, line 46: docs say a `tuple` is allowed here

pandas-stubs/io/sql.pyi
```diff
@@ -1,191 +1,139 @@
...
+    index_col: str | list[str] | None = ...,
     coerce_float: bool = ...,
-    params: Sequence[str] | tuple[str, ...] | Mapping[str, str] | None = ...,
-    parse_dates: Sequence[str]
-    | Mapping[str, str]
-    | Mapping[str, Mapping[str, Any]]
-    | None = ...,
-    columns: Sequence[str] = ...,
-    chunksize: int = ...,
+    params: list[str] | dict[str, str] | None = ...,
```
Dr-Irv, line 69: could be `tuple` here too

pandas-stubs/io/sql.pyi
```diff
@@ -1,191 +1,139 @@
...
-    def exists(self): ...
-    def sql_schema(self): ...
-    def create(self) -> None: ...
-    def insert_data(self): ...
-    def insert(self, chunksize=..., method=...) -> None: ...
-    def read(
-        self, coerce_float: bool = ..., parse_dates=..., columns=..., chunksize=...
-    ): ...
 
 class PandasSQL(PandasObject):
```
Dr-Irv, line 87: not in docs, so delete

pandas-stubs/io/sql.pyi
```diff
@@ -1,191 +1,139 @@
...
-        self,
-        sql,
-        index_col=...,
-        coerce_float: bool = ...,
-        parse_dates=...,
-        params=...,
-        chunksize=...,
-    ): ...
-    def to_sql(
+class SQLTable(PandasObject):
```
Dr-Irv, line 102: not in docs, so delete

Dr-Irv (changes requested):
For tests, you could test the types by surrounding the tests with `if TYPE_CHECKING`.   So then `pytest` won't execute them, but the type checking will still happen.

Dr-Irv:
One other idea.  Some of the methods have types that depend on `sqlalchemy`.  I'm pretty sure that the type checkers assume that `typeshed` is installed, which has `sqlalchemy` types defined, so it might be worth using `sqlalchemy` types where they are documented (e.g., the connection arguments).

Dr-Irv:
> One other idea. Some of the methods have types that depend on `sqlalchemy`. I'm pretty sure that the type checkers assume that `typeshed` is installed, which has `sqlalchemy` types defined, so it might be worth using `sqlalchemy` types where they are documented (e.g., the connection arguments).

Or maybe easier than worrying about `typeshed`, make an installation dependency on `sqlalchemy-stubs`

pandas-stubs/core/generic.pyi
bashtage: Fixed.

pandas-stubs/io/sql.pyi
bashtage: Fixed.

pandas-stubs/io/sql.pyi
bashtage: Fixed.

pandas-stubs/io/sql.pyi
bashtage: `SQLTable` is needed to provide typing for the `Callable` version of `to_sql`. Moreover, attributes of `SQLTable` are needed in this method. There is an example that uses `SQLTable` and some attributes.

pandas-stubs/core/generic.pyi
bashtage: Fixed.

pandas-stubs/io/sql.pyi
bashtage: Fixed.

bashtage:
> Or maybe easier than worrying about `typeshed`, make an installation dependency on `sqlalchemy-stubs`

I went with `types_SQLAlchemy` which seems to be actively maintained.  `sqlalchemy-stubs` hasn't updated in a year.

pandas-stubs/io/sql.pyi
bashtage: I think the same comment below also applies here since this is needed to type `SQLTable`, which is then used in the `Callable` version.

Dr-Irv:
Only issue is to add a typing test.  You can do
```python
def test_sql_types() -> None:
    df = # Some dataframe
    if TYPE_CHECKING:
        df.to_sql(# arguments)

        # additional tests
```

tests/test_io.py
```diff
@@ -337,3 +342,70 @@ def test_feather():
...
+                read_sql("select * from test", con=con, chunksize=1),
+                Generator[DataFrame, None, None],
+            ),
+            Generator,
+        )
+        con.close()
+
+
+def test_read_sql_table():
+    if TYPE_CHECKING:
```
Dr-Irv, line 373: It's a bit surprising that the other tests don't require `if TYPE_CHECKING` and this one does.  Do you know why?

bashtage: Because sqlite3 doesn't have read_table, so it can't be used in this test.  The other 2 work with sqlite3.

Dr-Irv: Can you add a comment about that?

pyproject.toml
```diff
@@ -51,6 +51,7 @@ openpyxl = ">=3.0.10"
 tables = ">=3.7.0"
 lxml = ">=4.7.1,<4.9.0"
 pyreadstat = ">=1.1.9"
+types_SQLAlchemy = ">=1.4.50"
```
Dr-Irv, line 54: I think this should go in the `tool.poetry.dependencies` section as it is needed for type checking

tests/test_io.py
Dr-Irv: Still would like a comment in the code as to why `TYPE_CHECKING` is used here

Dr-Irv (changes requested):

pyproject.toml
bashtage: Fixed.

bashtage:
> Still would like a comment in the code as to why TYPE_CHECKING is used here
Comment added.

tests/test_io.py
bashtage: Added a comment.

Dr-Irv (approved):
thanks @bastage

PR #230, opened 2022-08-26
Seris and DataFrame creation typing

amotzop:
* Cleaned up typing on `Series.__new__, DataFrame.__new__`.
* Allowed `DataFrame` to be created from a scalar and index/columns.

### DIFF
--- a/tests/test_frame.py
+++ b/tests/test_frame.py
@@ -42,7 +42,10 @@ def test_types_init() -> None:
         dtype=np.int8,
         copy=True,
     )
-    pd.DataFrame(0, index=[0, 1], columns=[0, 1])
+    check(
+        assert_type(pd.DataFrame(0, index=[0, 1], columns=[0, 1]), pd.DataFrame),
+        pd.DataFrame,
+    )
 
 
 def test_types_all() -> None:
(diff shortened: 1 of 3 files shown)


Dr-Irv (changes requested):
I'm a bit hesitant on these changes, because the type `Iterable[Hashable]` for `index` and `columns` might be too wide and include types that are not acceptable as an `index` or `columns`.   An example is a `set`, which will soon be deprecated.  There may be others as well. 
 In addition, the type used in `frame.pyi` for those arguments in the pandas source is `Axes`.  

Having said that, if you just want to make a change for allowing a `Scalar` argument to the `DataFrame.__new__()`, please add a test for that case.

tests/test_frame.py
```diff
@@ -42,6 +42,7 @@ def test_types_init() -> None:
         dtype=np.int8,
         copy=True,
     )
+    pd.DataFrame(0, index=[0, 1], columns=[0, 1])
```
Dr-Irv, line 45: can you do `check(assert_type(pd.DataFrame(0, index=[0, 1], columns=[0, 1]), pd.DataFrame), pd.DataFrame)` here?  

I know we don't have it above, but we're putting this in with all new tests.

Dr-Irv (changes requested):

Dr-Irv (approved):
Thanks @amotzop

PR #231, opened 2022-08-26
Add AnyArrayLike for merge on arguments

phofl:

These are all valid inputs for the on arguments


Dr-Irv (changes requested):
With any new tests, we're asking people to use `check` and `assert_type`.  So can you change these to be (for example):
```python
check(assert_type(df.merge(df2, on=pd.Series([1, 2, 3])), pd.DataFrame, pd.DataFrame)
```

phofl:
Added, wasn't aware that we should also add for argument checks

Dr-Irv (approved):
Thanks @phofl

PR #232, opened 2022-08-28
Add typing for dropna for Index and MultiIndex

phofl:


Dr-Irv (approved):
thanks @phofl

PR #233, opened 2022-08-28
Typ columns setter as Hashable

phofl:

Is there a useful way of applying check and assert_type here?


pandas-stubs/core/frame.pyi
```diff
@@ -1188,7 +1188,7 @@ class DataFrame(NDFrame, OpsMixin):
     @property
     def columns(self) -> Index: ...
     @columns.setter  # setter needs to be right next to getter; otherwise mypy complains
-    def columns(self, cols: list[_str] | Index[_str]) -> None: ...  # type: ignore[type-arg]
+    def columns(self, cols: Sequence[Hashable] | Index[_str]) -> None: ...  # type: ignore[type-arg]
```
Dr-Irv, line 1191: We shouldn't use `Sequence`, because a string will match.  Could use `AnyArrayLike | list[Hashable] | tuple[Hashable]`

Dr-Irv (changes requested):

pandas-stubs/core/frame.pyi
bashtage: Prolly list[HashableT]

phofl: Thx, added. Yes HashableT in that case.

Dr-Irv (approved):
Thanks @phofl

Issue #224, opened 2022-08-24
BaseOffset arithmetic methods (add, sub, etc.) accept a  _DatetimeT but a date is a valid parameter value


gandhis1:
An error will be given if attempting to subtract an offset from a `datetime.date`

Should be the `Union` of the two

Dr-Irv:
Please provide an example that illustrates a type checker showing an error.

gandhis1:
```
from datetime import date
from pandas.tseries.offsets import Day

x = date.today() - Day()
```

> test_case.py:4: error: No overload variant of "__sub__" of "date" matches argument type "Day"
test_case.py:4: note: Possible overload variants:
test_case.py:4: note:     `def __sub__(self, timedelta) -> date`
test_case.py:4: note:     `def __sub__(self, date) -> timedelta`
Found 1 error in 1 file (checked 1 source file)

Issue #223, opened 2022-08-24
All occurences of freq should be a Union[str, BaseOffset]


gandhis1:
The one bug I could find was the method `bdate_range` which is annotated as a `str`, but my suspicion is there are other occurrences as `freq` is a common parameter

Dr-Irv:
Please provide an example that illustrates a type checker showing an error.

gandhis1:
```
import pandas as pd
from pandas.tseries.offsets import Day, BusinessDay

x = pd.date_range("1/1/2022", "2/1/2022", freq="1D")
y = pd.date_range("1/1/2022", "2/1/2022", freq=Day())
z = pd.bdate_range("1/1/2022", "2/1/2022", freq=BusinessDay())
```

> test_case.py:6: error: Argument "freq" to "bdate_range" has incompatible type "BusinessDay"; expected "str"
Found 1 error in 1 file (checked 1 source file)

ojdo:
Sorry to piggy-back here, but this seems like the best place to ask: in a tiny script, mypy complains about

> Argument "freq" to "date_range" has incompatible type "timedelta"; expected "Union[str, BaseOffset]"

As far as I understand, and [offsets.pyi](https://github.com/pandas-dev/pandas-stubs/blob/07f9398a95d325dbf6302a321493d10aeb130aed/pandas-stubs/_libs/tslibs/offsets.pyi#L100) seems to agree, argument "freq" should also support `timedelta`. Is this a bug in mypy, or should one of the lines in this PR include another `timedelta |` somewhere?

If so, I would be happy to provide a PR. I am just very new to mypy with such heavy stubbing.

gandhis1:
The Pandas documentation for some functions that accept an offset does not seem to include `timedelta`. This interface may vary.

https://pandas.pydata.org/docs/reference/api/pandas.date_range.html

gandhis1:
I will note that this may be in error as the underlying code ultimately calls `to_offset`. So whatever that supports should be supported by the overlying function.

Dr-Irv:
> As far as I understand, and [offsets.pyi](https://github.com/pandas-dev/pandas-stubs/blob/07f9398a95d325dbf6302a321493d10aeb130aed/pandas-stubs/_libs/tslibs/offsets.pyi#L100) seems to agree, argument "freq" should also support `timedelta`. Is this a bug in mypy, or should one of the lines in this PR include another `timedelta |` somewhere?

[...]

Issue #222, opened 2022-08-24
CustomBusinessDay annotation is not complete


gandhis1:
The initializer does not appear to match as it is missing `holidays` and `calendar` https://pandas.pydata.org/docs/reference/api/pandas.tseries.offsets.CustomBusinessDay.html

Dr-Irv:
Please provide an example that illustrates a type checker showing an error.

gandhis1:
```
import pandas as pd
from pandas.tseries.offsets import CustomBusinessDay
from pandas.tseries.holiday import USFederalHolidayCalendar

x = CustomBusinessDay(holidays=USFederalHolidayCalendar())
```

> test_case.py:3: error: Skipping analyzing "pandas.tseries.holiday": found module but no type hints or library stubs
test_case.py:3: note: See https://mypy.readthedocs.io/en/stable/running_mypy.html#missing-imports
test_case.py:5: error: Unexpected keyword argument "holidays" for "CustomBusinessDay"
.../lib/python3.8/site-packages/pandas-stubs/_libs/tslibs/offsets.pyi:203: note: "CustomBusinessDay" defined here

Dr-Irv:
That's not valid code, but this is and will be fixed in a PR:
```python
from pandas.tseries.offsets import CustomBusinessDay
from pandas.tseries.holiday import USFederalHolidayCalendar

x = CustomBusinessDay(calendar=USFederalHolidayCalendar())
```

PR #235, opened 2022-08-28
fix various issues with offsets

Dr-Irv:
- [x] Closes 
  -  #222 
  - #223 
  - #224 
- [x] Tests added
  - `test_timefuncs.py: test_some_offsets()`


pandas-stubs/core/indexes/datetimes.pyi
```diff
@@ -76,25 +83,26 @@ class DatetimeIndex(DatetimeTimedeltaMixin, DatetimeIndexProperties):
...
-    periods=...,
-    freq=...,
-    tz=...,
-    normalize=...,
-    name=...,
-    closed=...,
+    start: str | DatetimeLike | None = ...,
+    end: str | DatetimeLike | None = ...,
+    periods: int | None = ...,
+    freq: str | BaseOffset | Tick = ...,
```
twoertwein, line 89: I think Tick is a sub-class of BaseOffset?

Dr-Irv: Yes, good catch.  Fixed in next commit

twoertwein:
Thank you @Dr-Irv !

PR #236, opened 2022-08-30
ENH: Correct typing for read/to_html

bashtage:


pandas-stubs/core/frame.pyi
```diff
@@ -326,23 +327,39 @@ class DataFrame(NDFrame, OpsMixin):
...
+            "unset",
+        ]
+        | None = ...,
         max_rows: int | None = ...,
         max_cols: int | None = ...,
         show_dimensions: _bool = ...,
         decimal: _str = ...,
         bold_rows: _bool = ...,
-        classes: _str | list | tuple | None = ...,
+        classes: _str | Sequence[str] | None = ...,
```
Dr-Irv, line 362: `Sequence[str]` will match `_str` so the `_str` is redundant.

pandas-stubs/core/frame.pyi
```diff
@@ -326,23 +327,39 @@ class DataFrame(NDFrame, OpsMixin):
     @overload
     def to_html(
         self,
-        buf: FilePathOrBuffer | None,
-        columns: Sequence[_str] | None = ...,
-        col_space: int | list[int] | dict[_str | int, int] | None = ...,
+        buf: FilePath | WriteBuffer[str],
+        columns: Sequence[HashableT] | None = ...,
```
Dr-Irv, line 331: This might be too broad (as was the previous version).  Will match a single string.  Use `list[HashableT] | tuple[HashableT,...]`

pandas-stubs/io/html.pyi
```diff
@@ -1,46 +1,46 @@
...
-    match: str = ...,
+    io: FilePath | ReadBuffer[str],
+    match: str | Pattern = ...,
     flavor: str | None = ...,
     header: int | Sequence[int] | None = ...,
-    index_col: int | Sequence[Any] | None = ...,
-    skiprows: int | Sequence[Any] | slice | None = ...,
+    index_col: int | Sequence[int] | list[HashableT] | None = ...,
+    skiprows: int | Sequence[int] | slice | None = ...,
     attrs: Mapping[str, str] | None = ...,
```
Dr-Irv, line 26: Is `Mapping` too wide?  Should we just use `dict` ?

pandas-stubs/io/html.pyi
```diff
@@ -1,46 +1,46 @@
...
     parse_dates: bool
-    | Sequence[int | str | Sequence[int | str]]
-    | dict[str, Sequence[int | str]] = ...,
+    | Sequence[int]
+    | list[HashableT]  # Cannot be Sequence[Hashable] to prevent str
+    | Sequence[Sequence[int]]
+    | Sequence[Sequence[Hashable]]
```
Dr-Irv, line 31: I think `Sequence[Sequence[int]]` is covered by `Sequence[Sequence[Hashable]]`

Dr-Irv (changes requested):

pandas-stubs/io/html.pyi
bashtage: Fixed.

pandas-stubs/io/html.pyi
bashtage: Fixed.

pandas-stubs/core/frame.pyi
bashtage: Fixed.

pandas-stubs/core/frame.pyi
bashtage: Fixed.

Dr-Irv:
So I would approve this, but concerned about the CI failures on Windows.  I can't figure out why that's happening all of a sudden.  Error message shows that `loguru` is not installed, but if you look at the complete log, it *is* installed.

bashtage:
Seems to be some sort of systematic failure.

Dr-Irv (approved):
thanks @bashtage

PR #238, opened 2022-08-31
run pytest against nightly

### LINKED ISSUE #218: TST: Add an unfailable CI run against pandas main
It would probably be a good idea to add a CI run against pandas main.  Ideally, this would use nightly wheels to save on time.  Not sure how easy this is with poetry and the current CI setup.

twoertwein:
- [x] Closes #218

Not sure whether it will succeed if pytest fails.

### DIFF
--- a/.github/setup/action.yaml
+++ b/.github/setup/action.yaml
@@ -1,0 +1,35 @@
+name: Install project dependencies
+
+inputs:
+  python-version:
+    required: true
+  os:
+    required: true
+
+runs:
+  using: composite
+  steps:
+
+    - name: Set up Python
+      uses: actions/setup-python@v4
+      with:
+        python-version: ${{ inputs.python-version }}
+
+    - name: Install Poetry
+      shell: bash
+      run: python -m pip install poetry
+
+    - name: Determine poetry version
+      shell: bash
+      run: echo "::set-output name=VERSION::$(poetry --version)"
+      id: poetry_version
+
+    - name: Cache poetry.lock
+      uses: actions/cache@v3
+      with:
+        path: poetry.lock
+        key: ${{ inputs.os }}-${{ inputs.python-version }}-poetry-${{ steps.poetry_version.outputs.VERSION }}-${{ hashFiles('pyproject.toml') }}
+
+    - name: Install project dependencies
+      shell: bash
+      run: poetry install -vvv --no-root
--- a/.github/setup/action.yaml
+++ b/.github/setup/action.yaml
@@ -16,11 +16,9 @@ runs:
         python-version: ${{ inputs.python-version }}
 
     - name: Install Poetry
-      shell: bash
       run: python -m pip install poetry
 
     - name: Determine poetry version
-      shell: bash
       run: echo "::set-output name=VERSION::$(poetry --version)"
       id: poetry_version
 
@@ -31,5 +29,4 @@ runs:
         key: ${{ inputs.os }}-${{ inputs.python-version }}-poetry-${{ steps.poetry_version.outputs.VERSION }}-${{ hashFiles('pyproject.toml') }}
 
     - name: Install project dependencies
-      shell: bash
       run: poetry install -vvv --no-root
--- a/.github/setup/action.yaml
+++ b/.github/setup/action.yaml
@@ -16,9 +16,11 @@ runs:
         python-version: ${{ inputs.python-version }}
 
     - name: Install Poetry
+      shell: bash
       run: python -m pip install poetry
 
     - name: Determine poetry version
+      shell: bash
       run: echo "::set-output name=VERSION::$(poetry --version)"
       id: poetry_version
 
@@ -29,4 +31,5 @@ runs:
         key: ${{ inputs.os }}-${{ inputs.python-version }}-poetry-${{ steps.poetry_version.outputs.VERSION }}-${{ hashFiles('pyproject.toml') }}
 
     - name: Install project dependencies
+      shell: bash
       run: poetry install -vvv --no-root
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -3,7 +3,7 @@ name: 'Test'
 on:
   [push, pull_request, workflow_dispatch]
 jobs:
-  test:
+  released:
     runs-on: ${{ matrix.os }}
     timeout-minutes: 10
     strategy:
@@ -13,34 +13,17 @@ jobs:
         python-version: ['3.8', '3.9', '3.10']
 
     steps:
-
     - uses: actions/checkout@v3
 
-    - name: Set up Python ${{ matrix.python-version }}
-      uses: actions/setup-python@v4
-      with:
-        python-version: ${{ matrix.python-version }}
-
-    - name: Install Poetry
-      run: pip install poetry
-
-    - name: Determine poetry version
-      run: echo "::set-output name=VERSION::$(poetry --version)"
-      id: poetry_version
-
-    - name: Cache poetry.lock
-      uses: actions/cache@v3
-      with:
-        path: poetry.lock
-        key: ${{ matrix.os }}-${{ matrix.python-version }}-poetry-${{ steps.poetry_version.outputs.VERSION }}-${{ hashFiles('pyproject.toml') }}
-
     - name: Install project dependencies
-      run: poetry install -vvv --no-root
+      uses: ./.github/setup
+      with:
+        os: ${{ matrix.os }}
+        python-version: ${{ matrix.python-version }}
 
     - name: Show poetry python location (Windows)
       shell: pwsh
-      run: |
-        poetry run where python
+      run: poetry run where python
       if: matrix.os == 'windows-latest'
 
     - name: Run mypy on 'tests' (using the local stubs) and on the local stubs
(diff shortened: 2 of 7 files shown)


twoertwein:
Might be good to file an issue with pyarrow to handle pandas nightly

> ImportError: pyarrow requires pandas 0.23.0 or above, pandas 0+unknown is installed

twoertwein:
> Might be good to file an issue with pyarrow to handle pandas nightly
> 
> > ImportError: pyarrow requires pandas 0.23.0 or above, pandas 0+unknown is installed

I guess only an issue on older pythons.....

Dr-Irv:
Looking at this, I don't see where you install the nightly build, but maybe I'm missing something or you're still working on that part??

twoertwein:
> Looking at this, I don't see where you install the nightly build, but maybe I'm missing something or you're still working on that part??

This install (and uninstalls nightly):
https://github.com/pandas-dev/pandas-stubs/blob/cc6ea9b7d65d6c1e0d039742d2a7f399396c2bae/scripts/test/__init__.py#L52

The issue is that I don't yet know how to make github ignore errors.

bashtage:
Add `"--upgrade","--use-deprecated=legacy-resolver"` to the pip command that installs the nightly.

Dr-Irv:
> > Looking at this, I don't see where you install the nightly build, but maybe I'm missing something or you're still working on that part??
> 
> This install (and uninstalls nightly):
> 
> https://github.com/pandas-dev/pandas-stubs/blob/cc6ea9b7d65d6c1e0d039742d2a7f399396c2bae/scripts/test/__init__.py#L52
> 
I think it was here, which I didn't remember was in the code:
https://github.com/pandas-dev/pandas-stubs/blob/80f561fc73f387f81f3876eeef4fe3bae82dae18/scripts/test/run.py#L83

twoertwein:
> Add `"--upgrade","--use-deprecated=legacy-resolver"` to the pip command that installs the nightly.

Thank you! I simplify the install/uninstall. Why is `--use-deprecated=legacy-resolver` needed?

bashtage:
It has to do with how pip check versions. Legacy takes the latest. Without this it will download every whl file to get Metadata before deciding on the best fit.

.github/workflows/test.yml
```diff
@@ -51,3 +51,8 @@ jobs:
 
     - name: Install pandas-stubs and run tests on the installed stubs
       run: poetry run poe test_dist
+
+    - if: matrix.python-version == '3.10' && matrix.os == 'ubuntu-latest'
+      name: Run pytest (against pandas nightly)
+      continue-on-error: true
```
twoertwein, line 57: this seems to hide errors but I don't think there is a warning/yellow option

Dr-Irv: When we were doing testing in the microsoft stubs project, we wanted to ignore warnings from pyright, and Graham Wheeler (owner of the repo) modified some github action to consume the pyright results and ignore the warnings.

Maybe there is an equivalent with pytest???

twoertwein: Warnings are easier to control (--disable-warnings and with filterwarnings). In our case, we want to see the warnings. I didn't find an option to always return 0.

Dr-Irv: That's why he created the action to consume errors and warnings (it was from pyright), and then return 0 if there were only warnings and no errors.

bashtage: I think you need condition: succeededOrFailed() on later steps. This is how we do never fail prerelease on statsmodels.

bashtage: We only have it on one step after pytest so might not be a good idea here.

bashtage: You can also have || true after the pytest command to make a never fail.

twoertwein: > You can also have || true after the pytest command to make a never fail.

That would be the least invasive option, but at least the following doesn't work:

`run: poetry run poe pytest --nightly ; exit 1 || true`

Dr-Irv: Since `pytest` is being run from a python script, couldn't you get or parse the results from running pytest within python, and then determine the exit status that gets returned?

twoertwein: I just did something like that :)

twoertwein:
I'm surprised we get these warnings:

>    /home/runner/work/pandas-stubs/pandas-stubs/tests/test_frame.py:290: FutureWarning: The 'inplace' keyword in DataFrame.set_index is deprecated and will be removed in a future version. Use `df = df.set_index(..., copy=False)` instead.
>     res6: None = df.set_index("col1", inplace=True)

I thought that change is already reverted on main

bashtage:
Reopen to trigger CI

.github/workflows/test.yml
```diff
@@ -57,3 +57,7 @@ jobs:
 
     - name: Install pandas-stubs and run tests on the installed stubs
       run: poetry run poe test_dist
+
+    - if: matrix.python-version == '3.10' && matrix.os == 'ubuntu-latest'
```
Dr-Irv, line 61: probably worth adding a comment here that says that this test has to be last, otherwise any additional test below it will use the nightly build

Dr-Irv (commented):
So while we don't exit on failure if the test against nightly builds fails, if there was a failure, would it show up red in the CI report?

twoertwein:
> So while we don't exit on failure if the test against nightly builds fails, if there was a failure, would it show up red in the CI report?

It will always be "green". As far as I know there is no "yellow".

.github/workflows/test.yml
twoertwein: Could probably run it on all platforms for 3.10.

twoertwein:
There might be a way to mark the CI run as skipped if it fails. Then we would have a visual indicator.

Dr-Irv:
> There might be a way to mark the CI run as skipped if it fails. Then we would have a visual indicator.

Without any visual indicator, I'm not sure how we would ever notice that something was revealed via this test.  Can you investigate that?

twoertwein:
I didn't find a way to visually indicate an error except by failing (failing now).

I also made pytest error on any warnings. This fails currently on nightly but #251 might fix that

Dr-Irv:
So I like what the PR is doing, but it is raising some questions in how we want to manage the project.  Opinions wanted!
1. If we see a failure in the job due to the nightly build, should we expect users to fix the code before accepting the PR?
2. If the answer to (1) is yes, then it's easy to manage.
3. If the answer to (2) is no, then that brings up a couple of questions (4) and (5)
4.  Should we create a separate job that only tests the nightly build, and then accept PR's where all the other tests pass?
5. Should we manage the project by having a "main_release" branch, corresponding to what we release weekly as we improve the stubs for the released version of pandas, and a "nightly" branch, corresponding to what is developed to work for a future version of pandas?  (Would have to decide which one of those is `main`)

twoertwein:
Could also make pytest-nightly run in its own workflow, then it is clear that ubuntu+3.10 did not fail for any other reason.

twoertwein:
>     1. If we see a failure in the job due to the nightly build, should we expect users to fix the code before accepting the PR?

If pandas nightly causes all new PRs to have errors - I would not expect people to fix that as part of their PR. But if people want to add new annotations that trigger future/deprecation warnings, I would be inclined to request changes :)

>     4. Should we create a separate job that only tests the nightly build, and then accept PR's where all the other tests pass?

In either case, it probably makes sense to have a separate workflow just for nightly.

>     5. Should we manage the project by having a "main_release" branch, corresponding to what we release weekly as we improve the stubs for the released version of pandas, and a "nightly" branch, corresponding to what is developed to work for a future version of pandas?  (Would have to decide which one of those is `main`)

I really like the current simplicity of pandas-stubs. If there are cases where we do not (yet) want to be in-line with nightly, we could simply have a pytest skip decorator based on the pandas version.

Dr-Irv:
> In either case, it probably makes sense to have a separate workflow just for nightly.
> 

So let's do that, and we'll operate on the principle that if the nightly fails, and everything else passes, then we'll accept the PR.

> I really like the current simplicity of pandas-stubs. 

Me too!!!

> If there are cases where we do not (yet) want to be in-line with nightly, we could simply have a pytest skip decorator based on the pandas version.

Good idea

twoertwein:
I will mark this PR as draft and then debug sharing steps between the general checks and pytest nightly.

.github/setup/action.yaml
```diff
@@ -0,0 +1,35 @@
...
+  using: composite
+  steps:
+
+    - name: Set up Python
+      uses: actions/setup-python@v4
+      with:
+        python-version: ${{ inputs.python-version }}
+
+    - name: Install Poetry
+      shell: bash
```
twoertwein, line 19: don't ask me why it needs a shell and why bash also works on windows. Just mirrored a similar file in the pandas repository :)

twoertwein:
Might be worth waiting for #251 to keep the nice green check mark :)

.github/setup/action.yaml
Dr-Irv: I don't think you need the shell here.  We aren't using it in `test.yaml`.  You only need the shell if you want to execute a unix command that is not available on Windows. 

Can you try deleting the references to the shell and see if it still works?

Dr-Irv (changes requested):

.github/setup/action.yaml
twoertwein: I think that is a difference between workflow files and "composite actions". Fails without it.

bashtage:
Can you make test/nightly orange for failure rather than red?

twoertwein:
> Can you make test/nightly orange for failure rather than red?

I did not find an option for that.

bashtage:
statsmodels has it on their pip pre run if pytest fails. If you merge this I'll see if I can figure out what is needed in a later PR.

.github/setup/action.yaml
Dr-Irv: > I think that is a difference between workflow files and "composite actions". Fails without it.

seems to be

Dr-Irv (approved):
thanks @twoertwein

Dr-Irv:
I think there is a way to mark a run as "allowed to fail".  At least I remember seeing it somewhere.

Hopefully @bashtage can figure it out!

twoertwein:
If it is not possible to make it warn-only but not fail, an option could be to run it only on main (after PRs are merged) or the reverse (only for PRs but not on main): then we have some more green :)

bashtage:
> If it is not possible to make it warn-only but not fail, an option could be to run it only on main (after PRs are merged) or the reverse (only for PRs but not on main): then we have some more green :)

Looks pretty hopeless for now.  Seems GHA doesn't have a yellow/warn mode despite the similarities to Azure (the I suspect that it runs on Azure).

bashtage:
I would always pass the run and have it there in CI so that interested parties could easily see what doing on.

The alternative would be to add 1 Azure run which could be configured to be yellow.  If you want the Azure run, I can set it up.

twoertwein:
Is there a way to customize the `failing/successful after n seconds` message:
image

We could then let it always succeed but at least change the message to `failed` :)

PR #239, opened 2022-08-31
Io excel

bashtage:


bashtage:
No new tests yet. Existing pass except for one buggy one, which was fixed.

pandas-stubs/io/excel/_base.pyi
```diff
@@ -1,188 +1,247 @@
...
     index_col: int | Sequence[int] | None = ...,
-    usecols: int | str | Sequence[int | str | Callable] | None = ...,
-    squeeze: bool = ...,
-    dtype: str | dict[str, Any] | Dtype = ...,
-    engine: str | None = ...,
-    converters: dict[int | str, Callable] | None = ...,
-    true_values: Sequence[Scalar] | None = ...,
-    false_values: Sequence[Scalar] | None = ...,
-    skiprows: Sequence[int] | int | Callable | None = ...,
+    usecols: str | Sequence[int] | Sequence[str] | Callable[[str], bool] | None = ...,
```
Dr-Irv, line 44: `Sequence[str]` will match `str` so no need to include both.

pandas-stubs/io/excel/_base.pyi
```diff
@@ -1,188 +1,247 @@
...
+    dtype: str | Dtype | dict[str, str | Dtype] | None = ...,
+    engine: Literal["xlrd", "openpyxl", "odf", "pyxlsb"] | None = ...,
+    converters: dict[int | str, Callable[[object], object]] | None = ...,
+    true_values: Iterable[Hashable] | None = ...,
+    false_values: Iterable[Hashable] | None = ...,
```
Dr-Irv, line 49: I think `Iterable[Hashable]` is too wide.  Docs say "list", so I think `list[str | int]` would be sufficient.

Dr-Irv (changes requested):

Dr-Irv:
> Is it OK to include optional dependencies in the typing stubs? Will this cause problems downstream?

The dependencies you added are there just for testing.  If there are typing stubs for the optional dependencies, then I think we should include them as required dependencies for pandas-stubs.

pandas-stubs/io/excel/_base.pyi
bashtage: The iterable comes from pandas main. I also tested using an iterator and it worked as expected.

Dr-Irv (commented):
I'm fine with all the changes, but wondering if we should also move the `test_read_excel()` function in `test_frame.py` to `test_io.py` .  Could do this as part of this PR or as a follow-up PR.

Dr-Irv:
Also need to resolve conflicts

pandas-stubs/io/excel/_base.pyi
bashtage: Fixed.

Dr-Irv (approved):
thanks @bashtage

Dr-Irv:
@bashtage can you resolve conflicts, then ping when green.

bashtage:
Green, ignoring the pre -test fail

Dr-Irv:
UGH.  I tried to resolve the conflicts via the web editor, and messed up black formatting.  Can you pull what I did and fix it?

Dr-Irv (approved):
thanks @bashtage

PR #242, opened 2022-09-01
Fix CI for Windows

### LINKED ISSUE #240: CI failing on Windows.
**Describe the bug**
loguru is breaking windows

**To Reproduce**
Run CI on windows

Should use standard logging for now since it is a minimal dependency.

Dr-Irv:
- [x] Closes #240 

### Summary

- Uses action to install poetry
- For the test of the distribution, avoids poetry because Windows CI isn't installing it in the right place
- Adds a couple of missing PYI files that pyright said were missing

One downside (at least for me).  The setting 
```
[tool.poe.executor]
type = "virtualenv"
```
makes a local test with poetry not work inside of a conda environment.

But at least the CI is working.


bashtage:
I was going to merge this since green.  Fix looks fine, although the source'ing is annoying. but maybe necessary. Could you instead call `poetry shell` before the `poe` commands.

I didn't merge because looks like you accidentally got some pyi changes in that aren't needed.

pyproject.toml
```diff
@@ -1,6 +1,7 @@
 [tool.poetry]
 name = "pandas-stubs"
-version = "1.4.3.220829"
+# Use the big version number so in CI, it will use the built wheel
+version = "1.4.3.999999"
```
twoertwein, line 4: Why is that needed? To invalidate the cache?

This should already pick the newest wheel:
https://github.com/pandas-dev/pandas-stubs/blob/5b70b46007c90e9cd76d9169e4f9d9dcd038d218/scripts/test/run.py#L49

.github/workflows/test.yml
```diff
@@ -22,32 +26,47 @@ jobs:
...
+        source $VENV
+        poetry run poe pytest
 
     - if: matrix.python-version == '3.8' && matrix.os == 'ubuntu-latest'
       uses: pre-commit/action@v3.0.0
 
     - name: Install pandas-stubs and run tests on the installed stubs
-      run: poetry run poe test_dist
+      run: |
+        source $VENV
```
twoertwein, line 64: In the above steps we can use `poe ...`, why can't we use it here?

twoertwein:
So the issue is that we have at least two python executables and for some reason, poetry/windows CI uses both of them in an inconsistent manner?

Is it possible to remove one of them from the PATH (or delete it) so that poetry/CI is not confused by it?

Dr-Irv:
> I didn't merge because looks like you accidentally got some pyi changes in that aren't needed.

That was intentional, as `pyright` was reporting that some imports were missing.

.github/workflows/test.yml
```diff
@@ -22,32 +26,47 @@ jobs:
...
-        key: ${{ matrix.os }}-${{ matrix.python-version }}-poetry-${{ steps.poetry_version.outputs.VERSION }}-${{ hashFiles('pyproject.toml') }}
+        virtualenvs-create: true
+        virtualenvs-in-project: true
+  
+    - name: Load cached venv
+      id: cached-pip-wheels
+      uses: actions/cache@v2
+      with:
+        path: ~/.cache
+        key: venv-${{ runner.os }}-${{ steps.setup-python.outputs.python-version }}-${{ hashFiles('**/poetry.lock') }}
```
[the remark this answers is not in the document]
Dr-Irv: I picked this up from https://github.com/snok/install-poetry#running-on-windows

There are comments there that say that caching doesn't work on Windows.

pyproject.toml
Dr-Irv: The problem here is that if I build and install the wheel from the run script, it gets installed into the wrong python environment.  So now with `test.yaml` doing a direct build of the wheel and installation, I have to force it to install the newly built wheel, which requires a higher version number than what is available on pypi.

.github/workflows/test.yml
Dr-Irv: As mentioned above, if I use the run script, it installs the wheel into the "system" environment, not poetry's virtual environment.

Dr-Irv:
> So the issue is that we have at least two python executables and for some reason, poetry/windows CI uses both of them in an inconsistent manner?
> 
> Is it possible to remove one of them from the PATH (or delete it) so that poetry/CI is not confused by it?

I don't know.  Prior to a few days ago, we always had 2 environments on Windows and things worked.  With the `setup-python` action, it creates a "system" python, and then poetry creates a virtual one.  I created an issue in the poetry repo thinking that the upgrade from poetry 1.1.15 to 1.2 was the issue, but the poetry people found a failed CI run on Windows using 1.1.15.

I'm now thinking that it is the `setup-python` action that messed things up.

Dr-Irv:
> I was going to merge this since green. Fix looks fine, although the source'ing is annoying. but maybe necessary. Could you instead call `poetry shell` before the `poe` commands.

See https://github.com/snok/install-poetry#running-on-windows as to why we have to do the sourcing.

pyproject.toml
twoertwein: > which requires a higher version number than what is available on pypi.

So pip is preferring the pypi.org wheel over the local wheel? There must be an option to ignore the remote wheel.

Could also simply bump the version number after each release so that the CI works but as soon as we forget it pip will install the wrong pandas-stubs - seems a bit fragile.

.github/workflows/test.yml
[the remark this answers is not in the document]
Dr-Irv: Don't know.  After spending hours on this last night, I was just happy to get it working.

pyproject.toml
Dr-Irv: Issue here is that we need this to work for PR's, which means that contributors have to change the version number.

In the `test.yaml`, I am now doing this:
```
python -m pip install --find-links=./dist --force-reinstall pandas-stubs
```
The `--find-links=./dist` makes it so that pip will look in the `dist` directory.  On Windows, you can't do `pip install dist/*`, so this is the workaround for that.

Dr-Irv:
closing in favor of #249

PR #243, opened 2022-09-01
ENH: to_csv and read_{fwf/csv/table} type improvements

bashtage:


bashtage:
@Dr-Irv Looks like the loguru stuff fixed itself?

bashtage:
Maybe not fixed, but random.  Might depend on the CI worker node.

Dr-Irv:
> Maybe not fixed, but random. Might depend on the CI worker node.

Yes, this randomness is part of the problem.  I don't know what to do about it.

pandas-stubs/core/generic.pyi
```diff
@@ -204,17 +205,19 @@ class NDFrame(PandasObject, indexing.IndexingMixin):
...
-        columns: Sequence[Hashable] | None = ...,
+        float_format: _str | Callable | None = ...,
+        columns: list[HashableT] | None = ...,
         header: _bool | list[_str] = ...,
         index: _bool = ...,
-        index_label: _bool | _str | Sequence[Hashable] | None = ...,
-        mode: _str = ...,
+        index_label: Literal[False] | _str | list[HashableT] | None = ...,
+        mode: Literal[
+            "a", "w", "x", "at", "wt", "xt", "ab", "wb", "xb", "w+", "w+b", "a+", "a+b"
```
Dr-Irv, line 217: Maybe create a type for this?

Dr-Irv (changes requested):
My only concern here is that I can't tell if you modified `read_csv()` or `read_table()` because they moved from one file to the other.

If you did change `read_csv()` or `read_table()`, can you do one PR that just modifies their signatures, but keeping them in the same file, so it is easy for me to compare what you changed, and then a second PR that moves them from where they are in `main` now to the new place you put them.

pandas-stubs/core/generic.pyi
bashtage: Fixed in other PR.

bashtage:
> If you did change `read_csv()` or `read_table()`, can you do one PR that just modifies their signatures, but keeping them in the same file, so it is easy for me to compare what you changed, and then a second PR that moves them from where they are in `main` now to the new place you put them.

Created #248 in place of this

PR #246, opened 2022-09-01
MAINT: Show python info on Windows

bashtage:

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -37,6 +37,12 @@ jobs:
     - name: Install project dependencies
       run: poetry install -vvv --no-root
 
+    - name: Show poetry python location (Windows)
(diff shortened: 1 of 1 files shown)


bashtage:
@Dr-Irv @twoertwein I think it would be good to get this in - it might help diagnose cache issues since we could see if the location of the virtualenv was changing, and it somehow windows was missing the location of the venv python.exe.

Dr-Irv (approved):
Thanks @bashtage

Dr-Irv:
Might be worth trying this without the pinning of the poetry version.  I think it worked because it picked up an "old" cache.

twoertwein:
Might also be good to make the cache key dependent on `poetry run python --version` instead of using the string from the matrix setup.

PR #248, opened 2022-09-01
ENH: Improve to_csv and from_{fwf,csv,table}

bashtage:


pandas-stubs/io/parsers/base_parser.pyi
```diff
@@ -0,0 +1,35 @@
+from typing import Any
+
+from pandas._typing import (
+    ArrayLike as ArrayLike,
+    DtypeArg as DtypeArg,
+    Scalar as Scalar,
+)
+
+class ParserBase:
```
Dr-Irv, line 9: I don't think we need this file

pandas-stubs/io/parsers/readers.pyi
```diff
@@ -0,0 +1,438 @@
...
+    npt,
+)
+
+from pandas.io.common import IOHandles
+
+@overload
+def read_csv(
+    filepath_or_buffer: FilePath | ReadCsvBuffer[bytes] | ReadCsvBuffer[str],
+    *,
+    sep: str | None | lib.NoDefault = ...,
```
Dr-Irv, line 33: We shouldn't use `lib.NoDefault` as an option in the stubs.  Users can't use it.  It's there for convenience in the internal typing.

pandas-stubs/_libs/lib.pyi
```diff
@@ -1,3 +1,8 @@
+from enum import Enum
+
 no_default = None
 
+class NoDefault(Enum):
```
Dr-Irv, line 5: remove this - it shouldn't be used by users.  It's acting as a sentinel for internal type checking

pandas-stubs/io/parsers/readers.pyi
```diff
@@ -0,0 +1,438 @@
...
+    names: list[str] | None | lib.NoDefault = ...,
+    index_col: int | str | Sequence[str | int] | Literal[False] | None = ...,
+    usecols: list[str]
+    | tuple[str, ...]
+    | Sequence[int]
+    | Series
+    | Index
+    | npt.NDArray
+    | Callable[[str], bool]
+    | None = ...,
```
Dr-Irv, line 45: may be missing `mangle_dupe_cols` here.  It's in the 1.4.4 docs

pandas-stubs/io/parsers/readers.pyi
```diff
@@ -0,0 +1,438 @@
...
+    dayfirst: bool = ...,
+    cache_dates: bool = ...,
+    iterator: bool = ...,
+    chunksize: int,
+    compression: CompressionOptions = ...,
+    thousands: str | None = ...,
+    decimal: str = ...,
+    lineterminator: str | None = ...,
+    quotechar: str = ...,
+    quoting: int = ...,
```
Dr-Irv, line 140: Same as in `to_csv()`, we can change `quoting`

pandas-stubs/io/parsers/readers.pyi
```diff
@@ -0,0 +1,438 @@
...
+    | Callable[[list[str]], list[str] | None] = ...,
+    delim_whitespace: bool = ...,
+    low_memory: bool = ...,
+    memory_map: bool = ...,
+    float_precision: Literal["high", "legacy", "round_trip"] | None = ...,
+    storage_options: StorageOptions | None = ...,
+) -> DataFrame: ...
+def read_fwf(
+    filepath_or_buffer: FilePath | ReadCsvBuffer[bytes] | ReadCsvBuffer[str],
+    colspecs: Sequence[tuple[int, int]] | str | None = ...,
```
Dr-Irv, line 409: Instead of `str`, should be `Literal["infer"]`

pandas-stubs/io/parsers/readers.pyi
```diff
@@ -0,0 +1,438 @@
...
+    widths: Sequence[int] | None = ...,
+    infer_nrows: int = ...,
+    **kwds: Any,
+) -> DataFrame | TextFileReader: ...
```
Dr-Irv, line 413: I am hazarding a guess here, but I'm guessing if you pass `chunksize` and/or `iterator` then it returns `TextFileReader` and if you don't, you get `DataFrame`.

So if `chunksize` and/or `iterator` are one of the `kwargs` that can be passed to `TextFileReader`, we might as well do the overload with those specific arguments as is done for `read_csv()` and `read_table()`.

Dr-Irv (changes requested):

pandas-stubs/core/generic.pyi
```diff
@@ -205,17 +207,17 @@ class NDFrame(PandasObject, indexing.IndexingMixin):
...
         header: _bool | list[_str] = ...,
         index: _bool = ...,
-        index_label: _bool | _str | Sequence[Hashable] | None = ...,
-        mode: _str = ...,
+        index_label: Literal[False] | _str | list[HashableT] | None = ...,
+        mode: FileWriteMode = ...,
         encoding: _str | None = ...,
-        compression: _str | Mapping[_str, _str] = ...,
+        compression: CompressionOptions = ...,
         quoting: int | None = ...,
```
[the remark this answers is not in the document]
bashtage: Says those values are not valid types for a literal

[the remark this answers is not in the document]
bashtage: Went with [0,1, 2, 3] which works.

pandas-stubs/_libs/lib.pyi
bashtage: Fixed.

pandas-stubs/io/parsers/readers.pyi
bashtage: Fixed.

pandas-stubs/io/parsers/readers.pyi
bashtage: This has been deprecated in 1.5.  Though we are so close might as well drop it now given how close 1.5 is.

pandas-stubs/io/parsers/readers.pyi
```diff
@@ -137,7 +138,7 @@ def read_csv(
     decimal: str = ...,
     lineterminator: str | None = ...,
     quotechar: str = ...,
-    quoting: int = ...,
+    quoting: CSVQuoting | None = ...,
```
Dr-Irv, line 141: I don't think `None` is valid here.

pandas-stubs/io/parsers/readers.pyi
Dr-Irv: This is a tough call, because I don't think there is an alternate available if you were still using the feature.  Docs say "Not implemented, and a new argument to specify the pattern for the names of duplicated columns will be added instead", meaning there is no workaround.  I think we should leave it in as I will be doing a release with 1.4.4 where it is supported, and we can remove it once 1.5 is released.

pandas-stubs/io/parsers/readers.pyi
Dr-Irv: This still has to be done.

Dr-Irv (changes requested):
still some open items.  Have to resolve conflicts, make it green.

pandas-stubs/io/parsers/readers.pyi
bashtage: readded this.

pandas-stubs/io/parsers/readers.pyi
bashtage: Fixed this.

Dr-Irv (approved):
thanks @bashtage

PR #250, opened 2022-09-01
unpin poetry

Dr-Irv:
We had pinned poetry to 1.1.15 when we were testing the Windows CI issues.  This removes that.


bashtage:
LGTM, and all green,.

PR #251, opened 2022-09-02
BUG: Correct read_stata

bashtage:
Update and silence for future warnings from pandas/main


tests/test_frame.py
```diff
@@ -1505,27 +1504,31 @@ def test_frame_scalars_slice() -> None:
 
     # Note: bool_ cannot be tested since the index is object and pandas does not
     # support boolean access using loc except when the index is boolean
-    check(assert_type(df.loc[str_], pd.Series), pd.Series)
+    with pytest.warns(FutureWarning, match="Comparison of Timestamp with datetime"):
```
twoertwein, line 1507: If we want to treat deprecation/future-warnings on main the same we do for the released pandas versions, we could change the type annotations to make mypy/pyright fail in these cases and make them negative tests.

pandas-stubs/core/frame.pyi
```diff
@@ -633,37 +633,6 @@ class DataFrame(NDFrame, OpsMixin):
         axis: AxisType = ...,
         fill_value: Hashable | None = ...,
     ) -> DataFrame: ...
-    @overload
-    def set_index(
```
twoertwein, line 637: I think this should wait until 1.5 is released. 1.5 deprecates `inplace` and adds the new `copy` argument.

Might be worth writing a `@skip_on_nightly` decorator for the tests.

Dr-Irv: Other option is we start maintaining a fork related to 1.5 changes.

twoertwein (approved):
`set_index` isn't ideal, but letting it always return a `DataFrame` is consistent with how it will be in the near future.

Dr-Irv:
I'm holding off on merging this until 1.5 is released.

twoertwein:
Definitely good to wait for 1.5 - especially since the `set_index` behavior might get reverted before 1.5 releases https://github.com/pandas-dev/pandas/issues/48141#issuecomment-1238505678

twoertwein:
Since https://github.com/pandas-dev/pandas/pull/48417 is merged, we can remove the `set_index` changes and probably merge this PR before 1.5 releases as it hopefully doesn't disrupt 1.4.x.

pandas-stubs/core/frame.pyi
bashtage: I reverted this since inplace is back for 1.5.0

tests/test_frame.py
bashtage: I agree with this.  We should go through the FutureWarnings after the next release and remove problematic code.

bashtage:
Doesn't do much any more, but does fix a bug in read_stata

Dr-Irv (approved):

Dr-Irv:
thanks @bashtage

PR #252, opened 2022-09-02
ENH: Improve remaining io

bashtage:

Still needs tests


pandas-stubs/core/frame.pyi
```diff
@@ -38,6 +38,7 @@ from pandas.core.window.rolling import (
     Rolling,
     Window,
 )
+import xarray as xr
```
Dr-Irv, line 41: `xarray` is an optional dependency for pandas.  Concerned that this will cause things to break for people who don't have `xarray` installed.

pandas-stubs/core/frame.pyi
```diff
@@ -257,8 +275,8 @@ class DataFrame(NDFrame, OpsMixin):
     def to_records(
         self,
         index: _bool = ...,
-        columnDTypes: _str | dict | None = ...,
-        indexDTypes: _str | dict | None = ...,
+        column_dtypes: _str | dict[HashableT, _str | DtypeNp] | None = ...,
+        index_dtypes: _str | dict[HashableT, _str | DtypeNp] | None = ...,
```
Dr-Irv, line 279: I think that `DTypeArg` is allowable for both of the above.  Docs say a "type" is OK.

tests/test_series.py
```diff
@@ -909,7 +910,7 @@ def test_types_to_list() -> None:
 
 def test_types_to_dict() -> None:
     s = pd.Series(["a", "b", "c"], dtype=str)
-    assert_type(s.to_dict(), Dict[Any, str])
+    assert_type(s.to_dict(), Dict[Hashable, str])
```
Dr-Irv, line 913: Might as well add a `check()` here.

Dr-Irv (changes requested):
Given the changes to `to_dict()` and `to_records()` typing with overloads, can you add tests for that?

Same goes for `to_xarray()`

pandas-stubs/core/frame.pyi
twoertwein: I think it should be fine: if it isn't installed, mypy/pyright will resolve it as Any/Unknown

Dr-Irv (approved):
thanks @bashtage

PR #253, opened 2022-09-02
__neg__ methods and bugfix

amotzop:
Added `__neg__` method to `Series, DataFrame and Index`.
Fixed a bug in `Index.difference` and added a relevant test.


amotzop:
There might be a better solution to the `Index.difference` problem, but nothing I tried seemed to satisfy both test cases other than getting rid of the `T1` inside the `list`.

pandas-stubs/core/generic.pyi
```diff
@@ -69,7 +69,7 @@ class NDFrame(PandasObject, indexing.IndexingMixin):
     def pop(self, item: _str) -> NDFrame: ...
     def squeeze(self, axis=...): ...
     def equals(self, other: Series[S1]) -> _bool: ...
-    def __neg__(self) -> None: ...
+    def __neg__(self: T) -> T: ...
```
bashtage, line 72: Is pos also wrong?

bashtage (changes requested):
Could you check pos too

pandas-stubs/core/generic.pyi
twoertwein: I think pyright is picky about using too wide TypeVar's for self. Would be better to use `NDFrameT` instead of `T`

Dr-Irv: Should be `def __neg__(self) -> NDFrame: ...`

pandas-stubs/core/indexes/base.pyi
```diff
@@ -142,11 +143,12 @@ class Index(IndexOpsMixin, PandasObject):
     def __and__(self, other) -> Index: ...
     def __or__(self, other) -> Index: ...
     def __xor__(self, other) -> Index: ...
+    def __neg__(self: T) -> T: ...
```
Dr-Irv, line 146: Should be `def __neg__(self) -> Index: ...`

Dr-Irv (changes requested):
For each of the tests you added,  can you add a comment at the beginning of each test function (or new test within an existing test function) pointing back to this PR.  Something like `# GH 253`

pandas-stubs/core/generic.pyi
amotzop: This means to behave like the `Self` typing in python 3.11 (it seems mypy doesn't support it).
So that you don't have to implement it for every subclass to return itself.
For me pyright worked fine (`T` will resolve to a subclass of `NDFrame` in any case because this definition is only for classes that derive from `NDFrame`)

Dr-Irv: If you use the `def __neg__(self) -> NDFrame: ...` pattern, it will be consistent with the rest of the stubs

amotzop: But then I'll have to also add it to `Series` and `DataFrame`. If you think its better than I'll do it, but I think this way is less error prune (no "code duplication").

Dr-Irv: Sorry, should be `def __neg__(self: NDFrameT) -> NDFrameT: ...` .  That should work and not require entries for `Series` and `DataFrame`

Dr-Irv (approved):
Thanks @amotzop

PR #258, opened 2022-09-03
update pre-commit

twoertwein:
flake8-pyi has a few more checks


Dr-Irv (approved):
thanks @twoertwein for staying on top of this

PR #259, opened 2022-09-03
Clean io

bashtage:
Big clean out of non-API classes


Dr-Irv (approved):
thanks @bastage.  Nice cleansing!

PR #260, opened 2022-09-04
Add py.typed

twoertwein:
`ABCExtensionArray` and `ABCSeries` did not exist in the stubs but this error might have been hidden because the pandas implementation provides it.

### DIFF
--- a/pandas-stubs/core/dtypes/generic.pyi
+++ b/pandas-stubs/core/dtypes/generic.pyi
@@ -1,0 +1,15 @@
+from pandas import (
+    DataFrame,
+    Index,
+    Series,
+)
+from pandas.core.arrays import ExtensionArray
+from pandas.core.generic import NDFrame
+
+ABCIndex = type[Index]
+
+ABCNDFrame = type[NDFrame]
+ABCSeries = type[Series]
+ABCDataFrame = type[DataFrame]
+
+ABCExtensionArray = type[ExtensionArray]
--- a/pandas-stubs/core/groupby/groupby.pyi
+++ b/pandas-stubs/core/groupby/groupby.pyi
@@ -7,7 +7,6 @@ from pandas.core.base import (
 from pandas.core.frame import DataFrame
 from pandas.core.generic import NDFrame
 from pandas.core.groupby import ops
-from pandas.core.groupby.indexing import GroupByIndexingMixin
 from pandas.core.indexes.api import Index
 from pandas.core.series import Series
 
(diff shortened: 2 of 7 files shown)


pandas-stubs/core/groupby/groupby.pyi
```diff
@@ -7,7 +7,6 @@ from pandas.core.base import (
 from pandas.core.frame import DataFrame
 from pandas.core.generic import NDFrame
 from pandas.core.groupby import ops
-from pandas.core.groupby.indexing import GroupByIndexingMixin
```
twoertwein, line 10: not defined in the stubs - but it provides only private methods, so I removed it.

twoertwein:
A much easier fix would be to simply have `pandas-stubs/py.typed` then we don't need renaming.

@Dr-Irv I now you removed it at some point because of a bug(?) in pyright. Is it now safe to enable it again?

Dr-Irv:
> A much easier fix would be to simply have `pandas-stubs/py.typed` then we don't need renaming.
> 
> @Dr-Irv I now you removed it at some point because of a bug(?) in pyright. Is it now safe to enable it again?

Don't know.  If that would work, I'd prefer it over having to rename `pandas` during testing.  Can you try that?

twoertwein:
> Don't know. If that would work, I'd prefer it over having to rename `pandas` during testing. Can you try that?

If I remove `ABCExtensionArray` and just have a py.typed, mypy/pyright notice the error :)

pandas-stubs/core/dtypes/generic.pyi
```diff
@@ -0,0 +1,15 @@
...
+from pandas.core.arrays import ExtensionArray
+from pandas.core.generic import NDFrame
+
+ABCIndex = type[Index]
+
+ABCNDFrame = type[NDFrame]
+ABCSeries = type[Series]
+ABCDataFrame = type[DataFrame]
+
+ABCExtensionArray = type[ExtensionArray]
```
twoertwein, line 15: Technically need only `ABCExtensionArray` and `ABCSeries` for the current stubs.

Dr-Irv (changes requested):
I think we should have `py.typed` with the word `partial` in it.

twoertwein:
https://peps.python.org/pep-0561/
> Many stub packages will only have part of the type interface for libraries completed, especially initially. For the benefit of type checking and code editors, packages can be “partial”. This means modules not found in the stub package SHOULD be searched for in parts four and five of the module resolution order above, namely inline packages and typeshed.


If we add `partial`, mypy does not find the missing `ABCExtensionArray` (and users's type checkers will also fallback on the pandas implementation). While pandas-stubs is not complete, I would slightly favor marking it as complete, so that people report any missing annotations to make pandas-stub more complete.

Dr-Irv:
> If we add `partial`, mypy does not find the missing `ABCExtensionArray` (and users's type checkers will also fallback on the pandas implementation). While pandas-stubs is not complete, I would slightly favor marking it as complete, so that people report any missing annotations to make pandas-stub more complete.

OK, I'm fine with that.

Dr-Irv (approved):
thanks @twoertwein

PR #261, opened 2022-09-04
remove append

twoertwein:
As long as a class as `__getattr__` exist, mypy and pyright assume that `.append` (and any other non-existing method) might still exists. Removing `__getattr__` has the disadvantage that users can no longer do `df.column_name`.


pandas-stubs/core/generic.pyi
```diff
@@ -359,7 +359,6 @@ class NDFrame(PandasObject, indexing.IndexingMixin):
         self, func: Callable[..., T] | tuple[Callable[..., T], str], *args, **kwargs
     ) -> T: ...
     def __finalize__(self, other, method=..., **kwargs) -> NDFrame: ...
-    def __getattr__(self, name: _str): ...
```
Dr-Irv, line 362: I'm a little wary about removing `__getattr__`.  Will the type checkers now complain to users if they do `df.column_name` ?

On the one hand, that's a good thing, because we can't type check it and we should steer people towards `df["column_name"]` for type checking purposes.  On the other hand, we might start getting lots of complaints on code that currently passes type checking.

twoertwein: Yes, this would complain about `df.column_name`. If we want mypy/pyright to error for `append`, we need to remove `__getattr__`. I don't have strong opinions about removing/keeping `__getattr__`, it has dis- and advantages.

Does  `__getattr__` always return a Series for DataFrames (and a scalar for Series?). If that is the case, we could at least type `__getattr__`. People would then get a error for `append` - but a rather weird one.

twoertwein: When called as `df.<name>` or `series.<name>`, `__getattr__` only ever returns a `Series`. One can however also directly call it: `df.__getattr__("index")` (or any other attribute), which might not be a `Series`.

I would be fine having `__getattr__` always return a `Series`.

twoertwein: When letting `__getattr__` return a `Series`, `.append()` will trigger these errors:

mypy: "Series[Any]" not callable
pyright: error: Object of type "Series[Unknown]" is not callable

Dr-Irv: Having`__getattr__()` return a `Series` will have some benefits.

But it needs to return `Series` when called on a `DataFrame` and `Scalar` when called on a `Series`

So this might actually be a big win.  We pick up when `append()` is wrong, and we get typing for `df.column_name`.  Only issue will be the odd error message, but we can see what happens if we start getting complaints.

twoertwein: > But it needs to return `Series` when called on a `DataFrame` and `Scalar` when called on a `Series`

Calling `__getattr__` on `Series` also seems to return just a `Series` :) but it should then also maintain the dtype!

Dr-Irv:
```python
>>> import pandas as pd
>>> s=pd.Series([1,2,3], index=["a", "b", "c"])
>>> s
a    1
b    2
c    3
dtype: int64
>>> s.a
1
```
Seems it could return a scalar too.  Can you show an example where it returns a `Series`?

twoertwein: You are right, I thought I could pass the column/series name - but that errors.

tests/test_series.py
```diff
@@ -1139,3 +1139,9 @@ def test_neg() -> None:
     sr_int = pd.Series([1, 2, 3], dtype=int)
     check(assert_type(-sr, pd.Series), pd.Series)
     check(assert_type(-sr_int, "pd.Series[int]"), pd.Series, int)
+
+
+def test_getattr() -> None:
+    # GH 261
+    series = pd.Series([1, 2, 3], index=["a", "b", "c"], dtype=int)
+    check(assert_type(series.a, int), np.int64)
```
twoertwein, line 1147: unfortunately, it doesn't return a builtin `int`

twoertwein:
The error on windows could be a pandas/numpy inconsistency. It seems that our windows CI is 64bit (np.intp is np.int64) but the observed dtype is np.int32.

pandas-stubs/core/generic.pyi
```diff
@@ -359,7 +360,7 @@ class NDFrame(PandasObject, indexing.IndexingMixin):
...
+    def __getattr__(self, name: _str) -> Series: ...
```
Dr-Irv, line 363: Move this to `core/frame.pyi` and you can avoid the `# type: ignore` in `core/series.pyi`

Dr-Irv (changes requested):

Dr-Irv (approved):
Thanks @twoertwein

PR #266, opened 2022-09-04
Fix DataFrame.xs and Series.dtype

### LINKED ISSUE #214: "key" parameter typing is incorrect in the DataFrame.xs method
**Describe the bug**
mypy reports an error when using the DataFrame.xs() method.  DataFrame MultiIndexes can easily accept non-string labels, such as ints.  In fact, this is very common.  

**To Reproduce**
Create a script like below and run mypy.  You will see mypy reporting an error about the "key" positional argument, which I assert is not a true error.

import pandas as pd
mi = pd.MultiIndex.from_product([[0, 1], [0, 1]], names=["foo", "bar"])
df = pd.DataFrame({"x": [10, 20, 30, 40], "y": [50, 60, 70, 80]}, index=mi)
df.xs(0, level="foo")

 - OS:  Windows
 - OS Version 10
 - python version 3.8
 - version of installed `pandas-stubs` 1.4.3.220724

### LINKED ISSUE #216: Series.dtype property return the wrong type
**Describe the bug**
The Series.dtype property is defined as "Union[ExtensionDtype, str, dtype[generic], Type[object]]".  However, I believe this is incorrect.  I noticed the problem when using the dtype.kind property.  There is no way that I can think of that series.dtype.kind should fail.  And there's no way I can think of that series.dtype returns a string.  When you are creating a Series, you can indeed input a string that resolves into a dtype (as shown in my code snippet below), but the Series.dtype property always returns a dtype, not a string.  Even pandas extension arrays, , for example those with the nullable ints or strings, have a "kind" property.

**To Reproduce**
Create a script like the below and run in mypy.  Even though it's valid use of series.dtype.kind, it reports errors.

import pandas as pd
s1 = pd.Series(['foo'], dtype='string')
s2 = pd.Series([1], dtype='Int64')
print(s1.dtype.kind == 'O')
print(s2.dtype.kind == 'i')

 - OS: Windows
 - OS Version 10
 - python version 3.8
 - version of type checker 0.971
 - version of installed `pandas-stubs`1.4.3.220724

Dr-Irv:
- [x] Closes #214 
- [x] Close #216
- [x] Tests added
  - `test_frame:test_xs()`
  - `test_series.test_dtype_type()`


twoertwein (approved):

twoertwein:
Thank you @Dr-Irv !

PR #268, opened 2022-09-05
Remove deprecated behaviors

twoertwein:


Dr-Irv (changes requested):
I'm fine with the changes, but just merged in changes on `read_csv()` that is now in a different file, so you'll have to move those changes to that file (`readers.pyi`)

twoertwein:
> I'm fine with the changes, but just merged in changes on `read_csv()` that is now in a different file, so you'll have to move those changes to that file (`readers.pyi`)

I believe I rebased (and forced pushed) while you were reviewing this PR - should already be compatible with main.

Dr-Irv (approved):
thanks @twoertwein 

I'll merge this, but your first commit had changes for `read_csv()`, and the final version does not, so you may want to look at that

Issue #237, opened 2022-08-30
pandas.Index.to_numpy() method's arguments not properly typed


jbiz25:
**Describe the bug**
The pandas.Index.to_numpy() method takes arguments, but mypy complains about them, whether they are positional or keyword.

**To Reproduce**

import pandas as pd
df = pd.DataFrame({"A": [1, 2, 3], "B": [4, 5, 6]})
array = df.index.to_numpy(dtype="int32")

This yields error: Unexpected keyword argument "dtype" for "to_numpy" of "Index"

import pandas as pd
df = pd.DataFrame({"A": [1, 2, 3], "B": [4, 5, 6]})
array = df.index.to_numpy(dtype="int32")

This yields error: Too many arguments for "to_numpy" of "Index

**Please complete the following information:**
 - OS: Windows
 - OS Version 10
 - python version 3.10
 - version of type checker: latest
 - version of installed `pandas-stubs`1.4.3.220829

Dr-Irv:
Thanks for the report.
Just need to fix:  https://github.com/pandas-dev/pandas-stubs/blob/80f561fc73f387f81f3876eeef4fe3bae82dae18/pandas-stubs/core/indexes/base.pyi#L207

PR and tests welcome!

LeMeteore:
What about something like:

```python
from typing import Any, Optional
from pandas._typing import npt
import pandas._libs.lib as lib

    def to_numpy(                                                                                                          
            self, dtype: npt.DTypeLike | None = None,                                                                      
            copy: bool = False,                                                                                            
            na_value: Optional[Any] = lib.no_default,                                                                      
            **kwargs                                                                                                       
    ) -> np.ndarray: ...    
```
If good enough I can send the PR.

jbiz25:
Thanks guys.  Also, FYI, it is probably worth investigating whether other the pandas objects that have to_numpy() methods are missing these, so that you can create similar annotations for them.

Dr-Irv:
> If good enough I can send the PR.

For a stub, you want to have `= ...` rather than specify the default arguments.  

If you can make that change there, and as suggested above, look at the other `to_numpy()` calls in `pandas-stubs`, and also add tests, that would be a welcome PR!

bashtage:
You can also run stubgen against pandas main to get the up to date list of
parameters and often correct type info.

LeMeteore:
Very good catch. According to grep, there are many places where the fix should be applied:

```
[...]

PR #270, opened 2022-09-05
[fix]: pandas.Index.to_numpy() method's arguments not properly typed

LeMeteore:
After cloning a fresh copy of Pandas.
After generating new stubs using stubtest.
After looking for all the type signature for to_numpy method.

There was nothing to update in the following files:
- ./_libs/tslibs/nattype.pyi
- ./_libs/tslibs/timedeltas.pyi

There was modification to do in the following files:
- ./core/arrays/masked.pyi
- ./core/arrays/numpy_.pyi
- ./core/arrays/base.pyi
- /core/frame.pyi
- ./core/base.pyi

The type signature was removed from the following file:
- ./core/indexes/base.pyi

- [x] Refs #237
- [x] tests added.


pandas-stubs/core/arrays/base.pyi
```diff
@@ -2,7 +2,11 @@ from typing import Sequence
 
 import numpy as np
 
-from pandas._typing import ArrayLike
+from pandas._typing import (
+    ArrayLike,
+    Scalar,
+    npt as npt,
```
twoertwein, line 8: We don't want to re-export `npt`. You can remove `as npt`

Dr-Irv:
@LeMeteore you can see the output from the CI on errors you need to fix.  You also need to add the tests.

pandas-stubs/core/arrays/masked.pyi
```diff
@@ -4,14 +4,22 @@ from pandas.core.arrays import (
     ExtensionOpsMixin,
 )
 
-from pandas._typing import Scalar
+from pandas._typing import (
+    Scalar,
+    npt as npt,
```
Dr-Irv, line 9: remove the `as npt` here and elsewhere

Dr-Irv (approved):
thanks @LeMeteore

PR #271, opened 2022-09-05
pytest nightly: use summary to indicate failures

twoertwein:


bashtage:
That looks like a reasonable fix.  Saw something like that on the massive thread asking for a yellow badge.

twoertwein:
Probably don't need `$GITHUB_STEP_SUMMARY` as github already generates a red warning in the summary
image

twoertwein:
It takes unfortunately a few steps to see that nightly errored:

1. wait for all tests to finish
2. click on any test
3. click on "summary"
4. Scroll down, if nightly errored you will see:
image

Or slightly faster (but less visual):

1. click on the nightly test (no need to wait for the other tests)
2. click on "run pytest (against pandas nightly)"
3. Scroll down to the pytest summary

Dr-Irv:
One thing that is a little bit of a pain is that in the Conversation view in GitHub, you don't see the annotation.  You have to go to the workflow.

twoertwein:
If we are able to achive (and keep) nightly to be successful, we don't need `continue-on-error` :)

I think except for the `set_index` `inplace` issue (which suggests using a new keyword that does not yet exist in 1.4.x), the other issues can be taken care of by removing the type annotations and by re-writing the tests as negative tests.

For `set_index`, we might need a small helper function `is_nightly() -> bool` to skip some of the tests on nightly.

twoertwein:
Closing this  - let's hope we can get nightly green when 1.5 releases

PR #272, opened 2022-09-05
CLN: FrameOrSeries(Union)

### LINKED ISSUE #255: CLEAN:  Replace FrameOrSeries TypeVar with NDFRameT, rename FrameOrSeriesUnion to be FrameOrSeries
We currently have this in `_typing.pyi`:
```
FrameOrSeries = TypeVar("FrameOrSeries", bound=NDFrame)
FrameOrSeriesUnion = Union[DataFrame, Series]
NDFrameT = TypeVar("NDFrameT", bound=NDFrame)
```

`FrameOrSeries` and `NDFrameT`  clearly are duplicates.

We should remove `FrameOrSeries` from the code and replace it with `NDFrameT`

If we do that, then we can also replace `FrameOrSeriesUnion` and call it `FrameOrSeries`

twoertwein:
- [x] Closes #255

Pandas does not define `FrameOrSeriesUnion` and instead uses `DataFrame | Series`.


bashtage:
Maybe move nightly to be first?  Sort of a pain to know that only nightly failed.

twoertwein:
Honestly, I wouldn't mind adding continue-on-error until nightly passes. Once it passes, we can remove it until it errors again and then either fix it/ignore it temporarily.

pandas-stubs/_typing.pyi
```diff
@@ -80,14 +79,10 @@ class WriteExcelBuffer(WriteBuffer[bytes], Protocol):
...
 
 FilePath = Union[str, PathLike[str]]
+FilePathOrBuffer = Union[
+    FilePath, IO[AnyStr], RawIOBase, BufferedIOBase, TextIOBase, TextIOWrapper, mmap
+]
 
-Buffer = Union[IO[AnyStr], RawIOBase, BufferedIOBase, TextIOBase, TextIOWrapper, mmap]
-FileOrBuffer = Union[str, Buffer[AnyStr]]
-FilePathOrBuffer = Union[PathLike[str], FileOrBuffer[AnyStr]]
-FilePathOrBytesBuffer = Union[PathLike[str], WriteBuffer[bytes]]
```
twoertwein, line 87: All these buffers do not exist in pandas. Only `FilePathOrBytesBuffer` is used in the stubs, so I merged them.

.github/workflows/test.yml
```diff
@@ -3,6 +3,33 @@ name: 'Test'
 on:
   [push, pull_request, workflow_dispatch]
 jobs:
+
+  nightly:
```
Dr-Irv, line 7: should the change in `test.yml` be in a separate PR?

Dr-Irv (changes requested):

.github/workflows/test.yml
twoertwein: removed

Dr-Irv (approved):
thanks @twoertwein

PR #273, opened 2022-09-05
TYP: Clean and improve config/options

bashtage:
Imrove types for options
Add tests


Dr-Irv (approved):
thanks @bashtage

PR #276, opened 2022-09-06
ENH: Finish typing utility functions

bashtage:


Dr-Irv (changes requested):
In `errors.pyi`, you have defined all these classes, but they are not all tested.  I think these are the ones that are missing (found via doing some `diff`, `sed`, and `sort` exercises):
```
AttributeConflictWarning
CSSWarning
CategoricalConversionWarning
ClosedFileError
DataError
DatabaseError
IncompatibilityWarning
IndexingError
InvalidColumnName
NumExprClobberingError
PossibleDataLossError
PossiblePrecisionLoss
PyperclipException
PyperclipWindowsException
SettingWithCopyError
SettingWithCopyWarning
SpecificationError
UndefinedVariableError
ValueLabelTypeMismatch
```

So I think tests should exist for those, or at least tell me why you chose to not test them.

tests/test_errors.py
```diff
@@ -1,97 +1,215 @@
...
+@pytest.mark.skipif(PD_LT_15, reason="Feature added in 1.5.0")
+def test_setting_with_copy_error() -> None:
+    with pytest.raises(errors.SettingWithCopyError):
+        raise errors.SettingWithCopyError()
+
+
+@pytest.mark.skipif(PD_LT_15, reason="Feature added in 1.5.0")
+def test_setting_with_copy_warning() -> None:
+    with pytest.warns(errors.SettingWithCopyWarning):
+        warnings.warn("", errors.SettingWithCopyWarning)
```
Dr-Irv, line 125: Not sure that the reason here is correct.  `SettingWithCopyWarning` exists in 1.4, but it is in `core/common.py` and not `errors`

Part of me wants this to be right since we are still with 1.4, but the other part says to let it go because this is probably not something people end up type checking.

So can you research these and see if the locations changed in 1.4 and then update the reasons accordingly?

bashtage: I fixed these and put in a lot of comments that should be 1.5.x easier.

Dr-Irv (approved):
Thanks @bashtage

PR #278, opened 2022-09-07
ENH: Improve plotting types

bashtage:
Add types for many inputs
Clean out non-public functions


pandas-stubs/plotting/_misc.pyi
```diff
@@ -1,85 +1,82 @@
...
-    data,
-    rowLabels=...,
-    colLabels=...,
-): ...
+    ax: Axes,
+    data: DataFrame | Series,
+    **kwargs,
+) -> Table: ...
 def register() -> None: ...
 def deregister() -> None: ...
```
Dr-Irv, line 30: I don't see `register()` and `deregister()` as documented, probably can be deleted.

pandas-stubs/plotting/_core.pyi
```diff
@@ -1,94 +1,250 @@
...
-    def pie(self, **kwargs) -> PlotAccessor: ...
-    def scatter(self, x, y, s=..., c=..., **kwargs) -> PlotAccessor: ...
+    @overload
+    def line(
+        self,
+        x: Hashable | None = ...,
+        y: Hashable | None = ...,
+        color: _PlotAccessorColor = ...,
+        subplots: Literal[False] | None = ...,
+        **kwargs,
```
Dr-Irv, line 88: All of these functions share the same `kwargs` defined in the docs for `DataFrame.plot()`, so I think we should include them in each of these stubs (unfortunately, lots of repeated code) and then not have `**kwargs`.

Dr-Irv (changes requested):

Dr-Irv (commented):
Might be good in the tests to have one test where we test all the `kwargs` documented in `DataFrame.plot()`

pandas-stubs/plotting/_misc.pyi
bashtage: These are short name implementations of `(de)register_matplotlib_converters`.  Keeping close to actual implementation to avoid confusion.  These are exported and then used in the `__init__`

pandas-stubs/plotting/_core.pyi
bashtage: Not sure we can get rid of kwargs since these are passed through to matplotlib, and some are plot specific.

Dr-Irv: > Not sure we can get rid of kwargs since these are passed through to matplotlib, and some are plot specific.

OK, but we can at least add all the arguments that are documented in `DataFrame.plot()`

bashtage: I've added these and tested them.

pandas-stubs/plotting/_core.pyi
```diff
@@ -77,7 +81,161 @@ def boxplot(
...
+            "pie",
+            "scatter",
+            "hexbin",
+        ] = ...,
+        ax: Axes | None = ...,
+        subplots: Literal[False] | None = ...,
+        sharex: bool = ...,
+        sharey: bool = ...,
+        layout: tuple[int, int] = ...,
+        figsize: tuple[int, int] = ...,
```
Dr-Irv, line 109: Can't `figsize` be a tuple of floats?

pandas-stubs/plotting/_core.pyi
```diff
@@ -77,7 +81,161 @@ def boxplot(
...
+        logx: bool | Literal["sym"] = ...,
+        logy: bool | Literal["sym"] = ...,
+        loglog: bool | Literal["sym"] = ...,
+        xticks: Sequence[float] = ...,
+        yticks: Sequence[float] = ...,
+        xlim: tuple[float, float] | list[float] = ...,
+        ylim: tuple[float, float] | list[float] = ...,
+        xlabel: str = ...,
+        ylabel: str = ...,
+        rot: float = ...,
```
Dr-Irv, line 124: docs show `int` for `rot`

Dr-Irv (changes requested):
I think all of the keywords in `DataFrame.plot()` can also be passed to things like `DataFrame.plot.line()`, etc.  So I think we should copy the keywords you included in `__call__()` over to each of the individual functions.     E.g., the following is valid:
```python
foo = pd.DataFrame({"x":[i for i in range(10)], "y":[i*i for i in range(10)]})
foo.plot.line("x", "y", figsize=(3,3))
```
You don't have to test each one.

pandas-stubs/plotting/_core.pyi
bashtage: Thanks, fixed.

pandas-stubs/plotting/_core.pyi
bashtage: I'm submitted a PR to pandas to standardized rot and fotnsize to be floats.

bashtage:
> I think all of the keywords in `DataFrame.plot()` can also be passed to things like `DataFrame.plot.line()`, etc. So I think we should copy the keywords you included in `__call__()` over to each of the individual functions. E.g., the following is valid:

I'm hesitant to do this since not all keywords are accepted by all plot types and some, even if accepted, have no effect. For example, `positions` and `colorbar` can't be used with `line`.  IMO, the general `plot()` function is a mess of a function but is there for quick analysis.

Dr-Irv:
> I'm hesitant to do this since not all keywords are accepted by all plot types and some, even if accepted, have no effect. For example, `positions` and `colorbar` can't be used with `line`. IMO, the general `plot()` function is a mess of a function but is there for quick analysis.

Fair point.  It's not clear to me if people would do `df.plot(x="x", y="y", kind="line", figsize=(5, 5))` or `df.plot.line(x="x", y="y",  figsize=(5, 5))`, although the examples at https://pandas.pydata.org/docs/user_guide/visualization.html?highlight=plotting seem to show the latter usage.  Having said that, I think many of the accepted keywords listed in `DataFrame.plot()` in the docs are potentially specific to the chosen backend, so getting that right from a typing perspective would be a nightmare.

Issue #264, opened 2022-09-04
error: No overload variant of "isna" matches argument type "NAType"  [call-overload]


randolf-scholz:
```python
import pandas as pd
print(pd.isna(pd.NA))
```

Dr-Irv:
Have to add `NaTType` to `isna()` and `notna()` overloads that specify `Scalar` in `core/dtypes/missing.pyi`

Dr-Irv:
Closed via #279

PR #279, opened 2022-09-07
[fix] No overload variant of "isna" matches argument type "NAType" [call-overload]

LeMeteore:
- Update `isna` & `notna` overloads inside pandas-stubs/core/dtypes/missing.pyi 
- Update `test_isna` function inside tests/test_pandas.py file

- [x] Refs #264  
- [x] Tests updated


LeMeteore:
When I tried to test with:
```python
check(assert_type(pd.isna(pd.NA), bool), bool)                                                                                              
check(assert_type(pd.notna(pd.NA), bool), bool)
```

I had the following errors:
```
tests/test_pandas.py:130: error: Expression is of type "Any", not "bool"  [assert-type]
tests/test_pandas.py:130: error: No overload variant of "isna" matches argument type "NAType"  [call-overload]
tests/test_pandas.py:130: note: Possible overload variants:
tests/test_pandas.py:130: note:     def isna(obj: DataFrame) -> DataFrame
tests/test_pandas.py:130: note:     def isna(obj: Series[Any]) -> Series[bool]
tests/test_pandas.py:130: note:     def isna(obj: Union[Index, List[Any], Union[ExtensionArray, ndarray[Any, Any]]]) -> ndarray[Any, Any]
tests/test_pandas.py:130: note:     def isna(obj: Union[Union[str, bytes, date, datetime, timedelta, bool, complex, Timestamp, Timedelta], NaTType]) -> bool
tests/test_pandas.py:131: error: Expression is of type "Any", not "bool"  [assert-type]
tests/test_pandas.py:131: error: No overload variant of "notna" matches argument type "NAType"  [call-overload]
tests/test_pandas.py:131: note: Possible overload variants:
tests/test_pandas.py:131: note:     def notna(obj: DataFrame) -> DataFrame
tests/test_pandas.py:131: note:     def notna(obj: Series[Any]) -> Series[bool]
tests/test_pandas.py:131: note:     def notna(obj: Union[Index, List[Any], Union[ExtensionArray, ndarray[Any, Any]]]) -> ndarray[Any, Any]
tests/test_pandas.py:131: note:     def notna(obj: Union[Union[str, bytes, date, datetime, timedelta, bool, complex, Timestamp, Timedelta], NaTType]) -> bool
```

I decided to use `np.nan` instead of `pd.NA`:
```python
check(assert_type(pd.isna(np.nan), bool), bool)                                                                                             
check(assert_type(pd.notna(np.nan), bool), bool) 
```

tests/test_pandas.py
```diff
@@ -112,13 +112,26 @@ def test_types_json_normalize() -> None:
...
 
 def test_isna() -> None:
-    s = pd.Series([1, np.nan, 3.2])
-    check(assert_type(pd.isna(s), "pd.Series[bool]"), pd.Series, bool)
-    b: bool = pd.isna(np.nan)
-    ar: np.ndarray = pd.isna(s.to_list())
-    check(assert_type(pd.notna(s), "pd.Series[bool]"), pd.Series, bool)
-    b2: bool = pd.notna(np.nan)
-    ar2: np.ndarray = pd.notna(s.to_list())
+    s1 = pd.Series([1, np.nan, 3.2])
```
Dr-Irv, line 115: Can you add a comment in this function that references the original issue?

tests/test_pandas.py
```diff
@@ -112,13 +112,26 @@ def test_types_json_normalize() -> None:
...
+    check(assert_type(pd.isna(idx1), np.ndarray), np.ndarray)
+
+    idx2 = pd.Index([1, 2])
+    check(assert_type(pd.notna(idx2), np.ndarray), np.ndarray)
+
+    check(assert_type(pd.isna(np.nan), bool), bool)
+    check(assert_type(pd.notna(np.nan), bool), bool)
+
+    check(assert_type(pd.isna(2.5), bool), bool)
+    check(assert_type(pd.notna(2.5), bool), bool)
```
Dr-Irv, line 134: Add a test for `pd.isna(pd.NA)` and `pd.notna(pd.NA)`

Dr-Irv (changes requested):

Dr-Irv:
> I decided to use `np.nan` instead of `pd.NA`:

I think you pulled the wrong type.  You need to use `NAType` and `NaTType`.  The first corresponds to `pd.NA`, the second to `pd.NaT` (which you should also test as well).
```python
>>> import pandas as pd
>>> type(pd.NA)
<class 'pandas._libs.missing.NAType'>
>>> type(pd.NaT)
<class 'pandas._libs.tslibs.nattype.NaTType'>
```

pandas-stubs/core/dtypes/missing.pyi
```diff
@@ -22,7 +24,7 @@ def isna(obj: Series) -> Series[bool]: ...
 @overload
 def isna(obj: Index | list | ArrayLike) -> np.ndarray: ...
```
twoertwein, line 25: Could change the return type of this overload to `npt.NDArray[np.bool_]`

LeMeteore: Without rerunning stubgen on a fresh clone of Pandas repository, was there a way for me to see this improvement ? Is there a way to develop the intuition ?

Dr-Irv: > Without rerunning stubgen on a fresh clone of Pandas repository, was there a way for me to see this improvement ? Is there a way to develop the intuition ?

Don't think even `stubgen` would figure that out.  We just learn as we go.  In this case, we know that when the results is an `np.ndarray` that contains only `boolean`, we can use a more specific type, which we always try to do whenever we can.

twoertwein: Writing type annotations is a mix of reading the documentation, looking at the implementation, and using automated tools like stubgen/pyright. In this case, we also have those annotations in the implementation (the `NaTType` is missing there too): https://github.com/pandas-dev/pandas/blob/ce143a2fb0a41e78c85d5311d0e159552d651b00/pandas/core/dtypes/missing.py#L99

pandas-stubs/core/dtypes/missing.pyi
```diff
@@ -20,9 +23,9 @@ def isna(obj: DataFrame) -> DataFrame: ...
 @overload
 def isna(obj: Series) -> Series[bool]: ...
 @overload
-def isna(obj: Index | list | ArrayLike) -> np.ndarray: ...
+def isna(obj: Index | list | ArrayLike) -> npt.NDArray[np.bool_]: ...
 @overload
-def isna(obj: Scalar) -> bool: ...
+def isna(obj: Scalar | NaTType | NAType) -> bool: ...
```
twoertwein, line 28: Probably not important: Could have an overload specific for `NaTType/NAType` as they should return `True`: `def isna(obj: NaTType | NAType) -> Literal[True]: ...`

pandas-stubs/core/dtypes/missing.pyi
```diff
@@ -1,12 +1,15 @@
...
 import numpy as np
+from numpy import typing as npt
 from pandas import (
     DataFrame,
     Index,
     Series,
 )
 
+from pandas._libs.missing import NAType
+from pandas._libs.tslibs.nattype import NaTType
```
Dr-Irv, line 12: can you use `from pandas._libs.tslibs import NaTType` here?  That's what we have used elsewhere

Dr-Irv (changes requested):
This all looks good except for one small thing.

tests/test_pandas.py
```diff
@@ -129,11 +130,11 @@ def test_isna() -> None:
     idx2 = pd.Index([1, 2])
     check(assert_type(pd.notna(idx2), np.ndarray), np.ndarray)
 
-    check(assert_type(pd.isna(pd.NA), bool), bool)
-    check(assert_type(pd.notna(pd.NA), bool), bool)
+    check(assert_type(pd.isna(pd.NA), Literal[True]), bool)
+    check(assert_type(pd.notna(pd.NA), Literal[False]), bool)
```
Dr-Irv, line 134: For completeness here, you should be able to do 
```python
assert check(assert_type(pd.isna(pd.NA), Literal[True]), bool)
assert not check(assert_type(pd.notna(pd.NA), Literal[False]), bool)
```
and likewise in the next 2 lines

This will make sure we are returning the right result.

Dr-Irv (approved):

Dr-Irv (changes requested):
hit the wrong button - make the changes with the tests and we should be good

pandas-stubs/core/dtypes/missing.pyi
```diff
@@ -34,6 +43,8 @@ def notna(obj: Series) -> Series[bool]: ...
 def notna(obj: Index | list | ArrayLike) -> np.ndarray: ...
```
bashtage, line 43: Should this also be npt.NDArray[np.bool_]

Dr-Irv: agreed.  Good catch

Dr-Irv (changes requested):

pandas-stubs/core/dtypes/missing.pyi
LeMeteore: You're definitely right \o/

Dr-Irv (approved):
Thanks @LeMeteore

marcglobality:
re this one, I'm getting errors in mypy due `pd.isna(None)`. Is this expected behaviour? Could we cover also `None` ?

Dr-Irv:
> re this one, I'm getting errors in mypy due `pd.isna(None)`. Is this expected behaviour? Could we cover also `None` ?

Can you create an issue, with the example, and the output of `mypy`?

marcglobality:
here you go https://github.com/pandas-dev/pandas-stubs/issues/318

PR #281, opened 2022-09-08
Accept `sqlalchemy.engine.Engine` for SQL IO API (`read_sql`, `to_sql`)

wakabame:
## Outline
The typecheck for sql io api developed in https://github.com/pandas-dev/pandas-stubs/commit/4c97ed3d8be74b6e982889f177dc6228b788e4da does not accept `sqlalchemy.engine.Engine` for the argument `con`. 

However, the original pandas implementation accepts them
* https://pandas.pydata.org/docs/reference/api/pandas.DataFrame.to_sql.html
* https://pandas.pydata.org/docs/reference/api/pandas.read_sql.html

In fact, `pandasSQL_builder`, which is used in [`to_sql`](https://github.com/pandas-dev/pandas/blob/main/pandas/io/sql.py#L687) and [`read_sql`](https://github.com/pandas-dev/pandas/blob/main/pandas/io/sql.py#L562), requires `sqlalchemy.engine.Connectable` not only `sqlalchemy.engine.Connection`:
https://github.com/pandas-dev/pandas/blob/50c119dce9005cb3e49c0cfb89f396aeecab94f1/pandas/io/sql.py#L757

So, in this PR, I improve the typecheck wider, and introduce testcodes for the case of sqlalchemy connection.
## Check List
    - No issues appeared

This is my first pull request for this repository, so I apologize if there are any problems. I would appreciate it if you could check it.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -57,6 +57,7 @@ odfpy = ">=1.4.1"
 xarray = ">=22.6.0"
 tabulate = ">=0.8.10"
 scipy = ">=1.9.1"
+SQLAlchemy = "^1.4.41"
 
 
 [build-system]
--- a/pandas-stubs/core/generic.pyi
+++ b/pandas-stubs/core/generic.pyi
@@ -160,7 +160,7 @@ class NDFrame(PandasObject, indexing.IndexingMixin):
     def to_sql(
         self,
         name: _str,
-        con: str | sqlalchemy.engine.Connection | sqlite3.Connection,
+        con: str | sqlalchemy.engine.Connectable | sqlite3.Connection,
         schema: _str | None = ...,
         if_exists: Literal["fail", "replace", "append"] = ...,
         index: _bool = ...,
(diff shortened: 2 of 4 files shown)


bashtage:
LGTM.  Let's make sure it passes.

pyproject.toml
```diff
@@ -58,6 +58,9 @@ xarray = ">=22.6.0"
 tabulate = ">=0.8.10"
 
 
+[tool.poetry.group.dev.dependencies]
+SQLAlchemy = "^1.4.41"
```
Dr-Irv, line 62: Only question is why put this in a separate group versus in the list above?

Otherwise looks good.

wakabame: Thank you for your comment. I didnt intended it, whereas the group created just by `poetry add --dev sqlalchemy`.

I'll it.

Dr-Irv (approved):
thanks @wakabame

wakabame:
@Dr-Irv 
Thank you for your quick and kind comment!

PR #283, opened 2022-09-09
ENH: Improve Resampler typing

bashtage:


Dr-Irv:
When this goes green, ping me to review

bashtage:
When a function runs any index type, and I just use `-> Index`?

Dr-Irv:
> When a function runs any index type, and I just use `-> Index`?

Not sure what you mean here???  If you're asking how to type a function that returns any `Index` type, then use `->Index`, but if the function has the index type as an argument, and returns an `Index` of the same type, you can use a `TypeVar` on the parameter so it returns the same type.

pandas-stubs/core/resample.pyi
```diff
@@ -1,56 +1,137 @@
...
+    def ax(self) -> Index: ...
+    def pipe(
+        self,
+        func: Callable | tuple[Callable, str],
+        *args,
+        **kwargs,
+    ) -> Scalar | Series | DataFrame | Resampler: ...
+    def aggregate(
+        self, func: _GroupByFuncArgs | None = ..., *args, **kwargs
+    ) -> Scalar | Series | DataFrame: ...
```
Dr-Irv, line 66: Not 100% sure, but with `resample()`, if you are doing `aggregate`, I think the return type will depend upon the function type passed.  See what is done in `SeriesGroupBy` and `DataFrameGroupBy` and see if we can do something similar here.  Same thing may be true for `pipe`

Dr-Irv (changes requested):

bashtage:
> Not sure what you mean here??? If you're asking how to type a function that returns any Index type, then use ->Index, but if the function has the index type as an argument, and returns an Index of the same type, you can use a TypeVar on the parameter so it returns the same type.

I meant the latter.  Basically it will return a subclass of `Index`, e.g., `DatetimeIndex`.

pandas-stubs/core/resample.pyi
bashtage: It is really hard to get it right since the output depends on the input type and the output type.  I think I'm not going to make any progress on refining this.

bashtage: The two `*GroupBy` have a big advantage since they know the input type. This is generic and so we don't know it here.

tests/test_frame.py
```diff
@@ -998,28 +999,31 @@ def test_types_from_dict() -> None:
...
+        .pipe(foo)
+    )
 
-        def foo(df: pd.DataFrame) -> pd.DataFrame:
-            return df
+    # Cant' check actual return type because it is DatetimeIndexResampler which is
+    # not part of the public API
+    assert_type(val, Union[Scalar, pd.Series, pd.DataFrame, Resampler])
 
+    if TYPE_CHECKING:  # skip pytest
```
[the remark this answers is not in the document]
bashtage: This is old code, not mine.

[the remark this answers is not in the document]
Dr-Irv: It was skipped because of `jinja2` not being installed, so you could make that a dev dependency and remove it (relates to the `Styler` PR you are doing)

pandas-stubs/core/resample.pyi
Dr-Irv: `Resampler` knows the input type because it is generic.  Not sure why we can't take advantage of that.

Unless the only solution is to do a `SeriesResampler` and `DataFrameResampler` akin to `SeriesGroupBy` and `DataFrameGroupBy`, which I realize would be painful

bashtage: They don't always return the same type.  For example, if using a Dataframe, and Callable[DataFrame, Scalar] you get a Series.  If using a Series, and the same callable, you get a Series.   Could always just ignore that case and say DataFrame always gets DataFrame, and Series always get Series. With that simplification that it could be `NDFramteT`

Dr-Irv: I think (but I am REALLY not sure) you might be able to do something like:
```python
@overload
def aggregate(
        self:Resampler[DataFrame], func: Callable[[DataFrame], Scalar]| None = ..., *args, **kwargs
    ) -> Series: ...
@overload
def aggregate(
        self:Resampler[Series], func: Callable[[DataFrame], Scalar]| None = ..., *args, **kwargs
    ) -> Scalar: ...

Comment (bashtage) [reply pandas-stubs/core/resample.pyi]: I think it is too complicated (for me).  Not saying it couldn't be cracked, but it is challenging.  I figured out you can have

DataFrame -> DataFrame | Series, depending on the callable and on the argument type (list or dict can taks a Callable[[DataFrame],Scalar] and get a DataFrame, while just the Callable gets a Series. Similarly you can have Series -> DataFrame | Series, depending on ehther using a string or simple callable (Series), or list of dict (DataFrame).

Comment (Dr-Irv) [reply pandas-stubs/core/resample.pyi]: OK.  I guess we'd have to create a bunch of examples that show all the possibilities to see if anything could really be done, or if it is just ends up being that the result is really dynamically typed, then we can't cover it with static typing.

File: pandas-stubs/core/resample.pyi [hunk 3]
```diff
@@ -1,56 +1,160 @@
...
         self,
-        method: str = ...,
-        axis: int = ...,
-        limit=...,
-        inplace: bool = ...,
-        limit_direction: str = ...,
-        limit_area=...,
-        downcast=...,
+        method: _Interpolation = ...,
+        axis: Axis = ...,
```
Dr-Irv, line 120: should be `AxisType`

tests/test_frame.py
Dr-Irv, line 1021: can you remove and add `jinja2` as a dev dependency?

Dr-Irv (changes requested):
a few more small things, then should be good to go

Dr-Irv (approved):
thanks @bashtage

PR #285, opened 2022-09-09
Added more possible types to groupby on

amotzop:
After the relevant changes to the pandas documentation, I added some missing types to the `by` argument of `groupby` functions.


pandas-stubs/_typing.pyi
```diff
@@ -203,7 +204,20 @@ XMLParsers = Literal["lxml", "etree"]
 # Any plain Python or numpy function
 Function = Union[np.ufunc, Callable[..., Any]]
 GroupByObjectNonScalar = Union[
-    tuple, list[HashableT], Function, Series, np.ndarray, Mapping[Label, Any], Index
+    tuple,
+    list[HashableT],
```
bashtage, line 208: Asiide from list[HashableT] these are not in the doc string.  Looks like they would, but would be best to update the pandas docs first to clarify that list of functions, arrays, etc are accepted.

pandas-stubs/_typing.pyi
```diff
@@ -203,7 +204,20 @@ XMLParsers = Literal["lxml", "etree"]
...
+    Series,
+    list[Series],
+    np.ndarray,
+    list[np.ndarray],
+    Mapping[Label, Any],
+    list[Mapping[Label, Any]],
+    Index,
+    list[Index],
+    Grouper,
+    list[Grouper],
```
Dr-Irv, line 220: If you are going to add all of these ones of the pattern `list[XXX]`, then create a test for each one.  e.g., for `list[Series]`, create an example that does a `groupby` on a list of series.

If all of those work, then also create an issue or PR in the pandas-repo that clarifies that those lists can be used as well.

Dr-Irv (changes requested):

pandas-stubs/_typing.pyi
amotzop: They are in the newest docs here: https://pandas.pydata.org/docs/dev/reference/api/pandas.DataFrame.groupby.html

pandas-stubs/_typing.pyi
amotzop: I did add the relevant tests to `test_series.py, test_frame.py`.
The pandas docs are already updated to include all these types, see here:
https://pandas.pydata.org/docs/dev/reference/api/pandas.DataFrame.groupby.html

Dr-Irv: I think you don't have tests for `list[np.ndarray]`, `list[Mapping[Label, Any]]` and `list[Index]`

Dr-Irv:
@amotzop There's a bug in mypy with respect to python 3.10.7, so we have pinned our CI to python 3.10.6.  Can you merge with main and push?  If the CI passes, I'll then approve and merge the PR.   

Thanks!

Dr-Irv (approved):
thanks @amotzop

PR #286, opened 2022-09-09
Fix inclusion of pandas.testing in __init__.pyi

asford:
The double-import of `pandas.testing` in `__init__.pyi` breaks detection of this module in pyright. Fixup by removing the unneeded absolute import in favor of the existing relative import used for other submodules.


Dr-Irv (approved):
thanks @asford

PR #299, opened 2022-09-15
MAINT: Fix nightly for test_errors

bashtage:


Dr-Irv (approved):
thanks @bashtage

bashtage:
Moving in the right direction, but not there yet.

PR #301, opened 2022-09-15
MAINT: Fix nightly by skipping and noting problematic code

bashtage:


tests/test_errors.py
```diff
@@ -1,46 +1,71 @@
...
-    PossibleDataLossError,
-)
-from pandas.io.sql import DatabaseError
-from pandas.io.stata import (
-    CategoricalConversionWarning,
-    InvalidColumnName,
-    PossiblePrecisionLoss,
-    ValueLabelTypeMismatch,
-)
+PD_LT_15 = parse(pd.__version__) < parse("1.5.0")
```
Dr-Irv, line 11: Move `PD_LT_15` to `tests/__init__.py`, then you don't have to `from packaging.version import parse` in all those files.

tests/test_resampler.py
```diff
@@ -297,7 +301,9 @@ def s2scalar(val: Series) -> float:
 
     check(S.resample("m").aggregate(np.sum), Series)
     check(S.resample("m").aggregate("sum"), Series)
-    check(S.resample("m").aggregate(s2series), Series)
+    if PD_LT_15:
+        # Warns on 1.5.0, add pytest.warns
```
Dr-Irv, line 305: Couldn't you just do something like
```python
        if PD_LT_15:
            check(S.resample("m").aggregate(s2series), Series)
        else:
             with pytest.warns(FutureWarning, match=" FutureWarning: Not prepending group keys to the result index of transform-like apply"):
                 check(S.resample("m").aggregate(s2series), Series)
```

Dr-Irv (changes requested):

tests/test_errors.py
bashtage: Fixed.

tests/test_resampler.py
bashtage: Fixed.

bashtage:
Should be good now.

Dr-Irv (approved):
thanks @bashtage

PR #303, opened 2022-09-16
MAINT: Pin MPL to avoid warnings

bashtage:


bashtage:
3.6.0 just dropped and creating warnings to occur on some builds.

twoertwein:
Could also remove https://github.com/pandas-dev/pandas-stubs/blob/f5cfdc4171bed8e4b7144d6a1443eca547e78ca2/scripts/test/run.py#L17 `Werror`. I added it to easily notice Future/DeprecationWarnings on nightly. If the warnings are causing too much trouble, we could try to have warnings only for nightly.

Dr-Irv (approved):
thanks @bashtage

PR #306, opened 2022-09-16
BUG: Generalize options

### LINKED ISSUE #294: Mypy complains when setting plotting backend
**Describe the bug**

Can not set the plotting backend

**To Reproduce**

```
pd.options.plotting.backend = "plotly"
```

errors with 

```
test.py:6: error: Item "str" of "Union[str, bool, int, DictWrapper, None]" has no attribute "backend"  [union-attr]
test.py:6: error: Item "bool" of "Union[str, bool, int, DictWrapper, None]" has no attribute "backend"  [union-attr]
test.py:6: error: Item "int" of "Union[str, bool, int, DictWrapper, None]" has no attribute "backend"  [union-attr]
test.py:6: error: Item "None" of "Union[str, bool, int, DictWrapper, None]" has no attribute "backend"  [union-attr]
```

after the latest release

According to the user guide this should work: https://pandas.pydata.org/docs/user_guide/options.html#overview

 - OS Version [e.g. 22] 12.5.1
 - python version 3.10
 - version of type checker mypy 0.960
 - version of installed `pandas-stubs` [v1.4.4.220906]

**Additional context**

bashtage:
Generalize using specific classes for each level

- [x] Closes #294

### DIFF
--- a/pandas-stubs/_config/config.pyi
+++ b/pandas-stubs/_config/config.pyi
@@ -22,7 +22,157 @@ class DictWrapper:
     def __getattr__(self, key: str) -> str | bool | int | DictWrapper | None: ...
     def __dir__(self) -> Iterable[str]: ...
 
-options: DictWrapper = ...
+class Compute(DictWrapper):
+    use_bottleneck: bool
+    use_numba: bool
+    use_numexpr: bool
+
+class DisplayHTML(DictWrapper):
+    border: int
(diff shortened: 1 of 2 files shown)


bashtage:
Not 100% about this.  Basically facing the class structure to be more specific.

Dr-Irv (approved):
thanks @bashtage

PR #309, opened 2022-09-17
Add accessors for TimedeltaIndex

### LINKED ISSUE #292: TimedeltaIndex missing accessors
**Describe the bug**
`TimedeltaIndex` is missing accessors

**To Reproduce**
```python 
import pandas as pd
tdi = pd.to_timedelta([3, 4, 5], "days")
reveal_type(tdi)
print(tdi.days)
```
mypy produces:
```text
tdi.py:3: note: Revealed type is "pandas.core.indexes.timedeltas.TimedeltaIndex"
tdi.py:4: error: "TimedeltaIndex" has no attribute "days"
```
pyright produces:
```text
tdi.py
  tdi.py:3:13 - information: Type of "tdi" is "TimedeltaIndex"
  tdi.py:4:11 - error: Cannot access member "days" for type "TimedeltaIndex"
    Member "days" is unknown (reportGeneralTypeIssues)
```

 - OS: Windows 10
 - python version:  3.10.4
 - version of type checker:  my 0.971 pyright 1.1.269
 - version of installed `pandas-stubs`   1.4.4.220906

**Additional context**

Will need to work with `core/indexes/accessors.pyi` to create appropriate `TimedeltaIndexProperties` and have that as a superclass of `TimedeltaIndex`, with appropriate tests, similar to the last set of tests in `test_timefuncs.py:test_series_dt_accessors`

Dr-Irv:
- [x] Closes #292 
- [x] Tests added:
    - `test_timefuncs.py:test_timedeltaindex_accessors()`

Notes:
1. For some reason, I couldn't use the existing classes to hook in the rounding methods, so had to duplicate them for `TimedeltaIndex`.  Maybe it is because `DatetimeIndex` and `TimedeltaIndex` have a common base class.  In `_DTRoundingMethodReturnType`, just changing the order of `DatetimeIndex` and `TimedeltaIndex` causes different failures with pyright.  Maybe doing `Generic[SomeTypeVar]` doesn't work if the classes for `SomeTypeVar` share some common base class.  mypy was OK with doing things this way.
2. Adding this feature caused mypy to complain about using `dtl` to represent a module in some of the array classes, so I removed it and explicitly imported the classes.

### DIFF
--- a/pandas-stubs/core/indexes/accessors.pyi
+++ b/pandas-stubs/core/indexes/accessors.pyi
@@ -13,6 +13,7 @@ from pandas import (
     Index,
     PeriodIndex,
     Timedelta,
+    TimedeltaIndex,
 )
 from pandas.core.accessor import PandasDelegate
 from pandas.core.arrays import (
@@ -156,6 +157,7 @@ _DTRoundingMethodReturnType = TypeVar(
     TimedeltaSeries,
     TimestampSeries,
     DatetimeIndex,
+    # TimedeltaIndex
 )
 
 class _DatetimeRoundingMethods(Generic[_DTRoundingMethodReturnType]):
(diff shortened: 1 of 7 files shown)


pandas-stubs/core/indexes/accessors.pyi
```diff
@@ -379,3 +388,32 @@ class DatetimeIndexProperties(
...
+    def round(
+        self,
+        freq: str | BaseOffset | None,
+        ambiguous: Literal["raise", "infer", "NaT"] | np_ndarray_bool = ...,
+        nonexistent: Literal["shift_forward", "shift_backward", "NaT", "raise"]
+        | Timedelta = ...,
+    ) -> TimedeltaIndex: ...
+    def floor(
+        self,
+        freq: str | BaseOffset | None,
```
twoertwein, line 408: Can be in a followup PR: `_typing` in pandas defines `Frequency` as `BaseOffset | str`, could use this here

twoertwein (approved):
I'm glad we have tests - reviewing these properties without them is almost impossible :)

twoertwein:
Thanks @Dr-Irv

PR #310, opened 2022-09-17
Support Callable conditions for DataFrame.where() and Series.where()

gandhis1:

### DIFF
--- a/tests/test_series.py
+++ b/tests/test_series.py
@@ -1192,3 +1192,20 @@ def test_types_to_numpy() -> None:
     check(assert_type(s.to_numpy(), np.ndarray), np.ndarray)
     check(assert_type(s.to_numpy(dtype="str", copy=True), np.ndarray), np.ndarray)
     check(assert_type(s.to_numpy(na_value=0), np.ndarray), np.ndarray)
+
+
+def test_where() -> None:
+    s = pd.Series([1, 2, 3], dtype=int)
+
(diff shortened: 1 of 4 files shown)


tests/test_series.py
```diff
@@ -1192,3 +1192,20 @@ def test_types_to_numpy() -> None:
...
     check(assert_type(s.to_numpy(na_value=0), np.ndarray), np.ndarray)
+
+
+def test_where() -> None:
+    s = pd.Series([1, 2, 3], dtype=int)
+
+    def cond1(x: int) -> bool:
+        return x % 2 == 0
+
+    check(assert_type(s.where(cond1), pd.Series[int]), pd.Series, int)
```
Dr-Irv, line 1203: Because `pd.Series[int]` is only known to the type checker, and not to the runtime, you need to surround that type in quotes, e.g.:
```python
    check(assert_type(s.where(cond1), "pd.Series[int]"), pd.Series, int)
```

Dr-Irv (changes requested):
This is a nice first-time  contribution.  Just have to fix the tests as indicated for `Series`.

tests/test_series.py
gandhis1: Amended previous commit. I was previously only running the `mypy` poetry workflow and so missed the actual test failures. I now confirmed everything is passing under `test`.

Dr-Irv (approved):
Thanks @gandhis1

PR #311, opened 2022-09-17
Index slice

### LINKED ISSUE #300: IndexSlice typing is no longer correct
**Describe the bug**
Prior to #282, `pd.IndexSlice` returned a tuple with specific types.  That is no longer the case.

**To Reproduce**
```python
import pandas as pd

df = (
    pd.DataFrame({"x": [1, 2, 2, 3, 4], "y": [10, 20, 30, 40, 10]})
    .assign(z=lambda df: df.x * df.y)
    .set_index(["x", "y"])
)
ind = pd.Index([2, 3])
reveal_type(pd.IndexSlice[ind, :])
```
With pandas-stubs 1.4.4.220911, we get:

pyright reports:
```text
isli.py:4:13 - information: Type of "pd.IndexSlice[ind, :]" is "tuple[Unknown, ...]"
```
mypy reports:
```text
isli.py:4: note: Revealed type is "builtins.tuple[Any, ...]"
```

With pandas-stubs 1.4.4.220906, we get:

pyright:
```text
isli.py:9:13 - information: Type of "pd.IndexSlice[ind, :]" is "tuple[str | str_ | bytes | date | datetime | timedelta | bool | complex | Timestamp | Timedelta | slice, ...]"
```

mypy:
```text
isli.py:9: note: Revealed type is "builtins.tuple[Union[Union[builtins.str, numpy.str_], Union[builtins.str, builtins.bytes, datetime.date, datetime.datetime, datetime.timedelta, builtins.bool, builtins.complex, pandas._libs.tslibs.timestamps.Timestamp, pandas._libs.tslibs.timedeltas.Timedelta], builtins.slice], ...]"
```

 - OS: Windows 10
 - python version:  3.10
 - mypy 0.971, pyright 1.269
 - version of installed `pandas-stubs`:  1.4.4.220911

**Additional context**

@bashtage needs to look at this, because change in #282 is what messed this up.

Dr-Irv:
- [x] Closes #300
- [x] Tests added
    - `test_frame.py:test_indexslice_getitem()`

### DIFF
--- a/pandas-stubs/core/indexing.pyi
+++ b/pandas-stubs/core/indexing.pyi
@@ -4,9 +4,14 @@ import numpy as np
 from pandas.core.indexes.api import Index
 
 from pandas._libs.indexing import _NDFrameIndexerBase
-from pandas._typing import IndexSliceTuple
+from pandas._typing import (
+    MaskType,
+    Scalar,
+)
 
-_IndexSliceTupleT = TypeVar("_IndexSliceTupleT", bound=IndexSliceTuple)
+_IndexSliceTuple = slice | tuple[Index | MaskType | Scalar | list[Scalar] | slice, ...]
+
+_IndexSliceTupleT = TypeVar("_IndexSliceTupleT", bound=_IndexSliceTuple)
 
 class _IndexSlice:
     def __getitem__(self, arg: _IndexSliceTupleT) -> _IndexSliceTupleT: ...
--- a/pandas-stubs/core/indexing.pyi
+++ b/pandas-stubs/core/indexing.pyi
@@ -1,12 +1,23 @@
-from typing import TypeVar
+from typing import (
+    TypeVar,
+    Union,
+)
 
 import numpy as np
 from pandas.core.indexes.api import Index
 
 from pandas._libs.indexing import _NDFrameIndexerBase
-from pandas._typing import IndexSliceTuple
+from pandas._typing import (
+    MaskType,
+    Scalar,
+    ScalarT,
+)
 
-_IndexSliceTupleT = TypeVar("_IndexSliceTupleT", bound=IndexSliceTuple)
+_IndexSliceTuple = Union[
+    slice, tuple[Union[Index, MaskType, Scalar, list[ScalarT], slice], ...]
+]
+
+_IndexSliceTupleT = TypeVar("_IndexSliceTupleT", bound=_IndexSliceTuple)
 
 class _IndexSlice:
     def __getitem__(self, arg: _IndexSliceTupleT) -> _IndexSliceTupleT: ...
--- a/pandas-stubs/core/indexing.pyi
+++ b/pandas-stubs/core/indexing.pyi
@@ -1,4 +1,7 @@
-from typing import TypeVar
+from typing import (
+    TypeVar,
+    Union,
+)
 
 import numpy as np
 from pandas.core.indexes.api import Index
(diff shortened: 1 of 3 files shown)


pandas-stubs/core/indexing.pyi
```diff
@@ -1,22 +1,20 @@
...
 
-_IndexSliceT = TypeVar("_IndexSliceT", bound=Union[StrLike, Scalar, slice])
+_IndexSliceTuple = slice | tuple[Index | MaskType | Scalar | list[Scalar] | slice, ...]
+
+_IndexSliceTupleT = TypeVar("_IndexSliceTupleT", bound=_IndexSliceTuple)
 
-class _IndexSlice(Generic[_IndexSliceT]):
-    def __getitem__(self, arg) -> tuple[_IndexSliceT, ...]: ...
+class _IndexSlice:
+    def __getitem__(self, arg: _IndexSliceTupleT) -> _IndexSliceTupleT: ...
```
bashtage, line 17: Isn't it always a tuple, even when slicing 1d?  In my experiments I always say some soft of tuple whether slicing 1 level 2 or 3.

bashtage: Or is slice a subclass of tuple somehow?

tests/test_frame.py
```diff
@@ -1278,6 +1278,18 @@ def test_indexslice_setitem():
...
+def test_indexslice_getitem():
+    # GH 300
+    df = (
+        pd.DataFrame({"x": [1, 2, 2, 3, 4], "y": [10, 20, 30, 40, 10]})
+        .assign(z=lambda df: df.x * df.y)
+        .set_index(["x", "y"])
+    )
+    ind = pd.Index([2, 3])
+    check(assert_type(pd.IndexSlice[ind, :], tuple[pd.Index, slice]), tuple)
+    check(assert_type(df.loc[pd.IndexSlice[ind, :]], pd.DataFrame), pd.DataFrame)
```
bashtage, line 1290: Ideally you would have a test that would show all possible types can be generated.

bashtage (changes requested):
Some questions and a test that shows all types allowed in the typevar can be used.

pandas-stubs/core/indexing.pyi
Dr-Irv: `_IndexSliceTuple` is either a `slice` or a `tuple`.  So if you do `pd.IndexSlice[1:2]`, that returns a `slice` not a `tuple`

tests/test_frame.py
Dr-Irv: Some of them are tested elsewhere, but adding more tests is a good idea.  Will work on that.

pandas-stubs/core/indexing.pyi
twoertwein: I must admit I have never directly used `pd.IndexSlice.__getitem__` - it seems to simply be an identity function.

```py
>>> obj = object()
>>> pd.IndexSlice[obj] is obj
True
```

Dr-Irv: If you are doing slicing on a `MultiIndex`), it really becomes necessary.  See the docs at https://pandas.pydata.org/docs/reference/api/pandas.IndexSlice.html?highlight=indexslice#pandas.IndexSlice for a good example.

tests/test_frame.py
Dr-Irv: I added a bunch of tests in the latest commit.

Dr-Irv:
@bashtage @twoertwein this is all green now, so should be good to go

twoertwein (approved):
LGTM but I'm not familiar with IndexSlice

twoertwein:
Thanks @Dr-Irv

PR #313, opened 2022-09-18
stubtest inconsistencies

twoertwein:
Was trying to go through the "differs from runtime argument" errors from stubtest. Stubtest reports too many (redundant/slightly misleading) errors https://github.com/python/mypy/issues/13676 - I wasn't motivated enough to parse its long output.


pandas-stubs/core/indexes/timedeltas.pyi
```diff
@@ -34,9 +34,9 @@ class TimedeltaIndex(DatetimeTimedeltaMixin, TimedeltaIndexProperties):
     def __truediv__(self, other: num) -> TimedeltaIndex: ...  # type: ignore[override]
     def astype(self, dtype, copy: bool = ...): ...
     def get_value(self, series, key): ...
-    def get_loc(self, key, tolerance=...): ...
+    def get_loc(self, key, method=..., tolerance=...): ...
```
phofl, line 37: Why are you adding this back in? This was effectively deprecated in 1.4

twoertwein: Thank you! I went through the error messages of stubtest - which unfortunately cannot know what is deprecated.

bashtage:
How are you running stubtest?

twoertwein:
> How are you running stubtest?

`poe stubtest` (see `poe --help` for more options) and then grep'ing for specific error phrases.

Could maybe start creating a small allowlist to cover deprecated arguments?

pandas-stubs/core/arrays/numpy_.pyi
```diff
@@ -62,6 +63,7 @@ class PandasArray(ExtensionArray, ExtensionOpsMixin, NDArrayOperatorsMixin):
     ): ...
     def prod(
         self,
+        *,
```
twoertwein, line 66: Many methods in PandasArray are keyword-only but some sub-classes allow also positional arguments for them.

Dr-Irv: I have to ask if we should be including `PandasArray` in the stubs.  While the class is documented, none of its methods are documented.

pandas-stubs/core/frame.pyi
```diff
@@ -1787,6 +1787,7 @@ class DataFrame(NDFrame, OpsMixin):
         min_periods: int | None = ...,
         center: _bool = ...,
         *,
+        win_type: None = ...,
```
Dr-Irv, line 1790: I think `win_type` should be `_str | None`

pandas-stubs/core/indexes/timedeltas.pyi
```diff
@@ -36,7 +36,7 @@ class TimedeltaIndex(DatetimeTimedeltaMixin, TimedeltaIndexProperties):
     def get_value(self, series, key): ...
     def get_loc(self, key, tolerance=...): ...
     def searchsorted(self, value, side: str = ..., sorter=...): ...
-    def is_type_compatible(self, typ) -> bool: ...
+    def is_type_compatible(self, kind: str) -> bool: ...
```
Dr-Irv, line 39: Maybe only fix this in `indexes/base.pyi` and not declare it here.  Not sure why we have it here when it is in the base class.

pandas-stubs/core/missing.pyi
```diff
@@ -1,7 +1,7 @@
 def mask_missing(arr, values_to_mask): ...
 def clean_fill_method(method, allow_nearest: bool = ...): ...
 def clean_interp_method(method, **kwargs): ...
-def find_valid_index(values, how: str): ...
+def find_valid_index(values, how: str = ...): ...
```
Dr-Irv, line 4: Not in the docs, so delete??

pandas-stubs/core/series.pyi
```diff
@@ -1528,6 +1527,7 @@ class Series(IndexOpsMixin, NDFrame, Generic[S1]):
         min_periods: int | None = ...,
         center: _bool = ...,
         *,
+        win_type: None = ...,
```
Dr-Irv, line 1530: `win_type` is `_str | None`

pandas-stubs/util/_test_decorators.pyi
```diff
@@ -13,9 +13,7 @@ skip_if_not_us_locale = ...
 skip_if_no_scipy = ...
 skip_if_no_ne = ...
 
-def skip_if_np_lt(
-    ver_str: str, reason: str | None = ..., *args, **kwds
-) -> Callable: ...
+def skip_if_np_lt(ver_str: str, *args, reason: str | None = ...) -> Callable: ...
```
Dr-Irv, line 16: don't think any of this is documented, so maybe we should just delete this file

Dr-Irv (changes requested):

pandas-stubs/core/arrays/numpy_.pyi
twoertwein: If none of its methods are documented to be public, it seems reasonable to remove the entire class. For now, I wouldn't want to do that since many sub-classes do not "overwrite" its methods in pandas-stubs (min/max/mean/median/sum/...).

pandas-stubs/core/series.pyi
twoertwein: see https://github.com/pandas-dev/pandas-stubs/pull/313#discussion_r976472297

twoertwein:
> Could maybe start creating a small allowlist to cover deprecated arguments?

I started to have a small allowlist for stubtest:

```
# deprecated arguments
pandas.testing.assert_series_equal
pandas.*.get_loc
# need keyword-only for overloads
pandas.*.rolling
```

There is obviously much more to add. I did not add it, but I'm happy to add it if other people also use stubtest and/or want to extent this list.

pandas-stubs/core/arrays/numpy_.pyi
twoertwein: I removed all methods except one which needs to be "implement", because it is defined as abstract in a parent class.

Dr-Irv: thanks

Dr-Irv (approved):

Dr-Irv:
thanks @twoertwein

PR #316, opened 2022-09-19
ENH: Allow TextClause

bashtage:
Allow sqlalchemy TextClause for sql select statements

### DIFF
--- a/pandas-stubs/io/sql.pyi
+++ b/pandas-stubs/io/sql.pyi
@@ -28,6 +28,10 @@ _SQLConnection = Union[
     sqlite3.Connection,
 ]
 
+_SQLStatement = Union[
+    str, sqlalchemy.sql.expression.Selectable, sqlalchemy.sql.expression.TextClause
+]
+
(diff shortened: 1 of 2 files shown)


pandas-stubs/io/sql.pyi
```diff
@@ -28,6 +28,10 @@ _SQLConnection = Union[
     sqlite3.Connection,
 ]
 
+_SQLStatement = Union[
+    str, sqlalchemy.sql.expression.Selectable, sqlalchemy.sql.elements.TextClause
```
gandhis1, line 32: need to `import sqlalchemy.sql.elements`

bashtage: Thanks.  I agree expression is the right endpoint, also based on https://docs.sqlalchemy.org/en/14/core/sqlelement.html#sqlalchemy.sql.expression.TextClause

Dr-Irv (approved):
thanks @bashtage

PR #326, opened 2022-09-24
TYP: tighten IO Protocols

twoertwein:
and to ensure that changes of https://github.com/pandas-dev/pandas/pull/48144 also work.


Dr-Irv:
It's not clear to me why we have to declare the methods of these various `Buffer` types for the public stubs, as users wouldn't instantiate those classes.

Is the goal to make it so that what we have in public pandas for these types in `_typing.py` is the same as in `_typing.pyi` in pandas-stubs ?

twoertwein:
> It's not clear to me why we have to declare the methods of these various `Buffer` types for the public stubs

Not having some of those mehods/properties can lead to errors such as https://github.com/pandas-dev/pandas/issues/45488 and more obvious errors (not having a `read` method for buffers used for reading - surprise). `pandas._typing` has comments for which feature each method is needed. We cannot use `typing.IO` as it requires way too many methods/properties (see https://github.com/python/typeshed/issues/4212) so I tried to add the minimal needed methods/properties in https://github.com/pandas-dev/pandas/pull/43951 (with an update in https://github.com/pandas-dev/pandas/pull/48144).

> as users wouldn't instantiate those classes.

These are protocols, they don't need to be instantiated (like `typing.IO`). The current tests of pandas-stubs show that all the current IO tests still work fine :)

> Is the goal to make it so that what we have in public pandas for these types in `_typing.py` is the same as in `_typing.pyi` in pandas-stubs ?

That's a secondary goal.

pandas-stubs/_typing.pyi
```diff
@@ -64,21 +64,32 @@ DtypeObj = Union[np.dtype[np.generic], ExtensionDtype]
...
+class ReadBuffer(BaseBuffer, Protocol[AnyStr_cov]):
+    def read(self, __n: int = ...) -> AnyStr_cov: ...
+
+class WriteBuffer(BaseBuffer, Protocol[AnyStr_con]):
+    def write(self, __b: AnyStr_con) -> Any: ...
+    def flush(self) -> Any: ...
 
 class ReadPickleBuffer(ReadBuffer[bytes], Protocol):
-    def readline(self, size: int | None = ...) -> bytes: ...
+    def readline(self) -> bytes: ...
```
twoertwein, line 82: All the buffers are used as input arguments (not as return values), so we can remove not needed optional arguments to be compatible with a wider class of IO functions that might or might not implement optional keywords.

Dr-Irv (approved):

Issue #227, opened 2022-08-25
Version 1.5 changes


Dr-Irv:
Need to update stubs for version 1.5 based on docs in RC:
https://pandas.pydata.org/pandas-docs/version/1.5.0rc0/whatsnew/v1.5.0.html

phofl:
How do you want to handle deprecations of arguments? Remove them from the stubs immediately?

bashtage:
I think once we shift to 1.5 as the target we should fully target 1.5 including all deprecations.

phofl:
So if argument x is deprecated we would just remove it from the stubs, correct?

bashtage:
That is what we have been doing.

Dr-Irv:
List of 1.5 things to handle.  As these are addressed, put a check mark and edit this comment to indicate which PR addressed the issue:
- [ ] `pd.ArrayDType`
- [ ] `[DataFrame.__dataframe__()]`
- [ ] `pandas.api.interchange.from_dataframe() `
- [ ] ` Styler.concat() `
- [ ] `Styler.to_string()`
- [ ] `Styler.relabel_index() `
- [ ] ` Styler.highlight_null()` now accepts `color `and deprecates `null_color`
- [ ] The argument `group_keys` has been added to the method `DataFrame.resample()`
- [ ] `pd.from_dummies`
- [ ] ` DataFrame.to_orc()`
- [ ] `read_xml` now supports `dtype`, `converters`, and `parse_dates`
- [ ] `read_xml` now supports large XML using `iterparse`
- [ ] `MultiIndex.to_frame()` now supports the argument `allow_duplicates`
- [ ] `DataFrame.plot()` will now allow the `subplots` parameter to be a list of iterables specifying column groups
- [ ] `Series.reset_index()` and `DataFrame.reset_index(`) now support the argument `allow_duplicates`
- [ ] `read_csv()` now supports `defaultdict` as a `dtype` parameter
- [ ] `DataFrame.rolling()` and `Series.rolling()` now support a `step `
- [ ] Add milliseconds field support for DateOffset 
- [ ] DataFrame.reset_index() now accepts a names argument 
[...]

PR #330, opened 2022-09-26
ENH: Add from_dummies

bashtage:
- [X] xref #227

### DIFF
--- a/pandas-stubs/core/reshape/encoding.pyi
+++ b/pandas-stubs/core/reshape/encoding.pyi
@@ -16,7 +16,7 @@ from pandas._typing import (
 
 def get_dummies(
     data: ArrayLike | DataFrame | Series,
-    prefix: str | Iterable[str] | dict[Hashable, str] | None = ...,
+    prefix: str | Iterable[str] | dict[HashableT, str] | None = ...,
     prefix_sep: str = ...,
     dummy_na: bool = ...,
     columns: list[HashableT] | None = ...,
--- a/pandas-stubs/core/reshape/encoding.pyi
+++ b/pandas-stubs/core/reshape/encoding.pyi
@@ -16,7 +16,7 @@ from pandas._typing import (
 
 def get_dummies(
     data: ArrayLike | DataFrame | Series,
-    prefix: str | Iterable[str] | dict[str, str] | None = ...,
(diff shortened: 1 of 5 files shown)


pandas-stubs/core/reshape/encoding.pyi
```diff
@@ -0,0 +1,31 @@
...
+
+from pandas._typing import (
+    ArrayLike,
+    Dtype,
+    HashableT,
+)
+
+def get_dummies(
+    data: ArrayLike | DataFrame | Series,
+    prefix: str | Iterable[str] | dict[str, str] | None = ...,
```
Dr-Irv, line 19: For the `dict` case, docs say "prefix can be a dictionary mapping column names to prefixes", so maybe this should be `dict[Hashable, str]`

pandas-stubs/core/reshape/reshape.pyi
```diff
@@ -1,16 +1,9 @@
...
-from pandas.core.frame import DataFrame
+from pandas import (
+    DataFrame,
+    Series,
+)
 
-def unstack(obj, level, fill_value=...): ...
-def stack(frame, level: int = ..., dropna: bool = ...): ...
+def unstack(obj: Series | DataFrame, level, fill_value: object | None = ...): ...
+def stack(frame: DataFrame, level: int = ..., dropna: bool = ...): ...
```
Dr-Irv, line 8: I don't think that `stack` or `unstack` are public in `reshape`.  We have `DataFrame.stack()`, `DataFrame.unstack()`, etc. so I think these 2 can be deleted.

Dr-Irv (changes requested):
Some small things to address

pandas-stubs/core/reshape/encoding.pyi
bashtage: Fixed.

pandas-stubs/core/reshape/reshape.pyi
bashtage: I removed the entire module.  CHecked other function, not in docs.

Dr-Irv: OK, good idea

pandas-stubs/core/reshape/encoding.pyi
Dr-Irv: thanks

PR #331, opened 2022-09-26
ENH: add from_dataframe and DataFrame

bashtage:
- [X] xref #227

### DIFF
--- a/pandas-stubs/core/interchange/dataframe_protocol.pyi
+++ b/pandas-stubs/core/interchange/dataframe_protocol.pyi
@@ -1,0 +1,117 @@
+import abc
+from abc import (
+    ABC,
+    abstractmethod,
+)
+import enum
+from typing import (
+    Any,
+    Iterable,
+    Sequence,
+    TypedDict,
+)
+
+class DlpackDeviceType(enum.IntEnum):
+    CPU: int
+    CUDA: int
+    CPU_PINNED: int
+    OPENCL: int
+    VULKAN: int
+    METAL: int
+    VPI: int
+    ROCM: int
+
+class DtypeKind(enum.IntEnum):
+    INT: int
+    UINT: int
+    FLOAT: int
+    BOOL: int
+    STRING: int
+    DATETIME: int
+    CATEGORICAL: int
+
+class ColumnNullType(enum.IntEnum):
+    NON_NULLABLE: int
+    USE_NAN: int
+    USE_SENTINEL: int
+    USE_BITMASK: int
+    USE_BYTEMASK: int
+
+class ColumnBuffers(TypedDict):
+    data: tuple[Buffer, Any]
+    validity: tuple[Buffer, Any] | None
+    offsets: tuple[Buffer, Any] | None
+
+class CategoricalDescription(TypedDict):
+    is_ordered: bool
+    is_dictionary: bool
+    categories: Column | None
+
+class Buffer(ABC, metaclass=abc.ABCMeta):
+    @property
+    @abstractmethod
+    def bufsize(self) -> int: ...
+    @property
+    @abstractmethod
+    def ptr(self) -> int: ...
+    @abstractmethod
+    def __dlpack__(self): ...
+    @abstractmethod
+    def __dlpack_device__(self) -> tuple[DlpackDeviceType, int | None]: ...
+
+class Column(ABC, metaclass=abc.ABCMeta):
+    @property
+    @abstractmethod
+    def size(self) -> int: ...
+    @property
+    @abstractmethod
+    def offset(self) -> int: ...
+    @property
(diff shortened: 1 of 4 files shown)


pandas-stubs/core/interchange/dataframe_protocol.pyi
```diff
@@ -0,0 +1,117 @@
...
+    @property
+    @abstractmethod
+    def bufsize(self) -> int: ...
+    @property
+    @abstractmethod
+    def ptr(self) -> int: ...
+    @abstractmethod
+    def __dlpack__(self): ...
+    @abstractmethod
+    def __dlpack_device__(self) -> tuple[DlpackDeviceType, int | None]: ...
```
Dr-Irv, line 60: Looking at https://data-apis.org/dataframe-protocol/latest/API.html, I don't think the second element of the tuple can be `None`

pandas-stubs/core/interchange/dataframe_protocol.pyi
```diff
@@ -0,0 +1,117 @@
...
+
+class Column(ABC, metaclass=abc.ABCMeta):
+    @property
+    @abstractmethod
+    def size(self) -> int: ...
```
Dr-Irv, line 65: Looking at https://data-apis.org/dataframe-protocol/latest/API.html, returns `Optional[int]`

pandas-stubs/core/interchange/dataframe_protocol.pyi
```diff
@@ -0,0 +1,117 @@
...
+
+class ColumnBuffers(TypedDict):
+    data: tuple[Buffer, Any]
+    validity: tuple[Buffer, Any] | None
+    offsets: tuple[Buffer, Any] | None
+
+class CategoricalDescription(TypedDict):
+    is_ordered: bool
+    is_dictionary: bool
+    categories: Column | None
```
Dr-Irv, line 48: Looking at https://data-apis.org/dataframe-protocol/latest/API.html, `categories` should be `dict | None`

Dr-Irv (changes requested):
Unclear to me whether we should add tests for this?  Other comments are related to checking the API definition at https://data-apis.org/dataframe-protocol/latest/API.html, which may reveal that we have issues in the pandas source in how this is typed.

pandas-stubs/core/interchange/dataframe_protocol.pyi
bashtage: All of the types here are straight from pandas main.

Dr-Irv: They may be wrong based on the docs, and are probably not tested within pandas itself, so can we fix here and do a PR there?

bashtage: Here https://github.com/pandas-dev/pandas/blob/396ad412cf4af59af199101d9b5e1e74f45c698a/pandas/core/interchange/dataframe_protocol.py#L273 it indicates that it should be a `Column`.

pandas-stubs/core/interchange/dataframe_protocol.pyi
bashtage: I think the spec type might also be wrong here.  The text indicates that the size is required, either the full DF size or the chunk size if a single chunk.

pandas-stubs/core/interchange/dataframe_protocol.pyi
bashtage: Here

https://github.com/pandas-dev/pandas/blob/a046792de912146582828de1baf3008750d93bd0/pandas/core/interchange/buffer.py#L60

it is `None`

pandas-stubs/core/interchange/dataframe_protocol.pyi
Dr-Irv: I used this to verify the API:  https://data-apis.org/dataframe-protocol/latest/API.html

I think we may have errors in the pandas implementation _if_ that is the "official" standard.

bashtage: I suspect the code in pandas is the de facto even if it differs from the docs.

bashtage: I would also add that I think pandas-stubs should match pandas, and if there is an issue, pandas should change first, then pandas-stubs keep up.

Dr-Irv: Did a little research, and it appears the docs I looked at are not up to date.  The interface changed in https://github.com/data-apis/dataframe-api/pull/69

Dr-Irv (approved):
thanks @bashtage

PR #332, opened 2022-09-26
ENH: Add step to rolling

bashtage:
- [ ] xref #227 

### DIFF
--- a/pandas-stubs/core/frame.pyi
+++ b/pandas-stubs/core/frame.pyi
@@ -56,6 +56,7 @@ from pandas._typing import (
     Axes,
     Axis,
     AxisType,
+    CalculationMethod,
     ColspaceArgType,
     CompressionOptions,
     Dtype,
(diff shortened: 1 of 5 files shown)


pandas-stubs/core/frame.pyi
```diff
@@ -1763,26 +1764,30 @@ class DataFrame(NDFrame, OpsMixin):
     @overload
     def rolling(
         self,
-        window,
+        window: int,
```
Dr-Irv, line 1767: docs say "int, offset, or BaseIndexer subclass"

Dr-Irv (changes requested):
Need to clean up the type of `window`

pandas-stubs/core/frame.pyi
bashtage: Fixed.

Dr-Irv (approved):
thanks @bashtage

PR #335, opened 2022-09-27
MAINT: Exclude OSX-3.10.6

bashtage:
Stop running for now


Dr-Irv:
I don't understand the need for this PR.  I'm seeing successful workflows with all 3 MacOS tests.

bashtage:
You don't see them because I usually rerun the failed tests when OSX 3.10 times out. Here is one where I didn't rerun

https://github.com/pandas-dev/pandas-stubs/actions/runs/3129422399

Dr-Irv:
> You don't see them because I usually rerun the failed tests when OSX 3.10 times out. Here is one where I didn't rerun
> 
> https://github.com/pandas-dev/pandas-stubs/actions/runs/3129422399

I've seen these "operations were cancelled" on Windows and MacOS, with different versions of python.  It's something funky on the github end, since it is so inconsistent.  What I usually do is just go in and rerun the failed job, and it goes through.

bashtage:
It has happened literally every other run on OSX 3.10  I don't think it adds much to have all 3 by 3 and it is a real pain.

Dr-Irv:
I just did a rerun of a failure, and it failed a LOT because mypy upgraded to 0.981 yesterday (9/26)

Dr-Irv:
So we might start seeing lots of failures everywhere.  Will try to look at this later today

bashtage:
I'm going to reopen this as an appeal.  My other PR has not had 3 times outs on OSX 3.10.  Something is just not right on OSX and 3.10.

Dr-Irv:
The fix is to edit `.github/workflows/test.yml` and increase the time limit from 10 minutes to 20 minutes here: https://github.com/pandas-dev/pandas-stubs/blob/46ee67b15d6689ccaff02c336475d8156820c75b/.github/workflows/test.yml#L12

Dr-Irv (approved):
thanks @bashtage 

I was also seeing timeouts on some Windows runs, so I think this will help alleviate this issue

Issue #336, opened 2022-09-27
mypy 0.981 compatibility


Dr-Irv:
mypy 0.981 released on 9/26/22.  If you do a manual re-run of a job, it picks that up, and we get some failures.  Could be mypy bugs, or settings, or errors we need to fix.

Might be able to remove the python 3.10.6 restriction, since I think 0.981 is supposed to fix the issue of using 3.10.7 or later.

twoertwein:
> pandas-stubs/core/series.pyi:440: error: Overloaded function signatures 1 and 2 overlap with incompatible return types  [misc]

This is an actual ambiguity/error as `None` is `Hashable`. Can probably be silenced in the stubs but should still work for users.

> error: No overload variant of "concat" matches argument types "Dict[str, Series[Any]]", "int"  [call-overload]

I think there might be two issues, something about using HashableT (replacing it with Any solves many of those errors - not sure what's going on there) and probably a mypy bug as `axis=1` seems to be infered as `int` and not as `Literal[1]`.

> error: Dict entry 0 has incompatible type "Tuple[str]": "str"; expected "str": "str"  [dict-item]

Could be a mypy bug with how it handles HashableT:  the typevar is used a second time in the declaration of get_dummies (so the two instances have to have the same type) but the second time it is not used because columns is not passed.

> tests/test_resampler.py:277: error: Expression is of type "Union[str, bytes, date, timedelta, complex]", not "Union[str, bytes, date, datetime, timedelta, bool, complex, Timestamp, Timedelta]"  [assert-type]

It seems that `_typing.Scalar` might contain redundant types: `pd.Timedelta` is part of `datetime.timedelta` and so on. It seems that mypy compacts `Scalar` in some cases but not always creating this discrepancy.


I will try to create minimal examples later and report them at mypy.

twoertwein:
@hauntsaninja it might be great to add pandas-stubs to mypy-primer. Pandas-stubs has many annotations and some of them might also be dubious - perfect test case for mypy-primer :)

hauntsaninja:
Thanks, added!

twoertwein:
Mypy 0.981 adds support for unpacking kwargs https://github.com/python/mypy/pull/13471

```py
from typing import TypedDict
from typing_extensions import Unpack
    
class Style(TypedDict, total=False):
    margin: int
    sticky: bool
    
def add_button(label: str, **kwds: Unpack[Style]) -> None:
    ...
```

[...]

PR #337, opened 2022-09-27
Remove use of removed mypy option

hauntsaninja:
I removed this in 0.981

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -156,7 +156,6 @@ warn_no_return = true
 warn_return_any = false # TODO
 warn_unreachable = false # GH#27396
 # Suppressing errors
-show_none_errors = true
 ignore_errors = false
 enable_error_code = "ignore-without-code" # same as in pandas
 # Miscellaneous strictness flags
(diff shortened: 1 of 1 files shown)


Dr-Irv:
So even without this change, we are now failing with `mypy` 0.981.  Tracking in #336

twoertwein:
> So even without this change, we are now failing with `mypy` 0.981. Tracking in #336

I think it will take some time to be compatible with 0.981.

@hauntsaninja For now, it would be good to limit the mypy version in pyproject.toml. Making any change to pyproject.toml invalidates the cached poetry.lock file which in this case installs the new mypy and lets everything fail.

Dr-Irv:
> @hauntsaninja For now, it would be good to limit the mypy version in pyproject.toml. Making any change to pyproject.toml invalidates the cached poetry.lock file which in this case installs the new mypy and lets everything fail.

Good idea

twoertwein (approved):

Dr-Irv (approved):
thanks @hauntsaninja

PR #338, opened 2022-09-27
ENH: Add more 1.5.0 features

bashtage:


bashtage:
More OSX timeout.

image

bashtage:
Second time out of OSX 3.10

image

bashtage:
image

bashtage:
3rd attempt time out.

image

bashtage:
Attempt 4 time out

image

Dr-Irv:
We need to increase the time limit.  Can you modify the other PR to do that?

bashtage:
> We need to increase the time limit. Can you modify the other PR to do that?

I'll try but I think it is stuck and any timeout won't fix it.  It should only take a couple of seconds to finish what it is doing when it stops working.

pandas-stubs/io/parsers/readers.pyi
```diff
@@ -44,7 +48,7 @@ def read_csv(
     | npt.NDArray
     | Callable[[str], bool]
     | None = ...,
-    dtype: DtypeArg | None = ...,
+    dtype: DtypeArg | defaultdict[str, Dtype] | None = ...,
```
Dr-Irv, line 51: doesn't this have to be changed for all the overloads in `read_csv()` ?

Dr-Irv (changes requested):
One small change on `read_csv()`

pandas-stubs/io/parsers/readers.pyi
bashtage: Probably, and also for read_clipboard.

bashtage: Fixed.

tests/test_frame.py
```diff
@@ -821,6 +821,19 @@ def test_types_to_feather() -> None:
             df.to_feather(file)
 
 
+def test_arrow_dtype() -> None:
```
Dr-Irv, line 824: this should go in `test_pandas.py` not `test_frame.py`

pandas-stubs/io/parsers/readers.pyi
Dr-Irv: I'm still seeing that only one of the overloads was changed.

Dr-Irv (changes requested):
1. move a test as indicated
2. Not seeing all overloads for `read_csv()` changed

Dr-Irv (approved):
thanks @bashtage

Dr-Irv:
@bashtage ping when green.  PR looks good, but testing is failing.

bashtage:
`defaultdict` typing is very problematic.  Best we can do for now seems to be just a plain `defaultdict` with no type info.  Using teh obvious choices `typing.DefautlDict[Hashable, Dtype]` of `[HashableT,Dtype]` do not work.  Even `str, Dtype` don't work.

Dr-Irv (approved):
thanks @bashtage

PR #341, opened 2022-09-29
TYP: de-duplicate some type unions

twoertwein:
This will avoid the new behavior/regression in mypy 0.981https://github.com/python/mypy/issues/13760

### DIFF
--- a/pandas-stubs/_typing.pyi
+++ b/pandas-stubs/_typing.pyi
@@ -53,7 +53,7 @@ PandasScalar = Union[bytes, datetime.date, datetime.datetime, datetime.timedelta
 DatetimeLike = Union[datetime.date, datetime.datetime, np.datetime64, Timestamp]
 
 # dtypes
-NpDtype = Union[str, np.dtype[np.generic], type[Union[str, complex, bool, object]]]
+NpDtype = Union[str, np.dtype[np.generic], type[object]]
 Dtype = Union[ExtensionDtype, NpDtype]
 AstypeArg = Union[ExtensionDtype, npt.DTypeLike]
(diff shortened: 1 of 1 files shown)


bashtage:
Lgtm.

twoertwein:
Should have maybe waited for more consensus on the mypy issue, luckily easy to revert :)

twoertwein:
Might also consider skipping flake8-pyi's check for `int | float | complex`. Personally, I would like to avoid redundant unions, but they serve also a documentation pupose for which it might be more explicit to keep redundant unions.

Dr-Irv:
> Might also consider skipping flake8-pyi's check for `int | float | complex`. Personally, I would like to avoid redundant unions, but they serve also a documentation pupose for which it might be more explicit to keep redundant unions.

I'm in favor of the redundancy, because it helps people understand what arguments are possible.  Trying to help the novices, not just the experts like us!

bashtage:
I also prefer redundancy, esp for int | float.

twoertwein:
Opened #342 to revert the change and disable the flake8-pyi check. Also opened a PR for the few pyi files in pandas itself.

AlexWaygood:
(flake8-pyi maintainer here)

In case it's helpful: we deliberately excluded type aliases from Y041 on the grounds that redundant unions often serve valuable documentation purposes. So flake8-pyi will complain about this code:

```python 
x: int | float
```

But it _deliberately won't_ complain about this code:

```python 
from typing_extensions import TypeAlias
_IntOrFloat: TypeAlias = int | float
x: _IntOrFloat
```

twoertwein:
> But it _deliberately won't_ complain about this code:

I think it did that manually :)

PR #342, opened 2022-09-29
Redundant unions

twoertwein:


pandas-stubs/_typing.pyi
```diff
@@ -135,8 +135,12 @@ Scalar = Union[
     str,
     bytes,
     datetime.date,
+    datetime.datetime,
     datetime.timedelta,
+    bool,
     complex,
```
Dr-Irv, line 141: Can we bring back `int` and `float` here as well?

twoertwein: Added

twoertwein: added ignore for

> pandas-stubs/core/groupby/generic.pyi:145: error: Overloaded function signatures 1 and 3 overlap with incompatible return types  [misc]

that was hidden by https://github.com/python/mypy/issues/13775

Dr-Irv (approved):
thanks @twoertwein

PR #343, opened 2022-09-29
make Series.apply() return Series or DataFrame based on callable

### LINKED ISSUE #293: Series.apply not callable when applying tolist()
**Describe the bug**

``Series().apply().tolist()`` errors with 
```
error: "Series[Any]" not callable  [operator]
```

after latest release. 

**To Reproduce**

```
s = pd.Series([1, 2, 3]).apply(lambda x: x).tolist()
```

```
error: "Series[Any]" not callable  [operator]
```
 - OS Version [e.g. 22] 12.5.1
 - python version 3.10
 - version of type checker mypy 0.960
 - version of installed `pandas-stubs` [v1.4.4.220906]

**Additional context**

Dr-Irv:
- [x] Closes #293
- [x] Tests added: 
  - Modified `test_series_apply()` to use `assert_type()` and added tests corresponding to the issue

There are probably other cases we should catch in testing, but this is a good change, IMHO.


pandas-stubs/core/series.pyi
```diff
@@ -663,9 +663,22 @@ class Series(IndexOpsMixin, NDFrame, Generic[S1]):
         *args,
         **kwargs,
     ) -> DataFrame: ...
+    @overload
     def apply(
-        self, func: Callable, convertDType: _bool = ..., args: tuple = ..., **kwds
-    ) -> Series | DataFrame: ...
+        self,
+        func: Callable[..., Scalar],
```
twoertwein, line 669: At least for these overloads it would be fine for the first `func` to return `Hashable`, as `Series` is not hashable. But I'm not sure whether that is true: does a function that returns a `tuple` (is hashable) return a `Series` or a `DataFrame` - what about a function returning a `list` (not hashable and not in `Scalar`). I don't mind not covering `list` if that is too messy (or having a fallback overload where `func` returns `object`).

Dr-Irv:
```python
import pandas as pd

s = pd.Series([-10, 2, 2, 3.4, 10, 10])


def retlist(x) -> list:
    return [x, 2 * x]


def rettuple(x) -> tuple:
    return (x, 2 * x)


r1 = s.apply(retlist)
print(r1)
print(type(r1))

r2 = s.apply(rettuple)
print(r2)
print(type(r2))
```
produces
```text
0    [-10.0, -20.0]
1        [2.0, 4.0]
2        [2.0, 4.0]
3        [3.4, 6.8]
4      [10.0, 20.0]
5      [10.0, 20.0]
dtype: object
<class 'pandas.core.series.Series'>
0    (-10.0, -20.0)
1        (2.0, 4.0)
2        (2.0, 4.0)
3        (3.4, 6.8)
4      (10.0, 20.0)
5      (10.0, 20.0)
dtype: object
<class 'pandas.core.series.Series'>
```
I'm guessing you could return any object, except a `Series` and it would still return a `Series`.  Problem here is that there really isn't a way to say "match everything except `Series`".  

I changed it to use `Hashable` in the next commit.

twoertwein: I rarely use apply: how are np.ndarray and pd.Index handled? If that is a common usecase, it might be nice to add them to the return values (same for list: `Hashable | list`).

Would also be happy to wait and add those non-Hashable types that seem popular (are reported in issues).

Dr-Irv: Typically, you are using `apply` to call a function on each item of a `Series`, so you return a single object that would be a member of the new `Series`.

Since you typically don't store any object in a `Series`, with the values being floats, ints, strs, etc., putting in the `Scalar` as values in a `Series` makes sense.  `Hashable` expands that, but it is probably only extending it to rarely used objects that would be stored in the `Series`.

twoertwein (approved):

PR #344, opened 2022-09-29
Permit generators (and any Iterable) in DataFrame initializer

### LINKED ISSUE #324: DataFrame initializer does not accept a Generator of a List as data
A two-dimensional array (list-of-list) is an acceptable value to pass to the `data` argument of the `pd.DataFrame` initializer. This works fine, however when the outer dimension is a `Generator` and not a `List`, this does not work. It seems `ListLikeU` should really be any `Iterable` (noting that this then would overlap somewhat with the `Iterable[Tuple]` annotation).

Example:

```
from typing import Any, List
import pandas as pd

data: List[List[Any]] = [
    [1, 2.0, "something", pd.Timestamp("1/1/2022")],
    [2, 4.0, "other", pd.Timestamp("7/4/2022")]
]
data_gen = (x for x in data)
df = pd.DataFrame(data=data_gen)
```

Note that actual types here are irrelevant, which is why I manually annotated as `Any`. The error remains even when you remove this annotation.

Output:

> error: Argument "data" to "DataFrame" has incompatible type "Generator[List[Any], None, None]"; expected "Union[Union[Sequence[Any], ndarray[Any, Any], Series[Any], Index], DataFrame, Dict[Any, Any], Iterable[Tuple[Hashable, Union[Sequence[Any], ndarray[Any, Any], Series[Any], Index]]], None]"  [arg-type]

gandhis1:
- [X] Closes #324 

### DIFF
--- a/pandas-stubs/core/frame.pyi
+++ b/pandas-stubs/core/frame.pyi
@@ -188,7 +188,7 @@ class DataFrame(NDFrame, OpsMixin):
         data: ListLikeU
         | DataFrame
         | dict[Any, Any]
-        | Iterable[tuple[Hashable, ListLikeU]]
+        | Iterable[ListLikeU | tuple[Hashable, ListLikeU]]
         | None = ...,
         index: Axes | None = ...,
         columns: Axes | None = ...,
--- a/tests/test_frame.py
+++ b/tests/test_frame.py
@@ -4,6 +4,7 @@ from collections import defaultdict
 import csv
 import datetime
 import io
+import itertools
 from pathlib import Path
 from typing import (
     TYPE_CHECKING,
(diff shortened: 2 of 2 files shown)


pandas-stubs/core/frame.pyi
```diff
@@ -188,7 +188,7 @@ class DataFrame(NDFrame, OpsMixin):
         data: ListLikeU
         | DataFrame
         | dict[Any, Any]
-        | Iterable[tuple[Hashable, ListLikeU]]
+        | Iterable[ListLikeU | tuple[Hashable, ListLikeU]]
```
twoertwein, line 191: Currently (on main), data is too wide and too narrow: it allows `str` but doesn't allow generators that do not yield `tuple[Hashable, ListLikeU]`. To be clear, some Generators are allowed as they are just Iterables.

If we want to be too wide, I would recommend:

```py
        data: DataFrame
        # column-wise (can enforce Hashable column 'name')
        | Mapping[Hashable, Iterable]
        # row-wise (cannot enforce Hashable column 'name')
        | Iterable[Iterable]
        # None
        | None = ...,
```
and add a test-case:
```py
from typing import Iterable

import pandas as pd

data: Iterable[Iterable] = iter(
    [
        iter([1, 2.0, "something", pd.Timestamp("1/1/2022")]),
    ]
)
pd.DataFrame(data=data)
```
and a similar one for column-wise.

Dr-Irv: > If we want to be too wide

I have a strong preference for being as narrow as possible (but not too narrow).  Prefer the whack-a-mole approach of adding types that are acceptable.

Dr-Irv (approved):
thanks @gandhis1

pandas-stubs/core/frame.pyi
twoertwein: Wouldn't mind a follow-up PR to separate row/column-wise initialization (and obviously also to make the annotations more narrow: `Iterable[Sequence]` can be a string - but that might be more tricky). The example from the issue was row-wise, but the added annotations are for what was previously meant for the column-wise case (with the current annotations `tuple[Hashable, ListLikeU]` can be removed as a `tuple` is a `Sequence` which is part of `ListLikeU`).

PR #346, opened 2022-09-30
CLEAN: use preferred pattern for index accessors

Dr-Irv:
pyright 1.1.271 had a bug https://github.com/microsoft/pyright/issues/3955 with my preferred way of handling some accessors, that was fixed in 1.1.272, so this PR uses the preferred implementation.


bashtage:
Does this need tests, or do some already exist?

Dr-Irv:
> Does this need tests, or do some already exist?

The tests already exist.  What this change does is make it so we have one set of declarations for `round`, `ceil`, etc., whereas I had to do two before because of the pyright bug.

pandas-stubs/core/indexes/accessors.pyi
```diff
@@ -157,7 +157,7 @@ _DTRoundingMethodReturnType = TypeVar(
     TimedeltaSeries,
     TimestampSeries,
     DatetimeIndex,
-    # TimedeltaIndex
```
twoertwein, line 160: looks like a leftover

twoertwein (approved):
Looks good to me - glad to re-use the definitions from `_DatetimeRoundingMethods[TimedeltaIndex]`!

pandas-stubs/core/indexes/accessors.pyi
Dr-Irv: It was a leftover from the "right" way to do it - that's what is put back in this PR!

twoertwein: oh, stupid me :)

twoertwein:
Thanks @Dr-Irv !

PR #347, opened 2022-09-30
ENH: Improve hashing and eval

bashtage:


Dr-Irv (approved):
thanks @bashtage

Issue #128, opened 2022-07-10
CLEAN:  Reconcile types in pandas._typing.py and pandas-stubs._typing.pyi


Dr-Irv:
There are some inconsistencies in types in `pandas._typing.py` and `pandas-stubs._typing.pyi` .  These should be reconciled.

In addition, types that are only in the stubs should probably be private with a leading underscore `_`

bashtage:
Are there any tools that will help keep stubs in sync with pandas?  When I went through io/stata.py there were a lot of changes since these were originally collatted.

Dr-Irv:
> Are there any tools that will help keep stubs in sync with pandas? When I went through io/stata.py there were a lot of changes since these were originally collatted.

Not that I'm aware of.  Maybe `stubtest` could help

twoertwein:
Might not make sense to fix all of them (some exist on main but not in 1.4.x) but it gives a quick overview what could be changed `poe stubtest | grep "pandas\._typing"`

> pandas._typing.AggFuncTypeDict is not a type
> pandas._typing.Axes is not a Union
> pandas._typing.AxisType is not present at runtime
> pandas._typing.BaseBuffer.__init__ is inconsistent, stub does not have *args argument "args"
> pandas._typing.Buffer is not present at runtime
> pandas._typing.CompressionDict is not a type
> pandas._typing.ConvertersArg is not present at runtime
> pandas._typing.DateTimeErrorChoices is not present at runtime
> pandas._typing.DatetimeLikeScalar is not a TypeVar
> pandas._typing.DtypeNp is not present at runtime
> pandas._typing.FileOrBuffer is not present at runtime
> pandas._typing.FilePathOrBuffer is not present at runtime
> pandas._typing.FilePathOrBytesBuffer is not present at runtime
> pandas._typing.FrameOrSeries is not present at runtime
> pandas._typing.FrameOrSeriesUnion is not present at runtime
> pandas._typing.FuncType is not a recognised type alias
> pandas._typing.Function is not present at runtime
> pandas._typing.GroupByObject is not present at runtime
> pandas._typing.GroupByObjectNonScalar is not present at runtime
> pandas._typing.HashableT is not present at runtime
> pandas._typing.IgnoreRaise is not present at runtime
> pandas._typing.IndexType is not present at runtime
> pandas._typing.IndexingInt is not present at runtime
> pandas._typing.IntervalClosedType is not present at runtime
> pandas._typing.KeysArgType is not present at runtime
[...]

PR #349, opened 2022-09-30
Y026: explicitly declare type aliases as TypeAlias

### LINKED ISSUE #345: Convert typing declarations to use TypeAlias
See discussion here:  https://github.com/pandas-dev/pandas-stubs/pull/341#issuecomment-1262928643

Should we change `_typing.pyi` to make all these types that we create be declared with `TypeAlias` (imported from `typing_extensions`) ?

@twoertwein - your opinion?

twoertwein:
Closes #345

(and remove `FilePathOrBuffer` - it does not exist anymore in pandas)


pandas-stubs/_version.pyi
```diff
@@ -4,4 +4,4 @@ version_json: str = ...
 
 def get_versions(): ...
 
-_stub_version = Literal["1.5.0.220926"]
+_stub_version: Literal["1.5.0.220926"]
```
twoertwein, line 7: Not sure why this is needed.

If we assign Literal as a value, we need a TypeAlias, but the type alias is unused  (private type aliases need to be used) which triggers another check from flake8-pyi.

twoertwein:
Would be nice to slowly move all the TypeAliases that are not in _typing to something like __typing (no need to make things worse for #128)

pandas-stubs/_version.pyi
Dr-Irv: Microsoft is copying the stubs over to the VS Code pylance distribution, so by having this, we can tell what version they copied over.

Dr-Irv (approved):
thanks @twoertwein

PR #350, opened 2022-10-01
Fix issue with Series.loc[x] where x is a tuple specifiying a specific index to get from a MultiIndex

Dr-Irv:
Following code failed:
```python
s = pd.Series(
        [1, 2, 3, 4], index=pd.MultiIndex.from_product([[1, 2], ["a", "b"]]), dtype=int
    )
assert_type(s.loc[1, "a"], int)
```
This PR fixes that.  Issue is that when a `Series` is backed by a `MultiIndex`, you should be able to get specific values by specifying a complete tuple.

Added test in `test_series.py:test_multiindex_loc`


pandas-stubs/core/series.pyi
```diff
@@ -129,6 +129,13 @@ class _iLocIndexerSeries(_iLocIndexer, Generic[S1]):
     ) -> None: ...
 
 class _LocIndexerSeries(_LocIndexer, Generic[S1]):
+    # ignore needed because of mypy.  Overlapping, but we want to distinguish
+    # having a tuple of just scalars, versus tuples that include slices or Index
+    @overload
+    def __getitem__(  # type: ignore[misc]
+        self,
+        idx: Scalar | tuple[Scalar, ...],
+    ) -> S1: ...
```
twoertwein, line 138: > but we want to distinguish having a tuple of just scalars, versus tuples that include slices or Index

Can you elaborate on which two cases this distunishes?

It might be nice to add inline comments which types/overloads are for index/multiindex.

I found two ways to use a tuple for a normal index, but I don't think this is an expected/typical usage:
```py
> type(pd.Series([1]).__getitem__((0, None)))
numpy.ndarray
> pd.Series([1], index=[("a", "b")]).__getitem__(("a", "b"))
1
```

Dr-Irv: If you have a `MultiIndex`, then you can specify a tuple of index values to get a specific element.  You can use a tuple that contains a slice to get more than one value:
```python
>>> s = pd.Series([1,2,3,4], index=pd.MultiIndex.from_product([[1,2], ["a","b"]], names=["nn","ab"]))
>>> s
nn  ab
1   a     1
    b     2
2   a     3
    b     4
dtype: int64
>>> s.loc[1,"b"]
2
>>> s.loc[pd.IndexSlice[1,:]]
ab
a    1
b    2
dtype: int64
```

Dr-Irv: next commit adds comments, and uses `_IndexSliceTuple`.  Note also that `s.loc[1,:]` works in example above.

twoertwein (approved):

twoertwein:
Thank you @Dr-Irv !

Do you think it would be helpful to make Series (and DataFrame) generic in regards to their index to help with such overloads?

Dr-Irv:
> Do you think it would be helpful to make Series (and DataFrame) generic in regards to their index to help with such overloads?

We could certainly try.  We'd have to track the `Index` versus `MultiIndex` in `__new__()` and also modify stubs for things like `set_index()` and `reset_index()`.  Might be tricky with `MultiIndex` a subclass of `Index`, and all the other `Index` subclasses.

To some extent, part of the issue here with `mypy` is that if you have 2 overloads that overlap, but the first overload is a subset of the second overload, you know the first one will match, and the second one is for the cases that don't match the first one.  pyright handles this just fine.

twoertwein:
> Might be tricky with `MultiIndex` a subclass of `Index`

I forgot that for a moment - yes that would create many overlapping overloads (I believe also in pyright). As long as all type checkers pick the first matching overload, it can be fine to use overlapping overloads deliberately.

PR #351, opened 2022-10-01
ENH: Imprpve interval_range and IntervalIndex

bashtage:

### DIFF
--- a/pandas-stubs/_typing.pyi
+++ b/pandas-stubs/_typing.pyi
@@ -44,9 +44,7 @@ PandasScalar: TypeAlias = Union[
 ]
 # Scalar: TypeAlias = Union[PythonScalar, PandasScalar]
 
-DatetimeLike: TypeAlias = Union[
-    datetime.date, datetime.datetime, np.datetime64, Timestamp
-]
+DatetimeLike: TypeAlias = Union[datetime.datetime, np.datetime64, Timestamp]
 
 # dtypes
 NpDtype: TypeAlias = Union[
--- a/pandas-stubs/core/indexes/interval.pyi
+++ b/pandas-stubs/core/indexes/interval.pyi
@@ -1,15 +1,25 @@
-from typing import Hashable
+import datetime as dt
+from typing import (
+    Any,
+    Hashable,
+    Sequence,
+)
 
 import numpy as np
+from pandas import Index
 from pandas.core.indexes.extension import ExtensionIndex
 
 from pandas._libs.interval import (
     Interval as Interval,
     IntervalMixin as IntervalMixin,
 )
+from pandas._libs.tslibs import BaseOffset
 from pandas._typing import (
+    DatetimeLike,
     DtypeArg,
     IntervalClosedType,
+    Label,
+    npt,
 )
 
 from pandas.core.dtypes.dtypes import IntervalDtype as IntervalDtype
@@ -28,7 +38,11 @@ class IntervalIndex(IntervalMixin, ExtensionIndex):
     @classmethod
     def from_breaks(
         cls,
-        breaks,
+        breaks: Sequence[int]
+        | Sequence[float]
+        | Sequence[DatetimeLike]
+        | npt.NDArray[np.int_]
+        | npt.NDArray[np.float_],
         closed: IntervalClosedType = ...,
         name: Hashable = ...,
         copy: bool = ...,
@@ -37,8 +51,16 @@ class IntervalIndex(IntervalMixin, ExtensionIndex):
     @classmethod
     def from_arrays(
         cls,
-        left,
-        right,
+        left: Sequence[int]
+        | Sequence[float]
+        | Sequence[DatetimeLike]
+        | npt.NDArray[np.int_]
+        | npt.NDArray[np.float_],
+        right: Sequence[int]
+        | Sequence[float]
+        | Sequence[DatetimeLike]
+        | npt.NDArray[np.int_]
+        | npt.NDArray[np.float_],
         closed: IntervalClosedType = ...,
         name: Hashable = ...,
         copy: bool = ...,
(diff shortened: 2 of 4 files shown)


pandas-stubs/core/indexes/interval.pyi
```diff
@@ -47,37 +69,57 @@ class IntervalIndex(IntervalMixin, ExtensionIndex):
...
         name: Hashable = ...,
         copy: bool = ...,
         dtype: IntervalDtype | None = ...,
     ) -> IntervalIndex: ...
+    def __contains__(self, key: Any) -> bool: ...
     def astype(self, dtype: DtypeArg, copy: bool = ...) -> IntervalIndex: ...
     @property
     def inferred_type(self) -> str: ...
     def memory_usage(self, deep: bool = ...) -> int: ...
+    def is_monotonic_decreasing(self) -> bool: ...
```
Dr-Irv, line 86: This is a `property` of `Index`, so I think you can delete this here.  The implementation is different, but we can just inherit from `Index`

pandas-stubs/core/indexes/interval.pyi
```diff
@@ -47,37 +69,57 @@ class IntervalIndex(IntervalMixin, ExtensionIndex):
...
+    def is_unique(self) -> bool: ...
```
Dr-Irv, line 87: inherited from `Index`, so delete

pandas-stubs/core/indexes/interval.pyi
```diff
@@ -47,37 +69,57 @@ class IntervalIndex(IntervalMixin, ExtensionIndex):
...
     @property
     def is_overlapping(self) -> bool: ...
-    def get_loc(self, key, tolerance=...) -> int | slice | np.ndarray: ...
+    # Note: tolerance removed as it has no effect
```
Dr-Irv, line 90: Should the pandas docs remove it from the signature?

pandas-stubs/core/indexes/interval.pyi
```diff
@@ -47,37 +69,57 @@ class IntervalIndex(IntervalMixin, ExtensionIndex):
...
+    def get_loc(
+        self,
+        key: Label,
+        method: str | None = ...,
```
Dr-Irv, line 94: `FillnaOptions | Literal['nearest'] | None`

Could fix it in `core/indexes/base.pyi` as well.

pandas-stubs/core/indexes/interval.pyi
```diff
@@ -47,37 +69,57 @@ class IntervalIndex(IntervalMixin, ExtensionIndex):
...
+    ) -> int | slice | np.ndarray: ...
     def get_indexer(
         self,
-        targetArrayLike,
+        target: Index,
         method: str | None = ...,
```
Dr-Irv, line 99: same comment on `method` being specific `Literal` values

pandas-stubs/core/indexes/interval.pyi
```diff
@@ -47,37 +69,57 @@ class IntervalIndex(IntervalMixin, ExtensionIndex):
...
         limit: int | None = ...,
         tolerance=...,
     ) -> np.ndarray: ...
```
Dr-Irv, line 102: `np.NDArray[np.intp]`

pandas-stubs/core/indexes/interval.pyi
```diff
@@ -47,37 +69,57 @@ class IntervalIndex(IntervalMixin, ExtensionIndex):
...
-    def __gt__(self, other): ...
-    def __ge__(self, other): ...
 
 def interval_range(
-    start=..., end=..., periods=..., freq=..., name=..., closed: str = ...
-): ...
+    start: int | float | DatetimeLike | None = ...,
+    end: int | float | DatetimeLike | None = ...,
+    periods: int | None = ...,
+    freq: int | str | BaseOffset | None = ...,
```
Dr-Irv, line 122: docs say `DateOffset` here, not `BaseOffset`

Dr-Irv (changes requested):

pandas-stubs/core/indexes/interval.pyi
bashtage: Removed.

pandas-stubs/core/indexes/interval.pyi
bashtage: I think it has to stay due to inheritance.  I changed the comment.

pandas-stubs/core/indexes/interval.pyi
bashtage: Fixed in both.

pandas-stubs/core/indexes/interval.pyi
bashtage: Fixed.

pandas-stubs/core/indexes/interval.pyi
bashtage: Fixed.

pandas-stubs/core/indexes/interval.pyi
bashtage: Fixed.

pandas-stubs/core/indexes/interval.pyi
bashtage: Removed.

bashtage:
Tests added, comments fixed, and green.

pandas-stubs/core/indexes/interval.pyi
```diff
@@ -25,6 +31,20 @@ from pandas._typing import (
 from pandas.core.dtypes.dtypes import IntervalDtype as IntervalDtype
 from pandas.core.dtypes.generic import ABCSeries
 
+_Edges: TypeAlias = Union[
```
Dr-Irv, line 34: nice idea

Dr-Irv (approved):
thanks @bashtage

pandas-stubs/_typing.pyi
```diff
@@ -44,9 +44,7 @@ PandasScalar: TypeAlias = Union[
 ]
 # Scalar: TypeAlias = Union[PythonScalar, PandasScalar]
 
-DatetimeLike: TypeAlias = Union[
-    datetime.date, datetime.datetime, np.datetime64, Timestamp
-]
+DatetimeLike: TypeAlias = Union[datetime.datetime, np.datetime64, Timestamp]
```
gandhis1, line 47: It appears that `date` is incompatible with this change. `datetime` is a sub class of `date`, so the opposite change - removing `datetime`, would probably work.

bashtage: Explainer for the narrowing is in #369

PR #352, opened 2022-10-03
ENH: Improve general datetime functions

bashtage:


bashtage:
Still a few TODOs to verify actual types

pandas-stubs/core/indexes/datetimes.pyi
```diff
@@ -87,23 +90,24 @@ def date_range(
...
     periods: int | None = ...,
+    # TODO: Test timedelta and Timedelta, update pandas docs
     freq: str | timedelta | Timedelta | BaseOffset = ...,
     tz: str | tzinfo = ...,
     normalize: bool = ...,
-    name: str | None = ...,
-    inclusive: IntervalClosedType = ...,
-    **kwargs,
+    name: Hashable | None = ...,
+    inclusive: Literal["left", "right"] | None = ...,
```
Dr-Irv, line 98: Based on the docs, this change is incorrect.  `IntervalClosedType` is the right type here.

pandas-stubs/core/indexes/datetimes.pyi
```diff
@@ -87,23 +90,24 @@ def date_range(
...
     normalize: bool = ...,
-    name: str | None = ...,
+    name: Hashable | None = ...,
     weekmask: str | None = ...,
-    holidays: list | None = ...,
-    inclusive: IntervalClosedType = ...,
-    **kwargs,
+    # TODO: Check if dt.date is allowed
+    holidays: list[str | DatetimeLike] | None = ...,
+    inclusive: Literal["left", "right"] | None = ...,
```
Dr-Irv, line 112: same comment as above about `inclusive`

pandas-stubs/core/indexes/timedeltas.pyi
```diff
@@ -42,5 +50,10 @@ class TimedeltaIndex(DatetimeTimedeltaMixin, TimedeltaIndexProperties):
     def to_series(self, index=..., name=...) -> TimedeltaSeries: ...
 
 def timedelta_range(
-    start=..., end=..., periods=..., freq=..., name=..., closed=...
+    start: TimedeltaConvertibleTypes = ...,
+    end: TimedeltaConvertibleTypes = ...,
```
Dr-Irv, line 54: `start` and `end` can be `"1 day"`, for example, so since the quantity can be anything, both of these should be `str`

pandas-stubs/core/indexes/datetimes.pyi
```diff
@@ -87,23 +92,24 @@ def date_range(
     start: str | DatetimeLike | None = ...,
     end: str | DatetimeLike | None = ...,
     periods: int | None = ...,
+    # TODO: Test timedelta and Timedelta, update pandas docs
```
gandhis1, line 93: These are tested, and the docs are updated (I did them).

tests/test_timefuncs.py
```diff
@@ -534,3 +534,104 @@ def test_types_to_numpy() -> None:
     check(assert_type(td_s.to_numpy(), np.ndarray), np.ndarray)
     check(assert_type(td_s.to_numpy(dtype="int", copy=True), np.ndarray), np.ndarray)
     check(assert_type(td_s.to_numpy(na_value=0), np.ndarray), np.ndarray)
+
+
+def test_to_timdelta_units() -> None:
+    pd.to_timedelta(1, "W")
+    pd.to_timedelta(1, "w")
```
Dr-Irv, line 541: For consistency, should do the `check(assert_type(` pattern here.

Dr-Irv (changes requested):

pandas-stubs/core/indexes/period.pyi
```diff
@@ -52,6 +54,6 @@ def period_range(
     start: str | pd.Period | None = ...,
     end: str | pd.Period | None = ...,
     periods: int | None = ...,
-    freq: str | pd.DateOffset | None = ...,
+    freq: str | BaseOffset | None = ...,
```
Dr-Irv, line 57: docs say `DateOffset` for `freq`, so concerned this could be too wide

Dr-Irv (changes requested):
still some unresolved stuff from prior comments

pandas-stubs/core/indexes/period.pyi
bashtage: I tested with Day() which passes but I'd only BaseOffset not DateOffset.   I think this is a common flaw in the docs.

Dr-Irv: What's not clear to me is whether there is anything that is a `BaseOffset` that is not a ` DateOffset`.  Looked at the pandas source and it does seem that `BaseOffset` is acceptable.

There is this code/comment in `offsets.pyx`:
```python
class OffsetMeta(type):
    """
    Metaclass that allows us to pretend that all BaseOffset subclasses
    inherit from DateOffset (which is needed for backward-compatibility).
    """
```

So I guess we're OK with your change here.

pandas-stubs/core/indexes/timedeltas.pyi
bashtage: `str` is a member of TimedeltaConvertibleTypes.

bashtage: All of the convertible types are tested.

tests/test_timefuncs.py
bashtage: These have been added

Dr-Irv (approved):
thanks @bashtage .  Good set of tests.

PR #354, opened 2022-10-04
ENH: Add typing for to_numeric

bashtage:


pandas-stubs/core/tools/numeric.pyi
```diff
@@ -1 +1,64 @@
...
+    errors: IgnoreRaiseCoerce = ...,
+    *,
+    downcast: Literal["integer", "signed", "unsigned", "float"],
+) -> npt.NDArray: ...
+@overload
+def to_numeric(
+    arg: list | tuple | npt.NDArray,
+    errors: Literal["ignore"],
+    downcast: None = ...,
+) -> npt.NDArray: ...
```
Dr-Irv, line 39: aren't these 2 overloads really the same?  They both return `npt.NDArray`. 

Based on the docs, if you use `errors = "ignore"`, then you get back the input (so maybe it's the same type)?

pandas-stubs/core/tools/numeric.pyi
```diff
@@ -1 +1,64 @@
...
+    errors: IgnoreRaiseCoerce = ...,
+    *,
+    downcast: Literal["integer", "signed", "unsigned", "float"],
+) -> pd.Series: ...
+@overload
+def to_numeric(
+    arg: pd.Series,
+    errors: Literal["ignore"],
+    downcast: None = ...,
+) -> pd.Series: ...
```
Dr-Irv, line 58: again, it seems to me that these 2 overloads don't need to be differentiated, because they return the same result, independent of the arguments, but maybe I'm missing something.

pandas-stubs/core/tools/numeric.pyi
```diff
@@ -1 +1,64 @@
...
+@overload
+def to_numeric(
+    arg: pd.Series,
+    errors: Literal["raise", "coerce"] = ...,
+    downcast: None = ...,
+) -> pd.Series[int] | pd.Series[float]: ...
```
Dr-Irv, line 64: We might be better off just returning `pd.Series` here, because when we return `Union` types, we then might require people to do `cast` operations.

pandas-stubs/core/tools/numeric.pyi
```diff
@@ -1 +1,64 @@
...
+    arg: list | tuple | npt.NDArray,
+    errors: Literal["ignore"],
+    downcast: None = ...,
+) -> npt.NDArray: ...
+@overload
+def to_numeric(
+    arg: list | tuple | npt.NDArray,
+    errors: Literal["raise", "coerce"] = ...,
+    downcast: None = ...,
+) -> npt.NDArray[np.intp] | npt.NDArray[np.float_]: ...
```
Dr-Irv, line 45: Maybe we should just return `npt.NDArray` here (or `np.ndarray`) so that people don't have to `cast` the result.

pandas-stubs/core/tools/numeric.pyi
```diff
@@ -1 +1,64 @@
...
+    Scalar,
+    npt,
+)
+
+@overload
+def to_numeric(
+    arg: Scalar,
+    errors: Literal["raise", "coerce"] = ...,
+    downcast: Literal["integer", "signed", "unsigned", "float"] | None = ...,
+) -> int | float: ...
```
Dr-Irv, line 20: Maybe return `float` here as it is the supertype, and then people don't have to `cast`

Dr-Irv (changes requested):
also need to resolve conflicts

pandas-stubs/_typing.pyi
```diff
@@ -195,7 +195,7 @@ IndexT = TypeVar("IndexT", bound=Index)
 
 IntervalClosedType: TypeAlias = Literal["left", "right", "both", "neither"]
 
-DateTimeErrorChoices: TypeAlias = Literal["ignore", "raise", "coerce"]
+IgnoreRaiseCoerce: TypeAlias = Literal["ignore", "raise", "coerce"]
```
twoertwein, line 198: I like this change but it might be good to have a quick consensus on how we want to name type aliases:

1. Should they reflect what they are (like here `IgnoreRaiseCoerce`)
2. or should they reflect where they are used (more semantic, might leed to duplicate aliases with the same content but different names)

I think currently we mostly use 1, but I can also see the point for 2: when DateTime accepts more error codes, we only want to widen its type and not of all the other IgnoreRaiseCoerce-like error codes in different functions.

I think I slightly prefer 1, as we already went down that path.

Dr-Irv: I prefer (1).

gandhis1: The `_typing` module has quite a few literals that follow 2 - or even more generally than that - the concept, rather than where they are used

pandas-stubs/core/tools/numeric.pyi
bashtage: Nearly the same, but the arguments differ to get the untyped. No complaints form mypy for overlap, and both are tested. 

Docs seem to be wrong:

```python
In [2]: pd.to_numeric([1,2,3],errors="ignore")
Out[2]: array([1, 2, 3], dtype=int64)

In [3]: pd.to_numeric([1,2,3,"none"],errors="ignore")
Out[3]: array([1, 2, 3, 'none'], dtype=object)
```


In either case, if you ignore and it works, then you get an intp or float array (w/o further args set).  If you hit a bad value, you seem to always get object.  In either case, can't tell which will happen statically.

pandas-stubs/core/tools/numeric.pyi
bashtage: Same issue here.  There are certain combinations where the dtype can be object or something numeric, and possibly something else (hard to know).  For standard cases, you either get int or float.

pandas-stubs/core/tools/numeric.pyi
bashtage: Feels like the enhanced specificity helps rather than Series[Any]

pandas-stubs/core/tools/numeric.pyi
bashtage: Do type checkers care about nested types?  mypy seems to accept int | float as just float.

pandas-stubs/core/tools/numeric.pyi
```diff
@@ -1 +1,64 @@
...
+def to_numeric(
+    arg: Scalar,
+    errors: Literal["ignore"],
+    downcast: Literal["integer", "signed", "unsigned", "float"] | None = ...,
+) -> Scalar: ...
+@overload
+def to_numeric(
+    arg: list | tuple | npt.NDArray,
+    errors: IgnoreRaiseCoerce = ...,
+    *,
```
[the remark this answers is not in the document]
bashtage: I only am doing it here so I can have a no-default after a default.

pandas-stubs/_typing.pyi
Dr-Irv: > The reason for this change was that type was first specific to DateTime functions, but then was needed for `to_numeric`, and so needed to have a less date-time-specific name since it seemed strange to have an argument labeled DateTeimErrorChoices in `to_numeric`..

I agree with this approach.

pandas-stubs/core/tools/numeric.pyi
Dr-Irv: With this simple code:
```python
def f2() -> Union[int, float]: ...

x = f2()
reveal_type(x)
```
`pyright` reveals the type as `int | float`.  mypy is making a change to no longer consolidate unions (based on issues we found with 0.981).

So if we can avoid returning a `Union` type it makes it better for people using the stubs.

pandas-stubs/core/tools/numeric.pyi
Dr-Irv: Maybe I'm missing something.  Why not do:
```python
def to_numeric(
    arg: list | tuple | npt.NDArray,
    errors: IgnoreRaiseCoerce = ...,
    downcast: Literal["integer", "signed", "unsigned", "float"] | None = ...
) -> npt.NDArray: ...
```

pandas-stubs/core/tools/numeric.pyi
Dr-Irv: Same response as above.  Could just do:
```python
def to_numeric(
    arg: pd.Series,
    errors: IgnoreRaiseCoerce = ...,
    downcast: Literal["integer", "signed", "unsigned", "float"] | None = ...
) -> pd.Series: ...
```

pandas-stubs/core/tools/numeric.pyi
Dr-Irv: I see your point.  Right now, we don't have any methods in pandas that differentiate between requiring `Series[int]` and `Series[float]` as a parameter, so I guess this is OK for now.

Dr-Irv: On the other hand, as a user, you could have:
```python
    ss = pd.to_numeric(pd.Series([1, 2, 3]))
    reveal_type(ss)
    q = ss[0]
    reveal_type(q)
    reveal_type(ss.loc[1])
```
here `q` and `ss.loc[1]` will have type `int | float` (with pyright), which could be inconvenient and require a `cast`.  But I think you'd have to do a `cast` if we declared it as `Series[float]` or just `Series`.

I'm willing to leave it as you have it.

pandas-stubs/core/tools/numeric.pyi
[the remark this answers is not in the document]
bashtage: I've simplified so removed this for now.

pandas-stubs/core/tools/numeric.pyi
bashtage: I have removed the union returns throughout.

pandas-stubs/core/tools/numeric.pyi
bashtage: I have removed the union returns throughout.

pandas-stubs/core/tools/numeric.pyi
bashtage: Removeing the union returns let me simplify the other cases.

pandas-stubs/core/tools/numeric.pyi
bashtage: Also simplified.

Dr-Irv (approved):
thanks @bashtage

PR #359, opened 2022-10-05
TYP: 'keep' parameter to index.duplicated can be bool

sebwills:
See https://pandas.pydata.org/docs/reference/api/pandas.Index.duplicated.html

> Parameters: keep: {‘first’, ‘last’, False}, default ‘first’


pandas-stubs/core/indexes/base.pyi
```diff
@@ -138,7 +138,7 @@ class Index(IndexOpsMixin, PandasObject):
     def drop_duplicates(
         self, keep: NaPosition | Literal[False] = ...
     ) -> IndexOpsMixin: ...
-    def duplicated(self, keep: _str = ...): ...
+    def duplicated(self, keep: bool | _str = ...): ...
```
twoertwein, line 141: You mentioned that the docs say `‘first’, ‘last’, False`.

how about: `Literal["first", "last", False]`

Dr-Irv: Also, can you add a test?

Dr-Irv (changes requested):

Dr-Irv (approved):
Thanks @sebwills

PR #363, opened 2022-10-06
Allow covariance in the agg dict passed to DataFrame or Series groupby.agg()

### LINKED ISSUE #361: GroupBy agg rejects a dict argument
I'll start with the code snippet and the errors:

```
    import pandas as pd
    df = pd.DataFrame(data=[range(10), ['a'] * 5 + ['b'] * 5]).T
    df.columns = ["value", "category"]
    agg_dict = {col: "sum" for col in df.columns}
    df.groupby("category").agg(agg_dict)
```

> tests/test_pandas.py:324: error: Incompatible types in assignment (expression has type "List[str]", variable has type "Index")  [assignment]
tests/test_pandas.py:326: error: Argument 1 has incompatible type "Dict[Any, str]"; expected "Union[Union[Callable[..., Any], str, ufunc], List[Union[Callable[..., Any], str, ufunc]], Dict[Hashable, Union[Union[Callable[..., Any], str, ufunc], List[Union[Callable[..., Any], str, ufunc]]]]]"  [arg-type]

So there are a several different things happening here:

- For starters, I am unable to assign a `List[str]` to `df.columns`. Not really related to the issue in the title, but might as well fix this at the same time.
- When I create `agg_dict`, I would expect the type to be `Dict[Hashable, str]`, since in a lot of places we are treating the column data type as `Hashable`. This may be dependent on the broader issue of an `Index` needing to be a generic type.
- Even if I override the type of `agg_dict` to `Dict[Hashable, str]` this still is not accepted by `agg()`. There may be a covariance thing at play here, it may need to be `Mapping` instead of `Dict`, but I haven't investigated this too closely yet.

There's enough things going on here that I didn't want to just start changing things around in the annotations before I could pinpoint exactly what we should be changing.

gandhis1:
- [X] Closes #361 


tests/test_frame.py
```diff
@@ -655,21 +660,19 @@ def test_types_groupby_agg() -> None:
         assert_type(df.groupby("col1").agg(["min", "max"]), pd.DataFrame), pd.DataFrame
     )
     check(assert_type(df.groupby("col1").agg([min, max]), pd.DataFrame), pd.DataFrame)
+    agg_dict1: dict[Hashable, str] = {"col2": "min", "col3": "max", 0: "sum"}
```
twoertwein, line 663: Does it also work when the type is `dict[str, str]`? Might need to change the keys to `HashableT` (or simply `Any`).

gandhis1: Fixed

I mentioned this in my other PR, but it wasn't clear to me what the consensus type was for a column or index label. I see `Hashable` in some places, and `Scalar` in others.

tests/test_frame.py
```diff
@@ -655,21 +660,23 @@ def test_types_groupby_agg() -> None:
...
```
Dr-Irv, line 663: Can you add a comment above this line indicating why `Hashable` has to be used?  Something along the idea that the type-checkers may or may not infer the type of the `dict` correctly.

Would `agg_dict1: dict[Union[int, str], str] = {"col2": "min", "col3": "max", 0: "sum"}` work?  (which might be more natural to a user than using `Hashable`)

tests/test_frame.py
```diff
@@ -655,21 +660,23 @@ def test_types_groupby_agg() -> None:
...
+    check(assert_type(df.groupby("col1").agg(agg_dict1), pd.DataFrame), pd.DataFrame)
+    agg_dict2: dict[Hashable, AggFuncTypeBase] = {"col2": min, "col3": max, 0: min}
+    check(assert_type(df.groupby("col1").agg(agg_dict2), pd.DataFrame), pd.DataFrame)
+    agg_dict3: dict[Hashable, str | AggFuncTypeBase] = {
```
Dr-Irv, line 667: same comment as above with respect to `AggFuncTypeBase`

tests/test_frame.py
```diff
@@ -655,21 +660,23 @@ def test_types_groupby_agg() -> None:
...
+        "col2": min,
+        "col3": "max",
+        0: lambda x: x.min(),
```
Dr-Irv, line 670: `lambda` problems are always a problem because type checkers can't infer them.  If `Callable` works, then it's OK

Dr-Irv (changes requested):

tests/test_frame.py
gandhis1: So of the various dict examples, only one actually needed to be annotated, and I used a union and a `Callable` as per your suggestion. The rest, the type checkers were able to successfully choose a default type that was compatible. I'm not sure exactly what the threshold is in terms of number of variables where MyPy chooses to graduate from an inferred `Union` to an inferred `object`, here it appeared to be 3 different types, but elsewhere I've seen more.

pandas-stubs/_typing.pyi
```diff
@@ -127,9 +127,9 @@ F = TypeVar("F", bound=FuncType)
 HashableT = TypeVar("HashableT", bound=Hashable)
 
 AggFuncTypeBase: TypeAlias = Union[Callable, str, np.ufunc]
-AggFuncTypeDictSeries: TypeAlias = dict[Hashable, AggFuncTypeBase]
-AggFuncTypeDictFrame: TypeAlias = dict[
-    Hashable, Union[AggFuncTypeBase, list[AggFuncTypeBase]]
+AggFuncTypeDictSeries: TypeAlias = Mapping[HashableT, AggFuncTypeBase]
```
twoertwein, line 130: Not for this PR but maybe for the future: It might be good to actually use a mapping that does not inherit from `dict` for tests that accept any mapping. Some pandas functions specifically check for dict/list/tuple, so it mightbe good to have Sequences/Mappings that do not inherit from list/tuple/dict.

tests/test_frame.py
```diff
@@ -655,21 +657,24 @@ def test_types_groupby_agg() -> None:
         assert_type(df.groupby("col1").agg(["min", "max"]), pd.DataFrame), pd.DataFrame
     )
     check(assert_type(df.groupby("col1").agg([min, max]), pd.DataFrame), pd.DataFrame)
+    agg_dict1 = {"col2": "min", "col3": "max", 0: "sum"}
+    check(assert_type(df.groupby("col1").agg(agg_dict1), pd.DataFrame), pd.DataFrame)
+    agg_dict2 = {"col2": min, "col3": max, 0: min}
+    check(assert_type(df.groupby("col1").agg(agg_dict2), pd.DataFrame), pd.DataFrame)
+    # Here, MyPy infers dict[object, object], so it must be explicitly annotated
```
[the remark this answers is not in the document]
gandhis1: That doesn't seem to make a difference:

> tests/test_frame.py:674: error: Argument 1 has incompatible type "Dict[object, object]"; expected "Union[Union[Callable[..., Any], str, ufunc], List[Union[Callable[..., Any], str, ufunc]], Mapping[Any, Union[Union[Callable[..., Any], str, ufunc], List[Union[Callable[..., Any], str, ufunc]]]]]"  [arg-type]

I changed it to a regular function, although the pattern of passing a `lambda` as a `dict` element in `agg()` is pretty common, I'd say.

Dr-Irv (approved):
Thanks @gandhis1

PR #367, opened 2022-10-06
Extend Scalar sub-type of Index to the __iter__() method

### LINKED ISSUE #365: DataFrame columns should return a type containing Hashable (or Scalar?) elements
Broke this out of #361 so that each issue is more single-purpose. Currently, `DataFrame.columns` returns an `Index`. If you cast it to a `List`, you get a `List[Unknown]`. I am thinking we need to override `__iter__()` on `Index` and make `Iterator` a generic (`Iterator[Hashable]`). Need to look into further, but here is the offending code:

```
    import pandas as pd
    df = pd.DataFrame(data={"col1": [1, 2], {2: ["a", "b"]})
    test = list(df.columns)
    reveal_type(test)   # List[Unknown]
```

gandhis1:
- [X] Closes #365 

As mentioned in that issue, I am a little bit confused about what the type of a column or index label is.....feels like in some places we decided it is a `Hashable` and in other places we decided it's a `Scalar`. But just to stay consistent with the way this was already annotated, I continued to use `Scalar`.


pandas-stubs/core/indexes/base.pyi
```diff
@@ -215,7 +215,7 @@ class Index(IndexOpsMixin, PandasObject):
     def shape(self) -> tuple[int, ...]: ...
     # Extra methods from old stubs
     def __eq__(self, other: object) -> np_ndarray_bool: ...  # type: ignore[override]
-    def __iter__(self) -> Iterator: ...
+    def __iter__(self) -> Iterator[Scalar]: ...
```
Dr-Irv, line 218: Can you try `Iterator[HashableT]` and see if that works?

gandhis1: So if I do nothing except change it to a `HashableT`, I get this:

> tests/test_indexes.py:65: error: Need type annotation for "column"  [var-annotated]
tests/test_indexes.py:66: error: Expression is of type "Any", not "Union[str, bytes, date, datetime, timedelta, bool, int, float, complex, Timestamp, Timedelta]"  [assert-type]
tests/test_indexes.py:67: error: Expression is of type "Any", not "Series[Any]"  [assert-type]
tests/test_indexes.py:74: error: Need type annotation for "column"  [var-annotated]
tests/test_indexes.py:76: error: Need type annotation for "column"  [var-annotated]
tests/test_indexes.py:84: error: Expression is of type "List[HashableT]", not "List[Union[str, bytes, date, datetime, timedelta, bool, int, float, complex, Timestamp, Timedelta]]"  [assert-type]
tests/test_frame.py:1338: error: Need type annotation for "s"  [var-annotated]

This could probably be viable if `Index` was itself a `Generic[HashableT]`. But without it, MyPy has no way over knowing what type `HashableT` is.

That said, one option is to make it `Hashable`, which goes back to my comment about whether we should be using `Hashable` everywhere or `Scalar` everywhere. Currently `__getitem__` returns `Scalar`, so if you want, I can change it to `Hashable`, and also change `__getitem__` as well.

gandhis1: My 2 cents about the general issue - I think sooner or later `Index` needs to become a generic class. You could get carried away with this though, e.g. if you really want exact typing I'd imagine you would want `Series[T, U, V]` where `T` is the value type, `U` is the index type, and `U` is the name type (i.e. corresponding to the column type on a `DataFrame`. Anyway I digress

Dr-Irv: So clearly `Iterator[HashableT]` won't work.  What if you use `Iterator[Hashable]` ?

We should be incremental on these changes.  Make a simple change - get it to work, and then try doing something more (e.g., changing `__getitem__ to return `Hashable`, or the bigger idea of making `Index` be `Generic`)

tests/test_frame.py
```diff
@@ -1342,7 +1342,8 @@ def test_getmultiindex_columns() -> None:
 
 def test_frame_getitem_isin() -> None:
     df = pd.DataFrame({"x": [1, 2, 3, 4, 5]}, index=[1, 2, 3, 4, 5])
-    check(assert_type(df[df.index.isin([1, 3, 5])], pd.DataFrame), pd.DataFrame)
+    bool_array = df.index.isin([1, 3, 5])
+    check(assert_type(df[bool_array], pd.DataFrame), pd.DataFrame)
```
gandhis1, line 1346: The type checkers think this is a `pd.Series`, but I can't see how that can be the case. I tried the `test_not_hashable` approach on `np_ndarray_bool` and it did not fail (so presumably it is not a `Hashable`. Any ideas?

Dr-Irv: Are you sure?  I just did this with the current version of the stubs:
```python
import pandas as pd

def test_frame_getitem_isin() -> None:
    df = pd.DataFrame({"x": [1, 2, 3, 4, 5]}, index=[1, 2, 3, 4, 5])
    bool_array = df.index.isin([1, 3, 5])
    sdf = df[bool_array]
    reveal_type(sdf)
```
Both `mypy` and `pyright` said that `sdf` was `DataFrame`

pandas-stubs/core/frame.pyi
```diff
@@ -487,7 +487,7 @@ class DataFrame(NDFrame, OpsMixin):
     def T(self) -> DataFrame: ...
     def __getattr__(self, name: str) -> Series: ...
     @overload
-    def __getitem__(self, idx: Scalar) -> Series: ...
+    def __getitem__(self, idx: Hashable) -> Series: ...
```
Dr-Irv, line 490: can't use `Hashable` here.  It overlaps with `tuple` in the 3rd overload.  Can you put this back to `Scalar` ?

Dr-Irv (changes requested):

tests/test_frame.py
gandhis1: That is after my change in this PR which updates `__getitem__`

pandas-stubs/core/frame.pyi
gandhis1: Ok, so what is the conclusion here? That we must use `Scalar` to represent the element of an index? Because if that is the case, then I would prefer reverting back all of my `Hashable` changes and only use `Scalar`. I think it would be a bad idea to leave this as `Scalar` and have `__iter__` produce a `Hashable` - that would be an inconsistent annotation and likely cause issues.

IMO - `Hashable` would be more correct. So I would suggest instead that I remove `tuple` from the 3rd overload. In fact, the 3rd overload appears to potentially be inaccurate:

```
import pandas as pd
x = pd.DataFrame([[1, 2, 3]], columns=[(0, 0), (1, 1), (2,2)])
type(x[(0, 0)])
```

> pandas.core.series.Series

It produces a `Series`, not a `DataFrame`, in this context. I can't say whether that is the case in every context.

Dr-Irv: But I think there are additional overlaps with `Hashable`  in the 3rd overload.  Maybe it is the `numpy` objects.  You'd need to do something like `Scalar | tuple` and maybe anything else in that union that is `Hashable`, but not in the 3rd overload.

gandhis1: MyPy failures:

> pandas-stubs/core/frame.pyi:495: error: Overloaded function signatures 1 and 3 overlap with incompatible return types  [misc]
tests/test_frame.py:1347: error: Expression is of type "Series[Any]", not "DataFrame"  [assert-type]

I need to do some more investigating. It isn't obvious to me which of the third overload parameters are `Hashable`...

Pyright does NOT report overlapping overloads - which makes me wonder whether it truly is the case or if it's a bug of some kind in MyPy. That said, it does report the second issue:

>   pandas-stubs/tests/test_frame.py:1359:23 - error: "assert_type" mismatch: expected "DataFrame" but received "Series[Unknown]" (reportGeneralTypeIssues)

Will take another look soon

pandas-stubs/core/frame.pyi
```diff
@@ -497,8 +497,7 @@ class DataFrame(NDFrame, OpsMixin):
     @overload
     def __getitem__(
         self,
-        idx: tuple
-        | Series[_bool]
+        idx: Series[_bool]
```
gandhis1, line 500: `x = pd.DataFrame([[1, 2, 3], [4, 5, 6], [7, 8, 9]], index=["a", "b", "c"])`

```   
   0  1  2
a  1  2  3
b  4  5  6
c  7  8  9
```

`x[[0, 1]]` works

`x[(0, 1)]` does not, because that is interpreted as a column label

gandhis1:
So it appears `NDArray` is being treated as a `Hashable`, which creates overlapping overloads per MyPy. I don't see any way to fix this currently.

So I took a different approach. Went back to annotating using `Scalar`, but then augmenting that `Union` with a `Tuple[Scalar, ...]`. I added some tests to illustrate some of the edge cases. There is one old test that is failing, and not sure what the best course of action there is, I added a note.

tests/test_frame.py
```diff
@@ -1368,8 +1368,10 @@ def test_getmultiindex_columns() -> None:
     res2: pd.DataFrame = df[li]
     res3: pd.DataFrame = df[
         [(i, s) for i in [1] for s in df.columns.get_level_values(1)]
-    ]
-    ndf: pd.DataFrame = df[[df.columns[0]]]
+    ]  # TODO
```
gandhis1, line 1371: This test is failing. I'm inclined to remove it as I do not believe we have any way to differentiate in annotations between a `MultiIndex` and a single-level index. This test only works if we do. Is that correct?

pandas-stubs/core/frame.pyi
```diff
@@ -115,6 +116,7 @@ from pandas.plotting import PlotAccessor
 
 _str = str
 _bool = bool
+_ScalarOrTupleT = TypeVar("ScalarOrTupleT", bound=Scalar | tuple[Scalar, ...])
```
Dr-Irv, line 119: Should be `"_ScalarOrTupleT" .  That's causing the failure right now.

tests/test_frame.py
Dr-Irv: I don't see why it is failing.  We are passing a list of tuples to `__getitem__()`

gandhis1: Here's the error by the way:

```
tests/test_frame.py:1369: error: Value of type variable "_ScalarOrTupleT" of "__getitem__" of "DataFrame" cannot be "Tuple[int, Union[Union[str, bytes, date, datetime, timedelta, bool, int, float, complex, Timestamp, Timedelta], Tuple[Union[str, bytes, date, datetime, timedelta, bool, int, float, complex, Timestamp, Timedelta], ...]]]"  [type-var]
```

The issue is that it was previously assumed that `s` is a `Scalar`, but my changes make it such that `s` may be a `Tuple`. - since `get_level_values` currently returns an `Index`, and I defined `__iter__` above. Not sure what else I can do besides removing the test, since we would need to differentiate between a `df.columns` that is a regular single index as the example is and one that is a `MultiIndex`, and we have no mechanism to do that right now.

pandas-stubs/core/frame.pyi
gandhis1: I updated it to match, but I didn't think the name of a `TypeVar` really mattered aside from debugging?

tests/test_frame.py
Dr-Irv: I think if you put back `list[ScalarT]` in the arguments to `DataFrame.__getitem__()`, this problem will go away.

Dr-Irv (approved):
thanks @gandhis1

PR #370, opened 2022-10-06
update README to say more about the stubs

### LINKED ISSUE #368: DOC: README should better explain why there is a separate stubs package and repo
The README doesn't adequately explain why there is a separate stubs package for pandas type hints (given that the pandas source already contains some type hints), and what the desired end state is regarding type annotations in pandas (e.g. same type hints in pandas and pandas-stubs, or type hints only in pandas-stubs, or something else).

As a user, this is confusing (should I consume pandas-stubs or not?). As a potential contributor it is also confusing (should I add/fix type hints in pandas, or pandas-stubs, or both?).

The README currently says:

> These are public type stubs for [pandas](http://pandas.pydata.org/), following the convention of providing stubs in a separate package, as specified in [PEP 561](https://peps.python.org/pep-0561/#stub-only-packages).

But PEP 561 says:

> For package maintainers wishing to ship stub files containing all of their type information, it is preferred that the *.pyi stubs are alongside the corresponding *.py files. However, the stubs can also be put in a separate package and distributed separately. 

which seems to be "prefer" putting type hints to be in the core pandas package rather than a separate stubs package. So some explanation of why there is a separate stubs package would be really helpful.

Dr-Irv:
- [x] Closes #368 

Just doc changes.


README.md
```diff
@@ -144,6 +144,14 @@ These stubs were initially forked from the Microsoft project <https://github.com
 
 We are indebted to Microsoft and that project for the initial set of public type stubs.  We are also grateful for the original pandas-stubs project at <https://github.com/VirtusLab/pandas-stubs> that created the framework for testing the stubs.
 
+## Differences between type declarations in pandas and pandas-stubs
+
+The <https://github.com/pandas-dev/pandas/> project has type declarations for some parts of pandas, both for the internal and public API's.  Those type declarations are used to make sure that the pandas code is _internally_ consistent.
+
+The <https://github.com/pandas-dev/pandas-stubs/> project provides type declarations for the pandas _public_ API.  The philsophy of these stubs can be found at <https://github.com/pandas-dev/pandas-stubs/blob/main/docs/philosophy.md/> While it would be ideal if the `pyi` files in this project would be part of the `pandas` distribution, this would require consistency between the internal type declarations and the public declarations, and the scope of a project to create that consistency is quite large.
```
bashtage, line 151: To me another goal is to iterate more rapidly that would be possible.  

Are these any eventual plans that these stubs would migrate back to pandas?  If so, might be worth mentioning.

bashtage (approved):
LGTM.  A small comment but doesn't need to be changed.

README.md
Dr-Irv: > To me another goal is to iterate more rapidly that would be possible.
> 
> Are these any eventual plans that these stubs would migrate back to pandas? If so, might be worth mentioning.

Both good points.  I did some edits to address those.  I do think there is a long term goal to make these part of pandas, but I could see that taking a loooonnnngggg time to accomplish.

twoertwein: > To me another goal is to iterate more rapidly that would be possible.
> 
> Are these any eventual plans that these stubs would migrate back to pandas? If so, might be worth mentioning.

I think for that, we really need to push to reduce the stubtest errors! Then we can follow https://github.com/python/typing/pull/1244 (I tried both of them at some point - one doesn't cope with some new annotation features from `__future__` and the other one presumably failed because the stubs and the implementation is not well aligned). I added autotyping at some point to pandas (mainly to force new code to have annotations :) - much easier than if someone unfamiliar with the code, like me, adds annotations)

twoertwein: I think it would definitely be great to think which aspects would we like to port to pandas: making Series generic is one of the major sources why pandas-stub's test fail with the inline annotations. Making Series in pandas generic would be a great step.

Dr-Irv: > I think it would definitely be great to think which aspects would we like to port to pandas: making Series generic is one of the major sources why pandas-stub's test fail with the inline annotations. Making Series in pandas generic would be a great step.

Two sides of this.
1. In the near term, we want to have stubs that people can use with their own code, so the stubs can "do more" than what pandas does.
2. In the longer term, we can figure out which parts of the stubs should be put into pandas.  I agree that having `Series` be `Generic` in pandas would be a great step, but my sense is that is a big effort, maybe could be a goal for pandas 3.0.  Could be worth a PDEP.

twoertwein: Just to be clear: I don't want to constrain pandas-stubs! But it would be great if pandas can benefit from pandas-stubs (sort of a one-way transfer).

Dr-Irv: @twoertwein If you're good with this change, can you approve and merge?

twoertwein (approved):
Could maybe itemize some of the points to make it easier to skim, but looks good to me.

PR #371, opened 2022-10-06
ENH: Add dt.date support for functions that accept it

### LINKED ISSUE #369: DatetimeLike annotation does not appear to work with dates
This example is `bdate_range`, but I suspect this issue is spread more widely.

```
    import pandas as pd
    from typing import Optional
    from datetime import date
    a: Optional[date] = date(2022, 1, 1)
    b: Optional[date] = date(2022, 1, 10)
    dr = pd.bdate_range(a, b, freq="d")
```

> error: Argument 2 to "bdate_range" has incompatible type "Optional[date]"; expected "Union[str, Union[datetime, datetime64, Timestamp], None]"  [arg-type]

Tagging @bashtage as I think he's done some work around this recently. It seems to me that we may have made an assumption somewhere that `date` is a subset of `datetime` (it's actually the opposite).

bashtage:
- [X] Closes #369

### DIFF
--- a/pandas-stubs/core/indexes/datetimes.pyi
+++ b/pandas-stubs/core/indexes/datetimes.pyi
@@ -1,5 +1,4 @@
 from datetime import (
-    date,
     timedelta,
     tzinfo,
 )


pandas-stubs/core/indexes/datetimes.pyi
```diff
@@ -113,15 +113,15 @@ def bdate_range(
...
+    end: str | DateAndDatetimeLike | None = ...,
     periods: int | None = ...,
     *,
     freq: str | timedelta | Timedelta | BaseOffset,
     tz: str | tzinfo = ...,
     normalize: bool = ...,
     name: Hashable | None = ...,
     weekmask: str | None = ...,
-    holidays: Sequence[str | DatetimeLike | date],
+    holidays: Sequence[str | DateAndDatetimeLike | date],
```
gandhis1, line 125: ` | date` could be removed here

Dr-Irv (changes requested):
Small change as pointed out by @gandhis1

Dr-Irv (approved):
thanks @bashtage

PR #373, opened 2022-10-07
ENH: Allow `None` for `converters` parameter in `read_csv` and `read_table` functions

skatsuta:
## Problem

The default value for `converters` parameter in [`read_csv`](https://pandas.pydata.org/docs/reference/api/pandas.read_csv.html) and [`read_table`](https://pandas.pydata.org/docs/reference/api/pandas.read_table.html) is `None`, but the current stub does not accept it when it is passed to `converters` explicitly like this:

```py
df = read_csv(path, converters=None)
```

Therefore, for example Pyright reports the following error for the above line:

```shell
error: Argument of type "None" cannot be assigned to parameter "converters" of type "dict[int | str, (str) -> Any]" in function "read_csv"
    Type "None" cannot be assigned to type "dict[int | str, (str) -> Any]" (reportGeneralTypeIssues)
```

## Solution

Add `None` to type annotation for the `converters` parameter.

Also add tests that type-check code that explicitly passes `None` to the parameter.


Dr-Irv (approved):
Thanks @skatsuta .  Nice first contribution

skatsuta:
@Dr-Irv Thanks for your quick review and merge!

Issue #348, opened 2022-09-30
bitwise operators on Series should support int


Dr-Irv:
**Describe the bug**
An integer should be allowed for the bitwise operators.  

**To Reproduce**
```python
import pandas as pd

s = pd.Series([1, 2, 3])

ll = 2

result = s & 2

print(result)
```

pyright reports: 
```text
Operator "&" not supported for types "Series[Unknown]" and "Literal[2]"
```
mypy reports:
```text
error: Unsupported operand types for & ("Series[Any]" and "int")  
```

**Please complete the following information:**
 - OS: Windows 10
 - python version:  3.10
 - version of type checker   mypy 0.971, pyright 1.1.273
 - version of installed `pandas-stubs`:  1.5.0.220926

Dr-Irv:
Add `int` argument to here:
https://github.com/pandas-dev/pandas-stubs/blob/cd62cb8f1853276aa690bb9c2218ec0183c0ff55/pandas-stubs/core/series.pyi#L1196

Also look at `__rand__()`, `__xor__()`, `__or__()`, `__rxor__`, `__ror__()`

HebiKotei:
Would you mind if I took a look at this over the next few days?

Dr-Irv:
> Would you mind if I took a look at this over the next few days?

Please go ahead!

HebiKotei:
take

ramvikrams:
sir I am a beginner so do we have to do
***
 def __and__(self, other: _ListLike | Series[S1,int]) -> Series[_bool]: ... 

***
int the __and__ function

Dr-Irv:
@ramvikrams no, you do something like this:
```python
 def __and__(self, other: int | _ListLike | Series[S1]) -> Series[_bool]: ... 
```
And for the other methods mentioned.

Then you have to write tests, e.g., 
```python
s = pd.Series([1,2,3,4])
check(assert_type(s & 3, pd.Series), pd.Series)
```

ramvikrams:
thank you sir for your explanation can I create a pull request for this

Dr-Irv:
> thank you sir for your explanation can I create a pull request for this

Yes, although @hebikotel might be already working on it.  Whoever creates one first is the one we'll concentrate on

ramvikrams:
Sir for adding  the tests in tests/test_series.py it should be added under some function or I can add solely at the end

Dr-Irv:
[...]

PR #374, opened 2022-10-07
added int to bitwise operators

ramvikrams:
For #348
added int to all bitwise operators


ramvikrams:
added the tests

PR #376, opened 2022-10-07
ENH: Improve typing of crosstab

bashtage:


tests/test_pandas.py
```diff
@@ -316,3 +316,100 @@ def test_eval():
...
+        assert_type(pd.crosstab(a, b, margins=True, dropna=True), pd.DataFrame),
+        pd.DataFrame,
+    )
+    check(
+        assert_type(pd.crosstab(a, b, colnames=["a"], rownames=["b"]), pd.DataFrame),
+        pd.DataFrame,
+    )
+    rownames: list[tuple] = [("b", 1)]
+    colnames: list[tuple] = [("a",)]
+    # pyright can't figure out that these are hashable
```
Dr-Irv, line 401: The issue here is that you have two arguments that are `list[HashableT]`, and `HashableT` is bound, so what happens is that one of them gets bound to a specific type, and the other one is different.
Consider this example:
```python
from typing import Hashable, TypeVar

HashableT = TypeVar("HashableT", bound=Hashable)
HashableT1 = TypeVar("HashableT1", bound=Hashable)
HashableT2 = TypeVar("HashableT2", bound=Hashable)


rownames: list[tuple] = [("b", 1)]
colnames: list[tuple] = [("a",)]

def f1(rn: list[HashableT], cn: list[HashableT]):
    pass

f1(rownames, colnames)


def f2(rn: list[HashableT1], cn: list[HashableT2]):
    pass

f2(rownames, colnames)
```
In this case, the call to `f1` fails because `rownames` and `colnames` are both `list[HashableT]`, but the call to `f2` works because the bindings are different, even though the `TypeVar` declarations are identical.

bashtage: Why would the type in the HashableT need to be the same?

bashtage: I mostly get it.  Clearly if you have f(x:T)->T: it has to be the same T, so something like f(a:list[T},b:list[T])->None also needs the two T to be the same.

Dr-Irv: > Why would the type in the HashableT need to be the same?

Not entirely sure, but I think it has to do with the `bound=Hashable` declaration.  If you have a function `g(x: HashableT) -> HashableT`, and then do `result = g("abc")` followed by `reveal_type(result)`, you will get `str`, because pyright said "you passed a string, which fits the expected type, but now I know you are returning a string, and not any other Hashable". 

The same thing is happening here.  You bound one argument, so it expects the second argument to have the same binding.

At least that is what I *think* is going on.

Dr-Irv: > I mostly get it. Clearly if you have f(x:T)->T: it has to be the same T, so something like f(a:list[T},b:list[T])->None also needs the two T to be the same.

Yes, that's how I see it as well.

bashtage: Fixed now.  

I think this is a wide spread problem with use of HashableT.  Pretty must any function that has 2 or more HashableT should have distinct HashableTs (e.g., HashableT0, HashableT1, ...)

Dr-Irv: > Fixed now.
> 
> I think this is a wide spread problem with use of HashableT. Pretty must any function that has 2 or more HashableT should have distinct HashableTs (e.g., HashableT0, HashableT1, ...)

I think @twoertwein knows more about this.  I recall some discussion we had about this in the past

pandas-stubs/core/reshape/pivot.pyi
```diff
@@ -33,12 +40,12 @@ def pivot(
...
 def crosstab(
-    index: Sequence | Series,
-    columns: Sequence | Series,
-    values: Sequence | None = ...,
-    rownames: Sequence | None = ...,
-    colnames: Sequence | None = ...,
-    aggfunc: Callable | None = ...,
+    index: list | npt.NDArray | Series | list[Sequence | npt.NDArray | Series],
+    columns: list | npt.NDArray | Series | list[Sequence | npt.NDArray | Series],
+    values: list | npt.NDArray | Series | None = ...,
```
Dr-Irv, line 45: I think all 3 of these should be `AnyArrayLike | list[AnyArrayLike]`

Both `Index` and `Categorical` work for `index` and `columns`.  Didn't try `values`.

Also, if `values` is not `None`, then `aggfunc` can't be `None`, and vice versa (either both `None` or both are not `None`)

pandas-stubs/core/reshape/pivot.pyi
```diff
@@ -33,12 +40,12 @@ def pivot(
...
+    rownames: list[HashableT] | None = ...,
+    colnames: list[_HashableT2] | None = ...,
+    aggfunc: str | np.ufunc | Callable[[Series], float] | None = ...,
```
Dr-Irv, line 48: Looking at the code, `aggfunc` is `AggFuncType`

Dr-Irv (changes requested):

pandas-stubs/core/reshape/pivot.pyi
bashtage: AggFuncType isn't specific enough (also AggFuncType is no longer part of pandas-stubs).  The input to the aggfunc is always a Series and the output is a scalar (int or float in the standard crosstab case).

tests/test_pandas.py
twoertwein: As far as I understand, the primary purpose of TypeVars is binding types (whether it is between input and output or only between inputs). We often use TypeVars as a workaround to get invariant(? - always mix those up) behaviors for generic classes such as list, dict, Mapping and so on.

pandas-stubs/core/reshape/pivot.pyi
Dr-Irv: OK, I see what you're saying.

pandas-stubs/core/reshape/pivot.pyi
```diff
@@ -4,20 +4,26 @@ from typing import (
...
 
 from pandas._typing import (
+    AnyArrayLike,
     HashableT,
     IndexLabel,
     Scalar,
-    npt,
 )
 
+_ExtendedAnyArrayLike: TypeAlias = Union[AnyArrayLike, pd.Categorical]
```
Dr-Irv, line 25: I think you could pass any `ExtensionArray`, not just `pd.Categorical`, so maybe this should be:
```python
_ExtendedAnyArrayLike: TypeAlias = Union[AnyArrayLike, ArrayLike]
```
since `ArrayLike` includes `ExtensionArray`

tests/test_pandas.py
```diff
@@ -1155,6 +1176,7 @@ def test_crosstab_args() -> None:
     )
 
     def m(x: pd.Series) -> float:
+        print(type(x))
```
Dr-Irv, line 1179: remove the `print`

Dr-Irv (changes requested):

tests/test_pandas.py
bashtage: Fixed.

pandas-stubs/core/reshape/pivot.pyi
bashtage: Fixed.

pandas-stubs/core/reshape/pivot.pyi
bashtage: These are all fixed now to allow the full set of AnyArrayLike | ArrayLike.

Dr-Irv (approved):
thanks @bashtage

PR #377, opened 2022-10-07
BUG: Correct use of HashableT

bashtage:
Ensure that distinct HashableTs are used within the same function. 

Not sure if these distinct instances should be common or private.  I think there is a case for at least 3 common ones, although I needed 5  to handle all uses.


Dr-Irv (approved):
thanks @bashtage .  Can't say that you caught them all, but good enough for now.

Before merging, would like opinion of @twoertwein

pandas-stubs/_typing.pyi
```diff
@@ -253,9 +253,11 @@ XMLParsers: TypeAlias = Literal["lxml", "etree"]
 
 # Any plain Python or numpy function
 Function: TypeAlias = Union[np.ufunc, Callable[..., Any]]
+# Use custom HashableT in shared types to avoid conflicts with shared HashableTs
+_GroupByObjectHashableT = TypeVar("_GroupByObjectHashableT", bound=Hashable)
```
twoertwein, line 257: How about moving it to `_typing` and calling it `HashableTa` or something like that? I'm sure we need it also in other places.

bashtage: The purpose of this one is that it will never be used outside for _typing.  This makes sure that it doesn't accidentally get used in a definition with one of the other HashableT instances.  I've renamed it to be simpler.

bashtage: I see your point, and have now used a public name.

twoertwein: I like your idea of giving it a private name, Could have something like `_HashableT` or `_HashableTa` in `_typing`

Dr-Irv (approved):
thanks @bashtage

PR #381, opened 2022-10-11
fix and, or, xor in logical operations

### LINKED ISSUE #380: `Series[int]` instead of `Series[bool]` for masks combined with binary operators
**Describe the bug**
When combining masks with a binary operator (e.g. &), the returned type is `Series[int]` instead of `Series[bool]`. This means `mypy` complains if you use the mask for indexing.

**To Reproduce**
```
import pandas as pd

df = pd.dataframe({"a": [1, 2, 3, 4, 5]})
mask = (df["a"] < 4) & (df["a"] > 2)
print(df[mask])
```
3. Indicate which type checker you are using (`mypy` or  `pyright`).
`mypy --strict`
```
bug.py:5: error: Invalid index type "Series[int]" for "DataFrame"; expected type "Union[Tuple[Any, ...], Series[bool], DataFrame, List[str], Index, ndarray[Any, dtype[str_]], ndarray[Any, dtype[bool_]], Sequence[Tuple[Union[str, bytes, date, datetime, timedelta, bool, int, float, complex, Timestamp, Timedelta], ...]]]"
```

 - OS: Ubuntu 20.04
 - python version 3.8.10
 - version of type checker: mypy 0.982
 - version of installed `pandas-stubs`: 1.5.0.221010

Dr-Irv:
- [x] Closes #380
- [x] Tests added: 
  - `test_series.py:test_logical_operators()`

Not sure that I am covering all the cases, but this improves where we were.


twoertwein (approved):

twoertwein:
Thanks @Dr-Irv !

PR #385, opened 2022-10-14
Allow str to be passing to rolling() window and use IntervalClosedType literal for series as well

gandhis1:


Dr-Irv (approved):
thanks @gandhis1 

As mentioned elsewhere, can you create an issue in the pandas repo that `DataFrame.rolling()` and `Series.rolling()` should show that a string is a valid argument?

PR #388, opened 2022-10-16
ENH: Improve typing for Timedelta

bashtage:
- [X] Tests added: Please use `assert_type()` to assert the type of any return value

xref #383

PR #389, opened 2022-10-16
ENH: Improve typing for Timestamp

bashtage:
xref #383


Dr-Irv (changes requested):
As noted in #388 , let's get that one right.  I think similar issues will exist here, so apply the same concepts from there (and `Period`) here, and then I will review this one.

pandas-stubs/_libs/tslibs/timestamps.pyi
```diff
@@ -111,17 +128,19 @@ class Timestamp(datetime):
     def date(self) -> _date: ...
     def time(self) -> _time: ...
     def timetz(self) -> _time: ...
-    def replace(
+    # Override since fold is more precise than datetime.replace(fold:int)
+    # Violation of Liskov substitution principle
```
Dr-Irv, line 132: Not sure if it is because of `fold` being `Literal` or because we use `int | None` for most of the arguments, or because in typeshed, there is a `*` before `fold` .  Can you check and update the comment appropriately?

pandas-stubs/_libs/tslibs/timestamps.pyi
```diff
@@ -131,44 +150,86 @@ class Timestamp(datetime):
...
+    def __eq__(self, other: object) -> Literal[False]: ...
+    @overload
+    def __ne__(self, other: Timestamp | datetime | np.datetime64) -> bool: ...  # type: ignore[misc]
+    @overload
+    def __ne__(self, other: TimestampSeries | Series[Timestamp]) -> Series[bool]: ...  # type: ignore[misc]
+    @overload
+    def __ne__(self, other: npt.NDArray[np.datetime64] | Index) -> np_ndarray_bool: ...  # type: ignore[misc]
+    @overload
+    def __ne__(self, other: object) -> Literal[True]: ...
     def __hash__(self) -> int: ...
```
Dr-Irv, line 233: should delete `__hash__`

tests/test_scalars.py
```diff
@@ -15,26 +22,568 @@
...
+def test_timestamp_add_sub() -> None:
+    ts = pd.Timestamp("2000-1-1")
+    np_td64_arr: npt.NDArray[np.timedelta64] = np.array([1, 2], dtype="timedelta64[ns]")
+
+    as_pd_timedelta = pd.Timedelta(days=1)
+    as_dt_timedelta = dt.timedelta(days=1)
+    as_offset = 3 * Day()
+    as_timedelta_index = pd.TimedeltaIndex([1, 2, 3], "D")
+    as_timedelta_series = pd.Series(as_timedelta_index)
+    check(assert_type(as_timedelta_series, TimedeltaSeries), pd.Series)
```
Dr-Irv, line 163: Use `check(assert_type(as_timedelta_series, TimedeltaSeries), pd.Series, pd.Timedelta)` here and elsewhere if checking for a `Series`

tests/test_scalars.py
```diff
@@ -15,26 +22,568 @@
...
+
+    check(assert_type(ts + as_timedelta_index, pd.DatetimeIndex), pd.DatetimeIndex)
+    check(assert_type(as_timedelta_index + ts, pd.DatetimeIndex), pd.DatetimeIndex)
+
+    check(assert_type(ts + as_timedelta_series, "TimestampSeries"), pd.Series)
+    check(assert_type(as_timedelta_series + ts, "TimestampSeries"), pd.Series)
+
+    check(assert_type(ts + as_np_ndarray_td64, npt.NDArray[np.datetime64]), np.ndarray)
+    # pyright and mypy disagree on the type of this expression
+    # pyright: Any, mypy: npt.NDArray[np.datetime64]
```
Dr-Irv, line 183: this is most likely fixed in latest `pyright` release, so can you remove above comment and `#pyright ignore` and test?

tests/test_scalars.py
```diff
@@ -15,26 +22,568 @@
...
+        ),
+        np.ndarray,
+    )
+
+    # Reverse order is not possible for all of these
+    check(assert_type(ts - as_pd_timedelta, pd.Timestamp), pd.Timestamp)
+    check(assert_type(ts - as_dt_timedelta, pd.Timestamp), pd.Timestamp)
+    check(assert_type(ts - as_offset, pd.Timestamp), pd.Timestamp)
+    check(assert_type(ts - as_timedelta_index, pd.DatetimeIndex), pd.DatetimeIndex)
+    check(assert_type(ts - as_timedelta_series, "TimestampSeries"), pd.Series)
```
Dr-Irv, line 197: I think you can remove the quotes here because you declared it above.

tests/test_scalars.py
```diff
@@ -15,26 +22,568 @@
...
+    check(assert_type(ts - as_np_ndarray_td64, npt.NDArray[np.datetime64]), np.ndarray)
```
Dr-Irv, line 198: add `np.datetime64` as last argument to `check`

tests/test_scalars.py
```diff
@@ -15,26 +22,568 @@
...
+    c_datetimeindex = pd.DatetimeIndex(["2000-1-1"])
+    c_np_ndarray_dt64 = np_dt64_arr
+    # Typing provided since there is no way to get a Series[Timestamp],
+    # which is a different type from a TimestampSeries
+    c_series_dt64: pd.Series[pd.Timestamp] = pd.Series(
+        [1, 2, 3], dtype="datetime64[ns]"
+    )
+    c_series_timestamp = pd.Series(pd.DatetimeIndex(["2000-1-1"]))
+    check(assert_type(c_series_timestamp, TimestampSeries), pd.Series)
+    # Use xor to ensure one is True and the other is False
```
Dr-Irv, line 220: don't understand this comment.  Where are you using `xor` ??

tests/test_scalars.py
```diff
@@ -15,26 +22,568 @@
...
+
+    gt = check(assert_type(ts > c_dt_datetime, bool), bool)
+    lte = check(assert_type(ts <= c_dt_datetime, bool), bool)
+    assert gt != lte
+
+    check(assert_type(ts > c_datetimeindex, np_ndarray_bool), np.ndarray)
+    check(assert_type(ts <= c_datetimeindex, np_ndarray_bool), np.ndarray)
+
+    check(assert_type(ts > c_np_ndarray_dt64, np_ndarray_bool), np.ndarray)
+    check(assert_type(ts <= c_np_ndarray_dt64, np_ndarray_bool), np.ndarray)
```
Dr-Irv, line 238: can add either `bool` or `np.bool_` (not sure which) as additional argument to `check` in above 4 tests.

tests/test_scalars.py
```diff
@@ -15,26 +22,568 @@
...
+
+    gt = check(assert_type(c_dt_datetime > ts, bool), bool)
+    lte = check(assert_type(c_dt_datetime <= ts, bool), bool)
+    assert gt != lte
+
+    check(assert_type(c_datetimeindex > ts, np_ndarray_bool), np.ndarray)
+    check(assert_type(c_datetimeindex <= ts, np_ndarray_bool), np.ndarray)
+
+    check(assert_type(c_np_ndarray_dt64 > ts, np_ndarray_bool), np.ndarray)
+    check(assert_type(c_np_ndarray_dt64 <= ts, np_ndarray_bool), np.ndarray)
```
Dr-Irv, line 257: can add `bool` or `np.bool_` as additional argument in above 4 tests

tests/test_scalars.py
```diff
@@ -15,26 +22,568 @@
...
+
+    gte = check(assert_type(ts >= c_dt_datetime, bool), bool)
+    lt = check(assert_type(ts < c_dt_datetime, bool), bool)
+    assert gte != lt
+
+    check(assert_type(ts >= c_datetimeindex, np_ndarray_bool), np.ndarray)
+    check(assert_type(ts < c_datetimeindex, np_ndarray_bool), np.ndarray)
+
+    check(assert_type(ts >= c_np_ndarray_dt64, np_ndarray_bool), np.ndarray)
+    check(assert_type(ts < c_np_ndarray_dt64, np_ndarray_bool), np.ndarray)
```
Dr-Irv, line 278: same comment about adding additional argument to `check`

tests/test_scalars.py
```diff
@@ -15,26 +22,568 @@
...
+    assert gte != lt
+
+    check(assert_type(c_np_dt64 >= ts, Any), np.bool_)
+    check(assert_type(c_np_dt64 < ts, Any), np.bool_)
+
+    check(assert_type(c_datetimeindex >= ts, np_ndarray_bool), np.ndarray)
+    check(assert_type(c_datetimeindex < ts, np_ndarray_bool), np.ndarray)
+
+    check(assert_type(c_np_ndarray_dt64 >= ts, np_ndarray_bool), np.ndarray)
+    check(assert_type(c_np_ndarray_dt64 < ts, np_ndarray_bool), np.ndarray)
```
Dr-Irv, line 297: same comment about additional argument to `check`

tests/test_scalars.py
```diff
@@ -15,26 +22,568 @@
...
+
+    eq = check(assert_type(ts == c_dt_datetime, bool), bool)
+    ne = check(assert_type(ts != c_dt_datetime, bool), bool)
+    assert eq != ne
+
+    check(assert_type(ts == c_datetimeindex, np_ndarray_bool), np.ndarray)
+    check(assert_type(ts != c_datetimeindex, np_ndarray_bool), np.ndarray)
+
+    check(assert_type(ts == c_np_ndarray_dt64, np_ndarray_bool), np.ndarray)
+    check(assert_type(ts != c_np_ndarray_dt64, np_ndarray_bool), np.ndarray)
```
Dr-Irv, line 318: same as above

tests/test_scalars.py
```diff
@@ -15,26 +22,568 @@
...
+    assert eq != ne
+
+    check(assert_type(c_datetimeindex == ts, np_ndarray_bool), np.ndarray)
+    check(assert_type(c_datetimeindex != ts, np_ndarray_bool), np.ndarray)
+
+    check(assert_type(c_np_ndarray_dt64 != ts, Any), np.ndarray)
+    check(assert_type(c_np_ndarray_dt64 == ts, Any), np.ndarray)
+
+    check(assert_type(c_series_dt64 == ts, "pd.Series[bool]"), pd.Series, bool)
+    check(assert_type(c_series_dt64 != ts, "pd.Series[bool]"), pd.Series, bool)
```
Dr-Irv, line 340: can remove the above 5 pairs of tests because `__eq__` and `__ne__` are always sent to the LHS object type method.

Dr-Irv (changes requested):

pandas-stubs/_libs/tslibs/timestamps.pyi
```diff
@@ -90,7 +107,7 @@ class Timestamp(datetime):
     def fromordinal(
         cls: type[_DatetimeT],
         ordinal: int,
-        freq: str | BaseOffset | None = ...,
+        *,
```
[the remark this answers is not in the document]
bashtage:
```
pd.Timestamp.fromordinal(730120, freq="D")
<ipython-input-7-28a2d2898818>:1: FutureWarning: The 'freq' argument in Timestamp is deprecated and will be removed in a future version.
  pd.Timestamp.fromordinal(730120, freq="D")
```

Seems to be deprecated. I'll leave it in but make it `None` only.

pandas-stubs/_libs/tslibs/timestamps.pyi
bashtage: It is because the super method is only int while here it is Literal[0,1]

bashtage: There is no * before fold in pandas main

bashtage: Comment updated.

pandas-stubs/_libs/tslibs/timestamps.pyi
bashtage: Removed.

tests/test_scalars.py
bashtage: Updated all

tests/test_scalars.py
bashtage: Added

tests/test_scalars.py
bashtage: In python when you know `a` and `b` are bool, then  `a != b` is the xor operator.  It ensures that 1 and only 1 is True and avoids directly testing pandas.

tests/test_scalars.py
bashtage: added

tests/test_scalars.py
bashtage: added

tests/test_scalars.py
bashtage: Shouldn't we just leave them in just for good measure to make sure something doesn't break in the future?

tests/test_scalars.py
bashtage: Fixed.

tests/test_scalars.py
bashtage: Added

tests/test_scalars.py
bashtage: It is.  I've bumped pytight to latest too in poetry.

bashtage:
#408 should go in first, but it is already in this one.

tests/test_scalars.py
Dr-Irv: OK, I never thought of `!=` as `xor` !

tests/test_scalars.py
Dr-Irv: OK to leave in, but I think they should be in a different test function, because they are really testing the LHS of `==` and `!=` as opposed to the RHS, so the func name `test_timestamp_cmp` isn't really reflective.  

So can you split it out, and make a comment that indicates that these are tests of the LHS against `Timestamp`

Dr-Irv (changes requested):
One small change about splitting out the `==` and `!=` tests when `Timestamp` is on RHS, otherwise looks good

Dr-Irv (approved):
thanks @bashtage

PR #390, opened 2022-10-16
ENH: Improve typing for Period

bashtage:
xref #383


Dr-Irv:
@bashtage For this and the other new PRs you created, can you click the "request review" item in the PR when you are ready for me to do the review?

pandas-stubs/_libs/tslibs/period.pyi
```diff
@@ -1,34 +1,137 @@
...
+    @overload
+    def __le__(self, other: Period) -> bool: ...
+    @overload
+    def __le__(self, other: PeriodIndex) -> npt.NDArray[np.bool_]: ...
+    @overload
+    def __lt__(self, other: Period) -> bool: ...
+    @overload
+    def __lt__(self, other: PeriodIndex) -> npt.NDArray[np.bool_]: ...
+    @overload  # type: ignore[override]
+    def __ne__(self, other: Period) -> bool: ...
```
Dr-Irv, line 118: same comment as on `__eq__()`

pandas-stubs/core/arrays/boolean.pyi
```diff
@@ -1,24 +1,11 @@
 import numpy as np
 
-from pandas._typing import (
-    Scalar,
-    type_t,
-)
-
```
Dr-Irv, line 7: shouldn't any changes to `core/arrays/boolean.pyi` not be in this PR?

pandas-stubs/core/dtypes/dtypes.pyi
```diff
@@ -93,8 +93,6 @@ class PeriodDtype(PandasExtensionDtype):
     def construct_from_string(cls, string: _str): ...
     @property
     def name(self) -> _str: ...
-    @property
-    def na_value(self): ...
```
Dr-Irv, line 97: should this be in a different PR?

tests/test_scalars.py
```diff
@@ -0,0 +1,234 @@
...
+def test_periof_add_subtract() -> None:
+    p = pd.Period("2012-1-1", freq="D")
+
+    as0 = pd.Timedelta(1, "D")
+    as1 = dt.timedelta(days=1)
+    as2 = np.timedelta64(1, "D")
+    as3 = np.int64(1)
+    as4 = int(1)
+    as5 = pd.period_range("2012-1-1", periods=10, freq="D")
+    as6 = pd.Period("2012-1-1", freq="D")
```
Dr-Irv, line 103: can you use more descriptive names here, e.g., instead of `as1`, call it `as_dt_td`, instead of `as4`, use `as_int`, etc.

tests/test_scalars.py
```diff
@@ -0,0 +1,234 @@
...
+    scale = 24 * 60 * 60 * 10**9
+    as7 = cast(TimedeltaSeries, pd.Series(pd.timedelta_range(scale, scale, freq="D")))
```
Dr-Irv, line 105: This cast shouldn't be needed, because I thought if you pass a `TimeDeltaIndex` to `pd.Series`, you get a `TimdeltaSeries`, and if not, we should fix that.

tests/test_scalars.py
```diff
@@ -0,0 +1,234 @@
...
+
+    check(assert_type(as5 - p, pd.Index), pd.Index)
+
+    check(assert_type(p.__radd__(as0), pd.Period), pd.Period)
+    check(assert_type(p.__radd__(as1), pd.Period), pd.Period)
+    check(assert_type(p.__radd__(as2), pd.Period), pd.Period)
+    check(assert_type(p.__radd__(as3), pd.Period), pd.Period)
+    check(assert_type(p.__radd__(as4), pd.Period), pd.Period)
+    check(assert_type(p.__radd__(as10), NaTType), NaTType)
+    check(assert_type(p.__radd__(p.freq), pd.Period), pd.Period)
```
Dr-Irv, line 153: let's not call `__radd__()` directly and just do things like `p.freq + p`, which should cause the type checker to look at `__radd__()`

Dr-Irv (changes requested):
I know you may have responded to some of these comments in the original PR, but better to restart the conversation here.

pandas-stubs/_libs/tslibs/period.pyi
```diff
@@ -1,34 +1,137 @@
...
+    @overload
+    def __add__(self, other: _PeriodAddSub) -> Period: ...
+    @overload
+    def __add__(self, other: NaTType) -> NaTType: ...
+    @overload
+    def __add__(self, other: Index) -> PeriodIndex: ...
+    @overload
+    def __add__(self, other: TimedeltaSeries) -> PeriodSeries: ...
+    @overload  # type: ignore[override]
+    def __eq__(self, other: Period) -> bool: ...
```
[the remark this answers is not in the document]
bashtage: Doesn't it override the default so `Period()== object()` isn't allowed except for listed types?  I thought it made sense to do this since the return type of generic objects() isn't always a `bool` - it can be a Series, Index, ndarray.

pandas-stubs/_libs/tslibs/period.pyi
```diff
@@ -1,34 +1,137 @@
...
+    def __eq__(self, other: PeriodIndex) -> npt.NDArray[np.bool_]: ...
+    @overload
+    def __ge__(self, other: Period) -> bool: ...
+    @overload
+    def __ge__(self, other: PeriodIndex) -> npt.NDArray[np.bool_]: ...
+    @overload
+    def __gt__(self, other: Period) -> bool: ...
+    @overload
+    def __gt__(self, other: PeriodIndex) -> npt.NDArray[np.bool_]: ...
     def __hash__(self) -> int: ...
```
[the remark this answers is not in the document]
bashtage: done.

pandas-stubs/_libs/tslibs/period.pyi
```diff
@@ -1,34 +1,137 @@
...
+    def __sub__(self, other: Period) -> BaseOffset: ...
+    @overload
+    def __sub__(self, other: NaTType) -> NaTType: ...
+    @overload
+    def __sub__(self, other: PeriodIndex) -> Index: ...
+    @overload
+    def __sub__(self, other: TimedeltaSeries) -> PeriodSeries: ...
+    @overload
+    def __sub__(self, other: TimedeltaIndex) -> PeriodIndex: ...
+    def __rsub__(self, other: PeriodIndex) -> Index: ...
```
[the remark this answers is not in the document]
bashtage: Agree.  Actual implementation is  `PeriodIndex.__sub__` . The Period.__rsub__ is NotImplemented.

pandas-stubs/core/arrays/boolean.pyi
bashtage: Removed these

pandas-stubs/core/dtypes/dtypes.pyi
bashtage: Removed these

tests/test_scalars.py
bashtage: I've fixed this.

tests/test_scalars.py
bashtage: changed these.

pandas-stubs/_libs/tslibs/period.pyi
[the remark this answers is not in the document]
Dr-Irv: There are two schools of thought here, and, tbh, I'm not sure which is best.  

1. Looking at typeshed, the huge majority of the libraries define `__eq__` using an `object` type for the `other` argument (if they define it at all).  That's because `__eq__()` should accept any object.  In the few cases where they specify another type, it is similar to how we need a special case for `def __eq__(self, other: PeriodIndex) -> npt.NDArray[np.bool_]: ...` but don't define it both ways.  So my suggestion here is that we just include `__eq__()` and `__ne__()` for when `other` is of type `PeriodIndex`.  For example, the following code is valid:  `pd.Period("4Q2005") == 10` so that would mean the type checker would accept it, even though we know the answer is `False`.
2. If we go your direction, we are saying that users should not do equality or inequality comparisons to an object of the wrong type.  But it is valid python.  So maybe we should consider something like this (although I'm not sure it would work without `# type: ignore` annotations:
```python
    def __eq__(self, other: Period) -> bool: ...
    @overload
    def __eq__(self, other: PeriodIndex) -> npt.NDArray[np.bool_]: ...
    @overload
    def __eq__(self, other: object) -> Literal[False]: ...
```

Using something like the second idea is a compromise (if you can get it to work), but I'm leaning towards the first way because it seems to be what others are doing.

Interestingly, the following happens with numpy:
```python
import numpy as np
from numpy import typing as npt
a: npt.NDArray[np.int64] = np.array([1,2,3])
reveal_type(a)
reveal_type(a == 4)
reveal_type(a >= 4)
```

With mypy, you get:
```text
goo.py:4: note: Revealed type is "numpy.ndarray[Any, numpy.dtype[numpy.signedinteger[numpy._typing._64Bit]]]"
goo.py:5: note: Revealed type is "Any"
goo.py:6: note: Revealed type is "numpy.ndarray[Any, numpy.dtype[numpy.bool_]]"
```

So for the `>=` operator, it gives the right type, but for the equality operator, it doesn't.

tests/test_scalars.py
Dr-Irv: Yes, I agree that's true for debugging purposes.  If you want, you could do it as `p.freq + p` and add a comment that you are testing `p.__radd__(p.freq)` .  The reason I want to use the `p.freq +  p` expression is that it directly tests how the rules for `__add__()` and `__radd__()` work.

pandas-stubs/core/frame.pyi
```diff
@@ -1776,7 +1776,7 @@ class DataFrame(NDFrame, OpsMixin):
     @overload
     def rolling(
         self,
-        window: int | BaseOffset | BaseIndexer,
+        window: int | str | BaseOffset | BaseIndexer,
```
Dr-Irv, line 1779: changes to `core/frame.pyi` should be in a different PR

pandas-stubs/core/series.pyi
```diff
@@ -1574,28 +1585,28 @@ class Series(IndexOpsMixin, NDFrame, Generic[S1]):
     @overload
     def rolling(
         self,
-        window: int | BaseOffset | BaseIndexer,
+        window: int | _str | BaseOffset | BaseIndexer,
```
Dr-Irv, line 1588: changes to `rolling` should be in a different PR

tests/test_scalars.py
```diff
@@ -0,0 +1,235 @@
...
+
+if TYPE_CHECKING:
+    from pandas.core.series import (
+        PeriodSeries,
+        TimedeltaSeries,
+    )
+
+    from pandas._typing import np_ndarray_bool
+else:
+    PeriodSeries = TimedeltaSeries = np_ndarray_bool = Any
```
Dr-Irv, line 27: delete the `else` part, and then just refer to `PeriodSeries` and `TimedeltaSeries` in quotes so it works at runtime.

See `test_timefuncs.py` for an example.

tests/test_scalars.py
```diff
@@ -0,0 +1,235 @@
...
+    as_timedelta_idx = pd.timedelta_range(scale, scale, freq="D")
+    as_nat = pd.NaT
+
+    check(assert_type(p + as_pd_td, pd.Period), pd.Period)
+    check(assert_type(p + as_dt_td, pd.Period), pd.Period)
+    check(assert_type(p + as_np_td, pd.Period), pd.Period)
+    check(assert_type(p + as_np_i64, pd.Period), pd.Period)
+    check(assert_type(p + as_int, pd.Period), pd.Period)
+    check(assert_type(p + p.freq, pd.Period), pd.Period)
+    check(assert_type(p + (p - as_period_index), pd.PeriodIndex), pd.PeriodIndex)
```
Dr-Irv, line 117: This seems like an odd test.  Maybe compute `p - as_period_index`, do a `check` and `assert_type` on that, and then do the test on the addition.

tests/test_scalars.py
```diff
@@ -0,0 +1,235 @@
...
+    check(assert_type(p + as_td_series, PeriodSeries), pd.Series, pd.Period)
+    check(assert_type(p + as_timedelta_idx, pd.PeriodIndex), pd.PeriodIndex)
+    check(assert_type(p + as_nat, NaTType), NaTType)
+    das8 = cast(TimedeltaSeries, (as_period_series - as_period_series))
```
Dr-Irv, line 121: get rid of the `cast`

tests/test_windowing.py
```diff
@@ -13,9 +13,13 @@
 
 from tests import check
 
+from pandas.tseries.frequencies import to_offset
```
Dr-Irv, line 16: changes to `test_windowing.py` should be in another PR

Dr-Irv (changes requested):

tests/test_windowing.py
bashtage: Something got messed up in a merge of main.  I had to rebase to get these to disappear.

tests/test_scalars.py
bashtage: There is a test of this 11 lines below in the subtraction section.

tests/test_scalars.py
bashtage: It is a lot nicer to get intellisense on these things and to no have to noqa F401 them.

pandas-stubs/core/series.pyi
bashtage: Fixed by rebase.

pandas-stubs/core/frame.pyi
bashtage: Fixed by rebase.

tests/test_scalars.py
Dr-Irv: Yes, the result is a `Series` of offsets.  But anything we write in a test function should be code that a user could write, and they wouldn't have access to `TimedeltaSeries`.  And it's not the right cast anyway.

My suggestion is as follows:
1. Create a new type in `core/series.pyi` that is `OffsetSeries`
2. For any of the subtraction operators in here that return a series of offsets, have them return `OffsetSeries` instead.
3. Then  you can modify `__add__()` in here so that if you are adding a `Period` and an `OffsetSeries`, you get a `PeriodSeries`
4. You could also modify `OffsetSeries` to handle addition/subtraction of `Period` to `OffsetSeries`

tests/test_scalars.py
Dr-Irv: So you can do:
```python
np_ndarray_bool = npt.NDArray[np.bool_]
PeriodSeries: TypeAlias = pd.Series
TimedeltaSeries: TypeAlias = pd.Series
```
instead and get those same benefits.

bashtage: I put these in the `else` bit but I can't remove the `if TYPE_CHECKING` since the true `TimedeltaSeries` is needed, not `Series[Any]`.  

The simplest solution would be to just make these object so they are defined without any imports

tests/test_scalars.py
bashtage: Changed.

tests/test_scalars.py
bashtage: I tried to do this but it broke things elsewhere in `Series`.  I think this is too complex a route, at least for this PR.  I've left the changes in but will likely just disable the test.

bashtage: Found this issue and now working.

tests/test_scalars.py
bashtage: I'm going to resolve but I think just `object` would be cleaner.

tests/test_scalars.py
bashtage: Added some notes and reorganized these.  Reog make is clear that operator comes with each call to `__radd/sub__`

Dr-Irv (commented):
I'm fine with everything here, but would like to change `__eq__()` and `__ne__()` as described in my comments.

pandas-stubs/_libs/tslibs/period.pyi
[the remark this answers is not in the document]
bashtage: I've gone down the router for `Literal[False]` for `__eq__(object)`

bashtage:
I think it is ready.  I also added `Series[Period`] and `PeriodSeries` for comparisons.

Dr-Irv (approved):
thanks @bashtage .  I think the `Literal[False]` idea for `__eq__()` looks good

PR #391, opened 2022-10-16
ENH: Improve typing for Interval

bashtage:

xref #383


pandas-stubs/core/indexes/interval.pyi
```diff
@@ -122,6 +217,30 @@ class IntervalIndex(IntervalMixin, ExtensionIndex):
     def get_value(self, series: ABCSeries, key): ...
     @property
     def is_all_dates(self) -> bool: ...
+    @overload  # type: ignore[override]
+    def __gt__(self, other: Interval | IntervalIndex) -> np_ndarray_bool: ...  # type: ignore[misc]
+    @overload
+    def __gt__(self, other: object) -> bool: ...
```
Dr-Irv, line 223: Don't think we need to allow `IntervalIndex` to be compared to any other object for `__gt__`, __ge__`, etc

pandas-stubs/core/indexes/interval.pyi
```diff
@@ -122,6 +217,30 @@ class IntervalIndex(IntervalMixin, ExtensionIndex):
...
+    @overload
+    def __le__(self, other: object) -> bool: ...
+    @overload  # type: ignore[override]
+    def __lt__(self, other: Interval | IntervalIndex) -> np_ndarray_bool: ...  # type: ignore[misc]
+    @overload
+    def __lt__(self, other: object) -> bool: ...
+    @overload  # type: ignore[override]
+    def __eq__(self, other: Interval | IntervalIndex) -> np_ndarray_bool: ...  # type: ignore[misc]
+    @overload
+    def __eq__(self, other: object) -> bool: ...
```
Dr-Irv, line 239: Do you want to use the pattern  you used in `Period`, where a comparison to any other object returns `Literal[False]` ?

tests/test_scalars.py
```diff
@@ -0,0 +1,287 @@
...
+    check(assert_type(interval_i + 1.5, "pd.Interval[float]"), pd.Interval)
+    check(assert_type(interval_f + 1.5, "pd.Interval[float]"), pd.Interval)
+    check(
+        assert_type(interval_ts + pd.Timedelta(days=1), "pd.Interval[pd.Timestamp]"),
+        pd.Interval,
+    )
+    check(
+        assert_type(interval_td + pd.Timedelta(days=1), "pd.Interval[pd.Timedelta]"),
+        pd.Interval,
+    )
```
Dr-Irv, line 221: For all of the arithmetic operations, should test the reverse operator, e.g., `pd.Timedelta(days=1) + interval_td`

tests/test_scalars.py
```diff
@@ -0,0 +1,287 @@
...
+    check(assert_type(0.5 in interval_i, bool), bool)
+    check(assert_type(1 in interval_i, bool), bool)
+    check(assert_type(1 in interval_f, bool), bool)
+    check(assert_type(pd.Timestamp("2000-1-1") in interval_ts, bool), bool)
+    check(assert_type(pd.Timedelta(days=1) in interval_td, bool), bool)
+
+    check(assert_type(hash(interval_i), int), int)
+    check(assert_type(hash(interval_f), int), int)
+    check(assert_type(hash(interval_ts), int), int)
+    check(assert_type(hash(interval_td), int), int)
```
Dr-Irv, line 232: No need to test `hash`

tests/test_scalars.py
```diff
@@ -0,0 +1,287 @@
...
+
+    interval_index_int = pd.IntervalIndex([interval_i])
+    interval_series_int = pd.Series(interval_index_int)
+
+    check(interval_series_int >= interval_i, pd.Series, bool)
```
Dr-Irv, line 237: All of these tests from here to the bottom need `assert_type`

Dr-Irv (changes requested):
Looks mostly good.  Ping when green and those issues have been addressed.

pandas-stubs/core/algorithms.pyi
```diff
@@ -19,12 +21,21 @@ from pandas._typing import AnyArrayLike
...
+    Interval[pd.Timedelta],
+)
+
 @overload
 def unique(values: PeriodIndex) -> PeriodIndex: ...  # type: ignore[misc]
 @overload
 def unique(values: CategoricalIndex) -> CategoricalIndex: ...  # type: ignore[misc]
 @overload
-def unique(values: IntervalIndex) -> IntervalIndex: ...  # type: ignore[misc]
+def unique(values: IntervalIndex[_IntervalT]) -> IntervalIndex[_IntervalT]: ...  # type: ignore[misc]
```
Dr-Irv, line 38: error you're getting is because `pd.interval_range()` is returning `IntervalIndex`, so you could change the return type there, or just remove the generic declaration here (i.e., leave it as it was).

Dr-Irv:
ping when green and when issues have been addressed.

Dr-Irv:
@bashtage checking on status of this.  It was in pretty good shape, just needs to go green and address a few issues.

tests/test_scalars.py
bashtage: Removed.

tests/test_scalars.py
bashtage: This is fixed.

pandas-stubs/core/indexes/interval.pyi
bashtage: Fixed.

pandas-stubs/core/indexes/interval.pyi
bashtage: Fixed.

pandas-stubs/core/algorithms.pyi
bashtage: I think this is fixed now.

tests/test_scalars.py
bashtage: Done, at least where allowed.

bashtage:
@Dr-Irv green ex nightly.

Dr-Irv:
> @Dr-Irv green ex nightly.

will review soon, but I fixed nightly and need you to review in #436

pandas-stubs/core/indexes/interval.pyi
```diff
@@ -216,46 +250,155 @@ class IntervalIndex(IntervalMixin, ExtensionIndex, Generic[_IntervalT]):
...
     @overload  # type: ignore[override]
-    def __gt__(self, other: Interval | IntervalIndex) -> np_ndarray_bool: ...  # type: ignore[misc]
+    def __gt__(  # type: ignore[misc]
+        self, other: IntervalT | IntervalIndex[IntervalT]
+    ) -> np_ndarray_bool: ...
+    @overload
+    def __gt__(self, other: pd.Series[IntervalT]) -> pd.Series[bool]: ...  # type: ignore[misc]
     @overload
-    def __gt__(self, other: object) -> bool: ...
+    def __gt__(self, other: object) -> Literal[False]: ...
```
Dr-Irv, line 262: I think this (and other comparisons to `object` for `gt`, `ge`, `lt`, `le` should be `Never`.  You can't compare an `Interval` to an arbitrary object.

pandas-stubs/core/indexes/interval.pyi
```diff
@@ -216,46 +250,155 @@ class IntervalIndex(IntervalMixin, ExtensionIndex, Generic[_IntervalT]):
...
+    periods: int | None = ...,
+    freq: str | BaseOffset | None = ...,
+    name: Hashable = ...,
+    closed: IntervalClosedType = ...,
+) -> IntervalIndex[Interval[pd.Timestamp]]: ...
 @overload
 def interval_range(
-    start: DatetimeLike | None = ...,
-    end: DatetimeLike | None = ...,
+    start: _TimedeltaLike,
```
Dr-Irv, line 397: shouldn't there be a more complete set of overloads for `_TimedeltaLIke` just like there are for the others (`int`, `float`, `_TimestampLike`)

tests/test_indexes.py
```diff
@@ -191,18 +191,25 @@ def test_range_index_union():
 
 
 def test_interval_range():
-    check(assert_type(pd.interval_range(0, 10), pd.IntervalIndex), pd.IntervalIndex)
+    check(
+        assert_type(pd.interval_range(0, 10), "pd.IntervalIndex[pd.Interval[int]]"),
+        pd.IntervalIndex,
```
Dr-Irv, line 196: could add `int` as last argument to `check` (similar for other tests that return `IntervalIndex`), with appropriate types

pandas-stubs/core/indexes/interval.pyi
Dr-Irv: Didn't see this as fixed?

Dr-Irv (changes requested):

tests/test_indexes.py
bashtage: They are actually `pd.Interval` not `int`, so it is a bit pointless since this is the only type IntervalIndex can have.

pandas-stubs/core/indexes/interval.pyi
bashtage: Added the one missing, and some tests.

pandas-stubs/core/indexes/interval.pyi
bashtage: Changed these.

bashtage:
> Didn't see this as fixed?

Pretty sure these are gone.  When I search for ` def __gt__(self, other: object) -> bool` there are no matches.  GH also says outdated.

pandas-stubs/core/indexes/interval.pyi
bashtage: Pretty sure this is gone from the latest.

bashtage:
@Dr-Irv Should be all fixed now.

pandas-stubs/core/indexes/interval.pyi
```diff
@@ -259,31 +262,31 @@ class IntervalIndex(IntervalMixin, ExtensionIndex, Generic[IntervalT]):
...
-    def __lt__(self, other: object) -> Literal[False]: ...
-    @overload  # type: ignore[override]
+    def __lt__(self, other: object) -> Never: ...
+    @overload
     def __eq__(self, other: IntervalT | IntervalIndex[IntervalT]) -> np_ndarray_bool: ...  # type: ignore[misc]
     @overload
     def __eq__(self, other: pd.Series[IntervalT]) -> pd.Series[bool]: ...  # type: ignore[misc]
     @overload
-    def __eq__(self, other: object) -> Literal[False]: ...
+    def __eq__(self, other: object) -> Never: ...
```
Dr-Irv, line 289: This should remain `Literal[False]`, because it is always valid code to do equality against another object.

.flake8
```diff
@@ -0,0 +1,3 @@
+[flake8]
+per-file-ignores =
+    tests/test_scalars.py: W503
```
Dr-Irv, line 3: I tested your PR with this removed, and it passed, so can you remove this file?

pandas-stubs/core/indexes/interval.pyi
```diff
@@ -11,80 +13,210 @@ import numpy as np
...
+    npt.NDArray[np.intp],
+    pd.Series[int],
+    pd.Int64Index,
+]
+_EdgesFloat: TypeAlias = Union[
     Sequence[float],
+    npt.NDArray[np.float64],
+    pd.Series[float],
+    pd.Float64Index,
+]
```
Dr-Irv, line 57: I think the use of `pd.Int64Index` and `pd.Float64Index` here will cause issues with the nightly builds once #436 is merged in, so we'll have to figure out how to deal with that when in a `Union` like this.

pandas-stubs/core/indexes/interval.pyi
```diff
@@ -11,80 +13,210 @@ import numpy as np
...
+            tuple[pd.Timedelta, pd.Timedelta]
+            | tuple[dt.timedelta, dt.timedelta]
+            | tuple[np.timedelta64, np.timedelta64]
+        ],
+        closed: IntervalClosedType = ...,
+        name: Hashable = ...,
+        copy: bool = ...,
+        dtype: IntervalDtype | None = ...,
+    ) -> IntervalIndex[pd.Interval[pd.Timedelta]]: ...
     def __contains__(self, key: Any) -> bool: ...
```
Dr-Irv, line 220: I think we could probably do `def __contains__(self: key: IntervalT) -> bool: ...` here or something like that.  Might have to deal with the `int/float` dichotomy with some overloads.

tests/test_scalars.py
```diff
@@ -51,6 +51,339 @@
...
+    )
+    check(
+        assert_type(
+            interval_i != interval_index_int,
+            np_ndarray_bool,
+        ),
+        np.ndarray,
+        np.bool_,
+    )
+
```
Dr-Irv, line 385: It would be nice to add tests that check that the typing is not too wide.  Don't have to do it for this PR.  E.g., `interval_i >= pd.Timestamp("2017-01-01")` should be disallowed.

Dr-Irv (changes requested):
A few more things

.flake8
bashtage: Removed

pandas-stubs/core/indexes/interval.pyi
bashtage: Updated this

Dr-Irv (approved):
thanks @bashtage

PR #398, opened 2022-10-25
(🎁) Support Python 3.11

KotlinIsland:

### DIFF
--- a/tests/test_io.py
+++ b/tests/test_io.py
@@ -70,10 +70,13 @@ DF = DataFrame({"a": [1, 2, 3], "b": [0.0, 0.0, 0.0]})
 CWD = os.path.split(os.path.abspath(__file__))[0]
 
 
-if sys.version_info >= (3, 11):
-    # This is only needed temporarily due to no wheels being available for arrow on 3.11
-    _arrow = pytest.importorskip("arrow")
+arrow_skip = pytest.mark.skipif(sys.version_info >= (3, 11), "pyarrow is not available for 3.11 yet")
+"""This is only needed temporarily due to no wheels being available for arrow on 3.11"""
 
+lxml_skip = pytest.mark.skipif(sys.version_info >= (3, 11), "pyarrow is not available for 3.11 yet")
+"""This is only needed temporarily due to no wheels being available for arrow on 3.11"""
+
+@arrow_skip
 @pytest.mark.skipif(WINDOWS, reason="ORC not available on windows")
 def test_orc():
     with ensure_clean() as path:
@@ -81,6 +84,7 @@ def test_orc():
         check(assert_type(read_orc(path), DataFrame), DataFrame)
 
 
+@arrow_skip
 @pytest.mark.skipif(WINDOWS, reason="ORC not available on windows")
 def test_orc_path():
     with ensure_clean() as path:
@@ -89,6 +93,7 @@ def test_orc_path():
         check(assert_type(read_orc(pathlib_path), DataFrame), DataFrame)
 
 
+@arrow_skip
 @pytest.mark.skipif(WINDOWS, reason="ORC not available on windows")
 def test_orc_buffer():
     with ensure_clean() as path:
@@ -101,6 +106,7 @@ def test_orc_buffer():
         file_r.close()
 
 
+@arrow_skip
 @pytest.mark.skipif(WINDOWS, reason="ORC not available on windows")
 def test_orc_columns():
     with ensure_clean() as path:
@@ -108,11 +114,12 @@ def test_orc_columns():
         check(assert_type(read_orc(path, columns=["a"]), DataFrame), DataFrame)
 
 
+@arrow_skip
 @pytest.mark.skipif(WINDOWS, reason="ORC not available on windows")
 def test_orc_bytes():
     check(assert_type(DF.to_orc(index=False), bytes), bytes)
 
-
+@lxml_skip
 def test_xml():
     with ensure_clean() as path:
         check(assert_type(DF.to_xml(path), None), type(None))
@@ -121,6 +128,7 @@ def test_xml():
             check(assert_type(read_xml(f), DataFrame), DataFrame)
 
 
+@lxml_skip
 def test_xml_str():
     with ensure_clean() as path:
         check(assert_type(DF.to_xml(), str), str)
@@ -287,6 +295,7 @@ def test_sas_xport() -> None:
         pass
 
 
+@lxml_skip
 def test_hdf():
     with ensure_clean() as path:
         check(assert_type(DF.to_hdf(path, "df"), None), type(None))
--- a/tests/test_io.py
+++ b/tests/test_io.py
@@ -5,6 +5,7 @@ import os.path
 import pathlib
 from pathlib import Path
 import sqlite3
+import sys
 from typing import (
     TYPE_CHECKING,
     Any,
@@ -68,6 +69,10 @@ from pandas.io.stata import StataReader
 DF = DataFrame({"a": [1, 2, 3], "b": [0.0, 0.0, 0.0]})
 CWD = os.path.split(os.path.abspath(__file__))[0]
 
+
+if sys.version_info >= (3, 11):
+    # This is only needed temporarily due to no wheels being available for arrow on 3.11
+    _arrow = pytest.importorskip("arrow")
 
 @pytest.mark.skipif(WINDOWS, reason="ORC not available on windows")
 def test_orc():
--- a/tests/test_pandas.py
+++ b/tests/test_pandas.py
@@ -20,10 +20,14 @@ from pandas._libs.missing import NAType
 from pandas._libs.tslibs import NaTType
 from pandas._typing import Scalar
 
+
 from tests import (
     TYPE_CHECKING_INVALID_USAGE,
     check,
+    importskip
 )
+
+pyarrow_skip = pytest.mark.skip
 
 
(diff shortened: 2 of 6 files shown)


bashtage:
Need to wait for lxml to have 3.11 wheels.

KotlinIsland:
https://github.com/lxml/lxml/pull/355
https://github.com/lxml/lxml/pull/356

bashtage:
There are 3.11 wheels on PyPu if 1.5.1, so it looks like 3.11 is supported by pandas.

BrenoJesusFernandes:
Be careful, looks like we still don't have  a stable version for python 3.11 for macos in github actions: https://github.com/actions/python-versions/releases/tag/3.11.0-3320066239

Dr-Irv:
CI is failing because no stable version of python 3.11 for MacOS yet and no `pyarrow` for python 3.11

Dr-Irv:
Issue is still that `pyarrow` doesn't work with python 3.11.  The CI sees no binary, then tries to build the wheel, and it fails.

There might be a workaround here to skip the requirement of pyarrow when testing on 3.11 (and then to adjust the tests accordingly to do a `pytest.skip`) but I don't know how to modify `pyproject.toml` to do that.

@BrenoJesusFernandes Any ideas on how to make the poetry requirement for `pyarrow` conditional on python <= 3.10 in `pyproject.toml` ?

Dr-Irv:
spoke with someone who is familiar with `pyarrow`, and they think 3.11 wheels will be available within a week, so maybe easiest to wait for that

BrenoJesusFernandes:
I don't get why pandas is available in python 3.11, when one requirement is not available. It's seems wrong for me. In poetry last versio, you can define multiples dependencies besides the dev-dependecies, but I think we should wait the dependecies to be available in 3.11.

Dr-Irv:
> I don't get why pandas is available in python 3.11, when one requirement is not available. It's seems wrong for me. In poetry last versio, you can define multiples dependencies besides the dev-dependecies, but I think we should wait the dependecies to be available in 3.11.

`pyarrow` is an _optional_ dependency.  So that's why `pandas` is available for python 3.11, because you can use `pandas` with python 3.11, you just can't use `pyarrow` with it.

So if there is a way to define the dependencies as one set for python <= 3.10 that we have now, and another set that is the same, except removes `pyarrow` for python 3.11, we can move forward.  Then we modify the tests that use `pyarrow` to test if `pyarrow` is available.  That's how the tests for pandas with python 3.11 are set up.

Or we wait a week (hopefully) until  `pyarrow` builds wheels for python 3.11

Dr-Irv:
> is there some test using pyarrow? if exits tests like that it is gonna fail in CI.

There are a couple of tests in `test_pandas.py` and `test_frame.py`.  They are protected with a `pytest.importorskip` .  

So we just need to modify the `pyproject.toml` file to indicate that `pyarrow` should not be installed when testing python 3.11.  Then the tests should work.

BrenoJesusFernandes:
hmm I see, so it's simple than I thought, just add something like that:

pyarrow = { version = "currently-version", python = "<3.11" }
]

Dr-Irv:
@KotlinIsland if you want to try what @BrenoJesusFernandes suggests above, that might make the CI work.

bellini666:
Since this is a stubs package, does it make sense to limit the upper bound of the python version? I would say that even libs shouldn't do that themselves, unless there's a very good reason.

This is very limiting for people using poetry for example due to [this issue](https://github.com/python-poetry/poetry/issues/697)

Dr-Irv:
> Since this is a stubs package, does it make sense to limit the upper bound of the python version? I would say that even libs shouldn't do that themselves, unless there's a very good reason.

The limit is there for now because we can't run a successful test with python 3.11.  We're not willing to release the stubs for a python version unless we can successfully test the stubs on that version.

In the past, we have had issues with the stubs being compatible with one version of python, but not another.  So our tests help make sure we are not making that mistake.

KotlinIsland:
Looks like `tables` and `lxml` don't have 3.11 wheels either.

KotlinIsland:
@Dr-Irv Okay, I have made `arrow`, `tables` and `lxml` conditional dependencies, and I have added a skip check for 3.11

tests/test_io.py
```diff
@@ -69,6 +70,10 @@
 CWD = os.path.split(os.path.abspath(__file__))[0]
 
 
+if sys.version_info >= (3, 11):
+    # This is only needed temporarily due to no wheels being available for arrow on 3.11
+    _arrow = pytest.importorskip("arrow")
```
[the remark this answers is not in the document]
KotlinIsland: Strange, it made the tests pass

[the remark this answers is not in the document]
Dr-Irv: > This is the output of pytest, it seems to me as if `test_io` is in some way using `pyarrow`?

I'm guessing there are dependencies.  So `test_orc` pattern needs `pyarrow`, `test_xml` pattern needs `lxml`, `test_hdf` pattern needs `pytables`, etc.  So we should do the skips based on the particular test (and underlying library) being done.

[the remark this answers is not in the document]
KotlinIsland: Hows that look?

KotlinIsland:
Looks like there is some issues with mypy: https://github.com/python/mypy/issues/13627

This is causing the ci to fail for 3.11 and will require updating mypy to 0.982

Dr-Irv:
@KotlinIsland Please see https://github.com/pandas-dev/pandas-stubs/pull/398#pullrequestreview-1161937232

We have to wait for mypy 0.990

Dr-Irv:
> @KotlinIsland Please see [#398 (review)](https://github.com/pandas-dev/pandas-stubs/pull/398#pullrequestreview-1161937232)
> 
> We have to wait for mypy 0.990

@KotlinIsland We are now testing with `mypy` 0.990 via #419 so can you merge with that, resolve conflicts, and see if the python 3.11 support will work?

KotlinIsland:
@Dr-Irv Can you trigger the CI?

Dr-Irv:
Currently two issues (but there might be more):
In `test_pandas.py`, have to use a pattern like this:
```python
import pandas.util._test_decorators as td  # at top of file somewhere

pyarrow_skip = td.skip_if_no("pyarrow")


@pyarrow_skip
def test_arrow_dtype() -> None:
    import pyarrow as pa

    check( ...
```
and in `pandas-stubs/core/arrays/arrow/dtype.pyi` (somewhere near the top):
```python
if sys.version_info < (3, 11):
    import pyarrow as pa
else:
    pa = Any
```

I'd also suggest setting up a local environment with python 3.11 and running the tests locally instead of waiting for me or someone else to do a CI approval.

Dr-Irv:
I played around a bit with this, and uncovered a pyright bug that doesn't ignore imports based on the python version test.  https://github.com/microsoft/pyright/issues/4182

So we have to wait for that to get fixed to be able to do conditional imports.

tests/test_pandas.py
```diff
@@ -20,11 +20,15 @@
...
 from pandas._typing import Scalar
 
+
 from tests import (
     TYPE_CHECKING_INVALID_USAGE,
     check,
+    importskip
 )
 
+pyarrow_skip = pytest.mark.skip
```
Dr-Irv, line 30: Shouldn't this be 
```python
pyarrow_skip = pytest.mark.skipif(sys.version_info >= (3, 11), "pyarrow is not available for 3.11 yet")
"""This is only needed temporarily due to no wheels being available for arrow on 3.11"""
```

tests/test_pandas.py
```diff
@@ -20,11 +20,15 @@
 from pandas._libs.tslibs import NaTType
 from pandas._typing import Scalar
 
+
 from tests import (
     TYPE_CHECKING_INVALID_USAGE,
     check,
+    importskip
```
Dr-Irv, line 27: this import is failing.

Dr-Irv (changes requested):

tests/test_io.py
```diff
@@ -69,13 +70,21 @@
 CWD = os.path.split(os.path.abspath(__file__))[0]
 
 
+arrow_skip = pytest.mark.skipif(sys.version_info >= (3, 11), "pyarrow is not available for 3.11 yet")
+"""This is only needed temporarily due to no wheels being available for arrow on 3.11"""
+
+lxml_skip = pytest.mark.skipif(sys.version_info >= (3, 11), "pyarrow is not available for 3.11 yet")
```
bashtage, line 76: Wrong reason in the explanation here. Should be lxml not arrow.

tests/test_pandas.py
```diff
@@ -461,9 +465,8 @@ def test_crosstab() -> None:
     )
 
 
+@pyarrow_skip
 def test_arrow_dtype() -> None:
-    pytest.importorskip("pyarrow")
```
bashtage, line 465: These is there for a reason to protect Windows.  I think it should stay.

tests/test_pandas.py
bashtage, line 30: Appears to be not needed.

bashtage (changes requested):
Some fixes needed.

Dr-Irv:
I created #435 to move this along.  Took some of your ideas from here.  Had to handle an issue with pandas that's a bit thorny and want to see if I can coordinate release of this with release of pandas 1.5.2 if a PR there is included.

Thanks for working on this.  It was a great head start.

PR #399, opened 2022-10-25
MAINT: Add coverage for mypy

bashtage:
Add HTML coverage reporting.


Dr-Irv:
Might want to add some documentation here:  https://github.com/pandas-dev/pandas-stubs/blob/main/docs/tests.md

pyproject.toml
```diff
@@ -93,6 +93,10 @@ script = "scripts.test.run:mypy_src"
 help = "Run mypy on 'tests' using the installed stubs"
 script = "scripts.test:test(dist=True, type_checker='mypy')"
 
+[tool.poe.tasks.mypy_coverage]
+help = "Run mypy on 'tests' using the local stubs and report HTML coverage"
+script = "scripts.test:test(coverage=True, type_checker='mypy')"
```
twoertwein, line 98: Could also make it an optional argument to `poe mypy`

bashtage: Not sure how I could do this.

twoertwein: something like this might work (haven't tested it):


```
[tool.poe.tasks.mypy]
help = "Run mypy on 'tests' (using the local stubs) and on the local stubs"
script = "scripts.test:test(src=True, coverage=coverage, type_checker='mypy')"
args = [{name = "coverage", positional = false, default = false, type = "boolean", required = false, help= "Whether to report coverage as HTML"}]
```

bashtage: Thanks.  I've gone this way.  I also reformatted the args into a more friendly TOML format.  Let me know if you don't like this style.

twoertwein: nice! I didn't use this style because it created some errors that I couldn't figure out :)

twoertwein: I currently can't test it because my python 3.10.7 installation triggers the `error: Positional-only parameters are only supported in Python 3.8 and greater  [syntax]`-bug

Dr-Irv: > I currently can't test it because my python 3.10.7 installation triggers the `error: Positional-only parameters are only supported in Python 3.8 and greater [syntax]`-bug

I think that's an issue with python 3.10.7 and mypy 0.971.  If you use python 3.10.6 (which is what we pinned in the github actions), you should be OK.

Dr-Irv:
> Might want to add some documentation here: https://github.com/pandas-dev/pandas-stubs/blob/main/docs/tests.md

@bashtage Can you add more documentation about what this option is doing?  I ran it locally and looked at the report, and am having a hard time seeing the value, but maybe I'm missing something???

Dr-Irv (changes requested):
small typo in docs

docs/tests.md
```diff
@@ -13,5 +13,9 @@ Here are the most important options. Fore more details, please use `poe --help`.
 - Run tests against the installed stubs (this will install and uninstall the stubs): `poe test_dist`
 - Optional: run pytest against pandas nightly (this might fail): `poe pytest --nightly`
 - Optional: Run stubtest to compare the installed pandas-stubs against pandas (this will fail): `poe stubtest`. If you have created an allowlist to ignore certain errors: `poe stubtest path_to_the_allow_list`
+- Optional: Run mypy with coverage: `poe mypy --coverage`. The reported coverage documents functions thay appear to be fully,
```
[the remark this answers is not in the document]
bashtage: Fixed.

Dr-Irv (approved):
thanks @bashtage

docs/philosophy.md
```diff
@@ -79,6 +79,11 @@ which is supported by `typing_extensions` version 4.2 and beyond makes it easier
 to test to validate the return types of functions and methods. Future work
 is intended to expand the use of `assert_type()` in the test code.
 
+After making changes to the type stubs and adding tests to verify these, please generate
+a coverage report to ensure that all changes are covered by the tests. You can generate
+the HTML report by running ``poe mypy_coverage``. The report will be generated in the
+directory ``coverage``.
```
Dr-Irv, line 85: I'm a bit hesitant on including this paragraph.  TBH, I can't make heads or tails out of the coverage report.  It's not clear what the yellow highlighted or red highlighted code areas mean in the test files or `pyi` files.  If we're saying to people "please generate a coverage report", then we should document how to interpret that report, i.e., what do the yellow and red areas mean within each file, and if the interpretation on the tests vs. `pyi` files is different, and what are the remedies for fixing those issues (assuming that `mypy` is not buggy here!)

So you could create a new section about generating the coverage report and describe how to use it, or we could leave out the above paragraph.

You have some of that material in `docs/tests.md`, but I think you have some knowledge here that isn't obvious to others (like me!)

Dr-Irv (changes requested):

docs/philosophy.md
bashtage: Red, no types in def
Yellow, some types in def
Green, full types in def

Dr-Irv: > Red, no types in def Yellow, some types in def Green, full types in def

When I look at something like this, I don't see how your description applies:
image

Maybe you have to ignore the overloads?

Why is `__setitem__()` in yellow?

That's why I think extra docs on specific things to check for, and which things to ignore, would be helpful.

docs/tests.md
```diff
@@ -10,8 +10,15 @@ Here are the most important options. Fore more details, please use `poe --help`.
   - Run only pyright: `poe pyright`
   - Run only pytest: `poe pytest`
   - Run only pre-commit: `poe style`
+  - Run mypy and generate a report of the code coverage from the tests: `poe mypy_coverage`
```
twoertwein, line 13: `poe mypy --coverage`. Could also remove it as we point to `poe --help` for more options.

docs/philosophy.md
twoertwein: I think the issue is that `Series` is generic but its type is not provided. 

Here is an example from _typing:

image

`tuple` should have an argument but doesn't (not sure about some of the others - is `Index` generic?)


I would be fine mentioning it just in `tests.md`. If it should also be in this file, `poe mypy --coverage`.

bashtage: Yes. It would prefer `Series[Any]` over `Series`.

bashtage: I don't fully understand why `@overload` is red TBH.

docs/tests.md
twoertwein: Probably better to remove this line and just have the optional block later.

docs/philosophy.md
Dr-Irv: Given the discussion above,  where we don't know everything that is going on, I'd like to soften the language in the above paragraph.  I'd be more comfortable changing "please generate a coverage report" to "you may wish to generate a coverage report" and changing "to ensure that all changes" to "to understand if all changes"

twoertwein: If we want to prominently push this coverage report by having it in `philosophy.md`, I wouldn't mind simply always enabling the coverage report (and not even having the option to skip it, assuming it doesn't take more time).

I don't have a strong opinion about how much to push this coverage report. For me, it seems like a similarly optional step as stubtest - in both cases, they should pass/be green, but it will take a long time.

If people try to get this report to green, we can maybe enable pyright's UnknownArgument/Type/... in the future.

Dr-Irv:
@bashtage if we soften up the wording on `philosophy.md`, I'd be ready to approve this.

twoertwein:
Closing it so that it is a tiny bit easier to focus on current PRs. Feel free to re-open.

Issue #393, opened 2022-10-18
`DataFrame.apply` stub doesn't reflect default value of `axis` parameter


debonte:
**Describe the bug**
If you call `DataFrame.apply` and pass parameters beyond `f` but not including `axis`, you'll get the following error from Pyright:

```text
No overloads for "apply" match the provided arguments
```

**To Reproduce**
```python
import pandas as pd

def gethead(s, y):
    return s.head(y)

df = pd.DataFrame({"A": [1, 2, 3, 4, 5, 6], "B": [11, 12, 13, 14, 15, 16]})

df2 = df.apply(gethead, args=tuple([4]))  # error
```

Pylance gives the following error on the `apply` call:

```text
No overloads for "apply" match the provided arguments
  Argument types: ((s: Unknown, y: Unknown) -> Unknown, tuple[int, ...])Pylance[reportGeneralTypeIssues](https://github.com/microsoft/pylance-release/blob/main/DIAGNOSTIC_SEVERITY_RULES.md#diagnostic-severity-rules)
```

**Please complete the following information:**
 - OS: Windows
 - OS Version: 10.0.25197 Build 25197
 - python version: 3.10.0
 - version of type checker: Pylance 2022.10.21
 - version of installed `pandas-stubs`: 1.5.0.220926

**Additional context**
Originally reported by @jonmooser at https://github.com/microsoft/pylance-release/issues/3491

I was originally thinking that the fix was to indicate in the stubs that `axis` has a default value (`axis: AxisType = ...`), but then the two overloads of `apply` overlap with each other. It was looking like the correct fix would require more knowledge of the innards of `apply` and Pandas in general than I have.

Dr-Irv:
The trick here is that the function `gethead()`  has to be properly typed to return a `Series`, and if default arguments are used, then we should return a `DataFrame` when we know that the function returns a `Series`.   See the description of the `result_type` argument which describes the default behavior.

So for your example to work, it would have to change, and we have to add an overload for `apply()`

jonmooser:
>  See the description of the `result_type` argument which describes the default behavior.
> 

Dr-Irv,
Could you clarify the significance of the result_type argument? The doc says that only applies when axis=1. In my example axis=0 (implicitly)

Dr-Irv:
> Dr-Irv, Could you clarify the significance of the result_type argument? The doc says that only applies when axis=1. In my example axis=0 (implicitly)

[...]

PR #401, opened 2022-10-26
Fix value_counts and apply.

danielroseman:
Fix a few issues.

* `Series.value_counts` was wrongly typed as returning a series of the same type as the original (`Series[S1]`), when it actually returns `Series[int]`.
* The result of the callable given to `Series.apply` does not need to be hashable; it can return any value valid for a Series, which can include non-hashable items like lists or dicts.
* `DataFrame.apply`, like `DataFrameGroupBy.apply`, returns a DataFrame or Series depending on the return value of the callable, not on the set of parameters given. The exception is when `result_type` is "broadcast", in which case the result is always a DataFrame.


Dr-Irv (changes requested):
This all looks good, and the changes for `apply` should fix #393 .  Can you add a test from that issue to this PR and then we can close that issue?

pandas-stubs/core/series.pyi
```diff
@@ -685,7 +685,7 @@ class Series(IndexOpsMixin, NDFrame, Generic[S1]):
     @overload
     def apply(
         self,
-        func: Callable[..., Hashable],
+        func: Callable[..., Scalar | Sequence | Mapping],
```
gandhis1, line 688: Don't we basically assume everywhere in this code base that the expectation is that dataframes and series only store hashable/scalar types, and NOT things like `dicts`, or arbitrary objects for the matter?

Dr-Irv: Yes, but in this case, it doesn't matter, because he is defining a `Callable` that returns types, that `apply` will then use to return a `Series`.

danielroseman:
@Dr-Irv test added as requested.

Dr-Irv (approved):
thanks @danielroseman

PR #404, opened 2022-10-31
Remove redundant PyPI link “Source code”

adamchainz:
[PyPI](https://pypi.org/project/pandas-stubs/) shows both “Repository” and “Source Code”:

Xnapper-2022-10-31-18 22 21

I guess “Repository” comes from Poetry, so you don’t need a second declaration.


Dr-Irv (approved):
thanks @adamchainz for noticing this

PR #406, opened 2022-11-02
(🎁) update mypy to 0.982

KotlinIsland:
# Problems:
- `Scalar` seems to be getting screwed up somehow when `pd.to_numeric` is called
- `Series.count` has an invalid overload due to `None` being `Hashable`.
  - Perhaps we could just put a type ignore on this, the real type would be something along the lines of `Hashable & Not[None]`, which is currently not supported.


Dr-Irv:
@KotlinIsland please see #336 and https://github.com/python/mypy/pull/13781#issuecomment-1265908878

We have to wait for mypy 0.990

PR #408, opened 2022-11-03
BUG: Remove unused imports to fix recent pyright

bashtage:


Dr-Irv (approved):
thanks @bashtage .  Nightly run is failing, but we can let that go

Issue #400, opened 2022-10-25
Python 3.11 support


nth10sd:
**Describe the bug**
https://github.com/pandas-dev/pandas-stubs/blob/581c1eb1d98916734f9378baa61d1446e227abd5/pyproject.toml#L34

Python 3.11 is not supported.

**To Reproduce**
Try installing in a `venv` from Python 3.11

**Additional context**
It seems pandas already has Python 3.11 support.

Dr-Irv:
Our tests are failing because `pyarrow` is used in the testing, and `pyarrow` doesn't support 3.11 yet.

PR created:  #398

PR #419, opened 2022-11-08
CI: mypy 0.990

### LINKED ISSUE #336: mypy 0.981 compatibility

twoertwein:
closes #336

xref #400


pandas-stubs/core/series.pyi
```diff
@@ -456,7 +456,7 @@ class Series(IndexOpsMixin, NDFrame, Generic[S1]):
         dropna: _bool = ...,
     ) -> _SeriesGroupByNonScalar[S1]: ...
     @overload
-    def count(self, level: None = ...) -> int: ...
+    def count(self, level: None = ...) -> int: ...  # type: ignore[misc]
```
Dr-Irv, line 459: can you add a comment here as to why the `# type: ignore` is needed?

tests/test_io.py
```diff
@@ -822,7 +822,7 @@ def test_sqlalchemy_selectable() -> None:
             class Base(metaclass=sqlalchemy.orm.decl_api.DeclarativeMeta):
                 __abstract__ = True
 
-            class Temp(Base):
+            class Temp(Base):  # type: ignore[misc]
```
Dr-Irv, line 825: can you add a comment here as to why the `# type: ignore` is needed?

Dr-Irv (changes requested):
Looks good.  Add some comments to explain some issues, and we should be good to go.

Can you also track which issues we can close because of this?  (if there are any such issues)

bashtage:
Are most of these due to inheritance?

twoertwein:
> Are most of these due to inheritance?

Yes, mypy seems now to be stricter about enforcing that a `property` is inherited by a `property`.

tests/test_io.py
twoertwein: I honestly don't understand why - I added the error reported by mypy

Dr-Irv (approved):
Thanks @twoertwein

PR #421, opened 2022-11-10
CI: fix tests for pandas nightly

twoertwein:
Matplotlib seems to be too old for nightly. The change for mypy is unrelated (just making it consistent with how the version is specified for pandas).


Dr-Irv (approved):
Thanks @twoertwein

PR #427, opened 2022-11-17
Allow tuple[str, ...] pattern to be passed to str.startswith / str.endswith

gandhis1:


Dr-Irv (approved):
thanks @gandhis1

PR #428, opened 2022-11-17
Support time start/end for BusinessHour & CustomBusinessHour

### LINKED ISSUE #396: CustomBusinessHour accepts a time as start and end
This requires updating the docs in pandas as well, because it is not accurate:  https://pandas.pydata.org/docs/reference/api/pandas.tseries.offsets.CustomBusinessHour.html

```
import pandas as pd
from datetime import time
a = time(6, 0)
b = time(14, 0)

In: pd.tseries.offsets.CustomBusinessHour(start=a, end=b)
Out: <CustomBusinessHour: CBH=06:00-14:00>
```

Example error:

> error: Argument "start" to "CustomBusinessHour" has incompatible type "Optional[time]"; expected "str"
error: Argument "end" to "CustomBusinessHour" has incompatible type "Optional[time]"; expected "str"

gandhis1:
- [X] Closes #396 

### DIFF
--- a/pandas-stubs/_libs/tslibs/offsets.pyi
+++ b/pandas-stubs/_libs/tslibs/offsets.pyi
@@ -1,6 +1,7 @@
 from datetime import (
     date,
     datetime,
+    time,
     timedelta,
 )
 from typing import (
@@ -131,8 +132,8 @@ class BusinessHour(BusinessMixin):
         self,
         n: int = ...,
         normalize: bool = ...,
-        start: str | Collection[str] = ...,
-        end: str | Collection[str] = ...,
+        start: str | time | Collection[str | time] = ...,
+        end: str | time | Collection[str | time] = ...,
         offset: timedelta = ...,
     ): ...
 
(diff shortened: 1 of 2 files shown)


pandas-stubs/_libs/tslibs/offsets.pyi
```diff
@@ -131,8 +132,8 @@ class BusinessHour(BusinessMixin):
         self,
         n: int = ...,
         normalize: bool = ...,
-        start: str | Collection[str] = ...,
-        end: str | Collection[str] = ...,
```
gandhis1, line 135: I see no substantiation anywhere for this now removed type, seems it was copied from older code which may have been auto-generated

Dr-Irv (approved):
thanks @gandhis1 .  Since the docs don't reflect that `time` is a valid argument, can you create an issue in the pandas repo that says that the docs should be updated?

gandhis1:
Created PR here: https://github.com/pandas-dev/pandas/pull/49774

PR #430, opened 2022-11-17
fix freqstr in BaseOffset being a property

theendlessriver13:
- [ ] Closes #xxxx (Replace xxxx with the Github issue number) **[does not apply]**

The `freqstr` is supposed to be a `property`, not a method returning a `str`

with this examples

```python
from pandas import DateOffset

offset = DateOffset(minutes=10)
reveal_type(offset.freqstr)
```

and running `mypy` on this results in:

```console
tt.py:4: note: Revealed type is "def () -> builtins.str"
Success: no issues found in 1 source file
```

with this patch it is properly captured as `str`:

```console
tt.py:4: note: Revealed type is "builtins.str"
Success: no issues found in 1 source file
```

I ran into to this while doing this:

```python
import pandas

delta = pandas.Timedelta(minutes=10)
offset = pandas.tseries.frequencies.to_offset(delta)

freq = None
if offset is not None:
    freq = offset.freqstr
    full_idx = pandas.date_range('2021', '2022', freq=freq)
```

```console
tt.py:9: error: Argument "freq" to "date_range" has incompatible type "Callable[[], str]"; expected "Union[str, timedelta, Timedelta, BaseOffset]"  [arg-type]
```

I'd be happy to add a test for this, but I'm not sure how - maybe you can give me a hint. I tried this, but it passed both times, so does not capture the error where `Offset` becomes a `Callable[[], str]`.

```python
def test_freqstr() -> None:
    offset = DateOffset(minutes=10)
    check(assert_type(offset.freqstr, str), str)
```


Dr-Irv:
> I'd be happy to add a test for this, but I'm not sure how - maybe you can give me a hint. I tried this, but it passed both times, so does not capture the error where `Offset` becomes a `Callable[[], str]`.
> 
> ```python
> def test_freqstr() -> None:
>     offset = DateOffset(minutes=10)
>     check(assert_type(offset.freqstr, str), str)
> ```

The above test should work.  If you remove your change, at least pyright picks up that `assert_type` will fail.  `mypy` should do that as well, although I didn't test that.

Can you add the test, then I'll enable the CI test.

theendlessriver13:
thanks for looking into this! I added the test, but it passes also on `main`, so not verifying my fix.

Dr-Irv:
> 

Can you indicate how you tested it on main?

I did `poetry run poe mypy` and got this:
```text
tests\test_timefuncs.py:1036: error: Expression is of type "Callable[[], str]", not "str"  [assert-type]
```

theendlessriver13:
oh I was just using `pytest`.

When using `poetry run poe mypy` I get an error on `main` but it's passing on `freqstr-property` so exactly what I want!

But with `poetry run poe pytest` everything is passing (like when I was just using `pytest`)
I was not aware how the tests are supposed to be run - sorry.

Dr-Irv:
> oh I was just using `pytest`.
> 
> When using `poetry run poe mypy` I get an error on `main` but it's passing on `freqstr-property` so exactly what I want!
> 
> But with `poetry run poe pytest` everything is passing (like when I was just using `pytest`) I was not aware how the tests are supposed to be run - sorry.

If you do `poetry run poe test`, it will run `mypy`, `pyright` and `pytest`.  The first two do the type checking, which is what you are fixing. The last one checks that what we expect in the types is what happens at runtime.  At runtime, `assert_type(x, a)` just returns `x` which is why it worked with `pytest` on `main`

Dr-Irv (approved):
thanks @theendlessriver13

PR #431, opened 2022-11-17
added np.timedelta64 in series

### LINKED ISSUE #423: Series arithmetic methods accepting a Timedelta should also accept a np.timedelta64
```
import numpy as np
import pandas as pd

s1 = pd.Series(data=pd.date_range("1/1/2020", "2/1/2020"))
s2 = pd.Series(data=pd.date_range("1/1/2021", "2/1/2021"))
s3 = (s1 - s2) / np.timedelta64(1, "M")  # pd.Series[float]
```
>`  error: No overload variant of "__truediv__" of "Series" matches argument type "timedelta64"  [operator]`

This applies to most arithmetic methods that accept a `Timedelta`

ramvikrams:
- [x] Closes #423 

@Dr-Irv   Sir, I actually need some help in writing the test for np.timedelta64.


Dr-Irv:
> @Dr-Irv Sir, I actually need some help in writing the test for np.timedelta64.

Look for the places where we test arithmetic with `pd.Timedelta`, then write similar tests for `np.timedelta64`.   The challenge here is that the tests are scattered between `test_timefuncs.py` and `test_scalars.py` .

There is a test in `test_timefuncs.py` that is called `test_timedelta_series_arithmetic(()` that almost does the right thing, but it is testing arithmetic on a `TimedeltaIndex` and not on a `TimedeltaSeries`.  

Ideally, write a new function that creates a `Series` of timestamps, `Series` of floats, and then does arithmetic operations for those objects on the left hand side of the operator and a `Timedelta` or `np.timedelta64` on the right-hand side

pandas-stubs/core/series.pyi
```diff
@@ -1739,6 +1739,8 @@ class TimestampSeries(Series[Timestamp]):
     # ignore needed because of mypy
     @property
     def dt(self) -> TimestampProperties: ...  # type: ignore[override]
+    @overload
+    def __truediv__(self, other) -> TimestampSeries: ...  # type: ignore
```
bashtage, line 1743: Why are there no types for other here?

ramvikrams: > Why are there no types for other here?

i had to add, i thought to complete it after writing tests for np.timedelta64

ramvikrams:
> 



> > @Dr-Irv Sir, I actually need some help in writing the test for np.timedelta64.
> 
> Look for the places where we test arithmetic with `pd.Timedelta`, then write similar tests for `np.timedelta64`. The challenge here is that the tests are scattered between `test_timefuncs.py` and `test_scalars.py` .
> 
> There is a test in `test_timefuncs.py` that is called `test_timedelta_series_arithmetic(()` that almost does the right thing, but it is testing arithmetic on a `TimedeltaIndex` and not on a `TimedeltaSeries`.
> 
> Ideally, write a new function that creates a `Series` of timestamps, `Series` of floats, and then does arithmetic operations for those objects on the left hand side of the operator and a `Timedelta` or `np.timedelta64` on the right-hand side

thanks sir I'll start with it.

ramvikrams:
> 

Sir writing the test but thet do not work on my system
```
import numpy as np
import pandas as pd

s2 = pd.Series(data=pd.date_range("1/1/2028", "2/1/2029"))
s3 = (s2 - np.timedelta64(1, "h") )
```

Dr-Irv:
> Sir writing the test but thet do not work on my system

I'm not sure what you mean by "not work".  That code executes just fine.  If it is about type checking, that is what you need to fix!

ramvikrams:
> > Sir writing the test but thet do not work on my system
> 
> I'm not sure what you mean by "not work".  That code executes just fine.  If it is about type checking, that is what you need to fix!
> 

Yes sir got it thanks

PR #432, opened 2022-11-18
added np.timedelta64 for series arithmatic methods

### LINKED ISSUE #423: Series arithmetic methods accepting a Timedelta should also accept a np.timedelta64

ramvikrams:
- [x] Closes #423 

sorry sir had to close that draft pr because for some reason I was not able to generate a pr from that


tests/test_timefuncs.py
```diff
@@ -1035,3 +1035,28 @@ def test_timedelta_range() -> None:
 def test_dateoffset_freqstr() -> None:
     offset = DateOffset(minutes=10)
     check(assert_type(offset.freqstr, str), str)
+
+
+def timedelta64_and_arithmatic_operator() -> None:
+    s1 = pd.Series(data=pd.date_range("1/1/2020", "2/1/2020"))
+    s2 = pd.Series(data=pd.date_range("1/1/2021", "2/1/2021"))
+    check(assert_type((s1 - np.timedelta64(1, "M")), np.timedelta64), np.timedelta64)
```
Dr-Irv, line 1043: Just create a variable `td = np.timedelta64(1, "M")` and reuse it

tests/test_timefuncs.py
```diff
@@ -1035,3 +1035,28 @@ def test_timedelta_range() -> None:
...
+    check(assert_type((s1 + np.timedelta64(1, "M")), np.timedelta64), np.timedelta64)
+    check(assert_type(((s1 * np.timedelta64(1, "M")), np.timedelta64), np.timedelta64))
+    check(assert_type(((s1 / np.timedelta64(1, "M")), np.timedelta64), np.timedelta64))
```
Dr-Irv, line 1046: The results here are wrong, because `s1` is a `TimeStampSeries`, so adding and subtracting a `np.timedelta64` will produce the same.  Have to try each one out locally and see what the right type is.

tests/test_timefuncs.py
```diff
@@ -1035,3 +1035,28 @@ def test_timedelta_range() -> None:
...
+    check(
+        assert_type(((s1 - s2) - np.timedelta64(1, "M")), np.timedelta64),
+        np.timedelta64,
+    )
```
Dr-Irv, line 1050: Just create a variable `s1 - s2`, which is a `TimedeltaSeries`, then add, subtract, multiply, divide by the timedelta to see what happens.

Dr-Irv (changes requested):
You also have to modify the operators in `TimedeltaSeries`

ramvikrams:
> You also have to modify the operators in `TimedeltaSeries`

yes sir will do

tests/test_timefuncs.py
ramvikrams: Sir it's worling fine with addition,substraction only problem is with multiplication,division any suggestions for that

Dr-Irv: > Sir it's worling fine with addition,substraction only problem is with multiplication,division any suggestions for that

Doesn't seem that it is working.  Look at the log files from the CI runs (or run locally) and you can then see the error messages and fix the code.

ramvikrams:
Done sir

tests/test_timefuncs.py
```diff
@@ -1035,3 +1035,30 @@ def test_timedelta_range() -> None:
...
     check(assert_type(offset.freqstr, str), str)
+
+
+def timedelta64_and_arithmatic_operator() -> None:
+    s1 = pd.Series(data=pd.date_range("1/1/2020", "2/1/2020"))
+    s2 = pd.Series(data=pd.date_range("1/1/2021", "2/1/2021"))
+    s3 = s2 - s1
+    td = np.timedelta64(1, "M")
+    check(assert_type((s1 - td), TimestampSeries), np.timedelta64)
+    check(assert_type((s1 + td), pd.Series[Any]), np.timedelta64)
```
Dr-Irv, line 1046: This is incorrect.  `s1` is `TimestampSeries`, so you need to add a stub for `__add__()` in `TimestampSeries` with arguments `Timedelta | np.timedelta64`.  Might also need a `@overload`

tests/test_timefuncs.py
```diff
@@ -1035,3 +1035,30 @@ def test_timedelta_range() -> None:
...
+    check(assert_type((s1 * td), TimedeltaSeries), np.timedelta64)
+    check(assert_type((s1 / td), pd.Series[float]), np.timedelta64)
```
Dr-Irv, line 1048: For the above 4 checks, I am surprised they got past `pytest`, because the ehecks should be of the form `check(assert_type((s1 - td), TimestampSeries), pd.Series, pd.Timestamp)` (for example)

tests/test_timefuncs.py
```diff
@@ -1035,3 +1035,30 @@ def test_timedelta_range() -> None:
...
+    check(
+        assert_type((s3 - td), TimedeltaSeries),
+        pd.Series,
```
Dr-Irv, line 1051: add `pd.Timedelta` to the last argument of `check` here

tests/test_timefuncs.py
```diff
@@ -1035,3 +1035,30 @@ def test_timedelta_range() -> None:
...
+    )
+    check(
+        assert_type((s3 + td), TimedeltaSeries),
+        pd.Series,
+    )
+    check(
+        assert_type((s3 * td), TimedeltaSeries),
+        pd.Series,
+    )
```
Dr-Irv, line 1060: for above 2 checks, add `pd.Timedelta` to the last argument of `check`

tests/test_timefuncs.py
```diff
@@ -1035,3 +1035,30 @@ def test_timedelta_range() -> None:
...
+    check(
+        assert_type((s3 / td), pd.Series[float]),
+        pd.Series,
+    )
```
Dr-Irv, line 1064: add `float` to last argument of `check` .  (Might need to be `np.float`)

Dr-Irv (changes requested):

pandas-stubs/core/series.pyi
```diff
@@ -1743,6 +1743,8 @@ class TimestampSeries(Series[Timestamp]):
     # ignore needed because of mypy
     @property
     def dt(self) -> TimestampProperties: ...  # type: ignore[override]
+    @overload
+    def __add__(self, other: Timedelta | np.timedelta64) -> TimedeltaSeries: ...
```
Dr-Irv, line 1747: result here of adding `TimestampSeries` plus a timedelta is `TimestampSeries`

tests/test_timefuncs.py
```diff
@@ -1042,23 +1042,11 @@ def timedelta64_and_arithmatic_operator() -> None:
...
-        assert_type((s3 * td), TimedeltaSeries),
-        pd.Series,
-    )
-    check(
-        assert_type((s3 / td), pd.Series[float]),
-        pd.Series,
-    )
+    check(assert_type((s1 - td), TimestampSeries), pd.Series, pd.Timestamp)
+    check(assert_type((s1 + td), TimestampSeries), pd.Series, pd.Timestamp)
+    check(assert_type((s1 * td), TimedeltaSeries), pd.Series, pd.Timestamp)
```
Dr-Irv, line 1047: multipying a `TimestampSeries` times a `Timedelta` is not allowed, so have to fix that.

In `TimestampSeries`, put in a method for `__mul__()` that takes the two timedelta arguments, and returns `Never` from `typing_extensions`

Change the test to assert the type, but do not do the `check` part.

tests/test_timefuncs.py
```diff
@@ -1042,23 +1042,11 @@ def timedelta64_and_arithmatic_operator() -> None:
...
+    check(assert_type((s1 / td), TimestampSeries), pd.Series, pd.Timestamp)
```
Dr-Irv, line 1048: dividing a `TimestampSEries` by a `Timedelta` is not allowed, so have to fix that

In `TimestampSeries`, put in a method for `__truediv__()` that takes the two timedelta arguments, and returns `Never` from `typing_extensions`

Change the test to assert the type, but do not do the `check` part.

tests/test_timefuncs.py
```diff
@@ -1042,23 +1042,11 @@ def timedelta64_and_arithmatic_operator() -> None:
...
+    check(assert_type((s3 - td), TimedeltaSeries), pd.Series, pd.Timedelta)
+    check(assert_type((s3 + td), TimedeltaSeries), pd.Series, pd.Timedelta)
+    check(assert_type((s3 * td), TimedeltaSeries), pd.Series, pd.Timedelta)
```
Dr-Irv, line 1051: multiplying a `TimedeltaSeries` by a timedelta fails, so have to fix that

In `TimedeltaSeries`, put in a method for `__mul__()` that takes the two timedelta arguments, and returns `Never` from `typing_extensions`

Change the test to assert the type, but do not do the `check` part.

Dr-Irv (changes requested):
Please do `poetry run poe test` before pushing new changes to make sure things work locally

tests/test_timefuncs.py
ramvikrams: after doing all this in mypy is having issues, it is saying `error: Expression is of type "NoReturn", not "TimestampSeries"  [assert-type]`

tests/test_timefuncs.py
ramvikrams: after doing all this in mypy is having issues, it is saying `error: Expression is of type "NoReturn", not "TimestampSeries"  [assert-type]`

tests/test_timefuncs.py
ramvikrams: And in `TimedeltaSeries ` should a `__trudiv__()` be added because it is showing errors

ramvikrams: and sir `__mul__()` it is still showing errors after adding  it.

pandas-stubs/core/series.pyi
ramvikrams: sir here there is no `TimestampSeries`

Dr-Irv: Look at line 1742

tests/test_timefuncs.py
Dr-Irv: > Sir in `TimedeltaSeries` should a `__trudiv__()` be added because it is showing errors

Yes, you should add that.  It is valid to divide `TimedeltaSeries` by `Timedelta` or `np.timedelta64` to get a `Series[float]`, and to divde by a `float` to get `TimedeltaSeries`

ramvikrams: Sir just not able to fix this `error: Name "__mul__" already defined on line 1761  [no-redef]` could not find about it over internet also

Dr-Irv: You probably have `__mul__()` defined in there twice, with different arguments.  You need to put a `@overload` before each `def __mul__()` statement

ramvikrams: Sir still mypy hows the same error

Dr-Irv: Without seeing your code, I can't see what you are doing wrong.  You can push and it will fail and I will tell you what you did wrong.

ramvikrams: yes sir

pandas-stubs/core/series.pyi
```diff
@@ -1755,10 +1757,14 @@ class TimedeltaSeries(Series[Timedelta]):
...
     def __add__(self, other: Timedelta | np.timedelta64) -> TimedeltaSeries: ...
     def __radd__(self, pther: Timestamp | TimestampSeries) -> TimestampSeries: ...  # type: ignore[override]
+    @overload
     def __mul__(self, other: num) -> TimedeltaSeries: ...  # type: ignore[override]
     def __sub__(  # type: ignore[override]
         self, other: Timedelta | TimedeltaSeries | TimedeltaIndex | np.timedelta64
     ) -> TimedeltaSeries: ...
+    def __truediv__(self, other: TimedeltaSeries | np.timedelta64) -> Series[float]: ...  # type: ignore[override]
+    @overload
+    def __mul__(self, other: TimestampSeries | np.timedelta64) -> Never: ...  # type: ignore[override]
```
Dr-Irv, line 1767: you need to put all the definitions of the same methods that have overloads together.  Move lines 1766-1767 to after line 1761

ramvikrams: thanks sir

ramvikrams: > you need to put all the definitions of the same methods that have overloads together. Move lines 1766-1767 to after line 1761

done sir

Dr-Irv: > `error: Signature of "__mul__" incompatible with supertype "Series" [override]` this is the last mypy error it doesn't go if I add `# type: ignore[override]` from line number 1760

It's OK to put in the `# type: ignore[override]` in this case.

ramvikrams: > > `error: Signature of "__mul__" incompatible with supertype "Series" [override]` this is the last mypy error it doesn't go if I add `# type: ignore[override]` from line number 1760
> 
> It's OK to put in the `# type: ignore[override]` in this case.

then it shows this error `error: Unused "type: ignore" comment`

Dr-Irv: check the line numbers that the error is reported on

ramvikrams: on  the same line where the incompatible with supertype error is

pandas-stubs/core/series.pyi
```diff
@@ -1739,20 +1744,27 @@ class TimestampSeries(Series[Timestamp]):
...
     def __add__(self, other: Timestamp | DatetimeIndex) -> TimestampSeries: ...
     @overload
-    def __add__(self, other: Timedelta) -> TimedeltaSeries: ...
+    def __add__(self, other: Timedelta | np.timedelta64) -> TimedeltaSeries: ...
     def __radd__(self, pther: Timestamp | TimestampSeries) -> TimestampSeries: ...  # type: ignore[override]
-    def __mul__(self, other: num) -> TimedeltaSeries: ...  # type: ignore[override]
+    @overload
+    def __mul__(self, other: TimestampSeries | np.timedelta64 | Timedelta | TimedeltaSeries) -> Never: ...
+    @overload
+    def __mul__(self, other: num) -> TimedeltaSeries: ...
```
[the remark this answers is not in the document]
ramvikrams: Thanks sir, it's complete it passed my local checks

tests/test_timefuncs.py
```diff
@@ -1043,11 +1043,11 @@ def timedelta64_and_arithmatic_operator() -> None:
...
-    assert_type((s3 * td), NoReturn)
-    assert_type((s3 / td), pd.Series[float])
+    assert_type((s1 - td), TimestampSeries)  # type: ignore
+    assert_type((s1 + td), TimestampSeries)  # type: ignore
+    assert_type((s1 * td), NoReturn)  # type: ignore
+    assert_type((s1 / td), NoReturn)  # type: ignore
+    assert_type((s3 - td), TimedeltaSeries)  # type: ignore
+    assert_type((s3 + td), TimedeltaSeries)  # type: ignore
+    assert_type((s3 * td), NoReturn)  # type: ignore
+    assert_type((s3 / td), pd.Series[float])  # type: ignore
```
Dr-Irv, line 1053: Do NOT put `#type: ignore` in the testing functions.

ramvikrams: ok sir

ramvikrams:
Done sir, I think it's complete

tests/test_timefuncs.py
```diff
@@ -1035,3 +1036,18 @@ def test_timedelta_range() -> None:
...
+    s2 = pd.Series(data=pd.date_range("1/1/2021", "2/1/2021"))
+    s3 = s2 - s1
+    td = np.timedelta64(1, "M")
+    assert_type((s1 - td), TimestampSeries)
+    assert_type((s1 + td), TimestampSeries)
+    assert_type((s1 * td), NoReturn)  # pyright: ignore
+    assert_type((s1 / td), NoReturn)  # pyright: ignore
+    assert_type((s3 - td), TimedeltaSeries)
+    assert_type((s3 + td), TimedeltaSeries)
+    assert_type((s3 * td), NoReturn)  # pyright: ignore
```
Dr-Irv, line 1052: Can you do the following:
1. for the 4 cases where you are expecting `TimestampSeries` or `TimedeltaSeries`, use `check(assert_type(  operation, expected_type), expected_type, expected_subtype)` where operation is the operation you are testing, expected_type is the type expected, and expected_subtype is `pd.Timestamp` or `pd.Timedelta` as appropriate
2. For the 3 cases where you have `NoReturn`, can you change that to `Never` , which you will need to import from `typing_extensions`
3. The `# pyright: ignore` is correct - pyright is doing the right thing here, but mypy is not

tests/test_timefuncs.py
```diff
@@ -1035,3 +1036,18 @@ def test_timedelta_range() -> None:
...
+    assert_type((s3 / td), pd.Series[float])
```
Dr-Irv, line 1053: use `check` on this one too

Dr-Irv (changes requested):

tests/test_timefuncs.py
ramvikrams: means on all three of these except the NoReturn one

tests/test_timefuncs.py
ramvikrams: Sir regarding point 2 and 3 when I changed it to `Never` pyrigth shows error, mypy works ok and if we add the `#type: ignore` both work fine so should I add it.
Because in the pyright error it says `"assert_type" mismatch: expected "Never" but received "Unknown"`

Dr-Irv: Yes, you will still need the `# pyright: ignore` where you had it before

ramvikrams: done sir

tests/test_timefuncs.py
```diff
@@ -1035,3 +1038,18 @@ def test_timedelta_range() -> None:
 def test_dateoffset_freqstr() -> None:
     offset = DateOffset(minutes=10)
     check(assert_type(offset.freqstr, str), str)
+
+
+def timedelta64_and_arithmatic_operator() -> None:
```
Dr-Irv, line 1043: the name of any test function must begin with `test_`.  If you make that change, you will see that the tests will fail.  They were never getting tested.

You will have to change the tests as well.  See example.

tests/test_timefuncs.py
```diff
@@ -1035,3 +1038,18 @@ def test_timedelta_range() -> None:
...
+    s1 = pd.Series(data=pd.date_range("1/1/2020", "2/1/2020"))
+    s2 = pd.Series(data=pd.date_range("1/1/2021", "2/1/2021"))
+    s3 = s2 - s1
+    td = np.timedelta64(1, "M")
+    check(assert_type((s1 - td), TimestampSeries), TimestampSeries, pd.Timestamp)
```
Dr-Irv, line 1048:
```
check(assert_type((s1 - td), "TimestampSeries"), pd.Series, pd.Timestamp)
```
All the tests should look like this.  The type `TimestampSeries` isn't known at runtime, so you have to surround it with quotes.  

you will have to do something similar for `TimedeltaSeries` and `pd.Series[float]`. 

Look elsewhere in the code to see examples

Dr-Irv (changes requested):

tests/test_timefuncs.py
ramvikrams: Sir even after adding the `test_` the `poe mypy` test doesn't fail on my system

Dr-Irv: Do `poe test` to run all tests or `poe pytest` to just run `pytest` which is where the failures will occur

ramvikrams: yes sir there were errors, I'll fix it then

tests/test_timefuncs.py
ramvikrams: Should we also change thte left 3 tests because they cannot pass pytets

Dr-Irv: Good point.  Move those 3 tests to the end and do
```python
     # test 1 here
     # test 2 here
     # test 3 here
```
you import `TYPE_CHECKING_INVALID_USAGE` from `tests`

pandas-stubs/core/series.pyi
```diff
@@ -1746,7 +1746,12 @@ class TimestampSeries(Series[Timestamp]):
     # ignore needed because of mypy
     @property
     def dt(self) -> TimestampProperties: ...  # type: ignore[override]
-    def __add__(self, other: TimedeltaSeries | np.timedelta64 | TimestampSeries | Timestamp) -> TimestampSeries: ...  # type: ignore[override]
+    @overload  # type: ignore[override]
+    def __add__(
+        self, other: TimedeltaSeries | np.timedelta64 | TimestampSeries
+    ) -> TimestampSeries: ...
+    @overload
+    def __add__(self, other: Timestamp) -> TimestampSeries: ...
```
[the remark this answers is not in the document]
ramvikrams: We have issue in line number 266 of `tests/time_funcs.py` if we change this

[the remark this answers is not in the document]
Dr-Irv: On that line, change `# TODO both: ignore[operator]` to `# pyright: ignore`

Dr-Irv (approved):
thanks @ramvikrams 

By the way, in the future, when you make commits, please make the commit message reflect the changes you make, not just use the word "update"

ramvikrams:
> thanks @ramvikrams
> 
> By the way, in the future, when you make commits, please make the commit message reflect the changes you make, not just use the word "update"

yes sir sure I will keep that in mind

PR #435, opened 2022-11-20
Python 3.11 support

### LINKED ISSUE #400: Python 3.11 support

Dr-Irv:
- [x] Closes #400 

Before merging, awaiting resolution of https://github.com/pandas-dev/pandas/pull/49736 which I hope will get into pandas 1.5.2, and, if so, we can then remove the warning in `test_frame.py`, otherwise we release with it.

Putting in draft status for now.  Should replace PR #398

Pinned pyright to 1.1.279 and 1.1.280 is reporting errors with relative imports
https://github.com/microsoft/pyright/issues/4212


tests/__init__.py
```diff
@@ -21,6 +22,21 @@
 WINDOWS = os.name == "nt" or "cygwin" in platform.system().lower()
 PD_LTE_15 = Version(pd.__version__) < Version("1.5.999")
 
+arrow_skip = pytest.mark.skipif(
+    sys.version_info >= (3, 11), reason="pyarrow is not available for 3.11 yet"
+)
+"""This is only needed temporarily due to no wheels being available for arrow on 3.11"""
```
bashtage, line 28: Should this be a comment rather than a song?

tests/__init__.py
```diff
@@ -21,6 +22,21 @@
...
+
+lxml_skip = pytest.mark.skipif(
+    sys.version_info >= (3, 11), reason="lxml is not available for 3.11 yet"
+)
+"""This is only needed temporarily due to no wheels being available for lxml on 3.11"""
```
bashtage, line 33: Same here and below

tests/test_pandas.py
```diff
@@ -464,14 +465,15 @@ def test_crosstab() -> None:
 def test_arrow_dtype() -> None:
     pytest.importorskip("pyarrow")
 
-    import pyarrow as pa
+    if sys.version_info < (3, 11):
+        import pyarrow as pa
```
bashtage, line 469: Is this because pa is available but didn't work?

bashtage (commented):
Some comments.  Would be good to explain why the Python version conditional test with a comment.

tests/__init__.py
Dr-Irv: I think you meant "string", not "song", but fixed in next commit

tests/test_pandas.py
Dr-Irv: No, it's because we don't want the type checker to do the type checking here with python 3.11, since pyarrow is not available yet.

gandhis1: I see a merged PR in the pyarrow repo that suggests 3.11 is supported? 

https://github.com/apache/arrow/pull/14499

bashtage:
LGTM.  Nightly is just a canary anyway.

tests/__init__.py
bashtage: Yes,

bashtage (approved):
LGTM.

Dr-Irv:
> I don't know if I should mention this as it might delay this being released slightly. But now that pyarrow 10.0.1 has been released, I believe that is supposed to have python 3.11 support.

Thanks for the heads up.  I saw your message in time and have pushed changes for `pyarrow` support with 3.11

.github/setup/action.yaml
```diff
@@ -21,7 +21,7 @@ runs:
 
     - name: Determine poetry version
       shell: bash
-      run: echo "::set-output name=VERSION::$(poetry --version)"
+      run: echo "{VERSION}=$(poetry --version)"
```
Dr-Irv, line 24: See https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/  .  We were getting warnings, and this gets rid of them

bashtage (approved):
LGTM.

PR #448, opened 2022-11-26
Fix annotations for DataFrame.apply() including additional overloads

gandhis1:

In #401 the annotations for `apply` was changed. This introduced false negatives in my code base, because the particular mix of return types and arguments I was using was not supported. Here I am still not sure it is exhaustive but I think it's an improvement. In particular the `result_type` needs to be accounted for, as well as differentiating between a true list-like, `str`, and `Series` in the return value of the function.

@Dr-Irv @bashtage @danielroseman

### DIFF
--- a/pandas-stubs/_typing.pyi
+++ b/pandas-stubs/_typing.pyi
@@ -8,6 +8,7 @@ from typing import (
     Iterator,
     Literal,
     Mapping,
+    MutableSequence,
     Optional,
     Protocol,
     Sequence,
@@ -155,6 +156,9 @@ AxisType: TypeAlias = Literal["columns", "index", 0, 1]
 DtypeNp = TypeVar("DtypeNp", bound=np.dtype[np.generic])
 KeysArgType: TypeAlias = Any
 ListLike = TypeVar("ListLike", Sequence, np.ndarray, "Series", "Index")
+ListLikeExceptSeriesAndStr = TypeVar(
+    "ListLikeExceptSeriesAndStr", MutableSequence, np.ndarray, tuple, "Index"
+)
 ListLikeU: TypeAlias = Union[Sequence, np.ndarray, Series, Index]
 StrLike: TypeAlias = Union[str, np.str_]
 Scalar: TypeAlias = Union[
--- a/pandas-stubs/_typing.pyi
+++ b/pandas-stubs/_typing.pyi
@@ -152,7 +152,9 @@ num: TypeAlias = complex
 SeriesAxisType: TypeAlias = Literal[
     "index", 0
 ]  # Restricted subset of _AxisType for series
-AxisType: TypeAlias = Literal["columns", "index", 0, 1]
+AxisTypeIndex: TypeAlias = Literal["index", 0]
+AxisTypeColumn: TypeAlias = Literal["columns", 1]
+AxisType: TypeAlias = AxisTypeIndex | AxisTypeColumn
 DtypeNp = TypeVar("DtypeNp", bound=np.dtype[np.generic])
 KeysArgType: TypeAlias = Any
 ListLike = TypeVar("ListLike", Sequence, np.ndarray, "Series", "Index")
--- a/pandas-stubs/core/frame.pyi
+++ b/pandas-stubs/core/frame.pyi
@@ -1090,7 +1090,7 @@ class DataFrame(NDFrame, OpsMixin):
         **kwargs,
     ) -> DataFrame: ...
 
-    # First set of apply() overloads is with defaults
+    # apply() overloads with default result_type of None, and is indifferent to axis
     @overload
     def apply(
         self,
@@ -1104,26 +1104,26 @@ class DataFrame(NDFrame, OpsMixin):
     @overload
     def apply(
         self,
-        f: Callable[..., Scalar],
+        f: Callable[..., S1],
         axis: AxisTypeIndex = ...,
         raw: _bool = ...,
         result_type: Literal[None] = ...,
         args=...,
         **kwargs,
-    ) -> Series: ...
+    ) -> Series[S1]: ...
 
-    # Second set of apply() overloads is with keyword result_type
+    # apply() overloads with keyword result_type, and axis does not matter
     @overload
     def apply(
         self,
-        f: Callable[..., ListLikeExceptSeriesAndStr],
+        f: Callable[..., S1],
         axis: AxisType = ...,
         raw: _bool = ...,
         args=...,
         *,
-        result_type: Literal["reduce"],
+        result_type: Literal["expand", "reduce"],
         **kwargs,
-    ) -> Series: ...
+    ) -> Series[S1]: ...
     @overload
     def apply(
         self,
@@ -1138,7 +1138,18 @@ class DataFrame(NDFrame, OpsMixin):
     @overload
     def apply(
         self,
-        f: Callable[..., ListLikeExceptSeriesAndStr | Series],
+        f: Callable[..., ListLikeExceptSeriesAndStr],
+        axis: AxisType = ...,
+        raw: _bool = ...,
+        args=...,
+        *,
+        result_type: Literal["reduce"],
+        **kwargs,
+    ) -> Series: ...
+    @overload
+    def apply(
+        self,
+        f: Callable[..., ListLikeExceptSeriesAndStr | Series | Scalar],
         axis: AxisType = ...,
         raw: _bool = ...,
         args=...,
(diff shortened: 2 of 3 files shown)


pandas-stubs/_typing.pyi
```diff
@@ -155,6 +156,9 @@ AxisType: TypeAlias = Literal["columns", "index", 0, 1]
 DtypeNp = TypeVar("DtypeNp", bound=np.dtype[np.generic])
 KeysArgType: TypeAlias = Any
 ListLike = TypeVar("ListLike", Sequence, np.ndarray, "Series", "Index")
+ListLikeExceptSeriesAndStr = TypeVar(
+    "ListLikeExceptSeriesAndStr", MutableSequence, np.ndarray, tuple, "Index"
+)
```
gandhis1, line 161: I agree, the name is awful, but do we have a better way to exclude `str` from list-likes?

Dr-Irv: I don't know of any

Dr-Irv (changes requested):
Can you add a test for `result_type="reduce"`, since you now differentiate between the possible values of `arg_type` ?

Otherwise looks good

pandas-stubs/core/frame.pyi
```diff
@@ -1086,36 +1089,178 @@ class DataFrame(NDFrame, OpsMixin):
...
+        f: Callable[..., ListLikeExceptSeriesAndStr | Series],
+        raw: _bool = ...,
+        args=...,
+        *,
+        axis: AxisTypeColumn = ...,
+        result_type: Literal["broadcast"],
+        **kwargs,
+    ) -> DataFrame: ...
+
+    # Add spacing between apply() overloads and remaining annotations
```
gandhis1, line 1263: @Dr-Irv Do you mind looking over what I did with the overloads here? I believe it is correct to have to enumerate all the permutations of keyword only and default parameters, but not sure if this has unnecessary overloads or not.

I will note the reason why I separated out `axis=0` and `axis=1` is because the returned type in fact does change (when `result_type` = `None`).

I will note, MyPy seems to have trouble with one of the test cases where it thinks that `apply()` returns an `Any`. That makes no sense and I'm inclined to silence that. Pyright passes just fine.

pandas-stubs/core/frame.pyi
```diff
@@ -1086,36 +1089,124 @@ class DataFrame(NDFrame, OpsMixin):
...
+    def apply(
+        self,
+        f: Callable[..., ListLikeExceptSeriesAndStr | Series],
+        raw: _bool = ...,
+        args=...,
+        *,
+        axis: AxisTypeColumn = ...,
+        result_type: Literal["broadcast"],
+        **kwargs,
+    ) -> DataFrame: ...
```
Dr-Irv, line 1207: You could combine the above two overloads with the first one using `result_type: Literal["expand", "broadcast"] | None = ...`

I also think you want `axis: AxisTypeColumn` here without the `...` because these apply when the axis=1

tests/test_frame.py
```diff
@@ -486,6 +486,182 @@ def gethead(s: pd.Series, y: int) -> pd.Series:
...
+    )
+
+    # While this call works in reality, it errors in the type checker, because this should never be called
+    # It does not make sense to pass a result_type of "expand" to a scalar return
+    # check(
+    #     assert_type(
+    #         df.apply(returns_scalar, result_type="expand"), pd.DataFrame
+    #     ),
+    #     pd.DataFrame,
+    # )
```
Dr-Irv, line 510: When I run the above, it returns `Series` so it seems that you typed it wrong when testing.  Can you put it back using `pd.Series` ?

tests/test_frame.py
```diff
@@ -486,6 +486,182 @@ def gethead(s: pd.Series, y: int) -> pd.Series:
...
+    )
+
+    # While this call works in reality, it errors in the type checker, because this should never be called
+    # It does not make sense to pass a result_type of "reduce" to a scalar or series return
+    # check(
+    #     assert_type(
+    #         df.apply(returns_scalar, result_type="reduce"), pd.DataFrame
+    #     ),
+    #     pd.DataFrame,
+    # )
```
Dr-Irv, line 529: above returns `Series`, so can you fix that?

tests/test_frame.py
```diff
@@ -486,6 +486,182 @@ def gethead(s: pd.Series, y: int) -> pd.Series:
...
+    # check(
+    #     assert_type(
+    #         df.apply(returns_series, result_type="reduce"), pd.Series
+    #     ),
+    #     pd.Series,
+    # )
```
Dr-Irv, line 535: The above should work, which means the typing is wrong.

tests/test_frame.py
```diff
@@ -486,6 +486,182 @@ def gethead(s: pd.Series, y: int) -> pd.Series:
...
+    )
+
+    # While this call works in reality, it errors in the type checker, because this should never be called
+    # It does not make sense to pass a result_type of "broadcast" to a scalar return
+    # check(
+    #     assert_type(
+    #         df.apply(returns_scalar, result_type="broadcast"), pd.DataFrame
+    #     ),
+    #     pd.DataFrame,
+    # )
```
Dr-Irv, line 548: The above should work if the typing is right.

tests/test_frame.py
```diff
@@ -486,6 +486,182 @@ def gethead(s: pd.Series, y: int) -> pd.Series:
...
+        pd.Series,
+    )
+
+    # While this call works in reality, it errors in the type checker, because this should never be called
+    # It does not make sense to pass a result_type of "expand" to a scalar return
+    # check(
+    #     assert_type(
+    #         df.apply(returns_scalar, axis=1, result_type="expand"), pd.DataFrame
+    #     ),
+    #     pd.DataFrame,
```
Dr-Irv, line 582: should work but returns a `Series`.

tests/test_frame.py
```diff
@@ -486,6 +486,182 @@ def gethead(s: pd.Series, y: int) -> pd.Series:
...
+    )
+
+    # While this call works in reality, it errors in the type checker, because this should never be called
+    # It does not make sense to pass a result_type of "reduce" to a scalar or series return
+    # check(
+    #     assert_type(
+    #         df.apply(returns_scalar, axis=1, result_type="reduce"), pd.DataFrame
+    #     ),
+    #     pd.DataFrame,
+    # )
```
Dr-Irv, line 604: above returns a `Series`, so fix the typing (and the test)

tests/test_frame.py
```diff
@@ -486,6 +486,182 @@ def gethead(s: pd.Series, y: int) -> pd.Series:
...
+    # check(
+    #     assert_type(
+    #         df.apply(returns_series, axis=1, result_type="reduce"), pd.DataFrame
+    #     ),
+    #     pd.DataFrame,
+    # )
```
Dr-Irv, line 610: above should work

tests/test_frame.py
```diff
@@ -486,6 +486,182 @@ def gethead(s: pd.Series, y: int) -> pd.Series:
...
+        pd.Series,
+    )
+
+    # While this call works in reality, it errors in the type checker, because this should never be called
+    # It does not make sense to pass a result_type of "broadcast" to a scalar return
+    # check(
+    #     assert_type(
+    #         df.apply(returns_scalar, axis=1, result_type="broadcast"), pd.DataFrame
+    #     ),
+    #     pd.DataFrame,
```
Dr-Irv, line 624: above should work

Dr-Irv (changes requested):
I Like the  additional tests, but all the ones you commented out do work, so we should accept them in typing.

tests/test_frame.py
Dr-Irv: It may make no sense, but I tried it and it worked.  In other words, it doesn't do nothing.  It does something!  So not sure what to do about that!

gandhis1: By nothing I mean that specifying `result_type="expand"` gives the same result as if you don't specify it (because it's not possible to expand a scalar).

The reason why it doesn't throw is because pandas has no way of knowing whether your function always returns a scalar or not, it has to assume that it's possible for some values to be scalars and some values to be list-like.

However, with annotations, we do know. Because a `Callable[[pd.Series], int]` can be assumed to always return an `int`.

I can certainly put it back, just felt that it might make sense to disallow it...

Dr-Irv: I see your point, but we've taken the position that if the code works with pandas, we can't disallow it in the typing.  If someone is using `expand` with a `Scalar`, while it doesn't make sense, one could argue that our documentation doesn't say it is invalid.

gandhis1: Fair enough, I will uncomment these test cases and adjust the overloads as needed

gandhis1:
I've updated the overloads to also handle the scalar value type when a `Series[T]` is returned. I've also updated the tests to include all cases which return a value.

Dr-Irv (approved):
thanks @gandhis1

PR #450, opened 2022-11-26
Permit covariance of key type in read_csv converters argument

gandhis1:

If it accepts a `Dict[int | str, ...]` it should accept a `Dict` of either key type.

### DIFF
--- a/pandas-stubs/io/parsers/readers.pyi
+++ b/pandas-stubs/io/parsers/readers.pyi
@@ -8,6 +8,7 @@ from typing import (
     Any,
     Callable,
     Literal,
+    Mapping,
     Sequence,
     overload,
 )
@@ -48,9 +49,9 @@ def read_csv(
     | None = ...,
     dtype: DtypeArg | defaultdict | None = ...,
     engine: CSVEngine | None = ...,
-    converters: dict[int | str, Callable[[str], Any]]
-    | dict[int, Callable[[str], Any]]
-    | dict[str, Callable[[str], Any]]
+    converters: Mapping[int | str, Callable[[str], Any]]
+    | Mapping[int, Callable[[str], Any]]
+    | Mapping[str, Callable[[str], Any]]
     | None = ...,
     true_values: list[str] = ...,
     false_values: list[str] = ...,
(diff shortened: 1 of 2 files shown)


tests/test_frame.py
```diff
@@ -1278,6 +1279,43 @@ def test_read_csv() -> None:
             pd.DataFrame,
         )
 
+        # Allow a variety of dict types for the converters parameter
+        converters1 = {"A": lambda x: str, "B": lambda x: str}
```
gandhis1, line 1283: Originally I wanted this dictionary to be `{"A": str, "B": str}`. But the type of `str` is `Type[str]`, which apparently isn't compatible with `Callable[[str], Any]`. Any idea if this is fixable?

Dr-Irv: Might be an issue with using a `lambda` function, which is untyped.  So if you did
```python
def convert_to_str(a: object) -> str:
    return str(a)
```
and then 
```python
converters1 = {"A":convert_to_str, "B": convert_to_str}
```
that might work

pandas-stubs/io/parsers/readers.pyi
```diff
@@ -48,7 +48,10 @@ def read_csv(
     | None = ...,
     dtype: DtypeArg | defaultdict | None = ...,
     engine: CSVEngine | None = ...,
-    converters: dict[int | str, Callable[[str], Any]] | None = ...,
+    converters: dict[int | str, Callable[[str], Any]]
+    | dict[int, Callable[[str], Any]]
+    | dict[str, Callable[[str], Any]]
```
Dr-Irv, line 53: Can you try `Mapping[int | str, Callable[[str], Any]]` instead of the union?

Dr-Irv (changes requested):
Try avoiding the `lambda` if you can.  (or at least add tests without it)

tests/test_frame.py
gandhis1: Oh, the current version works fine because it's actually a `Callable`. What isn't working is passing `Type[str]` directly. That makes the dict a `Dict[str, Type[str]]`. That's what I was wondering about.

pandas-stubs/io/parsers/readers.pyi
gandhis1: I tried that first. But it seems `Mapping` isn't covariant over the key, only the value: https://github.com/python/typing/issues/445

gandhis1: But mapping may work for that the `Type[str]` issue since that is the value...will try that

I also wonder if creating a covariant `TypeVar` and then a `Mapping[T_co, ...]` will work...I will also try that.

gandhis1:
Ok, the version I just pushed should fix all of that, and also fix the other parameters too - `na_values` and `parse_dates`.

This is something that is probably going to come up over and over again throughout this code base. I think as a rule of thumb all `Dict` inputs need to be a `Mapping` if we want value covariance, and we need to do unions if we want key covariance. And honestly I would guess the overwhelming majority of places we want both.

tests/test_frame.py
Dr-Irv: I guess the type checkers know that `str` could be either a type or a callable.

Dr-Irv:
> This is something that is probably going to come up over and over again throughout this code base. I think as a rule of thumb all `Dict` inputs need to be a `Mapping` if we want value covariance, and we need to do unions if we want key covariance. And honestly I would guess the overwhelming majority of places we want both.

I think we have to look at all of them and see.  Probably `dict[str, int]` is OK, but if we have two or more possibilities for the value, then we should switch to `Mapping`, and if  we have two or more possibilities for the key, then we have to use the pattern you used here by including each possible key with union.

Dr-Irv (approved):
thanks @gandhis1

PR #455, opened 2022-11-29
added return types for "SeriesGroupBy" and "DataFrameGroupBy" 

### LINKED ISSUE #438: groupby(...).transform is untyped
**Describe the bug**
`SeriesGroupBy.transform` and `DataFrameGroupBy.transform` are currently not typed:

https://github.com/pandas-dev/pandas-stubs/blob/001c0afb3ef0dfe21e96ced87c9e455c657a50c3/pandas-stubs/core/groupby/generic.pyi#L159

Running mypy with `--disallow-untyped-calls` reports an issue with any usage of this method.

ramvikrams:
- [x] Closes #438 


tests/test_frame.py
```diff
@@ -2063,3 +2063,25 @@ def test_setitem_none() -> None:
...
+    ser = pd.Series(
+        [390.0, 350.0, 30.0, 20.0],
+        index=["Falcon", "Falcon", "Parrot", "Parrot"],
+        name="Max Speed",
+    )
+    grouped = df.groupby("A")[["C", "D"]]
+    grouped1 = ser.groupby(ser > 100)
+    c1 = grouped.transform(lambda x: (x - x.mean()) / x.std())
+    c2 = grouped1.transform(lambda x: x.max() - x.min())
+    check(assert_type(c1, "pd.DataFrame"), pd.DataFrame)
```
Dr-Irv, line 2086: don't need quotes here

pandas-stubs/core/groupby/generic.pyi
```diff
@@ -61,7 +61,7 @@ class SeriesGroupBy(GroupBy, Generic[S1]):
     def agg(self, func: list[AggFuncTypeBase], *args, **kwargs) -> DataFrame: ...
     @overload
     def agg(self, func: AggFuncTypeBase, *args, **kwargs) -> Series: ...
-    def transform(self, func, *args, **kwargs): ...
+    def transform(self, func, *args, **kwargs) -> Series: ...
```
Dr-Irv, line 64: can you add `func: Callable` here (and below)?

Dr-Irv (changes requested):

pandas-stubs/core/groupby/generic.pyi
ramvikrams: Done sir

tests/test_frame.py
ramvikrams: Done sir

pandas-stubs/core/groupby/generic.pyi
ramvikrams: after adding Callable line number 643 and 645 showa this error `Argument 1 to "transform" of "DataFrameGroupBy" has incompatible type "str"; expected "Callable[..., Any]"  [arg-type]` but if we add `Callable[..., str]` it also shows the same error

pandas-stubs/core/groupby/generic.pyi
```diff
@@ -61,7 +61,7 @@ class SeriesGroupBy(GroupBy, Generic[S1]):
     def agg(self, func: list[AggFuncTypeBase], *args, **kwargs) -> DataFrame: ...
     @overload
     def agg(self, func: AggFuncTypeBase, *args, **kwargs) -> Series: ...
-    def transform(self, func, *args, **kwargs): ...
+    def transform(self, func: Callable, *args, **kwargs) -> Series: ...
```
Dr-Irv, line 64: Looks like it needs to be `Callable | str` (in both of them)

Can you also create an issue in the pandas repo saying that the docs for `transform` need to indicate that the function parameter in `groupby.transform()` can be a string function name, and that an example should be included?

ramvikrams: yes I'll do that

tests/test_frame.py
```diff
@@ -2063,3 +2063,25 @@ def test_setitem_none() -> None:
...
+        [390.0, 350.0, 30.0, 20.0],
+        index=["Falcon", "Falcon", "Parrot", "Parrot"],
+        name="Max Speed",
+    )
+    grouped = df.groupby("A")[["C", "D"]]
+    grouped1 = ser.groupby(ser > 100)
+    c1 = grouped.transform(lambda x: (x - x.mean()) / x.std())
+    c2 = grouped1.transform(lambda x: x.max() - x.min())
+    check(assert_type(c1, pd.DataFrame), pd.DataFrame)
+    check(assert_type(c2, pd.Series), pd.Series)
```
Dr-Irv, line 2087: can you add tests for `grouped.transform("sum")` and `grouped1.transform("cumsum")`

Dr-Irv (changes requested):

pandas-stubs/core/groupby/generic.pyi
```diff
@@ -166,7 +166,7 @@ class DataFrameGroupBy(GroupBy):
     ) -> DataFrame: ...
     def aggregate(self, arg: AggFuncTypeFrame = ..., *args, **kwargs) -> DataFrame: ...
     agg = aggregate
-    def transform(self, func, *args, **kwargs): ...
+    def transform(self, func: Callable | str, *args, **kwargs) -> DataFrame: ...
```
gandhis1, line 169: Can/should we make this a `Literal`?

From https://github.com/pandas-dev/pandas/blob/main/pandas/core/groupby/base.py#L60, the transformation kernels would be the ones applicable here:

```
transformation_kernels = Literal[
    "bfill",
    "cumcount",
    "cummax",
    "cummin",
    "cumprod",
    "cumsum",
    "diff",
    "ffill",
    "fillna",
    "ngroup",
    "pct_change",
    "rank",
    "shift",
]
```

gandhis1: Is there any way to do better than `Callable`? Like what about `Callable[Concatenate[pd.DataFrame, P], pd.DataFrame]`? Will `ParamSpec` work in that context? If not, can we use a `Protocol`? I know that we can't use `P.args` on `*args` and `P.kwargs` on `**kwargs`, though that would have been a very nice feature.

I'm just thinking, if we are going from an unannotated function to an annotated function, a loose annotation doesn't really do that much good versus not having an annotation at all.

tests/test_frame.py
ramvikrams: Yes sir

pandas-stubs/core/groupby/generic.pyi
ramvikrams: I think it would be better the ask Dr Irv about adding it.

Dr-Irv: I'd rather be incremental in terms of PR's where we could specify the `Literal` values (there are more than just the ones you listed), as well as making the `func` argument more specific.  @gandhis1 feel free to create an issue indicating that we could improve the typing on `groupby.transform` by being more specific about the allowed argument combinations.  It also would require a lot more tests.

Focus of this PR was just to get the return types right, and do a slight improvement in the `func` argument.

ramvikrams:
> thanks @ramvikrams 

Thanks @Dr-Irv sir

PR #457, opened 2022-12-01
Added __getattr__ in DataFrameGroupBy

### LINKED ISSUE #439: DataFrameGroupBy attribute access for columns
**Describe the bug**
When accessing a column on a DataFrameGroupBy object via attribute-access (see below), mypy complains that no such attribute exists.

**To Reproduce**

```
df = pd.DataFrame({'a': [1, 1, 2], 'b': [3, 4, 5]})
gb = df.groupby('a').b
```

mypy 0.991 says: `"_DataFrameGroupByNonScalar" has no attribute "b"`

ramvikrams:
- [x] Closes #439 (Replace xxxx with the Github issue number)
the error is saying expected `Series` but got `SeriesGroupBy`, I can't understand it

Sorry sir forgot to add the correct commit file name.


pandas-stubs/core/groupby/generic.pyi
```diff
@@ -359,3 +359,4 @@ class DataFrameGroupBy(GroupBy):
         ascending: bool = ...,
         dropna: bool = ...,
     ) -> Series[float]: ...
+    def __getattr__(self, name: str) -> Series: ...
```
Dr-Irv, line 362: This should be `SeriesGroupBy`
```python
>>> df = pd.DataFrame(
...         {
...             'C': [1, 5, 5, 2, 5, 5],
...             'D': [2.0, 5.0, 8.0, 1.0, 2.0, 9.0],
...         }
...     )
>>> gb = df.groupby('C').D
>>> gb
<pandas.core.groupby.generic.SeriesGroupBy object at 0x000002E6ED25CA00>
```

Dr-Irv (changes requested):

pandas-stubs/core/groupby/generic.pyi
ramvikrams: Done sir

tests/test_frame.py
```diff
@@ -2097,3 +2098,14 @@ def test_groupby_and_transform() -> None:
...
+
+def test_getattr_and_dataframe_groupby() -> None:
+    df = pd.DataFrame(
+        {
+            "C": [1, 5, 5, 2, 5, 5],
+            "D": [2.0, 5.0, 8.0, 1.0, 2.0, 9.0],
+        }
+    )
+    gb = df.groupby("C").D
+    check(assert_type(gb, SeriesGroupBy), SeriesGroupBy)
```
Dr-Irv, line 2111: I think I would prefer a test that does not import `SeriesGroupBy`  because that is not a documented class.

Borrowing from lines 669, but modified to test the getattr:
```python
    df = pd.DataFrame(
        data={"col1": [1, 1, 2], "col2": [3, 4, 5], "col3": [0, 1, 0], 0: [-1, -1, -1]}
    )
    check(assert_type(df.groupby("col1").col3.agg(min), pd.Series), pd.Series)
    check(
        assert_type(df.groupby("col1").col3.agg([min, max]), pd.DataFrame),
        pd.DataFrame,
    )
```

Dr-Irv (changes requested):

tests/test_frame.py
ramvikrams: i'll do that sir

ramvikrams: Sir I am failing to understand that even if we write tests ` check(assert_type((df.groupby("C").__getattr__("D")), pd.Series), pd.Series, pd.Series)` like this, how will it change because the return type of `__getattr__` will still be `SeriesGroupBy`

Dr-Irv: The idea is that you are not testing `__getattr__()` directly, but indirectly, because you always use a method in `SeriesGroupBy` in real code.  The example I gave is calling `agg()` on the `SeriesGroupBy` result.  Both `df.groupby("col1").col3` and `df.groupby("col1")["col3"]` return a `SeriesGroupBy` object.

See the test `test_types_groupby_agg` for an example where we use the second form, which is where I copied this from.

ramvikrams: sir I read the `test_types_groupby_agg` but here when I give the `str` to `__getattr__()` the error is there `about expected series but got seriesgroupby` pushing a commit so you can also check

Dr-Irv: Please implement the test as I wrote above.

pandas-stubs/core/groupby/generic.pyi
Dr-Irv: This should be `SeriesGroupBy`

tests/test_frame.py
```diff
@@ -2097,3 +2097,14 @@ def test_groupby_and_transform() -> None:
...
+
+def test_getattr_and_dataframe_groupby() -> None:
+    df = pd.DataFrame(
+        {
+            "col1": [1, 5, 5, 2, 5, 5],
+            "col2": [2.0, 5.0, 8.0, 1.0, 2.0, 9.0],
+            "col3": [9,8,7,5,6,1]
+        }
+    )
+    check(assert_type(df.groupby("col1").__getattr__("col3"), pd.Series), pd.Series)
```
Dr-Irv, line 2110: This test should be the same as https://github.com/pandas-dev/pandas-stubs/pull/457#discussion_r1039977959

Dr-Irv (changes requested):
Please follow instructions precisely

tests/test_frame.py
ramvikrams: Done

Dr-Irv (approved):
thanks @ramvikrams

Issue #456, opened 2022-11-30
Create more precise annotations for DataFrame/Series GroupBy operations (agg, apply, transform)


gandhis1:
All of these groupby operations take a function that can be a `str` or a `Callable`. They should ensure that:

- The `Callable` accepts a first parameter that is a `DataFrame` or `Series` as applicable
- The `Callable` returns the appropriate type - in some cases an `Iterable`, in other cases either a `DataFrame` or `Series`
- If a `str` is passed in, it is part of the set of supported functions, which are located here: https://github.com/pandas-dev/pandas/blob/main/pandas/core/groupby/base.py#L60
- If possible - and this may not be possible today - ensure that `args` and `kwargs` parameters are type checked against the passed-in `Callable`
- Comprehensive set of tests are needed here, to exercise all variations

I'll start with `transform` because that is a fairly simple API compared to the others and it is not very commonly used, so the potential impact of more narrow types there has a lower risk of an impactful regression.

Dr-Irv:
For `transform`, see this line, https://github.com/pandas-dev/pandas/blob/6b0ac938279d2baec210641bfaf3719872f9a187/pandas/core/groupby/base.py#L118 so you have to combine the `transform` specific ones and the `reduction` ones.

PR #459, opened 2022-12-05
Make Series.groupby.transform annotation more precise

gandhis1:
- [X] Closes partially #456

### DIFF
--- a/pandas-stubs/core/groupby/generic.pyi
+++ b/pandas-stubs/core/groupby/generic.pyi
@@ -22,15 +22,21 @@ from pandas.core.groupby.groupby import (  # , get_groupby as get_groupby
 )
 from pandas.core.groupby.grouper import Grouper
 from pandas.core.series import Series
-from typing_extensions import TypeAlias
+from typing_extensions import (
+    Concatenate,
+    TypeAlias,
+)
 
 from pandas._typing import (
     S1,
+    S2,
     AggFuncTypeBase,
     AggFuncTypeFrame,
     AxisType,
+    GroupByFuncStrs,
     Level,
     ListLike,
+    P,
     Scalar,
 )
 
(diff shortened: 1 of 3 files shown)


pandas-stubs/core/groupby/generic.pyi
```diff
@@ -61,7 +67,19 @@ class SeriesGroupBy(GroupBy, Generic[S1]):
...
     def agg(self, func: AggFuncTypeBase, *args, **kwargs) -> Series: ...
-    def transform(self, func: Callable | str, *args, **kwargs) -> Series: ...
+    @overload
+    def transform(
+        self,
+        func: Callable[Concatenate[Series[S1], P], Series[S2]],
+        *args: Any,  # Ideally we would use P.args here, but it does not work with engine/engine_kwargs present
+        engine: Literal["cython", "numba", None] = ...,
+        engine_kwargs: dict[str, Any] | None = ...,
+        **kwargs: Any,  # Ideally we would use P.kwargs here, but does not work with engine/engine_kwargs present
```
gandhis1, line 77: This was working great when I did not have `engine` and `engine_kwargs` in there because I could use `P.args` and `P.kwargs` as the annotation. However, I'm having trouble getting that working with these additional keyword arguments in there. Any ideas?

Dr-Irv: Or maybe place `engine` and `engine_kwargs` before `*args` and `**kwargs` ???

gandhis1: > Or maybe place `engine` and `engine_kwargs` before `*args` and `**kwargs` ???

So pyright no longer complains about that, but then the test doesn't work, because it thinks `engine` and `engine_kwargs` are positional parameters.

Still investigating...

gandhis1: Looks like it isn't possible. Here is a similar issue: https://github.com/microsoft/pyright/issues/4189

The relevant part of PEP 612:

> Inside of `bar`, we would get `TypeError: foo() got multiple values for argument 'x'`. Requiring these concatenated arguments to be addressed positionally avoids this kind of problem, and simplifies the syntax for spelling these types. Note that this also why we have to reject signatures of the form `(*args: P.args, s: str, **kwargs: P.kwargs)` (See [Concatenating Keyword Parameters](https://peps.python.org/pep-0612/#concatenating-keyword-parameters) for more details).

This is unfortunate. I'll have to revert back to using `Any`.

gandhis1:
I'm going to close this until I can dedicate the time to get this to the finish line.

PR #467, opened 2022-12-10
remove restriction on python 3.10 and fix pyright issue

Dr-Irv:
We restricted python 3.10 testing to 3.10.6 a while back, so removed that.

Pinned pyright to 1.1.282 because 1.1.283 causes CI failures due to some odd interaction between our installed stubs and installed matplotlib.  See https://github.com/microsoft/pyright/issues/4310


pyproject.toml
```diff
@@ -38,7 +38,7 @@ types-pytz = ">= 2022.1.1"
 mypy = "0.990"
 pyarrow = ">=10.0.1"
 pytest = ">=7.1.2"
-pyright = ">=1.1.281"
+pyright = "1.1.282"
```
twoertwein, line 41: Might be nice to add an inline TODO/comment citing the issue. Would also be fine merging it without it - but would at least wait until you get an initial response on the pyright issue (if it is going to be fixed soon, we could, for example, change the constraint to != 1.1.283).

twoertwein (approved):

PR #470, opened 2022-12-14
ENH: add dict to return type of func in `DataFrame#apply` when `result_type="expand"`

skatsuta:

## Problem

It is perfectly valid to pass a function with a return type of `dict` as the `func` argument to `DataFrame#apply` when `result_type="expand"`, but the current type stub does not allow it.

For example, the following code

```py
def returns_dict(x: pd.Series) -> dict[str, int]:
    return {'x': 1, 'y': 2}

df.apply(returns_dict, axis=1, result_type="expand")
```

results in the following error when type-checked by Pyright:

```shell
error: Argument of type "(x: Series[Unknown]) -> dict[str, int]" cannot be assigned to parameter "f" of type "(...) -> Series[Unknown]" in function "apply"
    Type "(x: Series[Unknown]) -> dict[str, int]" cannot be assigned to type "(...) -> Series[Unknown]"
      Function return type "dict[str, int]" is incompatible with type "Series[Unknown]"
        "dict[str, int]" is incompatible with "Series[Unknown]" (reportGeneralTypeIssues)
error: Argument of type "Literal['expand']" cannot be assigned to parameter "result_type" of type "Literal['reduce']" in function "apply"
    "Literal['expand']" cannot be assigned to type "Literal['reduce']" (reportGeneralTypeIssues)
```

## Solution

Add `dict` to the return type of `func` argument of `DataFrame#apply` when `result_type="expand"`.

Also, add tests for passing a function that returns a dict to `DataFrame#apply` in the cases where `axis` is 0 and 1, respectively.

### DIFF
--- a/pandas-stubs/core/frame.pyi
+++ b/pandas-stubs/core/frame.pyi
@@ -1126,7 +1126,7 @@ class DataFrame(NDFrame, OpsMixin):
     @overload
     def apply(
         self,
-        f: Callable[..., ListLikeExceptSeriesAndStr | Series],
+        f: Callable[..., ListLikeExceptSeriesAndStr | Series | dict],
         axis: AxisType = ...,
         raw: _bool = ...,
         args=...,
--- a/tests/test_frame.py
+++ b/tests/test_frame.py
@@ -477,6 +477,9 @@ def test_types_apply() -> None:
     def returns_listlike_of_3(x: pd.Series) -> tuple[int, int, int]:
         return (7, 8, 9)
 
+    def returns_dict(x: pd.Series) -> dict[str, int]:
+        return {"col4": 7, "col5": 8}
+
     # Misc checks
     check(assert_type(df.apply(np.exp), pd.DataFrame), pd.DataFrame)
     check(assert_type(df.apply(str), "pd.Series[str]"), pd.Series, str)
@@ -516,6 +519,10 @@ def test_types_apply() -> None:
         assert_type(
             df.apply(returns_listlike_of_3, result_type="expand"), pd.DataFrame
         ),
+        pd.DataFrame,
+    )
+    check(
+        assert_type(df.apply(returns_dict, result_type="expand"), pd.DataFrame),
         pd.DataFrame,
     )
 
@@ -597,6 +604,10 @@ def test_types_apply() -> None:
         assert_type(
             df.apply(returns_listlike_of_3, axis=1, result_type="expand"), pd.DataFrame
         ),
+        pd.DataFrame,
+    )
+    check(
+        assert_type(df.apply(returns_dict, axis=1, result_type="expand"), pd.DataFrame),
         pd.DataFrame,
     )
 


gandhis1:
I recently modified these `apply` overloads and the number of permutations of overloads is a lot higher now. So any changes here need to be very careful to make sure they apply new data types everywhere they need to be present.

The changes you made look good to me, I only have one question. During the last round of changes to `apply()` annotations, we decided that even if a particular combination of function return and `return_type` did not make logical sense, if `Pandas` allowed it without errors, then we would continue to permit it in the annotations.

I cannot test right now, but does `dict` work for any other `result_type` value (including the default of `None`)?

Dr-Irv (changes requested):
As suggested by @gandhis1 here: https://github.com/pandas-dev/pandas-stubs/pull/470#issuecomment-1351421981

you should modify all of the overloads to return `dict`, and create tests for all of the various combinations of `axis` and `result_type` as is already present in  the tests.

If some combination of `axis` and `result_type` doesn't work with `dict` as the return type, then you leave `dict` out as a possibility for the return type of the `Callable` in the `f` argument.

skatsuta:
@gandhis1 You are the one who have added those comprehensive overloads! Thanks for the hard work!
Will check other combinations and add support for them if they also work!

@Dr-Irv Thanks for your suggestion! Will add support for func returning dict to other overloads and also additional comprehensive test cases!

tests/test_frame.py
```diff
@@ -518,71 +514,47 @@ def gethead(s: pd.Series, y: int) -> pd.Series:
...
-        assert_type(df.apply(returns_series, result_type="broadcast"), pd.DataFrame),
-        pd.DataFrame,
-    )
-    check(
-        assert_type(
-            # Can only broadcast a list-like of 2 elements, not 3, because there are 2 rows
-            df.apply(returns_listlike_of_2, result_type="broadcast"),
-            pd.DataFrame,
-        ),
-        pd.DataFrame,
```
skatsuta, line 564: There assertions are moved to the bottom so that test cases for `result_type="broadcast"` are grouped together.

pandas-stubs/core/frame.pyi
```diff
@@ -1096,7 +1096,7 @@ class DataFrame(NDFrame, OpsMixin):
         f: Callable[..., ListLikeExceptSeriesAndStr | Series],
         axis: AxisTypeIndex = ...,
         raw: _bool = ...,
-        result_type: Literal[None] = ...,
+        result_type: None = ...,
```
skatsuta, line 1099: This does not necessarily have to be `Literal[None]`, just `None` is OK, and so are others.

skatsuta (commented):
@Dr-Irv I've added `Mapping` return type to all possible overloads, and comprehensive test cases for all `axis` and `result_type` parameter combinations, which are 2 x 4 = 8 assertions in total (2 axes and 4 result types).

I've also replaced `Literal[None]` with just `None`, and reformatted and reordered some existing assertions for readability and clarity, but none of them are removed.

Would appreciate if you could take a look when you have time, thanks!

Dr-Irv (approved):
thanks @skatsuta

PR #473, opened 2022-12-17
TST: run all tests

### LINKED ISSUE #444: CLN: Fix tests not executed by pytest
We have tests that are not beginning with the word `test_` so then `pytest` doesn't execute them.  Need to rename them and then see if they really work.
```
test_api_types.py:def check_extension_dtypes() -> None:
test_frame.py:def df_accepting_dicts_iterator() -> None:
test_frame.py:def series_added_in_astype() -> None:
test_io.py:def btest_read_fwf():
test_timefuncs.py:def fail_on_adding_two_timestamps() -> None:
```

twoertwein:
- [x] Closes #444

and update versions

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -3,22 +3,22 @@ ci:
     autofix_prs: false
 repos:
 -   repo: https://github.com/python/black
-    rev: 22.8.0
+    rev: 22.12.0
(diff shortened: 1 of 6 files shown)


Dr-Irv (approved):
thanks @twoertwein

PR #477, opened 2022-12-19
for gh-468

### LINKED ISSUE #468: nightly failing - possibly due to pandas bug
nightly is failing due to arithmetic between offsets and period index changing result.  

So either we see if that gets fixed, or change our tests.

See https://github.com/pandas-dev/pandas/issues/50162

ramvikrams:
- [ ] Closes #468
wanted some discussion about what could the return type for the func


tests/__init__.py
```diff
@@ -136,3 +136,21 @@ def pytest_warns_bounded(
             return nullcontext()
         else:
             return suppress(upper_exception)
+
+
+def nigthly_test_skip(
```
Dr-Irv, line 141: `nightly` is misspelled

Dr-Irv (changes requested):
I don't think we want to use a decorator on top of a test function, but rather have something that lets us skip running a particular line of code based on the version so that we can keep the offending lines rather isolated.

So, for example, this test in `test_scalars.py` is failing:
```python
check(assert_type(p + offset_index, pd.PeriodIndex), pd.PeriodIndex)
```
It is failing because in the future the result will be a `pd.Index`.  So we will leave the typing alone, but need to change what is passed to `check()` based on the version number.

Other tests are changing because of date/time parsing, in which case the warnings have to be checked based on the version number.

tests/__init__.py
ramvikrams: i'll correct it

ramvikrams: i'll correct it

ramvikrams:
> It is failing because in the future the result will be a `pd.Index`. So we will leave the typing alone, but need to change what is passed to `check()` based on the version number.
What we wan't is to check that test on the version speicifed and check other test normally, have I understood it correctly

Dr-Irv:
> What is to be done is check a test on the version speicifed and check other test normally, have I understood it correctly

Do the test as it is now if the version is 1.5.x, and use the new result if the version is later than that.

Also have to deal with the datetime failures, since that warning has changed.

ramvikrams:
Sir I was testing but i am facing a wired error if I run `poe commands` without entering the virtual envirnment it works, but if I enter the virtual envirnment by `poetry shell` the `poe commands` show `ModuleNotFoundError` error.

Dr-Irv:
closing in favor of #485

PR #479, opened 2022-12-20
flake8-pyi's Y027: use collections.abc in pyi files

twoertwein:
flake8-pyi encourages using `collections.abc` imports in pyi files.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -30,7 +30,7 @@ repos:
         - flake8-pyi==22.11.0
         types: [pyi]
         args: [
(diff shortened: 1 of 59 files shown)


Dr-Irv (approved):
I'm reluctantly approving this.  Given that many things like `Sequence`, `Mapping`, etc. were deprecated from `typing` in python 3.9, it makes sense to do this.  However, I don't fully get why things like `Callable` and `Hashable` are part of `collections.abc`.  So now if we add some typing somewhere, we have to remember which ones are in `collections.abc` versus `typing`.

twoertwein:
It is not ideal since we still have to use the old imports in py files. At least for pyi files, flake8-pyi will emit a warning when we import it from the "wrong" place.

PR #480, opened 2022-12-20
Make DataFrame drop allow sets. Fix failing tests with latest Numpy version.

MortGron:

The following example runs successfully in Pandas 1.5.2 but fails with Pandas-stubs 1.5.2.221124 and Mypy 0.991:
```
import pandas as pd
pd.DataFrame({"a": 1}, index=[1]).drop(columns={"a"})
```
Mypy error
```
error: No overload variant of "drop" of "DataFrame" matches argument type "Set[str]"  [call-overload]
note:     def drop(self, labels: Union[Hashable, Sequence[Hashable], Index] = ..., *, axis: Union[str, int] = ..., index: Union[Hashable, Sequence[Hashable], Index] = ..., columns: Union[Hashable, Sequence[Hashable], Index] = ..., level: Optional[Union[Hashable, int]] = ..., inplace: Literal[True], errors: Literal['ignore', 'raise'] = ...) -> None
note:     def drop(self, labels: Union[Hashable, Sequence[Hashable], Index] = ..., *, axis: Union[str, int] = ..., index: Union[Hashable, Sequence[Hashable], Index] = ..., columns: Union[Hashable, Sequence[Hashable], Index] = ..., level: Optional[Union[Hashable, int]] = ..., inplace: Literal[False] = ..., errors: Literal['ignore', 'raise'] = ...) -> DataFrame
note:     def drop(self, labels: Union[Hashable, Sequence[Hashable], Index] = ..., *, axis: Union[str, int] = ..., index: Union[Hashable, Sequence[Hashable], Index] = ..., columns: Union[Hashable, Sequence[Hashable], Index] = ..., level: Optional[Union[Hashable, int]] = ..., inplace: bool = ..., errors: Literal['ignore', 'raise'] = ...) -> Optional[DataFrame]
```
According to the Pandas documentation the labels, index and columns arguments should accept list-like types, and that includes sets accroding to this: https://pandas.pydata.org/docs/reference/api/pandas.api.types.is_list_like.html

Also some previously deprecated ways of calling `pivot_table` have been removed as of Numpy 1.24.0, so some tests had to be changed to succeed.

### DIFF
--- a/pandas-stubs/_typing.pyi
+++ b/pandas-stubs/_typing.pyi
@@ -161,7 +161,7 @@ ListLike = TypeVar("ListLike", Sequence, np.ndarray, "Series", "Index")
 ListLikeExceptSeriesAndStr = TypeVar(
     "ListLikeExceptSeriesAndStr", MutableSequence, np.ndarray, tuple, "Index"
 )
-ListLikeU: TypeAlias = Union[Sequence, np.ndarray, Series, Index]
+ListLikeU: TypeAlias = Union[Sequence, np.ndarray, Series, Index, set]
 StrLike: TypeAlias = Union[str, np.str_]
 Scalar: TypeAlias = Union[
     str,
--- a/pandas-stubs/core/frame.pyi
+++ b/pandas-stubs/core/frame.pyi
@@ -563,11 +563,11 @@ class DataFrame(NDFrame, OpsMixin):
     @overload
     def drop(
         self,
-        labels: Hashable | Sequence[Hashable] | Index = ...,
+        labels: Hashable | ListLikeU = ...,
         *,
         axis: Axis = ...,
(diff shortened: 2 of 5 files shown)


MortGron:
Numpy v 1.24.0 introduced a change breaking some tests. Therefore either the Numpy version has to be restricted or the test have to be rewritten to accomodate both Numpy versions less than 1.24.0 and greater than or equal to 1.24.0.

twoertwein:
Main already pins numpy to an older version. Before adjusting the tests, let's wait a bit for a new 1.5.x release of pandas to handle the incompatibility.

Using and adjusting `ListLikeU` seems like a good idea, but it might be worth to double check whether other places that currently accept `ListLikeU` also accept a `set` (for example, I believe `pd.DataFrame(columns=..., index=...)` cannot be a set).

PR #484, opened 2022-12-28
fix: test setter on df.columns with pyright, not mypy

Dr-Irv:
pyright 1.1.286 fixed an issue that didn't allow users to write `df.columns = ["abc", "def"]`.

Our testing code was not picking this up because we had `#type: ignore` comments on the testing code, which are needed because of a mypy bug.

So I've instrumented the test to be checked with `pyright`, but not `mypy`.  Unfortunately, `mypy` doesn't support a `mypy: ignore` comment.

- Issue reported to pyright  was here:  https://github.com/microsoft/pyright/issues/4293
- Discussion on how to see which type checker is being used is here:  https://github.com/microsoft/pyright/issues/4292

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -38,7 +38,7 @@ types-pytz = ">= 2022.1.1"
 mypy = "0.991"
 pyarrow = ">=10.0.1"
 pytest = ">=7.1.2"
-pyright = ">=1.1.284"
+pyright = ">=1.1.286"
 poethepoet = "0.16.0"
 loguru = ">=0.6.0"
 pandas = "1.5.2"
@@ -171,6 +171,7 @@ strict_equality = true
 show_error_context = false
 show_column_numbers = false
 show_error_codes = true
+always_true = "IS_TYPE_CHECKER_MYPY"
 
 [tool.pyright]
 typeCheckingMode = "strict"
@@ -192,3 +193,5 @@ reportUnusedVariable = false
 reportPrivateUsage = false
 # enable optional checks
 reportMissingModuleSource = true
+defineConstant = { IS_TYPE_CHECKER_MYPY = false }
+
--- a/tests/__init__.py
+++ b/tests/__init__.py
@@ -22,6 +22,7 @@ from pandas._typing import T
 TYPE_CHECKING_INVALID_USAGE: Final = TYPE_CHECKING
 WINDOWS = os.name == "nt" or "cygwin" in platform.system().lower()
 PD_LTE_15 = Version(pd.__version__) < Version("1.5.999")
+IS_TYPE_CHECKER_MYPY = True
 
 lxml_skip = pytest.mark.skipif(
     sys.version_info >= (3, 11), reason="lxml is not available for 3.11 yet"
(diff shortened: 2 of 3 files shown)


twoertwein (approved):
Does this new way of ignoring mypy-only errors allow enabling mypy's and pyright's check for unneeded ignores?

Dr-Irv:
> Does this new way of ignoring mypy-only errors allow enabling mypy's and pyright's check for unneeded ignores?

For mypy, the answer is no.  See https://github.com/python/mypy/issues/14365

I wanted to add a test like this:
```python
if TYPE_CHECKING_INVALID_USAGE and not IS_TYPE_CHECKER_MYPY:
     df.columns = "abc"  # type: ignore[assignment]
```
but mypy still checks it.  Although I guess I could use a `pyright: ignore` here and do that.  I will try that out.

twoertwein:
Could do something like at typeshed https://github.com/python/typeshed/commit/23ac9bff1913cec5bb507adc442b8b2bd380a3f0 Use `type: ignore` for only mypy and the pyright-specific codes only for pyright. I think this could be cleaner than the constant approach.

Dr-Irv:
> Could do something like at typeshed [python/typeshed@23ac9bf](https://github.com/python/typeshed/commit/23ac9bff1913cec5bb507adc442b8b2bd380a3f0) Use `type: ignore` for only mypy and the pyright-specific codes only for pyright. I think this could be cleaner than the constant approach.

We currently are reporting wrong ignores with `mypy`.  But for `pyright`, if we use the option `reportUnnecessaryTypeIgnoreComment`, we get a lot of reports.  The reason is that in the PYI files, we have `# type: ignore` comments that are needed for `mypy`, but not for `pyright`.  So the `typeshed` solution won't help here - we can't put something that says "ignore for mypy, but pyright should type check"

If we have a line of valid code, then I think this table summarizes things


| mypy reports | pyright reports | mypy fix | pyright fix |
| -------- | --------- | --------- | ---------------------------------- |
| Valid | Valid | Not needed | Not needed |
| Valid | Invalid | Not needed | Use #pyright: ignore |
| Invalid | Valid | #type: ignore | Not needed, but won't work with `reportUnnecessaryTypeIgnoreComment` |
| Invalid | Invalid | #type: ignore | #type: ignore |


It's the third case that is causing issues for us to be able to use `reportUnnecessaryTypeIgnoreComment` with pyright

gandhis1:
So there isn't a Mypy-only equivalent of `#pyright: ignore` (or a way to have Pyright disregard `type: ignore`)? I feel like the optimal solution would be to require separate ignores for Mypy and Pyright individually, but from the table it doesn't seem like that's even possible.

Dr-Irv:
> So there isn't a Mypy-only equivalent of `#pyright: ignore` (or a way to have Pyright disregard `type: ignore`)? I feel like the optimal solution would be to require separate ignores for Mypy and Pyright individually, but from the table it doesn't seem like that's even possible.

Unforunately, there is not a `#mypy: ignore` and there is a feature request here for that: https://github.com/python/mypy/issues/12358

PR #487, opened 2022-12-29
Enable reportUnnecessaryTypeIgnoreComment

twoertwein:
A potential alternative to #484.

If this PR gets merged:

- pyright ignores any `# type: ignore` comments
- if ignores are needed for pyright, use `# pyright: ignore[...]`
- if there are not needed `# pyright: ignore[...]`, pyright will let you know :)

Similar to https://github.com/python/typeshed/commit/23ac9bff1913cec5bb507adc442b8b2bd380a3f0 mypy ignores need to come before pyright ignores.


tests/test_frame.py
```diff
@@ -1780,16 +1779,15 @@ def test_set_columns() -> None:
...
-        df.columns = (1, "a")
+    df.columns = ["c", "d"]  # type: ignore[assignment]
+    df.columns = [1, 2]  # type: ignore[assignment]
+    df.columns = [1, "a"]  # type: ignore[assignment]
+    df.columns = np.array([1, 2])  # type: ignore[assignment]
+    df.columns = pd.Series([1, 2])  # type: ignore[assignment]
+    df.columns = np.array([1, "a"])  # type: ignore[assignment]
+    df.columns = pd.Series([1, "a"])  # type: ignore[assignment]
+    df.columns = (1, 2)  # type: ignore[assignment]
+    df.columns = (1, "a")  # type: ignore[assignment]
```
Dr-Irv, line 1790: Can you add this kind of test here?
```python
            df.columns = "abc"  # pyright: ignore[reportGeneralTypeIssues]
```
Want to make sure this works

Dr-Irv (changes requested):
I like this idea of telling pyright to don't parse the `# type: ignore` comments.  Then we know those are only there because of mypy

tests/test_frame.py
twoertwein: You added that in the other PR, or?

Dr-Irv: Yes.  So with my changes there, you'll have to update this PR to use the correct `#pyright ignore` or `#type: ignore` as needed.

twoertwein: Should be ready now

Dr-Irv (approved):
thanks @twoertwein 

Can you do a PR to update https://github.com/pandas-dev/pandas-stubs/blob/main/docs/philosophy.md to reflect how we are handling the ignore stuff, with respect to when we want to use the various ignore combinations, which I think are:
1. Use only `#type: ignore` when mypy incorrectly says there is an error, but pyright correctly says there is no error
2. Use only `#pyright: ignore` when mypy incorrectly says the code is OK, but pyright correctly finds an error
3.  Use both `#type: ignore`  and `#pyright: ignore` only in the test code when doing `TYPE_CHECKING_INVALID_USAGE`
4.  Use both `#type: ignore`  and `#pyright: ignore`  in the PYI files when there are overlapping overloads that both type checkers detect

PR #488, opened 2022-12-29
fix up ignores for invalid usage

### LINKED ISSUE #217: Review expected failures
In #213 @twoertwein marked typing tests that we expect to fail using `TYPE_CHECKING_INVALID_USAGE` .  The idea here is that we are testing code that we want the type checkers to pick up as being incorrect.  In many cases, `pyright` does figure that out.  `mypy` figures it out sometimes, and there are occasions where both type checkers see what we expect to fail as "correct" code.

At some point in time in the past, when the stubs were in the Microsoft repository, those tests were doing the "right" thing, i.e., both type checkers were marking those tests as "failures", and the `# type: ignore` comments were correct, in the sense that the type checkers saw that they were correct ignores.

Now, possibly due to changes in the settings used for `mypy` and `pyright`, as well as evolutions in the type checkers themselves, those tests are no longer failing as we expect them to.   Some investigation is required to determine whether this is due to something changing in the stubs, or to particular settings in the type checkers.

### LINKED ISSUE #461: Dividing a series of Timedeltas by a Timedelta is rejected
**Describe the bug**
Should allow `TimedeltaSeries / pd.Timedelta`

**To Reproduce**
```python
import pandas as pd

tds = pd.Series(pd.timedelta_range(start="1 day", periods=10))
td = pd.Timedelta("2 hours")

q = tds / td
```
pyright reports:
```text
tddiv.py
  c:\Code\pandas-stubs\tddiv.py:6:5 - error: Operator "/" not supported for types "TimedeltaSeries" and "Timedelta"
    Operator "/" not supported for types "TimedeltaSeries" and "Timedelta" (reportGeneralTypeIssues)
```

 - OS:Windows 10
 - python version:  3.9
 - version of type checker:  pyright 1.1.283
 - version of installed `pandas-stubs`:  1.5.2.221124

**Additional context**

Just need to add `Timedelta` here:
https://github.com/pandas-dev/pandas-stubs/blob/58416f2c8d3dd44f60e0b2a89da1328ae3fbb58b/pandas-stubs/core/series.pyi#L1789

Dr-Irv:
Goal is to make all the `#type: ignore` statements in the test files correspond to invalid usage and see that we are picking up the invalid usages that we had working in the tests in the past.  Picked up another issue in the meantime.   Closes out a bunch of TODO in the test files.

- [x] Closes #461
- [x] Closes #217


twoertwein (approved):

twoertwein:
Thanks @Dr-Irv !

PR #492, opened 2023-01-02
Fix issue for pandera allowing generic Series to work

### LINKED ISSUE #471: Incorrect "Type application has too few types" when constructing generic Series type
```
from pandera.typing import Series

x = Series[int]([1, 2, 3])
```
or
```
from pandera.typing import Series
import pandas as pd

df = pd.DataFrame([[1]], columns=['a'])
df['a'].pipe(Series[int])
```
cause mypy to produce errors like `error: Type application has too few types (2 expected)  [misc]`

 - OS: Windows + Linux
 - python version: 3.9
 - mypy==0.991
 - pandas-stubs==1.5.2.221213  (issue not present in 1.5.2.221124 and earlier)

**Additional context**

May be similar/same issue as #197 but for Series rather than DataFrames?

### LINKED ISSUE #497: mypy complaining on sorted, min and max builtins on Version 1.5.2.221213
**Describe the bug**

Starting with this commit https://github.com/pandas-dev/pandas-stubs/commit/bc615436797f4fd49b590428050fb9bd153a56df, I started to see complaints in my codebase about the `SupportsRichComparison` type.

One example is with the `sorted` method, but it also happens with `min` and `max`:

**To Reproduce**

```python
df = pd.DataFrame({'x': [1], 'y': [2], 'z': [3]})
sorted(df.columns)
reveal_types
```

```
>>> mypy test.py

error: Value of type variable "SupportsRichComparisonT" of "sorted" cannot be "Union[str, bytes, date, timedelta, int, float, complex, Tuple[Hashable, ...]]"  [type-var]
note: Revealed local types are:
note:     df: pandas.core.frame.DataFrame
```

Before that commit, the type from `df.columns` was `iterable[any]`, but after the commit, it now has this `complex` type, which doesn't have support for `__lt__` and `__gt__`

How do you think we should proceed on this case? 
`cast`/`ignore` on the code base is the indicated solution, or is there something we can "fix" on the stubs?

**Additional context**

Just to disclose, this is my "full use case" that started my investigation:
```
def acme(df: pd.DataFrame, metrics: set[str]) -> pd.DataFrame:
    df_metrics = set(df.columns) - set(Dimensions)
    if unknown_metrics := df_metrics - metrics:
        raise MyCustomException(f'My cool msg: {", ".join(sorted(metrics))}')

    return df
```

OS: MacOS M1
OS Version Ventura 13.0
python version: 3.9.12 
version of type checker: mypy 0.991 (compiled: yes)
version of installed pandas-stubs
pandas==1.5.2
pandas-stubs==1.5.2.221213

Dr-Irv:
- [x] Closes #471
- [x] Tests added: 
  - `test_series.py:test_pandera_generic()`

Fix here is to expand the `Series[IntervalT].__new__()` into each individual part.  Seems like a double `TypeVar` causes issues for `mypy`.  Had to do a similar fix for `Interval.__contains__()`

Either @twoertwein or @bashtage can approve or merge (both not required)

### DIFF
--- a/pandas-stubs/core/series.pyi
+++ b/pandas-stubs/core/series.pyi
@@ -70,6 +70,7 @@ from typing_extensions import (
 )
 import xarray as xr
 
+from pandas._libs.interval import Interval
 from pandas._libs.missing import NAType
 from pandas._libs.tslibs import BaseOffset
 from pandas._typing import (
@@ -94,7 +95,6 @@ from pandas._typing import (
     IgnoreRaise,
     IndexingInt,
     IntervalClosedType,
-    IntervalT,
     JoinHow,
     JsonSeriesOrient,
     Level,
@@ -216,13 +216,43 @@ class Series(IndexOpsMixin, NDFrame, Generic[S1]):
     @overload
     def __new__(
         cls,
-        data: IntervalIndex[IntervalT],
+        data: IntervalIndex[Interval[int]],
         index: Axes | None = ...,
         dtype=...,
         name: Hashable | None = ...,
         copy: bool = ...,
         fastpath: bool = ...,
-    ) -> Series[IntervalT]: ...
+    ) -> Series[Interval[int]]: ...
+    @overload
+    def __new__(
+        cls,
+        data: IntervalIndex[Interval[float]],
+        index: Axes | None = ...,
+        dtype=...,
+        name: Hashable | None = ...,
+        copy: bool = ...,
+        fastpath: bool = ...,
+    ) -> Series[Interval[float]]: ...
+    @overload
+    def __new__(
+        cls,
+        data: IntervalIndex[Interval[Timestamp]],
+        index: Axes | None = ...,
+        dtype=...,
+        name: Hashable | None = ...,
+        copy: bool = ...,
+        fastpath: bool = ...,
+    ) -> Series[Interval[Timestamp]]: ...
+    @overload
+    def __new__(
+        cls,
+        data: IntervalIndex[Interval[Timedelta]],
+        index: Axes | None = ...,
+        dtype=...,
+        name: Hashable | None = ...,
+        copy: bool = ...,
+        fastpath: bool = ...,
+    ) -> Series[Interval[Timedelta]]: ...
     @overload
     def __new__(
         cls,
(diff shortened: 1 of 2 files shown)


pandas-stubs/core/series.pyi
```diff
@@ -216,13 +216,43 @@ class Series(IndexOpsMixin, NDFrame, Generic[S1]):
...
-        data: IntervalIndex[IntervalT],
+        data: IntervalIndex[Interval[int]],
         index: Axes | None = ...,
         dtype=...,
         name: Hashable | None = ...,
         copy: bool = ...,
         fastpath: bool = ...,
-    ) -> Series[IntervalT]: ...
+    ) -> Series[Interval[int]]: ...
+    @overload
```
twoertwein, line 226: might be nice to add a TODO referencing the mypy issue

twoertwein (approved):

pandas-stubs/core/series.pyi
Dr-Irv: > might be nice to add a TODO referencing the mypy issue

I didn't research whether there exists a mypy issue, or how to create a small sample to replicate.  

I could add a comment indicating that we prefer the `Series[IntervalT]` notation, but that it doesn't seem to work and reference the original pandas-stubs issue.  Let me know if I should do that.

twoertwein: Would something like the following work?

```py
T = TypeVar("T", float, int, Timestamp, Timedelta)

        data: IntervalIndex[Interval[T]],
        ....
    ) -> Series[Interval[T]]: ...
```

Dr-Irv: > 

I can't see how this would work, because we have:
```python
class IntervalIndex(IntervalMixin, ExtensionIndex, Generic[IntervalT]):
```

Since `IntervalIndex` is generic, and `Interval` is generic, you can't make `IntervalIndex` depend on the "inner" generic type.

twoertwein: Okay, sorry for the trouble.

twoertwein:
Thanks @Dr-Irv !

PR #249, opened 2022-09-01
MAINT: Pin poe to let windows CI run correctly

bashtage:
- [X] Closes #240

PR #495, opened 2023-01-03
Update Poethepoet version

Dr-Irv:
In #249 we pinned poethepoet because Windows CI was failing as reported in #240

Based on https://github.com/nat-n/poethepoet/issues/88#issuecomment-1328330881, poethepoet has been fixed, so put the min version at 0.16.5 which fixed the issue.


pandas-stubs/core/series.pyi
```diff
@@ -217,13 +217,43 @@ class Series(IndexOpsMixin, NDFrame, Generic[S1]):
     @overload
     def __new__(
         cls,
-        data: IntervalIndex[IntervalT],
+        data: IntervalIndex[Interval[int]],
```
twoertwein, line 220: I assume these changes are from #492

Dr-Irv: Ugh. Same mistake again.  My local copy of `main` has the `Interval` change for some reason, and I keep branching off of that and making the same mistake.

This is now fixed for this PR.

twoertwein (approved):

twoertwein:
Thanks @Dr-Irv !

PR #501, opened 2023-01-07
Allow sorted() to work on Index

### LINKED ISSUE #497: mypy complaining on sorted, min and max builtins on Version 1.5.2.221213

Dr-Irv:
- [x] Closes #497
- [x] Tests added:
  - `test_indexes.py:test_sorted_and_list()`


Dr-Irv:
Either @twoertwein or @bashtage can review/approve/merge.  Approvals from both not needed.  Thanks!

tests/test_indexes.py
```diff
@@ -92,7 +96,11 @@ def test_column_contains() -> None:
 def test_column_sequence() -> None:
     df = pd.DataFrame([1, 2, 3])
     col_list = list(df.columns)
-    check(assert_type(col_list, List[Union[Scalar, Tuple[Hashable, ...]]]), list, int)
+    check(
+        assert_type(col_list, List[Union["IndexIterScalar", Tuple[Hashable, ...]]]),
```
twoertwein, line 100: Other tests (test_series.py) import `from __future__ import annotations` to use `tuple`/`list`.

Dr-Irv: It won't work at runtime on python 3.8.  CI failed for me when I did that.   You can do `foo: list[int]` and that will be accepted, but if you are doing ` assert_type(col_list, list[Union["IndexIterScalar", tuple[Hashable, ...]]])` that will fail at runtime on 3.8 because the runtime for 3.8 can't parse something like `list[int]` when it is not a typing annotation.

twoertwein (approved):

twoertwein:
Thanks @Dr-Irv !

PR #504, opened 2023-01-09
add operators for Index

### LINKED ISSUE #405: Various `Index` operators missing
**Describe the bug**
I discovered that `Index.__rpow__` is missing when trying to lint my code.

**To Reproduce**

```py
import pandas

df = pandas.DataFrame({"nums": [1, 2]})
print(2**df.index)
```

Runs:

```
$ python example.py
0    2
1    4
Name: nums, dtype: int64
```

But Mypy not happy:

```
$ pre-commit run mypy --file example.py
- hook id: mypy
- exit code: 1

example.py:4: error: No overload variant of "__pow__" of "int" matches argument type "Index"  [operator]
example.py:4: note: Possible overload variants:
example.py:4: note:     def __pow__(self, Literal[0]) -> Literal[1]
example.py:4: note:     def __pow__(self, Literal[0], None) -> Literal[1]
example.py:4: note:     def __pow__(self, Literal[1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15, 16, 17, 18, 19, 20, 21, 22, 23, 24, 25], None = ...) -> int
example.py:4: note:     def __pow__(self, Literal[-1, -2, -3, -4, -5, -6, -7, -8, -9, -10, -11, -12, -13, -14, -15, -16, -17, -18, -19, -20], None = ...) -> float
example.py:4: note:     def __pow__(self, int, None = ...) -> Any
example.py:4: note:     def __pow__(self, int, Literal[0]) -> NoReturn
example.py:4: note:     def __pow__(self, int, int) -> int
```

It seems that various dunder funcs on `Index` are missing.

 - OS: macOS
 - OS Version 12
 - python version: 3.10
 - version of type checker: 0.982
 - version of installed `pandas-stubs`: 1.5.1.221024 

**Additional context**
n/a

Dr-Irv:
- [x] Closes #405(
- [x] Tests added:
  - `test_indexes.py:test_index_operators()`

Some notes:

1. Ideally we would use `OpsMixin`, but we'd need to use return types of `Self` and `mypy` doesn't support that yet.  I put a comment in `arraylike.pyi` about that.
2. The idea here is to support the arithmetic operations on `Index`, where we don't know the dtype of the `Index`.  Could eventually improve this if we ever make `Index` a Generic type.
3. The boolean operators were there before, but they are now deprecated, so I removed them.

My philosophy is that this is "good enough for now", and if we get reports of issues on specific Index subclasses, we can then fix them going forward, rather than doing a complete set of tests now for all the Index subclasses.


Dr-Irv:
For this PR, would like BOTH @bashtage and @twoertwein to approve, and whomever does the second approval, can do the merge.  Thanks.

twoertwein:
> 1. Ideally we would use `OpsMixin`, but we'd need to use return types of `Self` and `mypy` doesn't support that yet.  I put a comment in `arraylike.pyi` about that.

I think `Self` is just a short form to avoid `TypeVar`s. Could define `_OpsMixinT = TypeVar("_OpsMixinT", OpsMixin)` and use it where you want to use `Self`? (haven't looked at the actual code changes.)

pandas-stubs/core/arraylike.pyi
```diff
@@ -7,6 +7,9 @@ from pandas import DataFrame
 
 class OpsMixinProtocol(Protocol): ...
 
+# We would like to use ->Self here so this could be used by Index, but mypy 0.991
+# doesn't support Self.  So we will have to add these operators directly to Index
+
 class OpsMixin:
     def __eq__(self: OpsMixinProtocol, other: object) -> DataFrame: ...  # type: ignore[override]
```
twoertwein, line 14: Can `OpsMixinProtocol` be removed? It's an empty protocol.

twoertwein: Could then replace it with:

```py
_OpsMixinT = TypeVar("_OpsMixinT", OpsMixin)
class OpsMixin:
    def __eq__(self: _OpsMixinT, other: object) -> _OpsMixinT: ...  # type: ignore[override]
```

Dr-Irv: OK, I have done this in https://github.com/pandas-dev/pandas-stubs/pull/504/commits/0a66721220084d8227fad798dd12e4ec61d553c8

Had to fix up some other things as a result, but the net result uses `OpsMixin` for `Index`, `DataFrame` and `Series`, which is good, with appropriately sprinkled mypy `# type: ignore[override]`

Dr-Irv:
Based on discussion here: https://github.com/python/mypy/issues/14424  I changed the way we detect invalid `Index` with logical operations in https://github.com/pandas-dev/pandas-stubs/pull/504/commits/3b3f2806f6ef747edcd3b0f001e4228f5e2bef09

tests/test_indexes.py
```diff
@@ -684,3 +688,72 @@ def test_sorted_and_list() -> None:
...
+    check(assert_type(i1 // i2, pd.Index), pd.Index)
+    check(assert_type(i1 // 10, pd.Index), pd.Index)
+    check(assert_type(10 // i1, pd.Index), pd.Index)
+    check(assert_type(i1**i2, pd.Index), pd.Index)
+    check(assert_type(i1**2, pd.Index), pd.Index)
+    check(assert_type(2**i1, pd.Index), pd.Index)
+    check(assert_type(i1 % i2, pd.Index), pd.Index)
+    check(assert_type(i1 % 10, pd.Index), pd.Index)
+    check(assert_type(10 % i1, pd.Index), pd.Index)
+    check(assert_type(divmod(i1, i2), Tuple[pd.Index, pd.Index]), tuple)
```
twoertwein, line 719: Not needed but could also check the tuple elements: `check(..., tuple, pd.Index)`

twoertwein (approved):

pandas-stubs/core/arraylike.pyi
twoertwein: Great, glad it worked!

Dr-Irv:
@bashtage awaiting your approval/merge

Dr-Irv:
@bashtage can you review?

Dr-Irv:
@twoertwein haven't heard from @bashtage in a few weeks so can you merge it?  Thanks.

tests/test_indexes.py
```diff
@@ -684,3 +688,72 @@ def test_sorted_and_list() -> None:
...
+        )
+        assert_type(  # type: ignore[assert-type]
+            10 & i1,  # type:ignore[operator] # pyright: ignore[reportGeneralTypeIssues]
+            Never,
+        )
+        assert_type(
+            i1 | i2,  # type:ignore[operator] # pyright: ignore[reportGeneralTypeIssues]
+            Never,
+        )
+        assert_type(  # type: ignore[assert-type]
```
twoertwein, line 740: Some of the checks have an ignore for `assert_type` and some don't. Why do some need an ignore?

Dr-Irv: It has to do with how `mypy` processes `Never` with certain operators and arguments.  If you look carefully, you'll see that the `# ignore[assert-type]` is occurring on any operator that has a numerical operand.  Without the `ignore`, `mypy` says the `assert_type()` is wrong, thinking the result is `Any`.  Probably some kind of `mypy` error, although the interpretation of the `Never` result is ambiguous.

twoertwein:
Thanks @Dr-Irv

PR #506, opened 2023-01-11
Remove use of Never

### LINKED ISSUE #441: FEATURE:  Start using `Never` as return types for disallowed operations
See #432 where methods were added to `TimestampSeries` and `TimedeltaSeries` that return `Never`, which `pyright` will pick up as invalid.

We could add more of these to disallow certain operations between incompatible types, and then use the `TYPE_CHECKING_INVALID_USAGE` pattern to check for these.

Dr-Irv:
- [x] Closes #441
- [ ] Tests added: None

Turns out that I misunderstood the use of `Never`, and we don't need to use it.  See https://github.com/pandas-dev/pandas-stubs/issues/441#issuecomment-1379450907 .

### DIFF
--- a/pandas-stubs/core/indexes/interval.pyi
+++ b/pandas-stubs/core/indexes/interval.pyi
@@ -18,10 +18,7 @@ from pandas.core.series import (
     TimedeltaSeries,
     TimestampSeries,
 )
-from typing_extensions import (
-    Never,
-    TypeAlias,
-)
+from typing_extensions import TypeAlias
 
 from pandas._libs.interval import (
     Interval as Interval,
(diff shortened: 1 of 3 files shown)


twoertwein (approved):
Nice clean up :) Happy to merge unless you want a second review.

Dr-Irv:
> Nice clean up :) Happy to merge unless you want a second review.

Let's give @bashtage a day to review, since I modified code from a previous PR of his.  If he doesn't respond, then go ahead and merge.

twoertwein:
Thanks @Dr-Irv !

PR #510, opened 2023-01-12
allow np.uint64 to be used in indexing.  Support numpy 1.24.1

### LINKED ISSUE #508: Add npt.NDArray[np.uint64] to IndexType?
Currently, `IndexType` contains `np_ndarray_int64` (a.k.a. `npt.NDArray[np.int64]`). Is that possible to add `npt.NDArray[np.uint64]` to `IndexType` as well?

**To Reproduce**

```
import numpy as np
import numpy.typing as npt
import pandas as pd

df = pd.DataFrame(dict(x=[1, 2, 3]), index=np.array([10, 20, 30], dtype="uint64"))
assert df.index.dtype == np.uint64

def get_NDArray(df: pd.DataFrame, key: npt.NDArray[np.uint64]):
    df2 = df.loc[key]
    reveal_type(df2)
    return df2
```

Both.

In `pyright`, the inferred type of `df2` is `Series[Unkown]`, which is incorrect (should be `DataFrame` instead). 

In `mypy`, it says

```
error: Invalid index type "ndarray[Any, dtype[unsignedinteger[_64Bit]]]" 
for "_LocIndexerFrame"; expected type 
"Union[slice, ndarray[Any, dtype[signedinteger[_64Bit]]], Index, List[int], 
Series[int], Series[bool], ndarray[Any, dtype[bool_]], List[bool], 
List[<nothing>], Tuple[Union[slice, ndarray[Any, dtype[signedinteger[_64Bit]]], 
Index, List[int], Series[int], Series[bool], ndarray[Any, dtype[bool_]], 
List[bool], List[<nothing>], Hashable], Union[List[<nothing>], slice, 
Series[bool], Callable[..., Any]]]]"
```

 - OS: MacOS
 - OS Version: 13.0.1 (22A400)
 - python version: 3.10.8
 - version of type checker: mypy: 0.991, Pylance language server 2023.1.10 (pyright bbf0ae78)
 - version of installed `pandas-stubs`: 1.5.2.230105 (build pyhd8ed1ab_2)

Dr-Irv:
- [x] Closes #508 
- [x] Tests added
  - `test_frame.py:test_npint_loc_indexer()`

Turns out that `np.uint64` is not a subclass of `np.int64`, so changed the indexing type to be `np.integer`, which is the parent class of `np.int64` and `np.uint64`.  Error only occurred with numpy 1.24.1, so modified things to allow us to test with that version.


tests/test_frame.py
```diff
@@ -2363,3 +2364,16 @@ def test_frame_dropna_subset() -> None:
...
     )
+
+
+def test_npint_loc_indexer() -> None:
+    # GH 508
+
+    df = pd.DataFrame(dict(x=[1, 2, 3]), index=np.array([10, 20, 30], dtype="uint64"))
+
+    def get_NDArray(df: pd.DataFrame, key: npt.NDArray[np.uint64]) -> pd.DataFrame:
+        df2 = df.loc[key]
```
twoertwein, line 2375: Wouldn't any `np.NDArray` work (not just integer) as long as the DataFrame index is of the same type?

twoertwein: could probably only enforce a tight dtype match if index was generic

tests/test_pandas.py
```diff
@@ -1705,7 +1706,7 @@ def test_pivot_table() -> None:
         ),
         pd.DataFrame,
     )
-    with pytest.warns(np.VisibleDeprecationWarning):
+    if Version(np.__version__) <= Version("1.23.5"):
```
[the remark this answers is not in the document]
Dr-Irv: We can remove this test once the bug is fixed in pandas.  There is a PR for that now at https://github.com/pandas-dev/pandas/pull/50682

tests/test_frame.py
Dr-Irv: > Wouldn't any `np.NDArray` work (not just integer) as long as the DataFrame index is of the same type?

Probably, but since we can't track the `dtype` of an `Index` in a `DataFrame`, I'm limiting this for now to the issue that was reported.  I think most people use arrays of `int` or arrays of `str` (which we probably could add), but I'd rather be incremental in adding support for all the possible types.

twoertwein (approved):

twoertwein:
Thanks @Dr-Irv !

PR #512, opened 2023-01-17
Optionally run tests with mypy nightly

twoertwein:
Add an option to easily test the stubs with mypy nightly. Might be important to do that whenever mypy is close to a new release (especially if mypy advertises pandas-stubs, see https://github.com/python/mypy/issues/14328).

Mypy nightly reports two unused ignores, so I didn't enable it on the CI.

`poe --help`:
image

### DIFF
--- a/scripts/test/run.py
+++ b/scripts/test/run.py
@@ -120,8 +120,24 @@ def nightly_mypy():
     ]
     subprocess.run(cmd, check=True)
 
+    # ignore unused ignore errors
+    config_file = Path("pyproject.toml")
+    config_file.write_text(
+        config_file.read_text().replace(
+            "warn_unused_ignores = true", "warn_unused_ignores = false"
+        )
+    )
+
 
 def released_mypy():
     version = _get_version_from_pyproject("mypy")
     cmd = [sys.executable, "-m", "pip", "install", f"mypy=={version}"]
     subprocess.run(cmd, check=True)
+
+    # check for unused ignores again
+    config_file = Path("pyproject.toml")
+    config_file.write_text(
+        config_file.read_text().replace(
+            "warn_unused_ignores = false", "warn_unused_ignores = true"
+        )
+    )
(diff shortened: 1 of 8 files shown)


twoertwein:
Enabled mypy nightly on the CI by disabling unused ignore errors.

Dr-Irv (changes requested):
I'm OK with this, but I think we should add something to the documentation section about how we are handling the failures in the nightly and mypy-nightly CI runs.  

My feeling is that if we have failures on either, we would still accept a PR, i.e., we are not expecting someone creating a PR for pandas-stubs that makes changes to the stubs to worry about the CI failing with those 2 jobs.

So can you add something to https://github.com/pandas-dev/pandas-stubs/blob/main/docs/tests.md about this new option, and also a paragraph related to what I wrote above?

Dr-Irv (approved):

scripts/test/run.py
```diff
@@ -95,13 +95,49 @@ def nightly_pandas():
...
+        "pip",
+        "install",
+        "--upgrade",
+        "git+https://github.com/python/mypy.git",
+    ]
+    subprocess.run(cmd, check=True)
+
+    # ignore unused ignore errors
+    config_file = Path("pyproject.toml")
+    config_file.write_text(
```
twoertwein, line 125: I don't really like changing pyproject.toml. I'll see whether there is a better way.

I'm also open to not run it on the CI - if there are errors that cannot be resolved by `warn_unused_ignores`, it would be quite annoying to get a failed email for each PR.

Dr-Irv:
Thanks @twoertwein

scripts/test/run.py
Dr-Irv: > I don't really like changing pyproject.toml. I'll see whether there is a better way.

We could put the options used for the nightly tests in a `mypy_nightly.ini` file and then do nightly with the `mypy` option `--config-file mypy_nightly.ini` .  Then this allows us to turn off/on different options for the nightly.

> 
> I'm also open to not run it on the CI - if there are errors that cannot be resolved by `warn_unused_ignores`, it would be quite annoying to get a failed email for each PR.

If we use a special config file, then we just have to modify that config file

PR #515, opened 2023-01-20
Change tests to support pandas 1.5.3

Dr-Irv:
We had some tests that had uncovered bugs in pandas 1.5.2 that are now fixed in 1.5.3.

Note - there is a nightly test that is now failing, but that's because it is using an old nightly build, so I added a comment about that.


tests/test_pandas.py
```diff
@@ -1706,7 +1705,8 @@ def test_pivot_table() -> None:
         ),
         pd.DataFrame,
     )
-    if Version(np.__version__) <= Version("1.23.5"):
+    if PD_LTE_15:
+        # Nightly builds failing since 1/12/2023, but this is fixed since then
```
twoertwein, line 1709: Might need to double-check why we install a more than 7-day old pandas (not needed for now).

If we expect that the CI will pull in a newer version very soon, I wouldn't mind simply letting it fail.

twoertwein (approved):

tests/test_pandas.py
Dr-Irv: > Might need to double-check why we install a more than 7-day old pandas (not needed for now).
> 
> If we expect that the CI will pull in a newer version very soon, I wouldn't mind simply letting it fail.

I put a message on slack and @lithomas1 said he'd look into it this weekend.  I'd rather not let it fail to avoid getting the failure emails!

Dr-Irv: https://github.com/pandas-dev/pandas/issues/50898

twoertwein:
Thanks @Dr-Irv

PR #518, opened 2023-01-23
gh-483: Changed the return type to `dict` in `to_dict`

### LINKED ISSUE #483: Problem with `.to_dict()` type inference
**Describe the bug**

I have a dataframe or pandas series and when I use the `.to_dict()` the inferred type by pandas-stubs is `Dict[Hashable, Any]`. In my case I know that that the index is an int but there is an error when I type the output as `Dict[int, Any]` saying that int is not a Hashable type.

Is there any solution for better inferring the type in this case?

**To Reproduce**

```
import pandas as pd

id=[1,2,3]
value=["a", "b", "c"]

df = pd.DataFrame(zip(id, value), columns=["id", "value"])
df.set_index("id")["value"].to_dict()

def funct(df_dict: Dict[int, Any]) -> None:
     pass
```

2. Indicate which type checker you are using (`mypy` or  `pyright`).3. 
Mypy

3. Show the error message received from that type checker while checking your example.4. 

`error: Argument 4 to "funct" has incompatible type "Dict[Hashable, Any]"; expected "Dict[int, Any]" `

**Additional context**
I am running it in `python version 3.10`

ramvikrams:
- [x] Closes #483 


tests/test_frame.py
```diff
@@ -2390,3 +2390,11 @@ def get_NDArray(df: pd.DataFrame, key: npt.NDArray[np.uint64]) -> pd.DataFrame:
...
     a: npt.NDArray[np.uint64] = np.array([10, 30], dtype="uint64")
     check(assert_type(get_NDArray(df, a), pd.DataFrame), pd.DataFrame)
+
+
+def test_change_to_dict_return_type() -> None:
+    id = [1, 2, 3]
+    value = ["a", "b", "c"]
+    df = pd.DataFrame(zip(id, value), columns=["id", "value"])
+    fd = df.set_index("id")["value"].to_dict()
+    check(assert_type(fd, dict[Hashable, Any]), dict)
```
Dr-Irv, line 2400: Should be `Dict[Any, Any]` instead of `dict[Hashable, Any]`

Dr-Irv (changes requested):

tests/test_frame.py
ramvikrams: Done Sir

tests/test_frame.py
```diff
@@ -2390,3 +2390,11 @@ def get_NDArray(df: pd.DataFrame, key: npt.NDArray[np.uint64]) -> pd.DataFrame:
...
     a: npt.NDArray[np.uint64] = np.array([10, 30], dtype="uint64")
     check(assert_type(get_NDArray(df, a), pd.DataFrame), pd.DataFrame)
+
+
+def test_change_to_dict_return_type() -> None:
+    id = [1, 2, 3]
+    value = ["a", "b", "c"]
+    df = pd.DataFrame(zip(id, value), columns=["id", "value"])
+    fd = df.set_index("id")["value"].to_dict()
+    check(assert_type(fd, dict[Any, Any]), dict)
```
Dr-Irv, line 2400: Need to capitalize `dict`

Dr-Irv (changes requested):

tests/test_frame.py
ramvikrams: Mypy still fails it shows expected type to be `Dict[Hashable, Any]`

ramvikrams: Should it be `dict[Any, S1]` because `flake8` complains about it `Y022 Use "dict[KeyType, ValueType]" instead of "typing.Dict[KeyType, ValueType]" (PEP 585 syntax)`

ramvikrams: > For `Series`, the type stub returns `dict[Any, S1]` , but for the test above, the type of the values of the series is unknown, so you should use `Dict[Any, Any]` in the `assert_type()`.
> 

oh, I got your point thanks

Dr-Irv (approved):
thanks @ramvikrams

ramvikrams:
> thanks @ramvikrams

thanks sir

PR #523, opened 2023-01-30
Update explode to handle lists

rchiodo:
Based on this feedback here that we received: 
https://github.com/microsoft/pylance-release/issues/3871

It looks like explode also takes a list.

- [x] Closes https://github.com/microsoft/pylance-release/issues/3871


bashtage:
Since str is included, why not Sequence[Hashable]

rchiodo:
> Since str is included, why not Sequence[Hashable]

Good idea, I'll switch it to that.

pandas-stubs/core/frame.pyi
```diff
@@ -1011,7 +1011,9 @@ class DataFrame(NDFrame, OpsMixin):
     def stack(
         self, level: Level | list[Level] = ..., dropna: _bool = ...
     ) -> DataFrame | Series[Any]: ...
-    def explode(self, column: _str | tuple, ignore_index: _bool = ...) -> DataFrame: ...
+    def explode(
+        self, column: _str | tuple | Sequence[Hashable], ignore_index: _bool = ...
```
Dr-Irv, line 1015: You can just use `Sequence[Hashable]` by itself, as that will match `_str` and `tuple` as well.

Dr-Irv (approved):
thanks @rchiodo

PR #526, opened 2023-02-03
Add RandomState support for df sampling

### LINKED ISSUE #525: The DataFrame.sample function does not accept Generator objects for random_state
**Describe the bug**
When calling [DataFrame.sample](https://pandas.pydata.org/docs/reference/api/pandas.DataFrame.sample.html) with a random_state argument, a type warning is
raised if a numpy.random.RandomState object is used, claiming that only an integer
can be inputted as an argument.

**To Reproduce**

```python
import numpy as np
import pandas as pd

rng = np.random.default_rng(0)

df = pd.DataFrame(data={'col1': [1, 2], 'col2': [3, 4]})
df.sample(n=1, random_state=rng)
```

Error message:
```
Argument of type "Generator" cannot be assigned to parameter "random_state" of type "int | None" in function "sample"
  Type "Generator" cannot be assigned to type "int | None"
    "Generator" is incompatible with "int"
```

 - OS: Manjaro Linux
 - python 3.10.9
 - version of type checker: pyright 1.1
 - version of installed `pandas-stubs`: 1.5.2

intermezzio:
- [ ] Closes #525
- [ ] Tests added: /core/frame.pyi L234

### DIFF
--- a/pandas-stubs/_typing.pyi
+++ b/pandas-stubs/_typing.pyi
@@ -374,7 +374,7 @@ ValidationOptions: TypeAlias = Literal[
     "m:m",
 ]
 
-RandomState = Union[
+RandomState: TypeAlias = Union[
     int,
     ArrayLike,
     np.random.Generator,
--- a/pandas-stubs/_typing.pyi
+++ b/pandas-stubs/_typing.pyi
@@ -373,4 +373,13 @@ ValidationOptions: TypeAlias = Literal[
     "many_to_many",
     "m:m",
 ]
+
+RandomState = Union[
+    int,
+    ArrayLike,
+    np.random.Generator,
+    np.random.BitGenerator,
+    np.random.RandomState,
+]
+
 __all__ = ["npt", "type_t"]
--- a/pandas-stubs/core/frame.pyi
+++ b/pandas-stubs/core/frame.pyi
@@ -94,6 +94,7 @@ from pandas._typing import (
     NaPosition,
     ParquetEngine,
     QuantileInterpolation,
+    RandomState,
     ReadBuffer,
     Renamer,
     ReplaceMethod,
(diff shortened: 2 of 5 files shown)


pandas-stubs/core/frame.pyi
```diff
@@ -1916,7 +1917,7 @@ class DataFrame(NDFrame, OpsMixin):
         frac: float | None = ...,
         replace: _bool = ...,
         weights: _str | ListLike | None = ...,
-        random_state: int | None = ...,
+        random_state: RandomState = ...,
```
Dr-Irv, line 1920: should be `RandomState | None`

pandas-stubs/_typing.pyi
```diff
@@ -373,4 +373,13 @@ ValidationOptions: TypeAlias = Literal[
     "many_to_many",
     "m:m",
 ]
+
+RandomState = Union[
```
Dr-Irv: Should be `RandomState:  TypeAlias = Union[`

(Sorry I should have caught that before - it's a difference between pandas and the stubs)

Dr-Irv (changes requested):
Can you also change the declarations of that parameter in `sample` in `core/series.pyi` and `core/groupby/generic.pyi` as well?  You don't need to add tests for those.

You need to run `poetry run poe style` on your code as well.

pandas-stubs/_typing.pyi
intermezzio: Done!

intermezzio:
I've added this to generic and series. Unfortunately I'm having issues running the testing suite (poetry install runs but the precommit hook tries to get python from an unrelated conda environment which is messing up `poetry run poe style`) although I've at least changed the styling so that it should fix the long line created in the first commit of the PR.

Dr-Irv (approved):
thanks @intermezzio

PR #528, opened 2023-02-06
fix NumericIndex for nightly

Dr-Irv:
Nightly builds were failing because `NumericIndex` is removed for 2.0, so this allows it to work for 1.5.x, but removes it from testing for 2.0. 

Once 2.0 is released, then we can remove `NumericIndex` in the PYI and tests.

### DIFF
--- a/tests/test_indexes.py
+++ b/tests/test_indexes.py
@@ -12,7 +12,6 @@ from typing import (
 import numpy as np
 from numpy import typing as npt
 import pandas as pd
-from pandas.core.indexes.numeric import NumericIndex
 from typing_extensions import (
     Never,
     assert_type,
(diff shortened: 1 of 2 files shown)


tests/test_indexes.py
```diff
@@ -12,23 +12,32 @@
...
 
+if TYPE_CHECKING:
+    from pandas.core.indexes.numeric import NumericIndex
+
+    from pandas._typing import IndexIterScalar
+else:
+    if not PD_LTE_15:
+        from typing_extensions import TypeAlias
+
+        NumericIndex: TypeAlias = pd.Index
```
twoertwein, line 37: `from pandas import Index as NumericIndex` could be slightly simpler

Dr-Irv: Good idea.  Fixed in commit https://github.com/pandas-dev/pandas-stubs/pull/528/commits/600ab56f654a20a7633847d766a13eca3faa32ab

twoertwein:
Thanks @Dr-Irv

PR #532, opened 2023-02-10
remove types from Index.__iter__()`

Dr-Irv:
  - `test_frame.py/test_in_columns()`

Without this fix, the following code reports errors:
```python
import pandas as pd
import numpy as np

df = pd.DataFrame(np.random.random((3,4)), columns=["cat", "dog", "rat", "pig"])
cols = [c for c in df.columns if "at" in c]
print(df[cols])
```
pyright then reports:
```text
colcheck.py:5:34 - error: Operator "in" not supported for types "Literal['at']" and "IndexIterScalar | tuple[Hashable, ...]"
    Operator "in" not supported for types "Literal['at']" and "bytes"
    Operator "in" not supported for types "Literal['at']" and "date"
    Operator "in" not supported for types "Literal['at']" and "datetime"
    Operator "in" not supported for types "Literal['at']" and "timedelta"
    Operator "in" not supported for types "Literal['at']" and "bool"
    Operator "in" not supported for types "Literal['at']" and "int"
    Operator "in" not supported for types "Literal['at']" and "float"
    Operator "in" not supported for types "Literal['at']" and "Timestamp"
    ... (reportGeneralTypeIssues)
```
mypy reports:
```text
colcheck.py:5: error: Unsupported operand types for in (likely involving Union)
 [operator]
colcheck.py:5: error: Unsupported right operand type for in ("Union[Union[str, bytes, date, datetime, timedelta, bool, int, float, Timestamp, Timedelta], Tuple[Hashable, ...]]")  [operator]
```

This is because of a change made in #367 by @gandhis1 .  Comments welcome from @gandhis1 if I missed something here.

Our application code uses this pattern a lot, and it would be really painful to keep having to do `cast` when selecting a subset of the columns.

### DIFF
--- a/pandas-stubs/core/indexes/base.pyi
+++ b/pandas-stubs/core/indexes/base.pyi
@@ -33,7 +33,6 @@ from pandas._typing import (
     DtypeObj,
     FillnaOptions,
     HashableT,
-    IndexIterScalar,
     IndexT,
     Label,
     Level,
@@ -223,7 +222,7 @@ class Index(IndexOpsMixin, PandasObject):
     def shape(self) -> tuple[int, ...]: ...
     # Extra methods from old stubs
     def __eq__(self, other: object) -> np_ndarray_bool: ...  # type: ignore[override]
-    def __iter__(self) -> Iterator[IndexIterScalar | tuple[Hashable, ...]]: ...
+    def __iter__(self) -> Iterator: ...
     def __ne__(self, other: object) -> np_ndarray_bool: ...  # type: ignore[override]
     def __le__(self, other: Index | Scalar) -> np_ndarray_bool: ...  # type: ignore[override]
     def __ge__(self, other: Index | Scalar) -> np_ndarray_bool: ...  # type: ignore[override]
--- a/tests/test_frame.py
+++ b/tests/test_frame.py
@@ -1813,7 +1813,6 @@ def test_frame_index_numpy() -> None:
 
 
 def test_frame_stack() -> None:
-
     multicol2 = pd.MultiIndex.from_tuples([("weight", "kg"), ("height", "m")])
     df_multi_level_cols2 = pd.DataFrame(
         [[1.0, 2.0], [3.0, 4.0]], index=["cat", "dog"], columns=multicol2
@@ -2406,6 +2405,10 @@ def test_npint_loc_indexer() -> None:
 
 
 def test_in_columns() -> None:
+    # GH 532 (PR)
     df = pd.DataFrame(np.random.random((3, 4)), columns=["cat", "dog", "rat", "pig"])
     cols = [c for c in df.columns if "at" in c]
     check(assert_type(cols, list), list, str)
+    check(assert_type(df.loc[:, cols], pd.DataFrame), pd.DataFrame)
+    check(assert_type(df[cols], pd.DataFrame), pd.DataFrame)
+    check(assert_type(df.groupby(by=cols).sum(), pd.DataFrame), pd.DataFrame)
--- a/tests/test_frame.py
+++ b/tests/test_frame.py
@@ -2403,3 +2403,9 @@ def test_npint_loc_indexer() -> None:
 
     a: npt.NDArray[np.uint64] = np.array([10, 30], dtype="uint64")
     check(assert_type(get_NDArray(df, a), pd.DataFrame), pd.DataFrame)
+
+
+def test_in_columns() -> None:
+    df = pd.DataFrame(np.random.random((3, 4)), columns=["cat", "dog", "rat", "pig"])
+    cols = [c for c in df.columns if "at" in c]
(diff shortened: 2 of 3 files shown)


pandas-stubs/core/indexes/base.pyi
```diff
@@ -223,7 +222,7 @@ class Index(IndexOpsMixin, PandasObject):
     def shape(self) -> tuple[int, ...]: ...
     # Extra methods from old stubs
     def __eq__(self, other: object) -> np_ndarray_bool: ...  # type: ignore[override]
-    def __iter__(self) -> Iterator[IndexIterScalar | tuple[Hashable, ...]]: ...
+    def __iter__(self) -> Iterator: ...
```
twoertwein, line 225: If we were to type the return type, shouldn't this simply have been `Iterator[Hashable]` (tuple and everything in IndexIterScalar) is hashable?

(Having mypy_primer could have been helpful to see the false-positives earlier.)

gandhis1: https://github.com/pandas-dev/pandas-stubs/pull/367#issuecomment-1304724134

gandhis1:
In the original issue: https://github.com/pandas-dev/pandas-stubs/issues/365

I had pointed out this was having issues:

```
    import pandas as pd
    df = pd.DataFrame(data={"col1": [1, 2], {2: ["a", "b"]})
    test = list(df.columns)
    reveal_type(test)   # List[Unknown]
```

That is a contrived example, but I believe the result of that is being unable to assign `df.columns` to a variable and then use said variable to index a DataFrame's columns.

If we make this change, are these common patterns still supported?

```
cols_to_keep = [c for c in df.columns if c.startswith("keep")]
df1 = df1.loc[:, cols_to_keep]
df2 = df2[cols_to_keep]
df3 = df3.groupby(by=cols_to_keep).sum()
```

I would have expected that I or someone else added tests for all of these but given that you didn't have to remove any tests, I'm not so sure that the test coverage is there.

pandas-stubs/core/indexes/base.pyi
Dr-Irv: > If we were to type the return type, shouldn't this simply have been `Iterator[Hashable]` (tuple and everything in IndexIterScalar) is hashable?
> 
> (Having mypy_primer could have been helpful to see the false-positives earlier.)

If I change it to `Iterator[Hashable]`, then I get this with the test (and example above) from pyright:
```text
Operator "in" not supported for types "Literal['at']" and "Hashable"
  Operator "in" not supported for types "Literal['at']" and "Hashable"
```

Dr-Irv:
> In the original issue: #365
> 
> I had pointed out this was having issues:
> 
> ```
>     import pandas as pd
>     df = pd.DataFrame(data={"col1": [1, 2], {2: ["a", "b"]})
>     test = list(df.columns)
>     reveal_type(test)   # List[Unknown]
> ```
> 
> That is a contrived example, but I believe the result of that is being unable to assign `df.columns` to a variable and then use said variable to index a DataFrame's columns.

Since we don't have a generic index yet, we don't know the types that are inside of `df.columns` or `df.index` .  We just know it is the base class `Index`.   But you can use the untyped `List` to index into a `DataFrame`.  

> 
> If we make this change, are these common patterns still supported?
> 
> ```
> cols_to_keep = [c for c in df.columns if c.startswith("keep")]
> df1 = df1.loc[:, cols_to_keep]
> df2 = df2[cols_to_keep]
> df3 = df3.groupby(by=cols_to_keep).sum()
> ```
> 
Yes, and I added that to the tests in commit https://github.com/pandas-dev/pandas-stubs/pull/532/commits/50f4f9f1d2072675318db09ac7b2fcc2e71953fe

> I would have expected that I or someone else added tests for all of these but given that you didn't have to remove any tests, I'm not so sure that the test coverage is there. I may have added tests checking the type when I should have added tests checking the usage pattern.

I guess we didn't have the tests when one tries to use `in` to restrict the columns that are chosen.

I did have an idea (but it would take some work to get right) to change pandas where we make `Index.tolist()` have an optional argument that would be used for typing purposes, so that if you knew what kind of type was in the index (or `df.columns`), you would write my example as follows:
```python
cols = [c for c in df.columns.tolist(str) if "at" in c]
```

The return type of `tolist()` would be `list[str]` in this case, whereas if you did `df.columns.tolist(int)`, the return type would be `list[int]`, but the values in the list would actually be strings (or whatever were the types in the index itself).

twoertwein (approved):

twoertwein:
Thanks @Dr-Irv

gandhis1:
Hm...I guess I'm surprised everything works, because when I originally submitted that bug, I did so in response to something that was broken in my firm's codebase. Maybe it was fixed elsewhere since then. When this is released I'll keep an eye out for any regressions.

PR #533, opened 2023-02-11
Update mypy

twoertwein:


Dr-Irv (approved):
Thanks @twoertwein .  Interesting things that mypy 1.0 fixed

PR #538, opened 2023-02-15
ENH: Add `to_tuples` method to `IntervalIndex`

skatsuta:

Thank you for always maintaining this stub!
It contributes greatly to the robustness of Python programs using pandas.

### Problem

`IntervalIndex` has [`to_tuples` method](https://pandas.pydata.org/docs/reference/api/pandas.IntervalIndex.to_tuples.html), but its stub is currently missing.

### Solution

Add a stub for `IntervalIndex#to_tuples` and a test for it.

Note that the documentation above states that the return type of that method is an ndarray,
but it is actually `Index` (the test to be added checks for this).
This appears to be a mistake in the pandas doc.
(Probably the doc for `IntervalArray#to_tuples` is also being used for `IntervalIndex#to_tuples`, and since `IntervalArray#to_tuples` does indeed return an ndarray, it is incorrectly showing up in the doc for `IntervalIndex#to_tuples` as well.)

### DIFF
--- a/tests/test_interval_index.py
+++ b/tests/test_interval_index.py
@@ -31,6 +31,11 @@ def test_from_tuples() -> None:
     )
 
 
+def test_to_tuples() -> None:
+    ind = pd.IntervalIndex.from_tuples([(0, 1), (1, 2)]).to_tuples()
+    check(assert_type(ind, pd.Index), pd.Index, tuple)
+
+
 def test_is_overlapping() -> None:
     ind = pd.IntervalIndex.from_tuples([(0, 2), (1, 3), (4, 5)])
     check(assert_type(ind.is_overlapping, bool), bool)
--- a/pandas-stubs/core/indexes/interval.pyi
+++ b/pandas-stubs/core/indexes/interval.pyi
@@ -213,6 +213,7 @@ class IntervalIndex(IntervalMixin, ExtensionIndex, Generic[IntervalT]):
         copy: bool = ...,
         dtype: IntervalDtype | None = ...,
     ) -> IntervalIndex[pd.Interval[pd.Timedelta]]: ...
+    def to_tuples(self, na_tuple: bool = ...) -> pd.Index: ...
     @overload
     def __contains__(self, key: IntervalT) -> bool: ...  # type: ignore[misc]
     @overload


tests/test_interval_index.py
```diff
@@ -31,6 +31,11 @@ def test_from_tuples() -> None:
     )
 
 
+def test_to_tuples() -> None:
+    ind = pd.IntervalIndex.from_tuples([(0, 1), (1, 2)]).to_tuples()
+    check(assert_type(ind, pd.Index), pd.Index)
```
Dr-Irv, line 36: Can you change this test to 
```python
check(assert_type(ind, pd.Index), pd.Index, tuple)
```

That will then check we get an `Index` of tuples.

Dr-Irv (changes requested):
Thanks for the PR.  One small change in the test.

Dr-Irv:
> Note that the documentation above states that the return type of that method is an ndarray,
> but it is actually `Index` (the test to be added checks for this).
> This appears to be a mistake in the pandas doc.
> (Probably the doc for `IntervalArray#to_tuples` is also being used for `IntervalIndex#to_tuples`, and since `IntervalArray#to_tuples` does indeed return an ndarray, it is incorrectly showing up in the doc for `IntervalIndex#to_tuples` as well.)

Can you create an issue in the pandas repo about this doc issue?  You're correct that the docs for `IntervalIndex.to_tuples()` is generated from `IntervalArray.to_tuples()` so that is something that should get fixed over there.

tests/test_interval_index.py
skatsuta: Actually I wrote such a test at first, but finally removed the last `tuple` argument because I think it is not necessary to check the type of elements in the `Index` in this test at this point.

If Indexes are generic now, then this test should be like this:

```py
check(assert_type(ind, "pd.Index[tuple]"), pd.Index, tuple)
```

in which case it is important to check that the inferred return type is *actually* an instance of an Index of tuples.

However, since Indexes are not generic at the moment, it is not the responsibility of this stub to ensure that a return value is actually an Index of tuples.
For now, this type stub layer should only guarantee the behavior of "returning an Index (of something)".

So, I think the current test is fine for now, and checking for the type of elements in an Index will make sense after generic support for indexes is added in the future (at that time this test should be modified to the code above).

Are you suggesting that this test should check for the type of elements in an Index even though it is currently a test for runtime behavior and not a test for this type stub?

skatsuta:
> Can you create an issue in the pandas repo about this doc issue? You're correct that the docs for IntervalIndex.to_tuples() is generated from IntervalArray.to_tuples() so that is something that should get fixed over there.

OK thanks, will open an issue.

tests/test_interval_index.py
Dr-Irv: >  I think the current test is fine for now, and checking for the type of elements in an Index will only make sense after generic support for Indexes is added in the future (at that time this test should be modified to the code above).
> 
> Are you suggesting that this test should check for the type of elements in the Index even though it is currently a test for runtime behavior and not a test for this type stub?

The `assert_type()` call checks the validity of the stub.  The `check()` call validates that the stub and runtime behavior are aligned.  By adding the `tuple` argument to `check`, it is just making sure that alignment is correct.  

We've done that (admittedly inconsistently) in other places, and it actually has revealed some errors in pandas itself, or in the stubs.  So the goal is to try to get more consistency in how we do our testing.

I agree that things will be better once we make `Index` generic in the stubs.

skatsuta (commented):
@Dr-Irv Thanks for your suggestion!
I have understood your intention now and added the `tuple` argument back.
I would appreciate it if you could re-review it when you have time.

Also, the test using mypy nightly is failing, but I don't think that is due to this change.

Dr-Irv (approved):
Thanks @skatsuta

Dr-Irv:
> Also, a test with mypy nightly is failing, but I don't think that is due to this change.

That was fixed in #541

PR #539, opened 2023-02-16
HDFStore.__iter__

twoertwein:
Without this PR

```py
for key in pd.HDFStore(...):
   ...
```
would cause `error: "List[str]" has no attribute "__next__"  [attr-defined]`


twoertwein:
Enabled Y034 which in theory should have caught this (https://github.com/PyCQA/flake8-pyi/issues/348). Also added Self wherever Y034 was suggesting it (could be used in many more places). 

xref https://github.com/pandas-dev/pandas/issues/51233

Dr-Irv (approved):
Glad you found the `Y034` thing.  Thanks @twoertwein

twoertwein:
The CI seems kind of slow (most runs are queued). Should we:

- only test on the latest and oldest python versions
- still test on all python versions but test them disjointly across different platforms
- same as above but always test oldest/newest on all platforms

Dr-Irv:
> The CI seems kind of slow (most runs are queued). Should we:
> 
> * only test on the latest and oldest python versions
> * still test on all python versions but test them disjointly across different platforms
> * same as above but always test oldest/newest on all platforms

I think we should leave it as is.  It seems the runs being queued depends on time of day, and happens more with Windows and MacOS than Linux.

PR #540, opened 2023-02-16
Enable Y019, Y037, and Y041

twoertwein:
I will try to enable a few more of the flake8-pyi checks in the next days.


Dr-Irv:
> I will try to enable a few more of the flake8-pyi checks in the next days.

It will be interesting to see if removing `Union` will work for older versions of python.  I think I did a test at some point, and for both `TypeAlias` and `TypeVar` declarations, you had to use `Union` instead of `|` .  But I could be mistaken.  Or maybe you have to use it in runtime code, but removing `Union` in stubs is OK.

twoertwein:
Could also enable Y041. I think we had a discussion about that in the past. Would need to change these cases:

> pandas-stubs/_libs/interval.pyi:151:45: Y041 Use "float" instead of "int | float" (see "The numeric tower" in PEP 484)
> pandas-stubs/core/series.pyi:1813:30: Y041 Use "float" instead of "int | float" (see "The numeric tower" in PEP 484)
> pandas-stubs/core/series.pyi:1813:83: Y041 Use "float" instead of "int | float" (see "The numeric tower" in PEP 484)
> pandas-stubs/core/series.pyi:1814:34: Y041 Use "float" instead of "int | float" (see "The numeric tower" in PEP 484)
> pandas-stubs/core/series.pyi:1814:87: Y041 Use "float" instead of "int | float" (see "The numeric tower" in PEP 484)
> pandas-stubs/core/tools/datetimes.pyi:84:19: Y041 Use "float" instead of "int | float" (see "The numeric tower" in PEP 484)
> pandas-stubs/core/tools/datetimes.pyi:86:13: Y041 Use "float" instead of "int | float" (see "The numeric tower" in PEP 484)

pandas-stubs/core/indexes/base.pyi
```diff
@@ -51,20 +53,18 @@ class InvalidIndexError(Exception): ...
...
+        self,
         idx: slice
         | np_ndarray_anyint
         | Sequence[int]
         | Index
         | Series[bool]
         | Sequence[bool]
         | np_ndarray_bool,
-    ) -> _IndexGetitemMixinT: ...
+    ) -> Self: ...
```
Dr-Irv, line 67: I thought I tried this here and it didn't work

Dr-Irv:
> Could also enable Y041. I think we had a discussion about that in the past. Would need to change these cases:
> 
> >

I looked at the cases and I would be OK with changing those ones.

The issue we were having before is removing `int` from `Scalar`, and we want to leave it there.

pandas-stubs/core/indexes/base.pyi
twoertwein: Locally it works for me (python 3.11.2 with mypy 1.0) but let's see whether it also works on the CI :)

Dr-Irv: seems to have worked

Dr-Irv (approved):

PR #541, opened 2023-02-17
make Timestamp.strptime() return None

Dr-Irv:
`mypy_nightly` started failing because `typeshed` got updated to have `datetime.strptime()` return `Self`, and since `Timestamp` is a subclass of `datetime`, the nightly mypy build was saying that the `datetime` result of `strptime()` was invalid since it should return `Timestamp` if the superclass returns `Self`.

But `Timestamp.strptime()` is not implemented, and we always raise, so changed the arguments to `Never` meaning that type checkers will always say it is invalid to call this method.


twoertwein (approved):
Has typeshed documented how to handle such cases (using Never, NotImplementd, ...)? I did a quick search but I didn't find a comment on that.

Dr-Irv:
> Has typeshed documented how to handle such cases (using Never, NotImplementd, ...)? I did a quick search but I didn't find a comment on that.

I didn't look.  But this all depends on whether you want to strongly enforce the Liskov substitution principle.  `mypy` does, while `pyright` does not.

In this particular case, we don't want people to call `Timestamp.strptime()` .  Right now, if you do, an exception is raised.  If we remove the declaration of `strptime()` from `timestamps.pyi`, then since `Timestamp` is a subclass of `datetime.datetime`, both `mypy` and `pyright` will say that a call to `Timestamp.strptime()` is valid, because it will find the stub for `datetime.datetime.strptime()`.  By using the `Never` trick, we are telling the type checker that this is an invalid function to call, and I think that is the best user experience.

twoertwein:
> I didn't look. But this all depends on whether you want to strongly enforce the Liskov substitution principle. `mypy` does, while `pyright` does not.

(I prefer pyright's approach, while mypy's approach is theoretically better - it is just not compatible with python: everything inherits from `object` - but some classes violate its `__hash__`). I definitely like informing users of the error (instead of adding a type ignore in pandas-stubs), but I wasn't sure whether there is a "typeshed-recommended" way to do so.

PR #543, opened 2023-02-17
gh-529 Fixing Dataframe.insert

### LINKED ISSUE #529: BUG: DataFrame.insert value is too strict
**Describe the bug**

**To Reproduce**

```
df = DataFrame({"a": [1, 2]})
df.insert(loc=0, column="b", value=None)
```

```
error: Value of type variable "ListLike" of "insert" of "DataFrame" cannot be "None"  [type-var]
```

 - OS Version [e.g. 22] Venturar 13.1
 - python version 3.10
 - version of type checker mypy 0.991
 - version of installed `pandas-stubs` latest

**Additional context**

According to the main repo, any Scalar is allowed

ramvikrams:
- [x] Closes #529(Replace xxxx with the Github issue number)


tests/test_frame.py
```diff
@@ -2410,3 +2410,15 @@ def test_in_columns() -> None:
...
+def test_insert_newvalues() -> None:
+    df = pd.DataFrame({"a": [1, 2]})
+    gh = df.insert(loc=0, column="b", value=None)
+    ab = pd.DataFrame({"col1": [1, 2], "col2": [3, 4]})
+    cd = ab.insert(loc=0, column="newcol", value=[99, 99])
+    ef = pd.DataFrame({"z": [4, 5, 6]})
+    hi = ef.insert(loc=0, column="g", value=4)
+    check(assert_type(gh, pd.DataFrame), pd.DataFrame)
+    check(assert_type(cd, pd.DataFrame), pd.DataFrame)
+    check(assert_type(hi, pd.DataFrame), pd.DataFrame)
```
Dr-Irv, line 2424: Since `insert()` returns `None`, you should change the references from `pd.DataFrame` to `None` in these 3 lines.

Dr-Irv (changes requested):

tests/test_frame.py
```diff
@@ -2410,3 +2410,15 @@ def test_in_columns() -> None:
...
+def test_insert_newvalues() -> None:
+    df = pd.DataFrame({"a": [1, 2]})
+    gh = df.insert(loc=0, column="b", value=None)
+    ab = pd.DataFrame({"col1": [1, 2], "col2": [3, 4]})
+    cd = ab.insert(loc=0, column="newcol", value=[99, 99])
+    ef = pd.DataFrame({"z": [4, 5, 6]})
+    hi = ef.insert(loc=0, column="g", value=4)
+    check(assert_type(gh, None), None)
+    check(assert_type(cd, None), None)
+    check(assert_type(hi, None), None)
```
Dr-Irv, line 2424: You're going to have to do this using the following style, since `insert()` returns `None`:
```python
assert assert_type(df.insert(loc=0, column="b", value=None), None) is None
```
Our `check` function doesn't accept `None`, and `mypy` looks at functions that return `None` as not returning a value, so you can't do the assignment.

Dr-Irv (changes requested):

tests/test_frame.py
ramvikrams: done sir

Dr-Irv (approved):
Thanks @ramvikrams

PR #545, opened 2023-02-19
gh-544: Fix pd.options.display.float_format

### LINKED ISSUE #544: Wrong type for Display.float_format
**Describe the bug**
`Display.float_format` is annotated as `str | None`. However, the correct type should be `Callable[[float], str]`. From [the documentation](https://pandas.pydata.org/docs/user_guide/options.html):
```
display.float_format : callable
    The callable should accept a floating point number and return
    a string with the desired format of the number. This is used
    in some places like SeriesFormatter.
    See formats.format.EngFormatter for an example.
    [default: None] [currently: None]
```

**To Reproduce**
MRE:
```
import pandas as pd

pd.options.display.float_format = '{:,.2f}'.format

df = pd.DataFrame([1.1, 2.2, 3.3, 4.4],
                  index=['a', 'b', 'c', 'd'],
                  columns=['value'])

print(df)
```
Mypy output:
```
$ mypy mre.py  
foo.py:3: error: Incompatible types in assignment (expression has type "Callable[[VarArg(object), KwArg(object)], str]", variable has type "Optional[str]")  [assignment]
```

 - OS: Arch Linux
 - OS Version: Linux 5.15.72
 - Python 3.10.7
 - Mypy and Pyright
 - `pandas-stubs` version 1.5.3.230214

m-pilia:
Set the correct type for pd.options.display.float_format.

- [x] Closes #544

### DIFF
--- a/tests/test_config.py
+++ b/tests/test_config.py
@@ -1,6 +1,8 @@
 from typing import (
     TYPE_CHECKING,
     Any,
+    Callable,
+    Optional,
 )
 
 import pandas as pd
@@ -36,3 +38,13 @@ def test_specific_option():
     check(assert_type(pd.options.plotting.backend, str), str)
     # Just check assignment
     pd.options.plotting.backend = "matplotlib"
+
+
+def test_display_float_format():
+    check(
+        assert_type(pd.options.display.float_format, Optional[Callable[[float], str]]),
+        type(None),
+    )
+    formatter = "{,.2f}".format
+    with pd.option_context("display.float_format", formatter):
+        assert pd.get_option("display.float_format") == formatter
--- a/pandas-stubs/_config/config.pyi
+++ b/pandas-stubs/_config/config.pyi
@@ -1,4 +1,7 @@
-from collections.abc import Iterable
+from collections.abc import (
+    Callable,
+    Iterable,
+)
 from contextlib import ContextDecorator
 from typing import (
     Any,
@@ -51,7 +54,7 @@ class Display(DictWrapper):
     date_yearfirst: bool
     encoding: str
     expand_frame_repr: bool
-    float_format: str | None
+    float_format: Callable[[float], str] | None
     html: DisplayHTML
     large_repr: str
     latex: DisplayLaTeX


tests/test_config.py
```diff
@@ -36,3 +38,13 @@ def test_specific_option():
     check(assert_type(pd.options.plotting.backend, str), str)
     # Just check assignment
     pd.options.plotting.backend = "matplotlib"
+
+
+def test_display_float_format():
+    check(
+        assert_type(pd.options.display.float_format, Optional[Callable[[float], str]]),
```
Dr-Irv, line 45: It seems that python 3.8 doesn't like `Optional[Callable[[float], str]]` at runtime.  

But if you do `assert_type(pd.options.display.float_format, "Optional[Callable[[float], str]])"`, (i.e., surround the type in quotes), then it should pass the type checkers for all python versions as well as `pytest` .  You'd also have to remove the imports of `Callable` and `Optional`

Dr-Irv (changes requested):
Thanks for the PR.  python 3.8 CI exhibited some funky behavior to work around.

tests/test_config.py
m-pilia: Hi Dr-Irv, thanks for the comment. Unfortunately if I specify the type as a string like this
```python
assert_type(pd.options.display.float_format, "Optional[Callable[[float], str]])"
```
and remove the imports, mypy produces some errors due to missing definitions (Python 3.10):
```
tests/test_config.py:43: error: Expression is of type "Optional[Callable[[float], str]]", not "Any"  [assert-type]
tests/test_config.py:43: error: Name "Optional" is not defined  [name-defined]
tests/test_config.py:43: note: Did you forget to import it from "typing"? (Suggestion: "from typing import Optional")
```
If I add back the imports then mypy succeeds, but then flake8 fails (because of the unused imports).

I think a solution could be to use the legacy `typing.Callable` class (that was replaced by PEP 585 in Python 3.9). This is deprecated, but should still work and can be changed back to `collections.abc.Callable` as soon as support for Python 3.8 is dropped (I would assume the legacy class will not be removed before then).

Dr-Irv: I ran the CI with the current version, and it passed, so using `typing.Callable` does seem to be the way to go.

Dr-Irv (approved):
Thanks @m-pilia

Issue #372, opened 2022-10-07
Fix  Series.astype() to return a Series[type] dependent on argument


Dr-Irv:
Deprecations in 1.5 on `Series.astype()` revealed that we have to work on that method.

See https://github.com/pandas-dev/pandas-stubs/issues/227#issuecomment-1260164389

Ideally, things like `s.astype(int)` and `s.astype("int")` would both return `Series[int]` .  Have to deal with each of the possible types, as well as the possible strings.

ramvikrams:
So we just have to change the return type to `Series[S1]`

Dr-Irv:
> So we just have to change the return type to `Series[S1]`

No, it's more nuanced than that, because you can pass strings to `astype()` that correspond to dtypes, and there are dtype arguments that are not in the `S1` typevar.  So you need a lot of overloads to appropriately do the mappings.  Have to handle a variety of numpy types as well.

I'd suggest making a table of values of `S1` in one column, and then in the second column, you list all the possible arguments for `astype()` that could generate that value of `S1`.  Then, from that table, you build a set of overloads.

There might be cases where there is no possible value for the `astype()` argument that give a particular value in `S1`.  We'll need to review them one by one.

ramvikrams:
> > So we just have to change the return type to `Series[S1]`
> 
> No, it's more nuanced than that, because you can pass strings to `astype()` that correspond to dtypes, and there are dtype arguments that are not in the `S1` typevar. So you need a lot of overloads to appropriately do the mappings. Have to handle a variety of numpy types as well.
> 
> I'd suggest making a table of values of `S1` in one column, and then in the second column, you list all the possible arguments for `astype()` that could generate that value of `S1`. Then, from that table, you build a set of overloads.
> 
> There might be cases where there is no possible value for the `astype()` argument that give a particular value in `S1`. We'll need to review them one by one.



i'll get started with it now then

PR #546, opened 2023-02-20
Change check() function to test actual types in Series and Index

Dr-Irv:
While reviewing #372 , it uncovered that by having `check()` use `next(iter())`, if you have a `Series([1,2,3], dtype=np.int8)`, it was not checking that the actual type of the first element was `np.int8`, since `Series.__iter__()` returns a python scalar.

So this corrects that, allowing us to be more precise in the type checking of the values inside the results, since we want to check that `astype()` is doing what we expect.

Also, was able to allow `pytables` to work with 3.11, since they released a compatible version.

Uncovered an odd issue with bitwise operations on lists of `int`

In some cases, used `np.integer` because it is a supertype that works across platforms.  Other times, `np.int_` worked fine.


tests/test_io.py
```diff
@@ -66,10 +66,7 @@
 from pandas.io.sas.sas_xport import XportReader
 from pandas.io.stata import StataReader
 
-from . import (
-    lxml_skip,
-    pytables_skip,
-)
+from . import lxml_skip
```
twoertwein, line 69: Could this be `from tests import lxml_skip` (if it is in `__init__.py`)?

twoertwein (approved):
Looks good to me! Hope you had a somewhat automated approach for all these changes :)

`iloc` vs `__iter__`: I would have been fine keeping the current approach as it is not clear what the "correct" type is.

tests/test_io.py
Dr-Irv: Probably, I just left it as it was.

Dr-Irv:
> Looks good to me! Hope you had a somewhat automated approach for all these changes :)

Unfortunately not!  It was a bit of wack-a-mole.
> 
> `iloc` vs `__iter__`: I would have been fine keeping the current approach as it is not clear what the "correct" type is.

Agreed, but I think we should be doing this level of checking when we get to things like `Int8Dtype()` .

twoertwein:
Thanks @Dr-Irv !

btw, the next mypy release (1.1?) will suggest using pandas-stubs https://github.com/python/mypy/issues/14328#event-8556645254

PR #550, opened 2023-02-21
Fixes mypy error when passing `np.datetime64/timedelta64` input to `pd.isna`

### LINKED ISSUE #549: mypy error when passing `np.datetime64` or `np.timedelta64` to `pd.isna`
**Describe the bug**
`pd.isna` does not support `np.datetime64` and `np.timedelta64` input types.

**To Reproduce**
```

import datetime as dt
from typing import TypeVar

import numpy as np
import pandas as pd

TV_TD = TypeVar('TV_TD', dt.datetime, np.datetime64, dt.timedelta, np.timedelta64)

def test_isna(value: TV_TD) -> bool:
    return pd.isna(value)
```

`mypy` output:
```
 error: No overload variant of "isna" matches argument type "datetime64"  [call-overload]
 note:     def isna(obj: DataFrame) -> DataFrame
 note:     def isna(obj: Series[Any]) -> Series[bool]
 note:     def isna(obj: Union[Index, List[Any], Union[ExtensionArray, ndarray[Any, Any]]]) -> ndarray[Any, dtype[bool_]]
 note:     def isna(obj: Union[Union[Union[str, bytes, date, datetime, timedelta, bool, int, float, Timestamp, Timedelta], complex], NaTType, NAType, None]) -> TypeGuard[Union[NaTType, NAType, None]]
 error: No overload variant of "isna" matches argument type "timedelta64"  [call-overload]
```

 - OS: MacOS
 - OS 12.6.3
 - python: 3.8.0
 - mypy: 1.0.1
 - version of installed `pandas-stubs`: 1.5.3.230214

**Additional context**
The problem seems to be that the [`IndexIterScalar`](https://github.com/pandas-dev/pandas-stubs/blob/main/pandas-stubs/_typing.pyi#L160) TypeAlias does not include `np.datetime64` and `np.timedelta64`. The `test_isna` test case is also missing tests for `np.datetime64` and `np.timedelta64` input. PR with a fix [here](https://github.com/pandas-dev/pandas-stubs/pull/550)

nph:
* Update `IndexIterScalar` to include `np.datetime64` and `np.timedelta64`
* Update `test_isna` to include tests for `np.datetime64` and `np.timedelta64`

- [x] Closes #549 

### DIFF
--- a/pandas-stubs/_typing.pyi
+++ b/pandas-stubs/_typing.pyi
@@ -163,6 +163,8 @@ IndexIterScalar: TypeAlias = (
     | datetime.date
     | datetime.datetime
     | datetime.timedelta
+    | np.datetime64
+    | np.timedelta64
     | bool
     | int
     | float
--- a/tests/test_pandas.py
+++ b/tests/test_pandas.py
@@ -266,6 +266,27 @@ def test_isna() -> None:
     assert not check(assert_type(pd.isna(2.5), bool), bool)
     assert check(assert_type(pd.notna(2.5), bool), bool)
 
+    # Checks for datetime, timedelta, np.datetime64 and np.timedelta64
+    py_dt = dt.datetime.now()
+    assert check(assert_type(pd.notna(py_dt), bool), bool)
+    assert not check(assert_type(pd.isna(py_dt), bool), bool)
+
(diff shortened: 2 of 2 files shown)


Dr-Irv (approved):
thanks @nph

PR #519, opened 2023-01-25
gh-372 :  Fixing Series.astype()

ramvikrams:
- [x] Closes #372 
- [x] Tests added: Please use `assert_type()` to assert the type of any return value

I have created this draft PR and I'll keep commiting to it, so you can also check whether the changes are acceptable or not.

PR #554, opened 2023-02-25
Support an ExtensionDtype and ExtensionArray

Dr-Irv:
While working on #519 , I wanted to test that we could pass a user-created `ExtensionDtype` and `ExtensionArray` .  So I copied the testing code from `pandas` that creates a `DecimalArray`, and it revealed a number of issues with our type declarations.

So this PR addresses that.  Not everything is really tested in terms of all the `ExtensionArray` declarations, but the goal here is to see that a user-defined `ExtensionArray` will pass typing checks.


pandas-stubs/_libs/lib.pyi
```diff
@@ -1,9 +1,27 @@
...
-def is_interval(val: object) -> bool: ...
-def is_complex(val: object) -> bool: ...
-def is_bool(val: object) -> bool: ...
-def is_integer(val: object) -> bool: ...
-def is_float(val: object) -> bool: ...
+def is_interval(val: object) -> TypeGuard[Interval]: ...
+def is_complex(val: object) -> TypeGuard[complex]: ...
+def is_bool(val: object) -> TypeGuard[bool]: ...
+def is_integer(val: object) -> TypeGuard[int]: ...
+def is_float(val: object) -> TypeGuard[float]: ...
```
twoertwein, line 27: I think most of these can also return the equivalent numpy version.

xref https://github.com/pandas-dev/pandas/pull/51309

Dr-Irv: I made them equivalent to what is in the PR. Tried for `is_float()` using `TypeGuard[float | np.floating]`, but that fails because `decimal.Decimal(x)` fails when `x` is a numpy float.

Also merged in recent PR for `astype()`, allowing that to now be tested.

pandas-stubs/core/arraylike.pyi
```diff
@@ -35,3 +40,6 @@ class OpsMixin:
     def __rdivmod__(self, other: Any) -> tuple[Self, Self]: ...
     def __pow__(self, other: Any) -> Self: ...
     def __rpow__(self, other: Any) -> Self: ...
+
+def dispatch_ufunc_with_out(self, ufunc: np.ufunc, method: str, *inputs, **kwargs): ...
+def dispatch_reduction_ufunc(self, ufunc: np.ufunc, method: str, *inputs, **kwargs): ...
```
twoertwein, line 45: Those two functions don't seem to be documented

pandas-stubs/core/arrays/base.pyi
```diff
@@ -37,22 +39,31 @@ class ExtensionArray:
     ) -> np.ndarray: ...
     def fillna(self, value=..., method=..., limit=...): ...
     def dropna(self): ...
-    def shift(
-        self, periods: int = ..., fill_value: object = ...
-    ) -> ABCExtensionArray: ...
+    def shift(self: Self, periods: int = ..., fill_value: object = ...) -> Self: ...
```
twoertwein, line 42: need `Self` only as a return value, not for `self` itself.

pandas-stubs/core/dtypes/base.pyi
```diff
@@ -1,23 +1,22 @@
...
-from pandas._libs.missing import NAType
 from pandas._typing import type_t
 
 class ExtensionDtype:
-    @property
-    def na_value(self) -> NAType | NaTType: ...
-    @property
-    def type(self) -> type_t: ...
+    type: type_t
+    na_value: Any
```
twoertwein, line 12: `na_value` seems to be a property (but the return value can be Any for some classes)

pandas-stubs/_libs/lib.pyi
twoertwein: The following seems to work just fine
```py
import numpy as np
from pandas.core.dtypes.common import is_float

is_float(np.float32(1))
```

pandas-stubs/core/arraylike.pyi
Dr-Irv: I took the implementation of `DecimalArray` and those functions were used, so to make the typing work, I had to document them.  I think we have a lot of such functions in the stubs that are not documented, so we can just leave this one for now.

pandas-stubs/core/dtypes/base.pyi
Dr-Irv: OK, had to fix the base class to make them abstract properties

pandas-stubs/core/arraylike.pyi
twoertwein: up to you, wouldn't mind having an ignore comment in the test if we are not sure whether these functions are supposed to be public.

pandas-stubs/core/dtypes/base.pyi
twoertwein: I wouldn't be surpsrised if they are used inconsistently within pandas :( in some places as properties, in some as normal attributes and in some as class variables. I just grepped through pandas/core/dtypes, I didn't find a single line that assigns a value to a class variable `na_value` or `type`. It seems consistently to be a property.

If they are class variables, can you add a quick `isinstance(ClassName.na_value, ...)` test.

If the pandas tests set it for simplicity/laziness to a class varaible (which is fine with me), we can just add an ignore to them.

Dr-Irv: > I wouldn't be surpsrised if they are used inconsistently within pandas :( in some places as properties, in some as normal attributes and in some as class variables. I just grepped through pandas/core/dtypes, I didn't find a single line that assigns a value to a class variable `na_value` or `type`. It seems consistently to be a property.

Take a look at `pandas/core/arrays/integer.py`.  You'll see assignments of `type = ` for the various `Integer` dtypes.

I believe it's declared as a property in `pandas/core/dtypes/base.py` so that if someone doesn't define it, it raises `AbstractMethodError`.  But it is used internally as a `ClassVar`.

For `na_value`, it is inconsistent, but based on the example for `DecimalDtype`, it was treated as a class variable there as well.

pandas-stubs/core/arraylike.pyi
Dr-Irv: > up to you, wouldn't mind having an ignore comment in the test if we are not sure whether these functions are supposed to be public.

OK, I did that.

pandas-stubs/core/dtypes/base.pyi
Dr-Irv: I left `name` and `type` as `ClassVar`, but changed `na_value` to be a property, because it has a default value.

twoertwein: > Take a look at `pandas/core/arrays/integer.py`. You'll see assignments of `type = ` for the various `Integer` dtypes.

I wounder why mypy/pyright didn't catch that. Maybe because a [parent class re-declares `type` as `Any`](https://github.com/pandas-dev/pandas/blob/ac70f230995aa1bbee97b5db7bfdc28e5c5ed14b/pandas/core/dtypes/dtypes.py#L82) or because of the decorators that mypy doesn't understand. Seems like something that should be fixed in pandas.

twoertwein:
Thanks @Dr-Irv !

pandas-stubs/core/dtypes/base.pyi
Dr-Irv: > I wounder why mypy/pyright didn't catch that. Maybe because a [parent class re-declares `type` as `Any`](https://github.com/pandas-dev/pandas/blob/ac70f230995aa1bbee97b5db7bfdc28e5c5ed14b/pandas/core/dtypes/dtypes.py#L82) or because of the decorators that mypy doesn't understand. Seems like something that should be fixed in pandas.

I was wondering the same thing.

In reality, `type` and `name` are abstract class properties of `ExtensionDtype`, but the type checkers aren't supporting that concept.  See long discussions here:  https://github.com/microsoft/pyright/issues/2601 and https://github.com/python/mypy/issues/8993

pandas-stubs/_typing.pyi
```diff
@@ -166,7 +166,8 @@ class WriteExcelBuffer(WriteBuffer[bytes], Protocol):
 
 FilePath: TypeAlias = str | PathLike[str]
 
-Axis: TypeAlias = str | int
+AxisInt: TypeAlias = Literal[0, 1]
+Axis: TypeAlias = AxisInt | Literal["index", "columns", "rows"]
```
aberres, line 170: @twoertwein @Dr-Irv While I understand the intention, this one causes a problem with the following types from Pandas:

`pandas.core.frame.DataFrame.reorder_levels(self, order: Sequence[Axis], axis: Axis = 0)`

The order parameter expects a sequence of ints or strings - with this change this is forbidden.

From my code:

```
reportquery.py:612: error: Argument 1 to
"reorder_levels" of "DataFrame" has incompatible type "Sequence[str]"; expected
"Sequence[Union[Literal[0, 1], Literal['index', 'columns', 'rows']]]"
[arg-type]
            df = df.reorder_levels(column_names, axis=1)
```

PS: Sorry for not creating an issue (yet). I am short on time.

Dr-Irv: > PS: Sorry for not creating an issue (yet). I am short on time.

When you do have the time, please create an issue with a full code sample.  

This has picked up that `reorder_levels()` has an incorrect type for `order`.  It should be `Sequence[Hashable]`

aberres: Here we go: https://github.com/pandas-dev/pandas-stubs/issues/560

PR #556, opened 2023-03-01
ENH: Add support for dtype string aliases to `Series#astype`

skatsuta:

# Problem

`Series#astype` has been improved in #519 to allow specific types to be inferred based on a value of the `dtype` argument.
I really appreciate the hard work that has been put into it.

However, that PR lacked support for aliases that specify dtype as a string.
Therefore, starting with version v1.5.3.230227, type checkers now reports the following errors for code like below:

#### Sample code
```py
s = pd.Series([1, 2, 3])
s_nullable_int = s.astype("Int64")
s_nullable_float = s.astype("Float64")
```

#### Errors reported by Pyright
```
error: Argument of type "Literal['Int64']" cannot be assigned to parameter "dtype" of type "ExtensionDtype" in function "astype"
    "Literal['Int64']" is incompatible with "ExtensionDtype" (reportGeneralTypeIssues)
error: Argument of type "Literal['Float64']" cannot be assigned to parameter "dtype" of type "ExtensionDtype" in function "astype"
    "Literal['Float64']" is incompatible with "ExtensionDtype" (reportGeneralTypeIssues)
```

Specifying dtype as a string is perfectly fine in pandas, so we need to fix the stub so that the errors are not reported for code like the above.

# Solution

Add support for various dtype string aliases to `(Boolean|Int|Str|Float|Complex)DtypeArg` types, and add tests for passing each of those aliases as arguments to `Series#astype`.

Along with this, I have also refactored the stubs and tests to group code for the same dtype together so that it is easier to see what kind of dtype arguments are supported.


Dr-Irv (approved):
Thanks @skatsuta .  I had a feeling that someone would pick up many of the possibilities that were missed when we put this together.

Dr-Irv:
I should be able to do a release by Friday 3/3, if not earlier.

Dr-Irv:
> I guess because `CategoryDtypeArg` is missing from `AstypeArg`.

Yes, you are correct.  Can you open up a new issue for `DataFrame.astype("category")` and maybe follow up with a PR?

PR #562, opened 2023-03-03
Support all possible time units for Series.astype literals

gandhis1:
@Dr-Irv FYI @ramvikrams @skatsuta 

### DIFF
--- a/tests/test_series.py
+++ b/tests/test_series.py
@@ -1571,12 +1571,22 @@ def test_updated_astype() -> None:
         np.complex128,
     )
 
+    # Check a couple of selected time units, but not all, to avoid excessive test cases
+    check(
+        assert_type(s.astype("timedelta64[M]"), TimedeltaSeries),
+        pd.Series,
+        Timedelta,
+    )
+    check(
+        assert_type(s.astype("datetime64[M]"), TimestampSeries),
+        pd.Series,
+        Timestamp,
+    )
     check(
         assert_type(s.astype("timedelta64[ns]"), TimedeltaSeries),
         pd.Series,
         Timedelta,
     )
-
     check(
         assert_type(s.astype("datetime64[ns]"), TimestampSeries),
         pd.Series,
--- a/tests/test_series.py
+++ b/tests/test_series.py
@@ -16,7 +16,6 @@ from typing import (
     Tuple,
     TypeVar,
     cast,
-    get_args,
 )
 
 import numpy as np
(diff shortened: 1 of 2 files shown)


tests/test_series.py
```diff
@@ -1571,12 +1571,22 @@ def test_updated_astype() -> None:
         np.complex128,
     )
 
+    # Check a couple of selected time units, but not all, to avoid excessive test cases
```
Dr-Irv, line 1574: I think we can do this as follows:
```python
if TYPE_CHECKING:
    from pandas._typing import TimedeltaDtypeArg

tdtypes: list["TimedeltaDtypeArg"] = [ "timedelta64[Y]",  # annotation ignored at runtime
    "timedelta64[M]", # etc ]
for tdt in tdtypes:
    check(
        assert_type(s.astype(tdt), TimedeltaSeries),
        pd.Series,
        Timedelta,
    )
```
The type checkers should then see that `tdt` is valid to pass, but, at least from my point of view, we can then see that all of those types are accepted at runtime when we do `pytest`.

Unfortunate part is that you have to copy all those values from `_typing.pyi` to `test_series.py`, but it still accomplishes the goal of testing them all without having to write a test for each one.

Can you try that out?

Dr-Irv (changes requested):

tests/test_series.py
gandhis1: So I just pushed an update which does this, and avoids the duplication by using `get_args` and casting to a `tuple[TimedeltaDtypeArg, ...]`. Unless I'm mistaken, I believe that would be functionally equivalent to what you have above.

That said, I'm skeptical this is a comprehensive test. All the type assertion is checking is the type alias, not what is part of that type alias. Don't we need to check standalone string literals in addition to whether the runtime behavior matches the annotation?

If that's the case and we want to check everything, then I can just create all the individual test cases. It seems like the # of test cases isn't really a concern.

Dr-Irv: > That said, I'm skeptical this is a comprehensive test. All the type assertion is checking is the type alias, not what is part of that type alias. Don't we need to check standalone string literals in addition to whether the runtime behavior matches the annotation?
> 
> If that's the case and we want to check everything, then I can just create all the individual test cases. It seems like the # of test cases isn't really a concern.

The problem with using `cast` and `get_args()` is that you have everything inside of a `if TYPE_CHECKING:` block, so we are missing the benefit of making sure that the types we have in the list work at runtime.  And we can't import `TimedeltaDtypeArg` at runtime, because it only exists in the stubs.  In other words, the code you wrote never gets executed by `pytest`.

So we have to either do each test explicitly for each string, or just duplicate the set of possible strings in `test_series.py`

Dr-Irv (changes requested):
You have `get_args()` inside of  a `if TYPE_CHECKING` block, so the code never gets executed.  Problem is that the types are not available at runtime.

Dr-Irv (approved):
Thanks @gandhis1

PR #581, opened 2023-03-16
gh:576 - fixed `.loc`

### LINKED ISSUE #576: Fix `.loc` to handle ambiguity if a single Scalar is first element of a tuple
From discussion in #575 

In https://github.com/pandas-dev/pandas-stubs/blob/41fe0f509e134109e2e7013632bba4a76fe56d52/pandas-stubs/core/frame.pyi#L163

Replace `Hashable` with `slice | _IndexSliceTuple | Callable`

If you make the suggested change, that will cause the test `test_frame.py:test_loc_slice()` to fail, but I now realize that the expression used there is ambiguous:
```python
>>> df1 = pd.DataFrame(
...         {"x": [1, 2, 3, 4]},
...         index=pd.MultiIndex.from_product([[1, 2], ["a", "b"]], names=["num", "let"]),
...     )
>>> df1.loc[1, :]
     x
let
a    1
b    2
>>> df2 = pd.DataFrame({"x": [1,2,3,4]}, index=[10, 20, 30, 40])
>>> df2.loc[10, :]
x    1
Name: 10, dtype: int64
```

So the first argument as an integer could return a `DataFrame` or `Series`, dependent on whether the underlying index is a regular `Index` or `MultiIndex`

The solution is then to add another overload in `_LocIndexerFrame.__getitem__()`:
```python
    @overload
    def __getitem__(self, idx: tuple[ScalarT, slice]) -> Series | DataFrame: ...
```
Then modify the test in `test_index_slice()` to check that the type is `Union[pd.Series, pd.DataFrame]`, and add another test corresponding to `df2` above.

_Originally posted by @Dr-Irv in https://github.com/pandas-dev/pandas-stubs/pull/575#discussion_r1135747714_

ramvikrams:
- [x] Closes #576 


tests/test_frame.py
```diff
@@ -2521,3 +2521,9 @@ def test_align() -> None:
     aligned_df0, aligned_df1 = df0.align(df1)
     check(assert_type(aligned_df0, pd.DataFrame), pd.DataFrame)
     check(assert_type(aligned_df1, pd.DataFrame), pd.DataFrame)
+
+
+def test_loc_new() -> None:
```
Dr-Irv, line 2526: Can you name this as `test_loc_returns_series()`

pandas-stubs/core/frame.pyi
```diff
@@ -188,6 +193,8 @@ class _LocIndexerFrame(_LocIndexer):
         | None,
     ) -> Series: ...
     @overload
+    def __getitem__(self, idx: tuple[ScalarT, slice]) -> Series | DataFrame: ...
```
Dr-Irv, line 196: I think this will work with `Scalar` instead of `ScalarT`.  Can you try that?

Dr-Irv (changes requested):

tests/test_frame.py
ramvikrams: done sir

pandas-stubs/core/frame.pyi
ramvikrams: done sir

Dr-Irv (approved):
thanks @ramvikrams

PR #584, opened 2023-03-16
CI: nightly builds from additional index

### LINKED ISSUE #579: nightly failing because it cannot find tzdata
See https://github.com/pandas-dev/pandas-stubs/actions/runs/4438156563/jobs/7789166798

nightly run is failing because it can't find `tzdata >= 22.1`, but that doesn't make sense as version 22.7 is available.

@twoertwein can you take a look?

twoertwein:
- [ ] Closes #579 (Replace xxxx with the Github issue number)

Seems to work for me locally. I think the issue is that the nightly index overwrote the normal pypi index so it didn't find any other packages.

### DIFF
--- a/scripts/test/run.py
+++ b/scripts/test/run.py
@@ -86,9 +86,10 @@ def nightly_pandas():
         "-m",
         "pip",
         "install",
+        "--pre",
         "--use-deprecated=legacy-resolver",
         "--upgrade",
-        "--index-url",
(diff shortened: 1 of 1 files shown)


Dr-Irv (approved):
Thanks @twoertwein .  Let's see what happens when I merge!

Dr-Irv:
Looks like it worked.

PR #587, opened 2023-03-22
Support mypy 1.1.1 and pyright 1.1.300

Dr-Irv:
Updating `mypy` and `pyright` versions.  Required a change to a test that `pyright` now correctly picks up.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -35,10 +35,10 @@ python = ">=3.8,<3.12"
 types-pytz = ">= 2022.1.1"
 
 [tool.poetry.dev-dependencies]
-mypy = "1.0"
+mypy = "1.1.1"
 pyarrow = ">=10.0.1"
 pytest = ">=7.1.2"
-pyright = "1.1.298"
+pyright = ">= 1.1.300"
 poethepoet = ">=0.16.5"
 loguru = ">=0.6.0"
 pandas = "1.5.3"
--- a/tests/test_frame.py
+++ b/tests/test_frame.py
@@ -87,22 +87,22 @@ def test_types_append() -> None:
     df = pd.DataFrame(data={"col1": [1, 2], "col2": [3, 4]})
     df2 = pd.DataFrame({"col1": [10, 20], "col2": [30, 40]})
     if TYPE_CHECKING_INVALID_USAGE:
-        res1: pd.DataFrame = df.append(df2)  # type: ignore[operator]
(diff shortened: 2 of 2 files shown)


twoertwein (approved):
Great that more and more of the negative tests are working as expected :)

twoertwein:
(pandas) nightly seems to be failing again:

>  FAILED tests/test_frame.py::test_types_window - FutureWarning: Support for axis=1 in DataFrame.expanding is deprecated and will be removed in a future version. Use obj.T.expanding(...) instead
> FAILED tests/test_series.py::test_types_window - FutureWarning: The 'axis' keyword in Series.expanding is deprecated and will be removed in a future version. Call the method without the axis keyword instead.
> FAILED tests/test_series.py::test_types_ewm - FutureWarning: The 'axis' keyword in Series.ewm is deprecated and will be removed in a future version. Call the method without the axis keyword instead.

Dr-Irv:
> (pandas) nightly seems to be failing again:
> 
> > FAILED tests/test_frame.py::test_types_window - FutureWarning: Support for axis=1 in DataFrame.expanding is deprecated and will be removed in a future version. Use obj.T.expanding(...) instead
> > FAILED tests/test_series.py::test_types_window - FutureWarning: The 'axis' keyword in Series.expanding is deprecated and will be removed in a future version. Call the method without the axis keyword instead.
> > FAILED tests/test_series.py::test_types_ewm - FutureWarning: The 'axis' keyword in Series.ewm is deprecated and will be removed in a future version. Call the method without the axis keyword instead.

So the question here is whether we make those changes now (because nightly is post 2.0), or wait until we start supporting 2.0, which is soon to be released.  Opinion?

twoertwein:
I think in the past we removed arguments when the currently released pandas emits warnings about the usage. In the current case, not even 2.0 emits a warning (didn't test it) so I'm not sure what the best way forward is.

Generally, I'm always happy to remove "soon"-to-be deprecated arguments as it makes the annotations and testing easier :) If we want to stick with the current "policy", we should guard these tests behind a pandas<2.1.

Dr-Irv:
> Generally, I'm always happy to remove "soon"-to-be deprecated arguments as it makes the annotations and testing easier :) If we want to stick with the current "policy", we should guard these tests behind a pandas<2.1.

I think that's the best solution.  Once 2.0 is released, and we start fixing the stubs with all the 2.0 features, we will remove the test code and tests for pandas <= 1.5, so this mirrors that practice.

PR #589, opened 2023-03-24
gh-583:added tolist()

### LINKED ISSUE #583: pd.array().tolist() is not recognized
**Describe the bug**

`.tolist()` is not recognized as a method for a pandas array. You can compare with generating the stubs using `stubgen -p pandas` which does provide it in `core/arrays/base.pyi`:

```
class ExtensionArray:
  ....
  def tolist(self) -> list: ...
``` 

Should pandas-stub support this? Could be something wrong with my typechecker, but I don't see `tolist` in the corresponding pandas-stubs' pyi.

**To Reproduce**

```
import pandas as pd
return pd.array([1,2,3]).tolist()
```

image

 - python version: 3.10.8
 - version of type checker: [fork of pyright](https://github.com/microbit-foundation/pyright). Should be compatible with pyright 1.1.279
 - version of installed `pandas-stubs`: 1.5.3.230304

ramvikrams:
- [x] Closes #583 


tests/test_pandas.py
```diff
@@ -1973,3 +1973,10 @@ def g(x: pd.Series) -> int:
         ),
         pd.DataFrame,
     )
+
+
+def test_tolist() -> None:
+    data = {"Courses": "pandas", "Fees": 20000, "Duration": "30days"}
+    s = pd.Series(data)
+    listObj = s.tolist()
```
twoertwein, line 1981: I think you meant to write `s.array.tolist()`

Could also have it one the same line `check(assert_type(s.array.tolist(), list), list)`

ramvikrams: Done

twoertwein (approved):
@Dr-Irv

tests/test_pandas.py
```diff
@@ -1973,3 +1973,9 @@ def g(x: pd.Series) -> int:
         ),
         pd.DataFrame,
     )
+
+
+def test_tolist() -> None:
+    data = {"Courses": "pandas", "Fees": 20000, "Duration": "30days"}
+    s = pd.Series(data)
+    check(assert_type(s.array.tolist(), list), list)
```
Dr-Irv, line 1981: Can you add a test for the original issue:  `pd.array([1,2,3]).tolist()`

Dr-Irv (changes requested):

tests/test_pandas.py
ramvikrams: Done

Dr-Irv: You didn't do what I asked.  I should see this snippet in the test:
```python
pd.array([1,2,3]).tolist()
```

Dr-Irv (changes requested):

tests/test_pandas.py
ramvikrams: actually we have to give it some positional argument

Dr-Irv: You still have not added the test  `pd.array([1,2,3]).tolist()`.  The test you had before was correct when you wrote `s.array().tolist()`.  But you need to do BOTH tests.

Dr-Irv (changes requested):

tests/test_pandas.py
ramvikrams: `check(assert_type(pd.array([1, 2, 3]).tolist(), list), list)` i have added thhis test but with warning as we need a positional argument but in this ex there is no positional argument

Dr-Irv: > Any idea how we should change the func so that the test runs

The issue here is that `pd.array()` is returning the wrong type.

If you go into `pandas-stubs\core\construction.pyi`, and change the return type of `array()` from `ABCExtensionArray` to `ExtensionArray` (and fix the imports, etc.), then that should fix the problem.

ramvikrams: done

Dr-Irv (approved):
thanks @ramvikrams

PR #596, opened 2023-03-29
add Hashable type to __get_item__ #592

### LINKED ISSUE #592: BUG: __get_item__ accepts more types than it's type signature says
### Pandas version checks

- [X] I have checked that this issue has not already been reported.

- [X] I have confirmed this bug exists on the [latest version](https://pandas.pydata.org/docs/whatsnew/index.html) of pandas.

- [X] I have confirmed this bug exists on the [main branch](https://pandas.pydata.org/docs/dev/getting_started/install.html#installing-the-development-version-of-pandas) of pandas.

### Reproducible Example

```python
from enum import Enum
import pandas as pd

class MyEnum(Enum):
    TAYYAR = "tayyar"
    HAYDAR = "haydar"

df = pd.DataFrame(data = [[12.2, 10], [8.8, 15], [22.1, 14]], columns=[MyEnum.TAYYAR, MyEnum.HAYDAR])
print(df[MyEnum.TAYYAR])
```

### Issue Description

The example code above takes an Enum as column to a DataFrame and uses that enum in `__get_item__`.

It works fine, which creates an inconsistency with the signature of  `__get_item__` method.

Here is the signature of `__get_item__`. According to this signature, enums are not accepted as index.
```python
def __getitem__(self, Union[str, bytes, date, datetime, timedelta, bool, int, float, complex, Timestamp, Timedelta], /) -> Series[Any]
```

When you use a static type checker e.g. mypy on the code example above, it fails.

```
mypy_enum.py:12: error: No overload variant of "__getitem__" of "DataFrame" matches argument type "MyEnum"  [call-overload]
mypy_enum.py:12: note: Possible overload variants:
mypy_enum.py:12: note:     def __getitem__(self, Union[str, bytes, date, datetime, timedelta, bool, int, float, complex, Timestamp, Timedelta], /) -> Series[Any]
mypy_enum.py:12: note:     def __getitem__(self, slice, /) -> DataFrame
mypy_enum.py:12: note:     def [ScalarT] __getitem__(self, Union[Tuple[Any, ...], Series[bool], DataFrame, List[str], List[ScalarT], Index, ndarray[Any, dtype[str_]], ndarray[Any, dtype[bool_]], Sequence[Tuple[Union[str, bytes, date, datetime, timedelta, bool, int, float, complex, Timestamp, Timedelta], ...]]], /) -> DataFrame
```

More info: https://stackoverflow.com/questions/75614405/mypy-indexing-pd-dataframe-with-an-enum-raises-no-overload-variant-error

### Expected Behavior

`__getitem__`'s signature should be updated to support the column type itself. The column type is often not limited to primitive types. In can be custom objects as well.

### Installed Versions

`assert '_distutils' in core.__file__, core.__file__`

anilbey:
- [x] Closes #592 (Replace xxxx with the Github issue number)


tests/test_frame.py
```diff
@@ -160,6 +161,17 @@ def test_types_getitem() -> None:
...
 
+def test_types_getitem_with_hashable() -> None:
+    # Testing getitem support for hashable types that are not scalar
+    # Due to the bug in https://github.com/pandas-dev/pandas-stubs/issues/592
+    class MyEnum(Enum):
+        FIRST = "tayyar"
+        SECOND = "haydar"
+
+    df = pd.DataFrame(data = [[12.2, 10], [8.8, 15]], columns=[MyEnum.FIRST, MyEnum.SECOND])
+    df[MyEnum.FIRST]
```
Dr-Irv, line 172: Can you use `check(assert_type(df[MyEnum.FIRST], pd.Series), pd.Series)`

Dr-Irv (changes requested):

tests/test_frame.py
anilbey: Thanks I just added it.

pandas-stubs/core/frame.pyi
```diff
@@ -516,7 +516,7 @@ class DataFrame(NDFrame, OpsMixin):
     def T(self) -> DataFrame: ...
     def __getattr__(self, name: str) -> Series: ...
     @overload
-    def __getitem__(self, idx: Scalar | tuple[Hashable, ...]) -> Series: ...
+    def __getitem__(self, idx: Scalar | Hashable | tuple[Hashable, ...]) -> Series: ...
```
twoertwein, line 519: Could reduce it to `def __getitem__(self, idx: Scalar | Hashable) -> Series: ...` as a `tuple` is declared as hashable in typeshed.

(Many members of `Scalar` are probably also hashable - not sure whether all of them are hashable.)

anilbey: Thanks I just removed `tuple[Hashable, ...]` from the signature.

anilbey: Similarly, should we include a `list[Hashable]` in the 3rd overload?

The 3rd overload:
```python

    @overload
    def __getitem__(
        self,
        idx: Series[_bool]
        | DataFrame
        | Index
        | np_ndarray_str
        | np_ndarray_bool
        | list[_ScalarOrTupleT],
    ) -> DataFrame: ...
```

anilbey: This is actually more complicated than I thought.

> Do we have tests for __getitem__ and slice? 

Yes, there is a slice test, but it is without assert types. 
 https://github.com/pandas-dev/pandas-stubs/blob/fbc5f20d35c0d031fc30dfe2cd8e1dfb21c8610a/tests/test_frame.py#L156

anilbey: I just added more assert_types in the test to make sure the other overloads are still working fine.

Dr-Irv (approved):
thanks @anilbey

anilbey:
I'm happy to help, thanks @Dr-Irv , @twoertwein for the quick reply and the feedback.

PR #603, opened 2023-03-30
BUG: Fix MultiIndex.dtypes (#597)

### LINKED ISSUE #597: MultiIndex stub lacks dtypes property
The stub from `MultiIndex` is missing the `dtypes` property.

**To Reproduce**

Runing `mypy` on a file with the following contents

```python
import pandas as pd

pd.MultiIndex.from_tuples([(1, 2.), (2, 3.)], names=['foo', 'bar']).dtypes
```

produces the error message ` error: "MultiIndex" has no attribute "dtypes"; maybe "dtype"?  [attr-defined]`, although `MultiIndex.dtypes` [is documented](https://pandas.pydata.org/docs/reference/api/pandas.MultiIndex.dtypes.html).

 - OS: Ubuntu 2020.04
 - python version: 3.11
 - mypy 1.1.1
 - pandas-stubs 1.5.3.230321

johannes-mueller:
- [x] Closes #597 

### DIFF
--- a/pandas-stubs/core/indexes/multi.pyi
+++ b/pandas-stubs/core/indexes/multi.pyi
@@ -4,7 +4,6 @@ from collections.abc import (
     Sequence,
 )
 from typing import (
-    Any,
     Literal,
     overload,
 )
@@ -16,6 +15,7 @@ from typing_extensions import Self
 
 from pandas._typing import (
     T1,
+    Dtype,
     DtypeArg,
     HashableT,
     np_ndarray_anyint,
@@ -70,7 +70,7 @@ class MultiIndex(Index):
     @property
     def dtype(self) -> np.dtype: ...
     @property
-    def dtypes(self) -> pd.Series[Any]: ...
+    def dtypes(self) -> pd.Series[Dtype]: ...
     def memory_usage(self, deep: bool = ...) -> int: ...
     @property
     def nbytes(self) -> int: ...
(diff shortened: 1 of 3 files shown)


pandas-stubs/core/indexes/multi.pyi
```diff
@@ -68,6 +69,8 @@ class MultiIndex(Index):
     def __contains__(self, key) -> bool: ...
     @property
     def dtype(self) -> np.dtype: ...
+    @property
+    def dtypes(self) -> pd.Series[Any]: ...
```
twoertwein, line 73: Could probably be tightened to `pd.Series[numpy.dtype | pd.api.extensions.ExtensionDtype]` but that probably causes trouble (I don't think those classes are part of S1)

johannes-mueller: With `pd.Series[numpy.dtype | pd.api.extensions.ExtensionDtype]` mypy says

```
pandas-stubs/core/indexes/multi.pyi:73: error: Invalid type argument value for "Series"  [type-var]
tests/test_indexes.py:862: error: Invalid type argument value for "Series"  [type-var]
```

Dr-Irv, line 73: When we can't infer the type of the `Series`, we just return `pd.Series`.  So drop the `Any` part.

johannes-mueller:
> You could try modifying `S1` to include `Dtype`, but if that doesn't work, then just have `dtypes` return `pd.Series`.

It seems to work in the sense that `poe test_all` passes in my setup.

Dr-Irv (approved):
Thanks @johannes-mueller

PR #604, opened 2023-03-30
Fix result type of `Series.sum(...)`

### LINKED ISSUE #595: `Series[bool].sum()` incorrectly returns `bool` instead of `int`
**Describe the bug**
As the title.

**To Reproduce**
```python
from pandas import Series

series = assert_type(Series([False, True] * 5, dtype=bool), "Series[bool]")
sum = assert_type(series.sum(), int)  # Fails.
assert sum == 5
```

pyright

image

 - OS: Windows
 - OS Version: 10
 - python version: 3.10.10
 - version of type checker: Pyright v1.1.300 (bundled in Pylance v2023.3.31)
 - version of installed `pandas-stubs`: 1.5.3.230321

**Additional context**
None.

Azureblade3808:
- [x] Closes #595
- [x] Tests added

### DIFF
--- a/pandas-stubs/core/series.pyi
+++ b/pandas-stubs/core/series.pyi
@@ -70,7 +70,10 @@ from pandas.core.window.rolling import (
     Rolling,
     Window,
 )
-from typing_extensions import TypeAlias
+from typing_extensions import (
+    Never,
+    TypeAlias,
+)
 import xarray as xr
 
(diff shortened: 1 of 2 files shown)


Dr-Irv (approved):
Thanks @Azureblade3808 .   That's a nice trick using `Series[Never]`

PR #608, opened 2023-03-31
gh-594: added upper bound

### LINKED ISSUE #594: Nightly failing with develop code
See https://github.com/pandas-dev/pandas-stubs/actions/runs/4546265885/jobs/8014680487

Seems like behavior changes from 2.0 to 2.1, moving where the warning no longer appears

So add an upper bound in https://github.com/pandas-dev/pandas-stubs/blob/fbc5f20d35c0d031fc30dfe2cd8e1dfb21c8610a/tests/test_timefuncs.py#L323
to add `upper="2.0.99"`

ramvikrams:
- [x] Closes #594 


Dr-Irv (approved):
I will merge this, and we will see if now `nightly` passes after the merge.  Thanks @ramvikrams

Dr-Irv:
So the good news is that this PR did close #594, but now we have a new nightly problem.
Created #610 for that.

PR #616, opened 2023-04-01
Fix mypy nightly

Dr-Irv:
I noticed that we had two references to `Series[Timestamp]` and `Series[Timedelta]` in methods, and those methods were `__sub__()`, so I removed them from `Series` and placed them in `TimestampSeries`, and I think this fixes the mypy nightly problem.  

Tested mypy nightly locally and it worked.  So if this passes CI, then mypy_nightly could be tested by doing a merge.

### DIFF
--- a/pandas-stubs/core/series.pyi
+++ b/pandas-stubs/core/series.pyi
@@ -1446,16 +1446,6 @@ class Series(IndexOpsMixin, NDFrame, Generic[S1]):
         self, other: Timestamp | datetime | TimestampSeries
     ) -> TimedeltaSeries: ...
     @overload
-    def __sub__(
-        self: Series[Timestamp],
-        other: Timedelta | TimedeltaSeries | TimedeltaIndex | np.timedelta64,
-    ) -> TimestampSeries: ...
-    @overload
-    def __sub__(
-        self: Series[Timedelta],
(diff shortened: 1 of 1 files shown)


pandas-stubs/core/series.pyi
```diff
@@ -1904,6 +1894,15 @@ class TimestampSeries(Series[Timestamp]):
...
     def __add__(self, other: TimedeltaSeries | np.timedelta64) -> TimestampSeries: ...  # type: ignore[override]
     def __radd__(self, other: TimedeltaSeries | np.timedelta64) -> TimestampSeries: ...  # type: ignore[override]
+    @overload  # type: ignore[override]
+    def __sub__(
+        self, other: Timestamp | datetime | TimestampSeries
+    ) -> TimedeltaSeries: ...
+    @overload
+    def __sub__(
+        self,
+        other: Timedelta | TimedeltaSeries | TimedeltaIndex | np.timedelta64,
```
twoertwein, line 1904: Intuitively I would expect that `Timedelta - Timedelta` is still a `Timedelta`? Might be good to have a test for this as this overload changes this behavior.

twoertwein: Oh, sorry I didn't realize this overload is in TimestampSeries

twoertwein (approved):

twoertwein:
Thanks @Dr-Irv

PR #617, opened 2023-04-01
updated message- PR 611

ramvikrams:


Dr-Irv (approved):
Hoping this will work.

Dr-Irv:
Now the message is right.  I think the lower bound needs to be "2.0.99", not "2.1", based on reading our own docs.  Can you try that @ramvikrams

ramvikrams:
> Now the message is right.  I think the lower bound needs to be "2.0.99", not "2.1", based on reading our own docs.  Can you try that @ramvikrams 

Yes

PR #619, opened 2023-04-02
updated the lower bound PR 611

ramvikrams:


Dr-Irv (approved):
Let's try with the lower bound

Dr-Irv:
We are getting closer.  It did test the warnings, but they didn't match because of the parentheses in the string.  So use this string: `r"Logical ops \(and, or, xor\) between Pandas objects and` .  Note the `r` at the beginning and the `\` inside.

ramvikrams:
> Let's try with the lower bound

Yes

PR #622, opened 2023-04-03
gh:552 - Corrected Series.mean return type

### LINKED ISSUE #552: Series.mean() does not return an float
**Describe the bug**
`Series.mean()` [is typed as float](https://github.com/pandas-dev/pandas-stubs/blob/main/pandas-stubs/core/series.pyi#L1533), but actually returns a `numpy.float64`.

This errors because I need to call `.item()` to get the actual float.

**To Reproduce**
```py
# pyright: strict
import pandas as pd

mean = pd.Series([1, 2, 3, 4]).mean()
print(type(mean)) # numpy.float64 instead of float
mean.item() # Cannot access member "item" for type "float"  Member "item" is unknown Pylance (reportGeneralTypeIssues)
```

 - OS: Linux
 - OS Version 20.04
 - python 3.8.10
 - version of type checker: 1.1.293
 - version of installed `pandas-stubs`: Pylance v2023.2.31

**Additional context**
I must have the actual float for `yaml.safe_dump`.

ramvikrams:
- [x] Closes #552

### DIFF
--- a/pandas-stubs/core/resample.pyi
+++ b/pandas-stubs/core/resample.pyi
@@ -88,8 +88,7 @@ class Resampler(BaseGroupBy, Generic[NDFrameT]):
         func: Callable[..., DataFrame]
         | tuple[Callable[..., DataFrame], str]
         | Callable[..., Series]
-        | tuple[Callable[..., Series], str]
-        | Callable[[DataFrame], np.float64],
+        | tuple[Callable[..., Series], str],
         *args,
         **kwargs,
     ) -> DataFrame: ...
--- a/pandas-stubs/core/resample.pyi
+++ b/pandas-stubs/core/resample.pyi
@@ -41,7 +41,10 @@ _FrameGroupByFuncArgs: TypeAlias = (
 )
 
 _SeriesGroupByFunc: TypeAlias = (
-    Callable[[Series], Scalar] | Callable[[Series], Series] | np.ufunc
+    Callable[[Series], Scalar]
+    | Callable[[Series], Series]
+    | Callable[[Series], np.float64]
+    | np.ufunc
 )
 _SeriesGroupByFuncTypes: TypeAlias = (
     _SeriesGroupByFunc | str | list[_SeriesGroupByFunc | str]
@@ -85,7 +88,8 @@ class Resampler(BaseGroupBy, Generic[NDFrameT]):
         func: Callable[..., DataFrame]
         | tuple[Callable[..., DataFrame], str]
         | Callable[..., Series]
-        | tuple[Callable[..., Series], str],
+        | tuple[Callable[..., Series], str]
+        | Callable[[DataFrame], np.float64],
         *args,
         **kwargs,
     ) -> DataFrame: ...
--- a/pandas-stubs/core/resample.pyi
+++ b/pandas-stubs/core/resample.pyi
@@ -41,10 +41,7 @@ _FrameGroupByFuncArgs: TypeAlias = (
 )
 
 _SeriesGroupByFunc: TypeAlias = (
-    Callable[[Series], Scalar]
-    | Callable[[Series], Series]
-    | Callable[[Series], np.float64]
-    | np.ufunc
+    Callable[[Series], Scalar] | Callable[[Series], Series] | np.ufunc
 )
 _SeriesGroupByFuncTypes: TypeAlias = (
     _SeriesGroupByFunc | str | list[_SeriesGroupByFunc | str]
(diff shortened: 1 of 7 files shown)


ramvikrams:
Now `MyPy` will fail because of tests in files  `test_windowing`,`test_resampler`,`test_frame` for which I think either we can suppress the warning or changing the DataFrame.mean() method

tests/test_series.py
```diff
@@ -310,10 +310,10 @@ def test_types_rank() -> None:
...
 def test_types_mean() -> None:
     s = pd.Series([1, 2, 3, np.nan])
-    f1: float = s.mean()
-    s1: pd.Series = s.groupby(level=0).mean()
-    f2: float = s.mean(skipna=False)
-    f3: float = s.mean(numeric_only=False)
+    check(assert_type(s.mean(), np.float64), np.float64)
+    s1 = s.groupby(level=0)
+    check(assert_type(s1.mean(skipna=False), pd.Series), np.float64)
+    check(assert_type(s1.mean(numeric_only=False), pd.Series), np.float64)
```
Dr-Irv, line 316: above two tests should be using `s.mean()` and checking result against `np.float64`

Dr-Irv (changes requested):
- For the errors in `test_windowing.py`, change the return types of `_mean()` to `np.float64`
- For the errors in `test_resampler.py`, change the return types of the functions `h()` in line 161 and `f()` in line 251 to `np.float.64`
- For the errors in `test_frame.py`, change the return types of these functions to `np.float64`:
  -  `sum_mean()` in line 2046
  - `lfunc` should have type `Callable[[pd.DataFrame], np.float64]` in line 2050

ramvikrams:
> * For the errors in `test_windowing.py`, change the return types of `_mean()` to `np.float64`
> * For the errors in `test_resampler.py`, change the return types of the functions `h()` in line 161 and `f()` in line 251 to `np.float.64`
Yes, I did that in my local system aready but the problem is after doing this change, but the problem is in `h()` has incompatible types and then same in `f()` 
> * For the errors in `test_frame.py`, change the return types of these functions to `np.float64`:
>  
>   * `sum_mean()` in line 2046
>   * `lfunc` should have type `Callable[[pd.DataFrame], np.float64]` in line 2050
Same for both of these 2 also they had incompatible types

ramvikrams:
> You may have to make changes to other parts of the tests to make this work, since the tests assumed `float` was the result of `Series.mean()`, and now you are changing it to `np.float64`.  Feel free to do that, and I will review.
> 
Thanks

ramvikrams:
Sir I actually what should we write  the `assert_type` when it is `unknown`. Means the assert_type is Unknown itself.

Dr-Irv:
> Sir I actually what should we write the `assert_type` when it is `unknown`. Means the assert_type is Unknown itself.

If the results are coming back as `Unknown` from the type checker, it means there is a bug somewhere in the stubs that you'll have to figure out.

ramvikrams:
> If the results are coming back as `Unknown` from the type checker, it means there is a bug somewhere in the stubs that you'll have to figure out.

thanks i'll do it

ramvikrams:
The `test_resampler` errors are now gone but could you suggest a fix for `test_frame` I could not get it

Dr-Irv:
> The `test_resampler` errors are now gone but could you suggest a fix for `test_frame` I could not get it

Can you try the following?

- Add `np.integer | np.float` to the type `IndexIterScalar` in `_typing.pyi` . 

If that creates new typing errors, then change the callables for `groupby.apply()` that declare `Scalar` as a return type to also allow `np.integer | np.float` as a return type.

ramvikrams:
Done Sir

pandas-stubs/core/resample.pyi
```diff
@@ -41,7 +41,10 @@ _FrameGroupByFuncArgs: TypeAlias = (
 )
 
 _SeriesGroupByFunc: TypeAlias = (
-    Callable[[Series], Scalar] | Callable[[Series], Series] | np.ufunc
+    Callable[[Series], Scalar]
+    | Callable[[Series], Series]
+    | Callable[[Series], np.float64]
```
Dr-Irv, line 46: you shouldn't need this change because you added `np.float_` to `IndexIterScalar`

Dr-Irv (changes requested):

pandas-stubs/core/resample.pyi
ramvikrams: done

Issue #624, opened 2023-04-03
Version 2.0 Compatibility Tracker


Dr-Irv:
This is the list of things that are in pandas 2.0 release notes that need to be addressed in pandas-stubs.  PR's welcome.  If you do a PR, check off the item and put a link to the PR that closed it.  One PR can address multiple issues.

Some of these may already have been taken care of, so if so, check them off and indicate with a comment such as "previously complete"

- [x] Int64Index, UInt64Index & Float64Index were deprecated in pandas version 1.4 and have now been removed. Instead [Index](https://pandas.pydata.org/docs/reference/api/pandas.Index.html#pandas.Index) should be used directly, and can it now take all numpy numeric dtypes, i.e. int8/ int16/int32/int64/uint8/uint16/uint32/uint64/float32/float64 dtypes:  #626
- [x] The various numeric datetime attributes of [DatetimeIndex](https://pandas.pydata.org/docs/reference/api/pandas.DatetimeIndex.html#pandas.DatetimeIndex) ([day](https://pandas.pydata.org/docs/reference/api/pandas.DatetimeIndex.day.html#pandas.DatetimeIndex.day), [month](https://pandas.pydata.org/docs/reference/api/pandas.DatetimeIndex.month.html#pandas.DatetimeIndex.month), [year](https://pandas.pydata.org/docs/reference/api/pandas.DatetimeIndex.year.html#pandas.DatetimeIndex.year) etc.) were previously in of dtype int64, while they were int32 for [arrays.DatetimeArray](https://pandas.pydata.org/docs/reference/api/pandas.arrays.DatetimeArray.html#pandas.arrays.DatetimeArray). They are now int32 on [DatetimeIndex](https://pandas.pydata.org/docs/reference/api/pandas.DatetimeIndex.html#pandas.DatetimeIndex) also  #632
- [x] Level dtypes on Indexes from [Series.sparse.from_coo()](https://pandas.pydata.org/docs/reference/api/pandas.Series.sparse.from_coo.html#pandas.Series.sparse.from_coo) are now of dtype int32, the same as they are on the rows/cols on a scipy sparse matrix. Previously they were of dtype int64  #632
[...]

PR #626, opened 2023-04-04
remove 1.5 tests and types

Dr-Irv:
Tested with pandas 2.0.  Removed references to pandas 1.5, and removed types (`NumericIndex` related), creating special types for typing so we can track `Index` with `int` and `Index` with `float`

Updated the version number, but don't intend to do a release until we handle issues in #624


twoertwein (approved):
Personally, I wouldn't mind removing the typing-only classes (less perceive typing but more maintanable).

twoertwein:
Thanks @Dr-Irv !

Dr-Irv:
> Personally, I wouldn't mind removing the typing-only classes (less perceive typing but more maintanable).

I started doing that.  But we have methods that in 1.5 were dependent on `Int64Index` and `Float64Index`.  If I switch to `Index`, then it creates all sorts of overload problems. 

This is one of these cases where removing `NumericIndex` and its subclasses was not good from a typing perspective.

The `_IntIndexType` and `_FloatIndexType` classes are a step towards a generic index implementation, similar to how we have `TimeStampSeries`, `TimeDeltaSeries`, etc.

PR #628, opened 2023-04-04
Remove cap on python version

### LINKED ISSUE #625: Please remove cap on python version in pyproject.toml
pandas-stubs's current upper bound on python version blocks every downstream package that also uses poetry from specifying python <4. 

Note that this is a separate issue from actually supporting python 3.12; pandas 2.0 does not yet support python 3.12, so actual support of python 3.12 isn't possible at this time. But the persnickety solver that pandas uses means that downstream users like me will have to wait on you to uncap your python version. By that time, though, many of us will have moved on to other things, while our packages will stop building even though they may work perfectly well under 3.12.

This issue is extensively discussed in Henry Schreiner's [blog post on poetry's decisions on capping](https://iscinumpy.dev/post/bound-version-constraints/). Unless you can affirmatively assert that pandas-stubs will not work on 3.12 at the time that pandas supports 3.12, you shouldn't dismiss all us downstream users in our optimism and force us to update our pyprojects.toml at some future time.  As someone pointed out in a [closed poetry issue](https://github.com/python-poetry/poetry/issues/2731) on this topic, it took 4 years to force a much-used dependency (isort) to change a single character misreporting its compatibility. 

Yes, this is really poetry's problem, not yours. Until they change things (and they have refused), the proper solution is to use a different build tool. But capping the python version contributes to a growing dependency hell.

twoertwein:
- [ ] Closes #625 (Replace xxxx with the Github issue number)

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -31,7 +31,7 @@ packages = [
 "Documentation" = "https://pandas.pydata.org/pandas-docs/stable"
 
 [tool.poetry.dependencies]
-python = ">=3.8,<3.12"
+python = ">=3.8"
 types-pytz = ">= 2022.1.1"
 
 [tool.poetry.dev-dependencies]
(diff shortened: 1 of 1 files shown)


pyproject.toml
```diff
@@ -49,7 +49,7 @@ pre-commit = ">=2.19.0"
 black = ">=22.12.0"
 isort = ">=5.10.1"
 openpyxl = ">=3.0.10"
-tables = { version = ">=3.7.0" }
+tables = { version = ">=3.7.0" , python = "<4"}  # 3.8.0 depends on blosc2 which caps python to <4
```
Dr-Irv, line 52: That's a dev dependency, so we should be OK.

Dr-Irv (approved):
Thanks @twoertwein

PR #631, opened 2023-04-04
update signature of Series.pop #627

### LINKED ISSUE #627: BUG: series.pop accepts more types than it's type signature says
**Describe the bug**
Similar to https://github.com/pandas-dev/pandas-stubs/issues/592, Series.pop also works on Hashable types but the signature does not mention them:

```python
def pop(self, item: _str) -> Series[S1]: ...
```

**To Reproduce**

Here is the code. Running it works fine.

```python
from enum import Enum
import pandas as pd

class MyEnum(Enum):
    TAYYAR = "tayyar"
    HAYDAR = "haydar"

df = pd.DataFrame(data = [[12.2, 10], [8.8, 15], [22.1, 14]], columns=[MyEnum.TAYYAR, MyEnum.HAYDAR])
series = df.loc[0]
print(series.pop(MyEnum.TAYYAR))

```

 Calling it with mypy gives the following error.
```
error: Argument 1 to "pop" of "Series" has incompatible type "MyEnum"; expected "str"  [arg-type]
```

 - OS: linux
 - python version: 3.9
 - version of type checker: mypy                 1.1.1
 - version of installed `pandas-stubs`: pandas-stubs         1.5.3.230321

anilbey:
Updates signature of Series.pop, adds test. The signature follows the corresponding signature in pandas: https://github.com/pandas-dev/pandas/blob/main/pandas/core/series.py#L5014

- [X] Closes #627


pandas-stubs/core/series.pyi
```diff
@@ -1021,7 +1021,7 @@ class Series(IndexOpsMixin, NDFrame, Generic[S1]):
     def droplevel(
         self, level: Level | list[Level], axis: AxisIndex = ...
     ) -> DataFrame: ...
-    def pop(self, item: _str) -> Series[S1]: ...
+    def pop(self, item: Hashable) -> Any: ...
```
Dr-Irv, line 1024: Change the result to `S1`, not `Any`

tests/test_series.py
```diff
@@ -226,6 +227,21 @@ def test_types_dropna() -> None:
...
+    class MyEnum(Enum):
+        FIRST = "tayyar"
+        SECOND = "haydar"
+
+    df = pd.DataFrame(
+        data=[[12.2, 10], [8.8, 15]], columns=[MyEnum.FIRST, MyEnum.SECOND]
+    )
+    series = df.loc[0]
+    res = series.pop(MyEnum.FIRST)
+    check(assert_type(res, Any), float)
```
Dr-Irv, line 242: Can you make this test only involve `Series`.  Something like:
```python
s = pd.Series([3.2, 4.3], index=[MyEnum.FIRST, MyEnum.SECOND], dtype=float)
res = s.pop(MyEnum.FIRST)
check(assert_type(res, float), np.float64)
```

Dr-Irv (changes requested):

tests/test_series.py
anilbey: Thanks, it is done in the next commit.

pandas-stubs/core/series.pyi
Dr-Irv: > Do you know what might be wrong here?

Can you remove `pop()` from `core/generic.pyi` ?   That should fix it.  I'll review your other changes once you make that change.  Thanks.

anilbey: Thanks, I just removed it. Tests pass now (at least locally).

tests/test_series.py
```diff
@@ -226,6 +227,21 @@ def test_types_dropna() -> None:
...
+    class MyEnum(Enum):
+        FIRST = "tayyar"
+        SECOND = "haydar"
+
+    s = pd.Series([3.2, 4.3], index=[MyEnum.FIRST, MyEnum.SECOND], dtype=float)
+    res = s.pop(MyEnum.FIRST)
+    check(assert_type(res, float), np.float64)
+
+    s2 = pd.Series([3, 5], index=["alibaba", "zuhuratbaba"], dtype=int)
+    check(assert_type(s2.pop("alibaba"), int), np.int64)
```
Dr-Irv, line 242:
```suggestion
    check(assert_type(s2.pop("alibaba"), int), np.int_)
```
You'll need to use `np.int_` so it passes on Windows.

Dr-Irv (changes requested):

tests/test_series.py
anilbey: Thanks, I didn't know. I just updated it.

Dr-Irv (approved):
Thanks @anilbey

PR #630, opened 2023-04-04
Broaden read csv param types

EFord36:
- [X] Closes #605, Closes #623
- [X] Tests added: Please use `assert_type()` to assert the type of any return value

Submitted as 1 PR for the 2 issues because the relevant code is so close that I thought it might be quicker to review together, plus trying to have both as separate PRs would likely result in merge conflicts if/when one was merged before the other. The commits addressing the two issues are separate though, so it would be easy for me to split into 2 separate PR if this is preferred.

As well as read_csv, modify read_table, read_clipboard, read_excel, which all have the relevant params (I searched for other functions but let me know if I missed any).

Closes #605 in a different way than suggested in that ticket, in order to allow broader types than just lists of string or ints, since list-likes are allowed here. Naming of the type alias UsecolsArgType was inspired by ColspaceArgType, happy to change if desired.

Please let me know if coverage of more different types for `names` and/or `usecols` is needed in the tests, or if there are stylistic details you would like me to change. There is a lot of repetition between the tests for the different functions that have these params - I'm happy to work on reducing it, but I thought it might mean restructuring some of the existing tests, so I didn't want to fiddle with it too much until I knew if it (and the actual functional change) was desired.

Passes tests (`poetry run poe test_all`) locally when the last commit before my changes is `7730abc` - the two commits after that (`6819e32` and `9c301b3`) result in test failures for me locally that seem unrelated to my change, and I get if I just run on `main` with no changes.

PR #634, opened 2023-04-05
Cleanup philosophy doc

EFord36:
- no issue closed (I could open one for the type, or others for the other changes, but it seems a small enough change to not be worth it - happy to if you would prefer though!
- no tests added - docs-only change.

While working on #630 and reading docs/philosophy, I noticed a 'too' typo for 'to', which I thought I'd fix (in the first commit in the PR).

While making that change though, I noticed a couple of inconsistencies in the same paragraph:

1. In the number of spaces after a '.' (full stop/period) at the end of a sentence - normally 2 spaces, but sometimes 1.
2. Some lines have trailing whitespace (while most don't)

I fixed these for the paragraph, but also the rest of the document (using the regex `\. ([^ ])` for the whitespace after '.', and `\s$` for the trailing whitespace.

### DIFF
--- a/docs/philosophy.md
+++ b/docs/philosophy.md
@@ -87,9 +87,9 @@ not too narrow and not too wide.  A type annotation to an argument to a function
 method is too narrow if it disallows valid arguments.  A type annotation to
 an argument to a function or method is too wide if
 it allows invalid arguments.  Testing for type annotations that are too narrow is rather
-straightforward. It is easy to create an example for which the type checker indicates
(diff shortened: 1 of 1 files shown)


Dr-Irv (approved):
Thanks @EFord36 

The "two spaces" after a period is because I learned to type by taking a typing class 45+ years ago on a typewriter, and back in those days, you always added two spaces after a period!  So it's a very hard habit to break.

PR #635, opened 2023-04-05
WeekOfMonth typing is incorrect (resolved)

### LINKED ISSUE #629: WeekOfMonth typing is incorrect
Typing should be what is in `pandas`:
```python
class WeekOfMonth(WeekOfMonthMixin):
    def __init__(
        self, n: int = ..., normalize: bool = ..., week: int = ..., weekday: int = ...
    ) -> None: ...
```
Currently looks like:
https://github.com/pandas-dev/pandas-stubs/blob/9c301b33b746e28401868b788364cecd00a34655/pandas-stubs/_libs/tslibs/offsets.pyi#L182

kacper3123:
- [x] Closes #629 (Replace xxxx with the Github issue number)


kacper3123:
are you talking about something like this?

Screenshot 2023-04-06 175148

Dr-Irv:
> are you talking about something like this?
> 
> Screenshot 2023-04-06 175148

No, I mean a test like the other tests we have for constructors.   

See the existing tests for examples.

PR #637, opened 2023-04-07
Update mypy to 1.2.0

twoertwein:
I'm getting some pyright issues locally. Let's see which version of pyright the CI picks.


Dr-Irv (approved):
Thanks @twoertwein

PR #643, opened 2023-04-12
update README to indicate level of 2.0 support

Dr-Irv:
At today's pandas core dev meeting, we decided to go ahead with releasing a 2.0 version of the stubs that is not fully compatible with pandas 2.0.  This PR updates the README to reflect a message indicating that fact.

Once merged, I will do a release of the stubs.


twoertwein (approved):
Sounds good to me.

Honestly, I would also be fine to simply drop the pandas version from pandas-stubs. We always strive to support only the latest version and guaranteeing to 100% reflect what's possible/intended is something we can never promise.

twoertwein:
Thanks @Dr-Irv !

PR #645, opened 2023-04-13
Cover pandas.tseries.holiday

adamchainz:
I've tried to add typse for everything in this module. I'm not too confident about my overloads or Series/DatetimeIndex usage, please check that!

- [n/a] Closes #xxxx (Replace xxxx with the Github issue number)


pandas-stubs/tseries/holiday.pyi
```diff
@@ -1,3 +1,148 @@
...
+class Holiday:
+    @overload
+    def __init__(
+        self,
+        name: str,
+        year: int | None = ...,
+        month: int | None = ...,
+        day: int | None = ...,
+        offset: BaseOffset | list[BaseOffset] = ...,
+        observance: None = ...,
```
Dr-Irv, line 36: I think `observance` should be `Callable[[datetime], datetime] | None`

pandas-stubs/tseries/holiday.pyi
```diff
@@ -1,3 +1,148 @@
...
+            np.integer | float | str | _date | datetime | np.datetime64 | None
+        ),
+        end_date: (np.integer | float | str | _date | datetime | np.datetime64 | None),
+        return_name: Literal[True] = ...,
+    ) -> Series: ...
+
+holiday_calendars: dict[str, type["AbstractHolidayCalendar"]]
+
+def register(cls: type["AbstractHolidayCalendar"]) -> None: ...
+def get_calendar(name: str) -> "AbstractHolidayCalendar": ...
```
Dr-Irv, line 86: If you put `from __future__ import annotations` at the top of this file, then you can remove the quotes around `"AbstractHolidayCalendar"` in this line and the above, as well as in places below.

pandas-stubs/tseries/holiday.pyi
```diff
@@ -1,3 +1,148 @@
...
+
+class AbstractHolidayCalendar:
+    rules: list[Holiday] = ...
+    start_date: Timestamp = ...
+    end_date: Timestamp = ...
+    _cache: list[tuple[Timestamp, Timestamp, Series[datetime]]] | None = ...
```
Dr-Irv, line 92: remove the `_cache` - it's not public

pandas-stubs/tseries/holiday.pyi
```diff
@@ -1,3 +1,148 @@
...
+        self,
+        other: "AbstractHolidayCalendar" | type["AbstractHolidayCalendar"],
+        inplace: Literal[False] = ...,
+    ) -> list[Holiday]: ...
+    @overload
+    def merge(
+        self,
+        other: "AbstractHolidayCalendar" | type["AbstractHolidayCalendar"],
+        inplace: Literal[True] = ...,
+    ) -> None: ...
```
Dr-Irv, line 130: Same as above - swap the overloads.  First one should be `inplace: Literal[True]`, returning `None`.  Second one is `inplace: bool = ...` returning `list[Holiday]`

Dr-Irv (changes requested):
Thanks for submitting this.  I would suggest setting up the environment for development (see https://github.com/pandas-dev/pandas-stubs/blob/main/docs/setup.md ) so you can test locally and not wait for the CI to identify issues.

Secondly, can you add a set of tests for these, using `check` and `assert_type()`.  Since we don't document these classes in the reference manual, just using the code shown in https://pandas.pydata.org/pandas-docs/stable/user_guide/timeseries.html?highlight=abstractholidaycalendar#holidays-holiday-calendars would be sufficient.  You could create a new file called `test_holidays.py` to accomplish this.

adamchainz (commented):
Thank you for the swift review! Working on fixes and tests now.

Dr-Irv:
ping when green and when you added the tests.

adamchainz:
I started on tests but already hit an error where the doc'd example fails type checking, with a class I'm not touching here:

```
tests/test_holidays.py:25: error: Argument "weekday" to "DateOffset" has incompatible type "weekday"; expected "int"  [arg-type]
```

`DateOffset`'s docs don't say it allows taking a dateutil `weekday` object, but that's what was documented. I'm a bit out of my depth, I'm not sure how much I should expect the pandas docs to be correct/complete...

Dr-Irv:
> `DateOffset`'s docs don't say it allows taking a dateutil `weekday` object, but that's what was documented. I'm a bit out of my depth, I'm not sure how much I should expect the pandas docs to be correct/complete...

The problem is twofold.  One is that the pandas docs aren't documenting that we would accept a `dateutil._common.weekday` type as a value for the `weekday` parameter to `DateOffset`.  The second issue is that the type specification for the `weekday` argument in `pandas-stubs/_libs/tslibs/offsets.pyi` says it is an `int`.  You could fix the latter, but then we have to make `types-python-dateutil` a dependency, which involves modifying `pyproject.toml` and takes some care.

So for your PR, just leave out the `weekday` argument, and I think the tests will then work.

Dr-Irv:
Created an issue in pandas for the docs:  https://github.com/pandas-dev/pandas/issues/52650

Dr-Irv (approved):
thanks @adamchainz

Issue #203, opened 2022-08-17
BUG: df.groupby(col, as_index=False).value_counts() returns a DataFrame but is annotated as Series


phofl:
**Describe the bug**

Using 
```
df = pd.DataFrame({"a": [1, 2, 3], "b": 1})

x: pd.DataFrame = df.groupby("a", as_index=False).value_counts()
```
raises an error, because mypy thinks the return is a Series

```
 Incompatible types in assignment (expression has type "Series[int]", variable has type "DataFrame")  [assignment]
``` 

**To Reproduce**
1. Provide a minimal runnable `pandas` example that is not properly checked by the stubs.
2. Indicate which type checker you are using (`mypy` or  `pyright`).
3. Show the error message received from that type checker while checking your example.

```
df = pd.DataFrame({"a": [1, 2, 3], "b": 1})

x: pd.DataFrame = df.groupby("a", as_index=False).value_counts()
```

In theory this should be ok, but not sure if this ``as_index=False`` can be typed?

**Please complete the following information:**
 - OS:  ubuntu
 - OS Version 22.04
 - python version 3.10
 - version of type checker mypy 0.971
 - version of installed `pandas-stubs` newest 20220815

**Additional context**

Dr-Irv:
Similar issue with `df.groupby('a",as_index=False).size()`, which returns a `DataFrame`, but we type it as `Series`

phofl:
Yep, this happens probably for every function where a Series is possible

Dr-Irv:
Two options here, one easy and one hard:
1. Easy way - just have `value_counts()` typed as `FrameOrSeriesUnion` as the result.
2. Hard way  (and may not be possible) - make `DataFrameGroupBy` and `SeriesGroupBy` Generic with a typing parameter that indices whether `Series` or `DataFrame` should be returned, based on the value of `is_index` in `DataFrame.groupby()` and `Series.groupby()`.  Requires overloads there, and careful typing.

Regarding (2), I recently did something like this to handle the case of `Series.str.split()` where the parameter `expand` changes the nature of the result.

phofl:
If doable, I would obviously prefer 2 :)

amotzop:
I think `as_index` is kind of special and changes the return type of many functions, so there is also a third option of breaking `SeriesGroupBy` and `DataFrameGroupBy` into 2 different classes depending on `as_index` and overloading the various functions which return them.

Dr-Irv:
[...]

PR #653, opened 2023-04-19
fixup groupby.size return type

### LINKED ISSUE #652: groupby.size may return DataFrame in as_index=False
**Describe the bug**

**To Reproduce**

`t.py`
```python
import pandas as pd

df = pd.DataFrame({'a': [1,2,3]})
reveal_type(df.groupby('a', as_index=False).size())
```

```console
$ mypy t.py
t.py:4: note: Revealed type is "pandas.core.series.Series[builtins.int]"
```

However, it is in fact a DataFrame

 - OS: [e.g. Windows, Linux, MacOS] linux
 - OS Version [e.g. 22] Linux DESKTOP-U8OKFP3 5.10.102.1-microsoft-standard-WSL2 #1 SMP Wed Mar 2 00:30:59 UTC 2022 x86_64 x86_64 x86_64 GNU/Linux
 - python version 3.8
 - version of type checker

```console
$ pip freeze | grep pandas
pandas==2.0.0
pandas-flavor==0.5.0
pandas-stubs==2.0.0.230412
```

**Additional context**

MarcoGorelli:
- [ ] Closes #652  (Replace xxxx with the Github issue number)


pandas-stubs/core/groupby/groupby.pyi
```diff
@@ -99,7 +99,7 @@ class GroupBy(BaseGroupBy[NDFrameT]):
     def last(self, **kwargs) -> DataFrame | Series: ...
     def max(self, **kwargs) -> DataFrame | Series: ...
     def min(self, **kwargs) -> DataFrame | Series: ...
-    def size(self) -> Series[int]: ...
+    def size(self) -> DataFrame | Series[int]: ...
```
twoertwein, line 102: I'm not too familiar with the non-existing helper classes `pandas.core.groupby.generic._SeriesGroupBy(Non)Scalar/_DataFrameGroupBy(Non)Scalar`. It might be possible to avoid this union by having more complex overloads for `groupby(..., as_index=False)`. not sure whether that is worth it @Dr-Irv

Dr-Irv: I'm not sure what to do here.  This is the same issue as #203 .  Ideally, we'd trace the use of `as_index=True` via some class manipulation to determine the right value of the result being `DataFrame` or `Series[int]` dependent on whether `as_index=True` was specified.  I started working on that about 6 months ago, but lost my working branch.  It's non-trivial.

The downside to the change proposed is that if people don't specify the `as_index` parameter (and I'm in that category of never using it because the default is `as_index=False`), then if you call `size()`, you will have to cast your result to `Series` when you know it is a `Series` because you used default arguments.

So let's say that the number of people using the default setting of `as_index=False` has A developers, and the number of people using `as_index=True` has B developers.  If we accept this PR, then all the A people have to make a code change and use `cast`.  If we don't accept this PR, then all the B people have to make a code change and use `cast` (the status quo).

I'm leaning towards not accepting it, because I think A is much larger than B, and leaving the issue open until we do it the "right" way, but my opinion can be changed.

gandhis1: `as_index = True` is the default, no? And even though it's the default I would still contend that more people use it with `as_index = False`

Dr-Irv: Ooops.  Got it backwards.  Have corrected it via editing.

MarcoGorelli: I don't really mind, I can always cast in my use-case where I'm using `as_index=False`

Dr-Irv: > I don't really mind, I can always cast in my use-case where I'm using `as_index=False`

So should I close the PR?

MarcoGorelli: i've done it, thanks for your review and explanation!

PR #654, opened 2023-04-19
Fix pandas-nightly

### LINKED ISSUE #640: Fix nightly due to a new deprecation in 2.1.x
See https://github.com/pandas-dev/pandas-stubs/actions/runs/4641384569/jobs/8214310541

Need to modify the associated test with a `pytest_warns_bounded` using `lower=2.0.99` and the appropriate message.

twoertwein:
- [x] Closes #640 (Replace xxxx with the Github issue number)

There is still a TypeError on nightly xref https://github.com/pandas-dev/pandas/issues/52790 (will mark as invalid - there is no warning but it will raise in 2.1: I think we should reflect that in our annotations)

### DIFF
--- a/tests/test_api_types.py
+++ b/tests/test_api_types.py
@@ -8,7 +8,10 @@ from typing_extensions import assert_type
 
 from pandas._typing import DtypeObj
 
-from tests import check
+from tests import (
+    check,
+    pytest_warns_bounded,
+)
 
 nparr = np.array([1, 2, 3])
 arr = pd.Series([1, 2, 3])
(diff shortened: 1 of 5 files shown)


twoertwein:
Cannot have something like 

class PeriodDtype(PandasExtensionDtype):
    @overload
    def __init__(self, freq: str | BaseOffset = ...) -> None: ...
    @overload
    def __init__(self, freq: CustomBusinessDay) -> Never: ...

As mypy then thinks pd.PeriodDtype is not type[ExtensionType] anymore. I think that would have wider implications than indicating it is allowed to pass CustomBusinessDay

Dr-Irv (approved):
Thanks @twoertwein  .  Let's see if it works after the merge.

PR #661, opened 2023-04-26
reorder DataFrame.__getitem__() overloads to maybe support numpy 1.23.5

### LINKED ISSUE #656: Filtering with a boolean mask incorrectly changes type from `DataFrame` to `Series`
**Describe the bug**

When a dataframe is filtered with a boolean mask, e.g. `df[mask]`, the result is a DataFrame. MyPy throws an error, as it thinks the result is a Series.

**To Reproduce**

Checking this function with `mypy` throws an error:

```python
import pandas as pd

def bar(df: pd.DataFrame) -> pd.DataFrame:
    mask = df.index == df.index
    df2 = df[mask]  # mypy thinks df2 is a Series, but it's a DataFrame
    return df2    # error: Incompatible return value type (got "Series[Any]", expected "DataFrame")  [return-value]
```

**System info**

 - OS: Linux Ubuntu 20.04.5
 - `python` 3.9.10
 - `mypy` 1.2.0
 - `pandas-stubs` 2.0.0.230412
 - `pandas` `2.0.1`

Dr-Irv:
- [x] Closes #656 (Replace xxxx with the Github issue number)

I tested locally and this change fixes an issue with numpy 1.23.5.  I don't want to add another test to the CI for this, but the change works with our current tests, and I think it fixes the issue with numpy 1.23.5 by putting the narrower types first.


pandas-stubs/core/frame.pyi
```diff
@@ -558,6 +554,10 @@ class DataFrame(NDFrame, OpsMixin):
         | np_ndarray_bool
         | list[_ScalarOrTupleT],
     ) -> DataFrame: ...
+    @overload
+    def __getitem__(self, idx: Scalar | Hashable) -> Series: ...
+    @overload
+    def __getitem__(self, rows: slice) -> DataFrame: ...
```
twoertwein, line 560: `slice` is also `Hashable`, so it should probably before the second overload.

If possible it would also be great to rename the argument names - there is no argument named `idx` or `rows`, only `key`

https://github.com/pandas-dev/pandas/blob/d182a3495a95c7067c13690cfe037597c3714d5d/pandas/core/frame.py#L3733

Dr-Irv: Thanks.  I made those 2 changes in https://github.com/pandas-dev/pandas-stubs/pull/661/commits/d300c2d48afa08862901750f8ac13bcfc521b774

twoertwein (approved):

twoertwein:
Thanks @Dr-Irv !

PR #665, opened 2023-04-29
add support for `np.recarray` in `pd.factorize`

taoufik07:
Needed to pass `df.to_records` which returns an `np.recarray` directly to `pd.factorize`


Dr-Irv (approved):
thanks @taoufik07

PR #667, opened 2023-04-30
update pyright, numpy. fix issues 664 and 662

### LINKED ISSUE #662: numpy 1.24.3 support for comparisons with `pd.Timedelta` and `pd.Timestamp`
I did a `poetry update` and my environment got updated to numpy 1.24.3.  In that version of numpy, comparisons of numpy `timedelta` or `timestamp` objects with pandas `Timedelta` and `Timestamp` objects now return a `bool` instead of `np.bool_`, when the numpy object is on the LHS of the comparison.  The type declarations didn't change - they still return `Any`.

So we need to do the following:
1. Update the `pyproject.toml` to point to numpy 1.24.3
2. Update tests of the following form in `test_scalars.py` to have `check` compare the result to `bool` instead of `np.bool_`:
```python
check(assert_type(c_timedelta64 > td, Any), np.bool_)
```
I think there are about 14 such tests.

### LINKED ISSUE #664: Cannot access member "contains" for type "ExtensionArray"
**Describe the bug**

```
Cannot access member "contains" for type "ExtensionArray"
  Member "contains" is unknownPylance[reportGeneralTypeIssues](https://github.com/microsoft/pyright/blob/main/docs/configuration.md#reportGeneralTypeIssues)
(function) contains: Unknown
```

**To Reproduce**

from https://github.com/pandas-dev/pandas/issues/52616#issuecomment-1506195632:

```py
df = pd.DataFrame({'A': range(1, 10)})
obj = pd.Interval(1,4)
ser = pd.Series(obj, index=df.index)
arr = ser.array
arr.contains(df["A"])
```

 - OS: Linux
 - OS Version  Ubuntu 20.04.4 LTS
 - python version 3.10.9
 - version of type checker Pylance v2023.4.40
 - version of installed `pandas-stubs` 2.0.0.230412-py3-none-any

Dr-Irv:
- [x] Closes #662 
- [x] Closes #664
- [x] Tests added: 
  - `tests/test_interval.py:test_interval_array_contains()`

Update to numpy 1.24.3, pandas 2.0.1, pyright 1.1.305


pandas-stubs/core/series.pyi
```diff
@@ -1997,3 +2007,7 @@ class OffsetSeries(Series):
     def __radd__(self, other: Period) -> PeriodSeries: ...
     @overload
     def __radd__(self, other: BaseOffset) -> OffsetSeries: ...
+
+class IntervalSeries(Series, Generic[_OrderableT]):
+    @property
+    def array(self) -> IntervalArray: ...
```
twoertwein, line 2013: Why doesn't it work to add overloads instead of creating a helper class here? Not sure whether mypy finally accepts overload+property.

Dr-Irv: > Why doesn't it work to add overloads instead of creating a helper class here? Not sure whether mypy finally accepts overload+property.

I couldn't see how to get even pyright to support `overload` with `property` .

twoertwein: I thought at least pyright supports that - but I haven't tried it.

I think even if mypy+pyright support property+overload, we probably wouldn't want to use it as the first overload would need to be the more specific overload returning `IntervalArray` but we shouldn't return that for `Series[Any/Unknown]`.

twoertwein:
Thanks @Dr-Irv !

PR #670, opened 2023-05-01
close out a few first issues

### LINKED ISSUE #557: Series.apply() should accept `None` as Callable return type
**Describe the bug**
Pandas is fine if the Callable passed to `series.apply(...)` returns `None`, but `mypy` throws an error. Passing a null-returning Callable to `df.apply()` does not throw an error.

**To Reproduce**
```python
import pandas as pd

s = pd.Series([{'a': 1, 'b':2}])
s['c'] = s.apply(lambda x: None)
```

mypy fails with

```
a.py:4: error: Argument 1 to "apply" of "Series" has incompatible type "Callable[[Any], None]"; expected "Callable[..., Union[str, bytes, date, datetime, timedelta, datetime64, timedelta64, bool, int, float, Timestamp, Timedelta, complex, Sequence[Any], Mapping[Any, Any]]]"  [arg-type]
a.py:4: error: Incompatible return value type (got "None", expected "Union[str, bytes, date, datetime, timedelta, datetime64, timedelta64, bool, int, float, Timestamp, Timedelta, complex, Sequence[Any], Mapping[Any, Any]]")  [return-value]
```

 - MacOS
 - 12.6
 - python3.10
 - mypy 1.0.1
 - pandas-stubs 1.5.3.230227

**Additional context**
Using a data frame instead of a series results in no error:

```python
import pandas as pd

df = pd.DataFrame([{'a': 1, 'b':2}])
df['c'] = df.apply(lambda x: None)
```

### LINKED ISSUE #586: `Series.loc()` typecheck fails with a lambda (`Callable[[Any], Any]`)
**Describe the bug**
`Series.loc` does not accept a lambda (or more generally a `Callable[[Any], Any]`) to filter the rows. of __getitem__.

I suspected this is the case for `pd.DataFrame` too, but I haven't tested it.

**To Reproduce**

```python
import pandas as pd
s = pd.Series([1, 2])
filtered = s.loc[lambda x: x > 1]
```

Using `mypy`.

Error message:
```text
test_loc_with_callable.py:7: error: No overload variant of "__getitem__" of "_LocIndexerSeries" matches argument type "Callable[[Any], Any]"  [call-overload]
    filtered = s.loc[lambda x: x > 1]
```

 - OS: Linux
 - OS Version: output of `uname -a`: `5.14.0-1058-oem #66-Ubuntu SMP Fri Feb 10 09:46:18 UTC 2023 x86_64 x86_64 x86_64 GNU/Linux`
 - python version: Python 3.9.5
 - version of type checker: `mypy 1.1.1`
 - version of installed `pandas-stubs`: `1.5.3.230304`

### LINKED ISSUE #600: timedelta_range or interval_range should accept Timedelta or timedelta as freq
**Describe the bug**
The `timedelta_range` and `interval_range` functions should accept a `pandas.Timedelta` or `datetime.timedelta` object as the `freq` argument.

**To Reproduce**
Example:
```
import pandas

def foo(step: float):
    return pandas.timedelta_range(
        start=pandas.Timedelta(0),
        periods=10,
        freq=pandas.Timedelta(seconds=step),
    )
```

Using `mypy foo.py`:
```
foo.py:7: error: Argument "freq" to "timedelta_range" has incompatible type "Timedelta"; expected "Union[str, DateOffset, None]"  [arg-type]
```

 - OS: Debian 11.6
 - Python: 3.9.2
 - mypy: 1.1.1
 - pandas: 1.5.3
 - pandas-stubs: 1.5.3.230321

**Additional context**
https://github.com/pandas-dev/pandas-stubs/issues/223 mentions `date_range` and `bdate_range`.

https://github.com/pandas-dev/pandas/pull/48631 updates the documentation for `date_range`, `bdate_range`, and `interval_range`.

https://github.com/pandas-dev/pandas-stubs/pull/322 updates the stubs for `date_range` and `bdate_range`.

However, using `Timedelta` or `timedelta` with `timedelta_range` seems to work as well.  (I'm not sure I fully understand the difference between Timedelta and DateOffset, but I don't see why a Timedelta *shouldn't* be accepted here.)

Dr-Irv:
Had a little time so closed out a few "good first issues"

- [x] Closes #557
- [x] Closes #586
- [x] Closes #600

- [x] Tests added: 
  - `test_series.py: test_apply_returns_none()`
  - `test_series.py: test_loc_callable()`
  - Various tests for `timedelta_range()` and `interval_range()` with `timedelta` arguments.

### DIFF
--- a/pandas-stubs/core/indexes/interval.pyi
+++ b/pandas-stubs/core/indexes/interval.pyi
@@ -371,7 +371,7 @@ def interval_range(
     start: _TimestampLike,
     end: _TimestampLike = ...,
     periods: int | None = ...,
-    freq: str | BaseOffset | None = ...,
+    freq: str | BaseOffset | pd.Timedelta | dt.timedelta | None = ...,
     name: Hashable = ...,
     closed: IntervalClosedType = ...,
 ) -> IntervalIndex[Interval[pd.Timestamp]]: ...
@@ -381,7 +381,7 @@ def interval_range(
     start: None = ...,
     end: _TimestampLike,
     periods: int | None = ...,
-    freq: str | BaseOffset | None = ...,
+    freq: str | BaseOffset | pd.Timedelta | dt.timedelta | None = ...,
     name: Hashable = ...,
     closed: IntervalClosedType = ...,
 ) -> IntervalIndex[Interval[pd.Timestamp]]: ...
@@ -391,7 +391,7 @@ def interval_range(
     *,
     end: None = ...,
     periods: int | None = ...,
-    freq: str | BaseOffset | None = ...,
+    freq: str | BaseOffset | pd.Timedelta | dt.timedelta | None = ...,
     name: Hashable = ...,
     closed: IntervalClosedType = ...,
 ) -> IntervalIndex[Interval[pd.Timestamp]]: ...
@@ -400,7 +400,7 @@ def interval_range(
     start: _TimedeltaLike,
     end: _TimedeltaLike = ...,
     periods: int | None = ...,
-    freq: str | BaseOffset | None = ...,
+    freq: str | BaseOffset | pd.Timedelta | dt.timedelta | None = ...,
     name: Hashable = ...,
     closed: IntervalClosedType = ...,
 ) -> IntervalIndex[Interval[pd.Timedelta]]: ...
@@ -410,7 +410,7 @@ def interval_range(
     start: None = ...,
     end: _TimedeltaLike,
     periods: int | None = ...,
-    freq: str | BaseOffset | None = ...,
+    freq: str | BaseOffset | pd.Timedelta | dt.timedelta | None = ...,
     name: Hashable = ...,
     closed: IntervalClosedType = ...,
 ) -> IntervalIndex[Interval[pd.Timedelta]]: ...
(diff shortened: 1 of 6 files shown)


pandas-stubs/core/indexes/interval.pyi
```diff
@@ -371,7 +371,7 @@ def interval_range(
     start: _TimestampLike,
     end: _TimestampLike = ...,
     periods: int | None = ...,
-    freq: str | BaseOffset | None = ...,
+    freq: str | BaseOffset | pd.Timedelta | dt.timedelta | None = ...,
```
twoertwein, line 374: is `pd.Timedelta` a subclass of `datetime.timedelta`? Could simplify this union (or keep, if it is more explicit).

pandas-stubs/core/series.pyi
```diff
@@ -179,7 +179,13 @@ class _LocIndexerSeries(_LocIndexer, Generic[S1]):
...
     def __getitem__(
         self,
-        idx: MaskType | Index | Sequence[float] | list[str] | slice | _IndexSliceTuple,
+        idx: MaskType
+        | Index
+        | Sequence[float]
+        | list[str]
+        | slice
+        | _IndexSliceTuple
+        | Callable,
```
twoertwein, line 188: Happy to keep it as is - it is probably not worth to try to enumerate all allowed arguments/return values.

pandas-stubs/core/indexes/interval.pyi
Dr-Irv: Yes, it is a subclass.  We're inconsistent in the stubs.  Some places we just say `pd.Timedelta`, others we have the  union.  I like the explicit aspect to make it clear that `dt.timedelta` is accepted.

pandas-stubs/core/series.pyi
Dr-Irv: Yes.  We have it this way in `DataFrame.loc` as well.

twoertwein:
Thanks @Dr-Irv !

PR #679, opened 2023-05-03
Extends `DataFrame.groupby` overloads to recognize some scalar index types

### LINKED ISSUE #674: `DataFrame.groupby` should return `_DataFrameGroupByScalar` when grouping by a PeriodIndex
**Describe the bug**
When performing a groupby on a dataframe with a `PeriodIndex` the group identifier should be a `Period`, i.e. a scalar (On that note `Period` appears to be missing from the definition of `Scalar` in `_typing.pyi`)

**To Reproduce**
```python
import pandas as pd
df = pd.DataFrame({'date': pd.date_range('2020-01-01', '2020-12-31'), 'days': 1})
index = pd.PeriodIndex(df.date, freq='M')
for period, group in df.groupby(index):
    period.start_time  # mypy: attr-defined ­— Tuple[Any, ...] has no attribute "start_time"
```

 - Python 3.10
 - mypy 1.0.1
 - pandas-stubs 2.0.1.230501

**Additional comments**
This should probably extend to most other specialized Index types apart from the base `Index` and `MultiIndex`, for the latter it's obvious that it would be non-scalar, for the former it would depend on what Python object is referenced, since converting `MultiIndex` to and from an `Index` on a `builtins.tuple` is quite common, it's probably the right move to treat `Index` as non-scalar by default or return an `Any` variant of `DataFrameGroupBy` that could either be scalar or non-scalar.

Daverball:
- [x] Closes #674


Daverball:
Initially I had also added the Series version of each Index to the type unions and added a separate new test for those, however it seems a little less useful, since you would have to use the series from the dataframe, which does not have a specific type, so you would need to manually cast to a more specific type to get the benefit there.

It certainly can be done, but it felt so awkward writing the test that I decided against it for now. It's also not a pattern I personally have ever used before on a groupby. This also decreases the surface area of the patch, so it should be a less controversial change overall.

pandas-stubs/core/groupby/generic.pyi
```diff
@@ -149,6 +150,9 @@ class SeriesGroupBy(GroupBy, Generic[S1]):
 class _DataFrameGroupByScalar(DataFrameGroupBy):
     def __iter__(self) -> Iterator[tuple[Scalar, DataFrame]]: ...
 
+class _DataFrameGroupByPeriod(DataFrameGroupBy):
```
twoertwein, line 153: If the only difference between these many helper classes is `__iter__`, it might be nicer to have one Generic class:

```py
class _DataFrameGroupBy(DataFrameGroupBy, Generic[T]):
    def __iter__(self) -> Iterator[tuple[T, DataFrame]]: ...
```

@Dr-Irv

Dr-Irv: I think that's a good idea.  Would need to create a type alias (ideally prefixed with an underscore and only in this file - not sure if it would require to be declared in `_typing.py`) that is a `TypeVar` that includes `Scalar`, `Period`, and `tuple` and then modify `DataFrame.apply()` to return the correct generic version.

tests/test_frame.py
```diff
@@ -1980,6 +1980,46 @@ def test_groupby_result() -> None:
         pass
 
 
+def test_groupby_result_for_scalar_indeces() -> None:
```
Dr-Irv, line 1983:
```suggestion
def test_groupby_result_for_scalar_indexes() -> None:
```

tests/test_frame.py
```diff
@@ -1980,6 +1980,46 @@ def test_groupby_result() -> None:
...
+    check(assert_type(value, pd.DataFrame), pd.DataFrame)
+
+    dt_index = pd.DatetimeIndex(df.date)
+    iterator2 = df.groupby(dt_index).__iter__()
+    assert_type(iterator2, Iterator[Tuple[Scalar, pd.DataFrame]])
+    index2, value2 = next(iterator2)
+    assert_type((index2, value2), Tuple[Scalar, pd.DataFrame])
+
+    check(assert_type(index2, Scalar), pd.Timestamp)
+    check(assert_type(value2, pd.DataFrame), pd.DataFrame)
```
Dr-Irv, line 2002: When I look at this sequence, since you know the `Scalar` will be a `Timestamp`, might as well create an overload for that argument and have `__iter__()` then return `tuple[pd.Timestamp, pd.DataFrame]`

This is also another reason to make it `Generic`

tests/test_frame.py
```diff
@@ -1980,6 +1980,46 @@ def test_groupby_result() -> None:
...
+
+    tdelta_index = pd.TimedeltaIndex(df.date - pd.Timestamp("2020-01-01"))
+    iterator3 = df.groupby(tdelta_index).__iter__()
+    assert_type(iterator3, Iterator[Tuple[Scalar, pd.DataFrame]])
+    index3, value3 = next(iterator3)
+    assert_type((index3, value3), Tuple[Scalar, pd.DataFrame])
+
+    check(assert_type(index3, Scalar), pd.Timedelta)
```
Dr-Irv, line 2010: same comment as above.  If you're going to have `TimedeltaIndex` as a possible argument and we know the iterator will contain `Timedelta`, might as well have a separate overload for that.

Dr-Irv (changes requested):
Thanks for the contribution.  Can you make the change to use `Generic` and then just have separate overloads for the different `Index` types you are including?

pandas-stubs/core/frame.pyi
```diff
@@ -52,6 +56,7 @@ import xarray as xr
 
 from pandas._libs.missing import NAType
 from pandas._libs.tslibs import BaseOffset
+from pandas._libs.tslibs.period import Period
```
Dr-Irv, line 59: This import is unused so you need to remove it.

Strongly suggest setting up a dev environment as described at https://github.com/pandas-dev/pandas-stubs/blob/main/docs/setup.md and testing locally before doing next commit.

Daverball: I did do that, so not sure why pre-commit didn't catch that one. That's a leftover from when I tried to add Series to the overloads as well.

pandas-stubs/core/groupby/generic.pyi
Daverball: I also think a generic is a good idea, although there is a slightly higher potential that it will break people's code, it's a lot better than pretending there are subclasses of `GroupByDataFrame`, which comes with its own semantic problems.

Putting the `TypeVar` into `_typing,py` seems sensible, if anyone creates a function that operates on a `GroupByDataFrame`  and returns a modified version, you probably want them to be able to adjust the signature so the bound type is preserved, so they would need to either redefine the `TypeVar` or import/access it in a type checking context from `pandas.core.groupby.generic` which is a lot more verbose than `pandas._typing`.

tests/test_frame.py
Daverball: If we're going to add a bunch of overloads anyways, shall I also make ones for the rest of the index types that are neither `Index` nor `MultipleIndex`?

Making `GroupByDataFrame` generic we could also be a little bit more careful about where we put the `Index` overload. Since we don't really know the `dtype` of the index we should probably just return `GroupByDataFrame[Any]`.

- `CategoricalIndex` is in the same boat as `Index`, since we don't really know the type of the categories, so that could be part of the same overload.
- `MultiIndex` in the overload that originally had `Index` should be fine, since the group identifier will be a tuple.
- `IntervalIndex` would need its own overload with `pd.Interval`, it's currently also not part of `Scalar` just like `pd.Period`.

Daverball: I did end up going with exactly that

pandas-stubs/_typing.pyi
```diff
@@ -404,6 +404,29 @@ Function: TypeAlias = np.ufunc | Callable[..., Any]
...
+    datetime.datetime,
+    datetime.timedelta,
+    np.datetime64,
+    np.timedelta64,
+    bool,
+    int,
+    float,
+    complex,
+    Timestamp,
+    Timedelta,
```
Dr-Irv, line 421: You can delete all of these as they are in `Scalar`

pandas-stubs/core/frame.pyi
```diff
@@ -1011,7 +1016,85 @@ class DataFrame(NDFrame, OpsMixin):
...
+        sort: _bool = ...,
+        group_keys: _bool = ...,
+        squeeze: _bool = ...,
+        observed: _bool = ...,
+        dropna: _bool = ...,
+    ) -> DataFrameGroupBy[IntervalT]: ...
+    @overload
+    def groupby(
+        self,
+        by: MultiIndex,
```
Dr-Irv, line 1075: Might as well just combine this with the last overload that returns `DataFrameGroupBy[tuple]`

Dr-Irv (changes requested):
This looks really good.  Just some small changes to make.

tests/test_frame.py
Dr-Irv: I like the path that you took with this

pandas-stubs/core/frame.pyi
Dr-Irv: Our pre-commit doesn't catch overloads that aren't used.  It's picked up by `pyright` which is not in the pre-commit - just the tests

pandas-stubs/_typing.pyi
Daverball: That's what I had at the start, but then mypy will complain about the `Scalar` overload on `DataFrame.groupby`. There is a subtle difference between `Union[DataframeGroupBy[str], ...]` and `DataFrameGroupBy[Union[str, ...]]` when matching overloads. This seemed like the most ergonomic way to still allow that to happen.

pandas-stubs/core/frame.pyi
Daverball: This will not work due to `MultiIndex` being a subclass of `Index` so then it would match the incorrect overload first.

Daverball: If we wanted to reorder the overloads to combine these two cases, we would have to stop using `GroupByObjectNonScalar`, since that also contains `Index`.

Daverball: Although it seems like `GroupByObjectNonScalar` is currently only used in `series.py` and `frame.py`, so maybe the answer is to just do the refactor and remove `Index` from this type union and add an overload to `Series.groupby` for the ambiguous case.

The logical next step would be to do exactly the same change for `SeriesGroupBy` anyways, so the type of the group identifier gets preserved if you access a column on `DataFrameGroupBy`, although making it a two parameter `Generic` might make it slightly more confusing to use, so I opted to leave that transformation for another PR in the future.

pandas-stubs/_typing.pyi
Dr-Irv: Yes, I see what you're saying.  Apparently, if you have a parameter of `T` on a method that returns `X[T]` where `X` is `Generic` and `T` is a `TypeVar`, `T` doesn't get bound correctly.

Oh, well.

pandas-stubs/core/frame.pyi
Dr-Irv: Yes, I see what you're saying, and I'm fine to let it be.

Dr-Irv (approved):
Thanks @Daverball .  Really nice contribution.

PR #680, opened 2023-05-03
GH-676 : Corrected `DtypeArg` for accepting strings

### LINKED ISSUE #676: `read_sql_*` methods should accept `dtype` argument with same type as `DataFrame.astype()`
For the `read_sql_*` methods, the argument `dtype` is just passed to `DataFrame.astype()`, where we allow `dtype: AstypeArg | Mapping[Any, Dtype] | Series`, so the `read_sql_*()` methods should do the same.

In particular, doing something like `dtype: {"col": "str"}` is totally allowed, but not permitted in the stubs.

pandas-stubs version 2.0.1.230501

ramvikrams:
- [x] Closes #676 

### DIFF
--- a/tests/test_io.py
+++ b/tests/test_io.py
@@ -1390,7 +1390,7 @@ def test_read_with_lxml_dtype_backend() -> None:
         )
 
 
-def test_read_sql_str_dtype() -> None:
+def test_read_sql_dict_str_value_dtype() -> None:
     with ensure_clean() as path:
         con = sqlite3.connect(path)
         check(assert_type(DF.to_sql("test", con), Union[int, None]), int)
--- a/tests/test_io.py
+++ b/tests/test_io.py
@@ -1388,3 +1388,31 @@ def test_read_with_lxml_dtype_backend() -> None:
         check(
             assert_type(read_xml(path, dtype_backend="pyarrow"), DataFrame), DataFrame
         )
+
+
+def test_read_sql_str_dtype() -> None:
+    with ensure_clean() as path:
+        con = sqlite3.connect(path)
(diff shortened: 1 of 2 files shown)


tests/test_io.py
```diff
@@ -1388,3 +1388,22 @@ def test_read_with_lxml_dtype_backend() -> None:
...
+    with ensure_clean() as path:
+        con = sqlite3.connect(path)
+        check(assert_type(DF.to_sql("test", con), Union[int, None]), int)
+        check(
+            assert_type(
+                read_sql_query(
+                    "select * from test",
+                    con=con,
+                    index_col="index",
+                    dtype="int",
```
Dr-Irv, line 1403: Based on the issue, this should be `dtype={"a": "int"}`

tests/test_io.py
```diff
@@ -1388,3 +1388,22 @@ def test_read_with_lxml_dtype_backend() -> None:
         check(
             assert_type(read_xml(path, dtype_backend="pyarrow"), DataFrame), DataFrame
         )
+
+
+def test_read_sql_str_dtype() -> None:
```
Dr-Irv, line 1393:
```suggestion
def test_read_sql_dict_str_value_dtype() -> None:
```

Dr-Irv: Also can you add a comment here referring back to #676 ?

Dr-Irv (changes requested):

tests/test_io.py
ramvikrams: Done

Dr-Irv (approved):
Thanks @ramvikrams

PR #683, opened 2023-05-04
allow passing lists as values to `locframe.__setitem__`

taoufik07:

### DIFF
--- a/pandas-stubs/core/frame.pyi
+++ b/pandas-stubs/core/frame.pyi
@@ -198,7 +198,7 @@ class _LocIndexerFrame(_LocIndexer):
     def __setitem__(
         self,
         idx: MaskType | StrLike | _IndexSliceTuple | list[ScalarT],
-        value: S1 | ArrayLike | Series | DataFrame | None,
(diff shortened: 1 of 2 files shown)


pandas-stubs/core/frame.pyi
```diff
@@ -198,7 +198,7 @@ class _LocIndexerFrame(_LocIndexer):
     def __setitem__(
         self,
         idx: MaskType | StrLike | _IndexSliceTuple | list[ScalarT],
-        value: S1 | ArrayLike | Series | DataFrame | None,
+        value: S1 | ArrayLike | Series | DataFrame | list | None,
```
twoertwein, line 201: I'm not sure how strict we want to be. Could replace `list` with `Sequence[S1]`

Dr-Irv: `Sequence[S1]` might be too narrow.  I think we would want to allow something that had mixed types.  E.g., `df.loc["a"] = [3.2, "abc"]`

Dr-Irv (approved):
Thanks @taoufik07

PR #687, opened 2023-05-07
upgrade SQLAlchemy and pyright

Dr-Irv:
Based on https://github.com/pandas-dev/pandas/issues/50558#issuecomment-1537477148, we could upgrade SqlAlchemy to 2.0.  Doing that revealed an issue in those tests and some of the typing.

Also bumped pyright to 1.1.306, but the default value for `useLibraryCodeForTypes` changed in that version, so added that in `pyproject.toml`.  Without that, some of the plotting tests fail because `matplotlib` doesn't have typing yet.

### DIFF
--- a/pandas-stubs/io/sql.pyi
+++ b/pandas-stubs/io/sql.pyi
@@ -13,6 +13,7 @@ from typing import (
 from pandas.core.base import PandasObject
 from pandas.core.frame import DataFrame
 import sqlalchemy.engine
+from sqlalchemy.orm import FromStatement
 import sqlalchemy.sql.expression
 from typing_extensions import TypeAlias
 
@@ -26,7 +27,11 @@ from pandas._typing import (
 _SQLConnection: TypeAlias = str | sqlalchemy.engine.Connectable | sqlite3.Connection
 
 _SQLStatement: TypeAlias = (
-    str | sqlalchemy.sql.expression.Selectable | sqlalchemy.sql.expression.TextClause
+    str
+    | sqlalchemy.sql.expression.Selectable
+    | sqlalchemy.sql.expression.TextClause
+    | sqlalchemy.sql.Select
+    | FromStatement
 )
 
 @overload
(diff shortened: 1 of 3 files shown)


tests/test_io.py
```diff
@@ -1095,6 +1095,7 @@ def test_read_sql_via_sqlalchemy_connection():
                 assert_type(read_sql("select * from test", con=conn), DataFrame),
                 DataFrame,
             )
+        engine.dispose()
```
twoertwein, line 1098: Can `engine` be used as a context manager? Might be slightly nicer if that's possible

pandas-stubs/io/sql.pyi
```diff
@@ -26,7 +27,11 @@ from pandas._typing import (
 _SQLConnection: TypeAlias = str | sqlalchemy.engine.Connectable | sqlite3.Connection
 
 _SQLStatement: TypeAlias = (
-    str | sqlalchemy.sql.expression.Selectable | sqlalchemy.sql.expression.TextClause
+    str
+    | sqlalchemy.sql.expression.Selectable
+    | sqlalchemy.sql.expression.TextClause
+    | sqlalchemy.sql.Select
+    | FromStatement
```
twoertwein, line 34: Do we need to add those libraries (or their stubs) as a dependency?

tests/test_io.py
Dr-Irv: No, it is not possible.

pandas-stubs/io/sql.pyi
Dr-Irv: I don't think so.  I just added the two new types that seem to be valid for the test we had.  We have `sqlalchemy` in our dev environment, but we don't want to make it something you have to install to use the stubs.

twoertwein (approved):

twoertwein:
Thanks @Dr-Irv !

PR #690, opened 2023-05-10
Refactors `Series.groupby` to match `DataFrame.groupby`

Daverball:
- [x] Follow-up to #679  (Related Issue #674 has already been closed)


Daverball:
I also lifted `Index` out of `GroupByObjectNonScalar` to avoid the unsafe overload errors and to get rid of one of the redundant overloads.

tests/test_series.py
```diff
@@ -565,6 +565,153 @@ def test_types_groupby_methods() -> None:
...
+        pd.DataFrame(
+            {"a": [0, 1, 1], "b": [0, 1, 0]},
+            # avoid ambiguous integer size on Windows
+            dtype=np.int64,
+        )
+    )
+    s = pd.Series([0, 1, 2], index=multi_index, dtype=int)
+    # workaround different default integer size on Windows, the series
+    # constructor does not work yet with IntDtypeArg, but astype does
+    s = s.astype(np.int64)
```
Dr-Irv, line 583: I think you can avoid this by using `np.integer` in the `check` tests below instead of `np.int64`

Dr-Irv (changes requested):
Thanks for the nice contribution.  Can you try `np.integer` rather than using `astype(np.int64)` ?

Dr-Irv (approved):
Thanks @Daverball

PR #693, opened 2023-05-12
CI: unpin lxml and update mypy

### LINKED ISSUE #688: lxml - try upgrade to 4.9.1
See dependabot alert.  https://github.com/pandas-dev/pandas-stubs/security/dependabot/1

Probably not an issue for us because it's not executable code, but we should see if things work if we upgrade.

twoertwein:
- [x] Closes #688 (Replace xxxx with the Github issue number)


pyproject.toml
```diff
@@ -35,13 +35,13 @@ python = ">=3.8"
 types-pytz = ">= 2022.1.1"
 
 [tool.poetry.dev-dependencies]
-mypy = "1.2.0"
+mypy = "1.3.0"
+pandas = "2.0.1"
```
twoertwein, line 39: moves pinned dependecies to the top

pyproject.toml
```diff
@@ -50,15 +50,15 @@ black = ">=23.3.0"
 isort = ">=5.12.0"
 openpyxl = ">=3.0.10"
 tables = { version = ">=3.7.0" , python = "<4"}  # 3.8.0 depends on blosc2 which caps python to <4
-lxml = { version = ">=4.7.1,<4.9.0", python = "<3.11" }
+lxml = ">=4.7.1"
```
Dr-Irv, line 53:
```suggestion
lxml = ">=4.9.1"
```

The original issue was to try `lxml` to 4.9.1 due to some security vulnerability, so can we change that here?

Dr-Irv (changes requested):

PR #694, opened 2023-05-12
Fix inplace overload for DataFrame.interpolate

### LINKED ISSUE #691: DataFrame.interpolate method return type does not honor default inplace argument
Using pandas-stubs-2.0.1.230501.

inplace defaults to False, so the below shoudl return a new DataFrame.  But pyright thinks the return type is DataFrame | None. Here is some code showing the problem:

import pandas as pd
df: pd.DataFrame = -pd.DataFrame()
df = df.interpolate(method="linear")   # pyright reports the below message for this line

Expression of type "DataFrame | None" cannot be assigned to declared type "DataFrame"
  Type "DataFrame | None" cannot be assigned to type "DataFrame"
    Type "None" cannot be assigned to type "DataFrame"

But if I repalce the line with the below, then pyright is happy:

df = df.interpolate(method="linear", inplace=False)

twoertwein:
- [x] Closes #691 (Replace xxxx with the Github issue number)

### DIFF
--- a/pandas-stubs/core/groupby/generic.pyi
+++ b/pandas-stubs/core/groupby/generic.pyi
@@ -210,7 +210,6 @@ class DataFrameGroupBy(GroupBy, Generic[ByT]):
     def cumsum(self, axis: Axis = ..., **kwargs) -> DataFrame: ...
     def describe(self, **kwargs) -> DataFrame: ...
     def ffill(self, limit: int | None = ...) -> DataFrame: ...
-    @overload
     def fillna(
         self,
         value,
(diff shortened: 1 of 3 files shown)


pandas-stubs/core/groupby/generic.pyi
```diff
@@ -210,38 +210,15 @@ class DataFrameGroupBy(GroupBy, Generic[ByT]):
...
     def fillna(
         self,
         value,
         method: str | None = ...,
         axis: Axis = ...,
         limit: int | None = ...,
         downcast: dict | None = ...,
-        *,
-        inplace: Literal[False],
+        inplace: Literal[False] = ...,
```
Dr-Irv, line 220: Need to preserve the ordering of the params.  `inplace` comes before `limit`,  then `downcast`

Dr-Irv (changes requested):

Dr-Irv (approved):
thanks @twoertwein

PR #696, opened 2023-05-13
Add numpy  as a 'runtime' dependency

### LINKED ISSUE #672: Adding numpy as requirement for installation. (was adding numpy as requirement)
**Describe the bug**
Hey,
we have a environment where only linters are installed. After running `mypy .` we install all missing subs with `mypy --install-types --non-interactive`.  
After running `mypy --ignore-missing-imports --disallow-any-generics --disallow-untyped-defs --no-implicit-optional --disallow-incomplete-defs .` we geht an error for accessing a series:
```
dummy = test_df["dummy"].tolist()
```
Raises:
bug_minimal_example.py:5: error: "Series[Any]" not callable  [operator]

Error vanishes after numpy or pandas is installed.
**To Reproduce**
```python
import pandas as pd

test_df = pd.DataFrame()
test_df["dummy"] = [1, 2, 3, 4, 5]
dummy = test_df["dummy"].tolist()
print(dummy)
```
bug_minimal_example.py:5: error: "Series[Any]" not callable  [operator]

 - OS: Ubuntu [WSL]
 - OS Version: 22.04.2 LTS
 - python version: 3.8
 - version of type checker: mypy 1.2.0

**Additional context**
Suggested fix:
Adding pandas to requirements

twoertwein:
- [x] Closes #672 (Replace xxxx with the Github issue number)

I'm not keen on adding numpy as a dependency but typeshed also adds numpy as a dependency for some of their stubs.

I think we should only add runtime dependencies if all of the below applies:

- used in the stubs
- a py-typed libary (or has stubs)
- the package is either a required dependecy of pandas (or someone creates a bug report?)

I re-wrote a few imports to make it easier to grep for non-pandas imports.


Dr-Irv (approved):
thanks @twoertwein .  I agree with your assessment of rules to determine when we make it a dependency.

PR #699, opened 2023-05-19
Fix WeekOfMonth constructor

### LINKED ISSUE #629: WeekOfMonth typing is incorrect

twoertwein:
- [x] Closes #629 (Replace xxxx with the Github issue number)


tests/test_timefuncs.py
```diff
@@ -635,7 +635,7 @@ def test_some_offsets() -> None:
         pd.DatetimeIndex,
     )
     # GH 320
-    tswm1 = pd.Timestamp("9/23/2022") + pd.offsets.WeekOfMonth(2, 3)
+    tswm1 = pd.Timestamp("9/23/2022") + pd.offsets.WeekOfMonth(2, True)
```
twoertwein, line 638: Pandas probably treated 3 as True.

tests/test_timefuncs.py
```diff
@@ -1141,3 +1141,8 @@ def test_timestamp_strptime_fails():
                 "%Y-%M-%D",  # type: ignore[arg-type] # pyright: ignore[reportGeneralTypeIssues]
             )
         )
+
+
+def test_weekofmonth_init():
+    # GH 629
+    pd.offsets.WeekOfMonth(n=1, week=1, weekday=1, normalize=True)
```
twoertwein, line 1148: Does this need a check/assert_type statement?

Dr-Irv (changes requested):
Add `check(assert_type(...`

Dr-Irv (approved):
thanks @twoertwein

PR #702, opened 2023-05-22
SeriesGroupBy.diff

### LINKED ISSUE #658: error: "_SeriesGroupByScalar[Any]" has no attribute "diff"  [attr-defined]
```python
import pandas as pd

index = pd.MultiIndex.from_product([["a", "b"], [1, 2, 3]], names=["foo", "bar"])
s = pd.Series(list(range(len(index))), index=index)
s.groupby("foo").diff()  # error: "_SeriesGroupByScalar[Any]" has no attribute "diff"  [attr-defined]
```

 - Ubuntu 22.04
 - python 3.10.6
 - mypy  1.2.0
 - pandas                              2.0.1
 - pandas-stubs                        1.5.3.230304

twoertwein:
- [x] Closes #658 (Replace xxxx with the Github issue number)


pandas-stubs/core/groupby/generic.pyi
```diff
@@ -141,6 +141,7 @@ class SeriesGroupBy(GroupBy, Generic[S1, ByT]):
     def idxmax(self, axis: Axis = ..., skipna: bool = ...) -> Series: ...
     def idxmin(self, axis: Axis = ..., skipna: bool = ...) -> Series: ...
     def __iter__(self) -> Iterator[tuple[ByT, Series[S1]]]: ...
+    def diff(self, periods: int = ..., axis: Axis = ...) -> Series: ...
```
twoertwein, line 144: Could return `Series[S1]` but that would be incorrect for timestamps (and maybe more types).

Dr-Irv: In `series.pyi`, we have that doing subtraction on `TimestampSeries` yields `TimedeltaSeries`, and subtraction on `PeriodSeries` yields `OffsetSeries` .  

So it might be the case that you could do overloads here on `self: SeriesGroupBy[Timestamp, ByT]` to make that work, but it might be tricky to get it to work with `mypy` without creating the other kinds of `Series` like we do in `series.pyi`.

We can leave that as a new issue.  Do you want to do that?

Dr-Irv (approved):
Thanks @twoertwein

PR #705, opened 2023-05-23
Allow Timedelta for `Series.rolling` and `DataFrame.rolling`

### LINKED ISSUE #704: `.rolling` should accept `Timedelta` as values.
Timedelta is not an allowed argument type on `pd.Series.rolling` nor `pd.DataFrame.rolling`, even though they are supported and [official docs](https://pandas.pydata.org/docs/reference/api/pandas.DataFrame.rolling.html) states so too.

**To Reproduce**
The following was run with `mypy`
```python
import pandas as pd

idx = pd.date_range('2023-01-01', '2023-02-01')

pd.Series(range(32) index=idx).rolling(pd.Timedelta('7D')).mean()
pd.DataFrame(range(32), index=idx).rolling(pd.Timedelta('7D')).mean()
```

This will produce the error messages

> No overload variant of "rolling" of "Series" matches argument type "Timedelta"  [call-overload]
> No overload variant of "rolling" of "DataFrame" matches argument type "Timedelta"  [call-overload]

 - OS: MacOS Monterey
 - OS Version: 12.5.1
 - python version: 3.10.11
 - version of type checker: mypy 1.3.0

pinealan:
- [x] Closes #704 

### DIFF
--- a/pandas-stubs/core/frame.pyi
+++ b/pandas-stubs/core/frame.pyi
@@ -1955,7 +1955,7 @@ class DataFrame(NDFrame, OpsMixin):
     @overload
     def rolling(
         self,
-        window: int | str | Timedelta | BaseOffset | BaseIndexer,
+        window: int | str | _dt.timedelta | BaseOffset | BaseIndexer,
         min_periods: int | None = ...,
         center: _bool = ...,
         on: Hashable | None = ...,
@@ -1969,7 +1969,7 @@ class DataFrame(NDFrame, OpsMixin):
     @overload
     def rolling(
         self,
-        window: int | str | Timedelta | BaseOffset | BaseIndexer,
+        window: int | str | _dt.timedelta | BaseOffset | BaseIndexer,
         min_periods: int | None = ...,
         center: _bool = ...,
         on: Hashable | None = ...,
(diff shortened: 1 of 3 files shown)


pandas-stubs/core/frame.pyi
```diff
@@ -1955,7 +1955,7 @@ class DataFrame(NDFrame, OpsMixin):
     @overload
     def rolling(
         self,
-        window: int | str | BaseOffset | BaseIndexer,
+        window: int | str | Timedelta | BaseOffset | BaseIndexer,
```
twoertwein, line 1958: If `datetime.timdelta` is also allowed (I'm not sure), you could replace `pd.Timedelta` with `datetime.timdelta`. I think `pd.Timedelta` inherits from `datetime.timdelta`

Dr-Irv: Yes, this is correct, so can you change all the`Timedelta` declarations you added to `rolling` to `timedelta` and change test as indicated below?

Dr-Irv (changes requested):

Dr-Irv (approved):
thanks @pinealan

PR #708, opened 2023-05-24
add proper type when grouping by a Series

### LINKED ISSUE #706: Series.groupby(by: Series) should not return SeriesGroupBy[S1, tuple]
**Describe the bug**

When using a `Series` in a `groupby`, the `byT` type is supposed to be `tuple` because `Series` is included in the `TypeVar` `GroupByObjectNonScalar`, but since there is only one `Series`, it should not be a tuple, but rather the type of the Series we use in the groupby.

See how the `GroupByObjectNonScalar` is declared here : https://github.com/pandas-dev/pandas-stubs/blob/70c412c84ec70fabdeb6b774b06811758b1b4b2b/pandas-stubs/_typing.pyi#L435 (note how it includes both `Series` and `list[Series]`, same goes with `Function` or `Grouper` probably.

See how the resulting `groupby` is decided to be `SeriesGroupBy[S1, tuple]` here when `by` is either a `MultiIndex` or a `GroupByObjectNonScalar` : https://github.com/pandas-dev/pandas-stubs/blob/70c412c84ec70fabdeb6b774b06811758b1b4b2b/pandas-stubs/core/series.pyi#L627

It seems to me that the `GroupByObjectNonScalar` is a bit too broad and includes both objects representing only one grouper and multiple groupers (where the `SeriesGroupBy[S1, tuple]` is consistent). So maybe we should divide this `TypeVar` in two and change the `groupby` overloads so that when only one grouper is present, the type is `Any` ?

_typing.pyi
```python
MultiGroupByObjectNonScalar: TypeAlias = (
    tuple
    | list[_HashableTa]
    | list[Function]
    | list[Series]
    | np.ndarray 
    | list[np.ndarray]
    | list[Mapping[Label, Any]]
    | list[Index]
    | list[Grouper]
)
MonoGroupByObjectNonScalar: TypeAlias = (
    Function
    | Series
    | Mapping[Label, Any]
    | Grouper
)
```

core/series.pyi
```python
    @overload
    def groupby(
        self,
        by: MultiIndex | MultiGroupByObjectNonScalar = ...,
        axis: AxisIndex = ...,
        level: Level | None = ...,
        as_index: _bool = ...,
        sort: _bool = ...,
        group_keys: _bool = ...,
        squeeze: _bool = ...,
        observed: _bool = ...,
        dropna: _bool = ...,
    ) -> SeriesGroupBy[S1, tuple]: ...
    @overload
    def groupby(
        self,
        by: CategoricalIndex | Index | MonoGroupByObjectNonScalar,
        axis: AxisIndex = ...,
        level: Level | None = ...,
        as_index: _bool = ...,
        sort: _bool = ...,
        group_keys: _bool = ...,
        squeeze: _bool = ...,
        observed: _bool = ...,
        dropna: _bool = ...,
    ) -> SeriesGroupBy[S1, Any]: ...
```

Note that I did not include the `np.ndarray` object in `MonoGroupByObjectNonScalar` since it can be 2D. Not sure how this one should be treated.

Of course this proposition is just an illustration, I suppose such division would call for many other modification elsewhere, especially within core/frame.pyi and that there might be better solution.

**To Reproduce**

The following example shows that the type of output is deduced to be `list[tuple[tuple, Series]]` while it should be `list[tuple[bool, Series]]`

```python
import pandas as pd
import numpy as np

series = pd.Series(np.random.randn(10))
above_0 = series > 0

output = list(series.groupby(above_0))

print(output)
reveal_type(output)
```
output when running with python :
```
[(False, 0   -0.666977
2   -1.586030
3   -0.862037
5   -0.617085
6   -1.553402
8   -0.694892
9   -0.492148
dtype: float64), (True, 1    0.848925
4    0.603415
7    0.938546
dtype: float64)]
```

Output when running with pyright
```
example.py:11:13 - information: Type of "output" is "list[tuple[tuple[Unknown, ...], Series[Unknown]]]"
```

This happens because the `list` function calls the `__iter__` method which is typed here : https://github.com/pandas-dev/pandas-stubs/blob/70c412c84ec70fabdeb6b774b06811758b1b4b2b/pandas-stubs/core/groupby/generic.pyi#L143

 - OS: Ubuntu
 - OS 22.04
 - python version : 3.10
 - version of type checker : pyright v1.1.298
 - version of installed `pandas-stubs` : 2.0.1.230501

ClementPinard:
Work in Progress, No test yet

I added a new TypeVar that is the intersection of S1 and ByT.
The type hint for S1 types outside of ByT is not covered anymore, but I suppose that it would theoretically fail in pandas, hence its absence of ByT

- [X] Closes #706


ClementPinard:
Just uploaded tests for iteration over groupby objects.
Note that test are falling now, probably because the typer fails to identify a `Series[bool]` and thus the correspondig. Would love some guidance over this.

twoertwein:
>  Would love some guidance over this.

It looks as if it should work :) I would start with adding `reveal_type(...)` on each line (and break complex statements up) to see which type mypy and pyright infer. The CI stopped after the mypy failure, locally, you can run `poe pyright` to see whether pyright is happy.

ClementPinard:
My vscode was happy so I guess pyright was happy.

However, mypy do fail on local too.

I'll do some additional investigation.

ClementPinard:
Interesting take with reveal_types :

```python
def test_types_groupby_iter() -> None:
    s = pd.Series([1, 1, 2])
    series_groupby: pd.Series[bool] = pd.Series([True, True, False])
    first_group = next(iter(s.groupby(series_groupby)))
    reveal_type(s.groupby(series_groupby))
    reveal_type(s.groupby(series_groupby).__iter__())
    reveal_type(iter(s.groupby(series_groupby)))
```

Outputs

```
tests/test_series.py:738: note: Revealed type is "pandas.core.groupby.generic.SeriesGroupBy[Any, builtins.bool]"
tests/test_series.py:739: note: Revealed type is "typing.Iterator[Tuple[builtins.bool, pandas.core.series.Series[Any]]]"
tests/test_series.py:740: note: Revealed type is "typing.Iterator[Any]"
```

How can the type of `iter(something)` be different from `something.__iter__` ? Is there some double check I should do ?

Even more puzzling is that the type is right for DataFrames :

```python
def test_types_groupby_iter() -> None:
    df = pd.DataFrame(data={"col1": [1, 1, 2], "col2": [3, 4, 5]})
    series_groupby: pd.Series[bool] = pd.Series([True, True, False])
    first_group = next(iter(df.groupby(series_groupby)))
    reveal_type(df.groupby(series_groupby))
    reveal_type(df.groupby(series_groupby).__iter__())
    reveal_type(iter(df.groupby(series_groupby)))
```

output

```
tests/test_frame.py:992: note: Revealed type is "pandas.core.groupby.generic.DataFrameGroupBy[builtins.bool]"
tests/test_frame.py:993: note: Revealed type is "typing.Iterator[Tuple[builtins.bool, pandas.core.frame.DataFrame]]"
tests/test_frame.py:994: note: Revealed type is "typing.Iterator[Tuple[builtins.bool, pandas.core.frame.DataFrame]]"
```

But I don't see any particular additional type hint `DataFrameGroupBy` regarding iteration compared to `SeriesGroupBy` ...

Dr-Irv:
> How can the type of `iter(something)` be different from `something.__iter__` ? Is there some double check I should do ?

I think the issue is that when you are doing `s.groupby(series_groupby)`, `s` is `Series[Any]` because there was no `dtype` information supplied.  If you had created `s=pd.Series([1, 1, 2], dtype=int)`, then you'd get the results you expect.

NOTE - my ability to help out will be pretty limited between now (5/24) and 6/4 due to vacation.

ClementPinard:
You are right, mypy now works !

Did not know the dtype was useful to help the type checker.

twoertwein:
> NOTE - my ability to help out will be pretty limited between now (5/24) and 6/4 due to vacation.

I will merge low-risk PRs during that timeframe.

pandas-stubs/_typing.pyi
```diff
@@ -427,12 +427,28 @@ ByT = TypeVar(
     Interval[Timedelta],
     tuple,
 )
+SeriesByT = TypeVar(
```
twoertwein, line 430: Can you please add a comment here that `SeriesByT` is the intersection of `S1` and `ByT`. If either of them is adjusted in the future, it is then also clear that `SeriesByT` needs to be adjusted.

gandhis1:
> > How can the type of `iter(something)` be different from `something.__iter__` ? Is there some double check I should do ?
> 
> I think the issue is that when you are doing `s.groupby(series_groupby)`, `s` is `Series[Any]` because there was no `dtype` information supplied. If you had created `s=pd.Series([1, 1, 2], dtype=int)`, then you'd get the results you expect.
> 
> NOTE - my ability to help out will be pretty limited between now (5/24) and 6/4 due to vacation.

Maybe this is deserving of a separate issue, but shouldn't we be able to create an initializer overload that takes a `list[T]` and returns a `pd.Series[T]`?

Dr-Irv:
> Maybe this is deserving of a separate issue, but shouldn't we be able to create an initializer overload that takes a `list[T]` and returns a `pd.Series[T]`?

Can you create a separate issue?  I think this might work.

tests/test_frame.py
```diff
@@ -985,6 +985,20 @@ def test_types_groupby_any() -> None:
...
+def test_types_groupby_iter() -> None:
+    df = pd.DataFrame(data={"col1": [1, 1, 2], "col2": [3, 4, 5]})
+    series_groupby = pd.Series([True, True, False], dtype=bool)
+    first_group = next(iter(df.groupby(series_groupby)))
+    check(
+        assert_type(first_group[0], "bool"),
+        bool,
+    )
+    check(
+        assert_type(first_group[1], "pd.DataFrame"),
```
Dr-Irv, line 997: You don't need quotes around the types in the two `assert_type()` statements, because those are valid types at runtime.

tests/test_series.py
```diff
@@ -731,6 +731,20 @@ def test_types_group_by_with_dropna_keyword() -> None:
     s.groupby(level=0).sum()
 
 
+def test_types_groupby_iter() -> None:
+    s = pd.Series([1, 1, 2], dtype=int)
+    series_groupby = pd.Series([True, True, False], dtype=bool)
+    first_group = next(iter(s.groupby(series_groupby)))
+    check(
+        assert_type(first_group[0], "bool"),
```
Dr-Irv, line 739: no quotes needed here

tests/test_series.py
```diff
@@ -731,6 +731,20 @@ def test_types_group_by_with_dropna_keyword() -> None:
...
+        bool,
+    )
+    check(
+        assert_type(first_group[1], "pd.Series[int]"),
+        pd.Series,
```
Dr-Irv, line 744:
```suggestion
        assert_type(first_group[1], "pd.Series[int]"),
        pd.Series, np.integer
```

You do need quotes here because `pd.Series[int]` is not a type at runtime.

Dr-Irv (changes requested):
A couple small things from me, plus address issue to add comment about `SeriesByT`

ClementPinard:
Thank you for the review, I do believe I have addressed the comments

pandas-stubs/_typing.pyi
```diff
@@ -427,12 +427,30 @@ ByT = TypeVar(
     Interval[Timedelta],
     tuple,
 )
+# Use a distinct SeriesByT when using groupby with Series of known dtype.
+# Essentially, an intersection between Series S1 TypeVar, and ByT TypeVar
+SeriesByT = TypeVar(
+    "SeriesByT",
+    str,
```
Dr-Irv, line 434: We've been recently advised to no longer use constrained `TypeVars`, so can you change this to `SeriesByT = TypeVar("SeriesByT", bound=str | bytes | ...)`  We made the change for `S1` in a recent PR

Dr-Irv (changes requested):
Because of a recent change to the definition of `S1`, can you merge with `main`, address the comment below, and if the tests pass, we should be good to go.

pandas-stubs/_typing.pyi
ClementPinard: Hi,

What is the difference between `TypeVar("SeriesByT", bound = str | bytes ...)` and `TypeVar("SeriesByT, str, byts, ...)`  ?

I tried to replace the SeriesByT with the `bound` keyword, but then the mypy did not pass anymore :

```
Poe => mypy

pandas-stubs/core/series.pyi:648: error: Type variable "SeriesByT" not valid as type argument value for "SeriesGroupBy"  [type-var]
pandas-stubs/core/frame.pyi:1100: error: Type variable "SeriesByT" not valid as type argument value for "DataFrameGroupBy"  [type-var]
Found 2 errors in 2 files (checked 224 source files)

Step: 'Run mypy on 'tests' (using the local stubs) and on the local stubs' failed!

```

twoertwein: `bound` allows the type to be a union of all the listed types. Without it, the type has to be one of the listed (sub-)types.

You probably also need to adjust the type used for `SeriesGroupBy` and `DataFrameGroupBy` to use `bound`.

ClementPinard: Thanks I actually tried that and got rid of the errors.

Dr-Irv (approved):
thanks @ClementPinard

PR #686, opened 2023-05-07
pd.RangeIndex should accept range as the first argument

kshitiz305:
- [x] Closes #677  (Replace xxxx with the Github issue number)
- [x] Tests added: Please use `assert_type()` to assert the type of any return value

PR #709, opened 2023-05-24
pin pyright to 1.1.308

Dr-Irv:
A PR #686 revealed that pyright 1.1.309 picks up a variety of bugs that we need to fix, some of which may be bugs `pyright` introduced, so this pins pyright to version 1.1.308.  

I've pushed a branch `fixespyright309` that does some fixes for this, but I won't have time to work on this for a few weeks due to an impending vacation.  With that branch, I did one `#pyright ignore` for something I couldn't figure out in `test_frame.py`, but `test_series.py` is reporting 76 errors on that branch.


twoertwein:
Thanks @Dr-Irv !

PR #716, opened 2023-05-29
Support for pyright 1.1.310

### LINKED ISSUE #711: pyright above 1.1.308 has failures on our tests
See https://github.com/microsoft/pyright/issues/5178

Using the released version from https://github.com/pandas-dev/pandas-stubs/releases/tag/v2.0.1.230501

- Version 1.1.308 reported zero errors
- Version 1.1.309 reported 171 errors
- Version 1.1.310 reported 36 errors

Until https://github.com/microsoft/pyright/issues/5178 is resolved, need to keep pyright at 1.1.308.  There is the possibility that the newer versions are picking up real issues with our stubs that we need to address, but I'm doubtful.

We may see people using pyright in VS Code reporting issues that are pyright bugs, not our bugs.

Dr-Irv:
- [x] Closes #711 
- [ ] Tests added:  None

See discussion in https://github.com/microsoft/pyright/issues/5178

Recommendation was to return `Self` instead of `Series[S1]` .  But if you do that when we don't know the type, it breaks `mypy`.  So if we don't know the type of the `Series`, we return `Series` and that only broke a test related to pandera sublcassing, but only for `pyright`, so I modified that test to make `pyright` happy.


pandas-stubs/_typing.pyi
```diff
@@ -310,25 +310,26 @@ UsecolsArgType: TypeAlias = (
...
-    Dtype,
-    Timestamp,
-    Timedelta,
-    Period,
-    Interval[int],
-    Interval[float],
-    Interval[Timestamp],
-    Interval[Timedelta],
-    CategoricalDtype,
+    bound=str
```
Dr-Irv, line 316: See point #3 at https://github.com/microsoft/pyright/issues/5178#issuecomment-1566190533 ,  Recommended to use a `Union` rather than a constrained type var.

pandas-stubs/core/series.pyi
```diff
@@ -302,16 +303,21 @@ class Series(IndexOpsMixin, NDFrame, Generic[S1]):
...
-        | dict[int, S1]
-        | dict[_str, S1]
-        | None = ...,
+        data: Series[S1] | dict[int, S1] | dict[_str, S1] = ...,
+        index: Axes | None = ...,
+        dtype=...,
+        name: Hashable | None = ...,
+        copy: bool = ...,
+        fastpath: bool = ...,
+    ) -> Self: ...
```
Dr-Irv, line 316: This overload was previously connected to the next one in terms of the `data` argument, but I split it into two to have it return `Self` when `S1` is known, and to return `Series` if not.

pandas-stubs/core/series.pyi
```diff
@@ -2008,16 +2014,16 @@ class TimestampSeries(Series[Timestamp]):
     # ignore needed because of mypy
     @property
     def dt(self) -> TimestampProperties: ...  # type: ignore[override]
-    def __add__(self, other: TimedeltaSeries | np.timedelta64) -> TimestampSeries: ...  # type: ignore[override]
-    def __radd__(self, other: TimedeltaSeries | np.timedelta64) -> TimestampSeries: ...  # type: ignore[override]
+    def __add__(self, other: TimedeltaSeries | np.timedelta64 | timedelta) -> TimestampSeries: ...  # type: ignore[override]
+    def __radd__(self, other: TimedeltaSeries | np.timedelta64 | timedelta) -> TimestampSeries: ...  # type: ignore[override]
```
Dr-Irv, line 2018: Apparently pyright 1.1.310 had bugs that revealed we were missing some types here

tests/test_frame.py
```diff
@@ -2598,15 +2598,15 @@ def test_align() -> None:
         columns=["A", "B", "C"],
     )
 
-    s0 = pd.Series(data={1: "1", 3: "3", 5: "5"})
+    s0 = pd.Series(data={0: "1", 3: "3", 5: "5"})
```
Dr-Irv, line 2601: Without this change, the first element is `np.nan`, so this allows us to see that the first element is a string 3 lines below.

tests/test_series.py
```diff
@@ -1604,12 +1605,14 @@ def test_pandera_generic() -> None:
     T = TypeVar("T")
 
     class MySeries(pd.Series, Generic[T]):
-        ...
+        def __new__(cls, *args, **kwargs) -> Self:
+            return object.__new__(cls)
```
Dr-Irv, line 1609: This makes `pyright` not complain.

tests/test_series.py
```diff
@@ -1604,12 +1605,14 @@ def test_pandera_generic() -> None:
...
 
     def func() -> MySeries[float]:
         return MySeries[float]([1, 2, 3])
 
-    func()
+    result = func()
+    assert result.iloc[1] == 2
```
Dr-Irv, line 1615: I modified the test to make sure that defining `__new__()` didn't change the result.

pandas-stubs/_typing.pyi
twoertwein: Does this also affect other TypeVars (could be done later)?

twoertwein: Can we use `S1 = TypeVar("S1", bound=Scalar)`? You replaced `S1` with `Scalar` in some places

twoertwein:
```py
 S1 = TypeVar(
     "S1",
    bound=Scalar
     | Interval[int]
     | Interval[float]
     | Interval[Timestamp]
     | Interval[Timedelta]
    | Dtype
    | Period
    | CategoricalDtype
    | datetime.time,
 )
```

Seems to work (but Scalar includes a few more types!)

pandas-stubs/core/series.pyi
twoertwein: Can the two Self overloads be combined?

tests/test_frame.py
twoertwein: I don't understand how is `1` inferred as `np.nan`?

pandas-stubs/_typing.pyi
Dr-Irv: > Does this also affect other TypeVars (could be done later)?

Yes, and I created an issue for that #719

tests/test_frame.py
Dr-Irv: In the code just below the above, we call `DataFrame.align()`.  The specified `DataFrame` does not have the index `1`, so the `align` call puts `np.nan` in the first element of the result.

Part of the reason this change was needed is because I changed the test to check the type of the `Series` to `Series[str]` and I wanted `check()` to check the type of the first element.

twoertwein:
Thanks @Dr-Irv !

PR #721, opened 2023-05-31
Fix Series.to_string

### LINKED ISSUE #720: Mypy reports issues with `to_string` on `Series` that seem erroneous
**Describe the bug**
When calling `to_string` on `Series` with `index=False`, mypy reports that this is not a valid variant, although it looks ok.

**To Reproduce**

Minimal example:

```
import pandas as pd

s = pd.Series(["A", "B", "C"])
print(s.to_string())
print(s.to_string(index=False))
```

`mypy example.py`

```
example.py:5: error: No overload variant of "to_string" of "Series" matches argument type "bool"  [call-overload]
example.py:5: note: Possible overload variants:
example.py:5: note:     def to_string(self, buf: Union[str, PathLike[str], WriteBuffer[str]], na_rep: str = ..., formatters: Any = ..., float_format: Any = ..., sparsify: Optional[bool] = ..., index_names: bool = ..., justify: Optional[str] = ..., max_rows: Optional[int] = ..., min_rows: Optional[int] = ..., max_cols: Optional[int] = ..., show_dimensions: bool = ..., decimal: str = ..., line_width: Optional[int] = ..., max_colwidth: Optional[int] = ..., encoding: Optional[str] = ...) -> None
example.py:5: note:     def to_string(self, buf: None = ..., na_rep: str = ..., formatters: Any = ..., float_format: Any = ..., sparsify: Optional[bool] = ..., index_names: bool = ..., justify: Optional[str] = ..., max_rows: Optional[int] = ..., min_rows: Optional[int] = ..., max_cols: Optional[int] = ..., show_dimensions: bool = ..., decimal: str = ..., line_width: Optional[int] = ..., max_colwidth: Optional[int] = ..., encoding: Optional[str] = ...) -> str
```

 - OS: Linux
 - OS Version: Ubuntu 20.04
 - python version: 3.9.14
 - version of type checker: mypy 1.3.0
 - version of pandas: 2.0.1

**Additional context**
Nothing else to add

twoertwein:
- [x] Closes #720 (Replace xxxx with the Github issue number)


pandas-stubs/core/series.pyi
```diff
@@ -453,38 +453,28 @@ class Series(IndexOpsMixin, NDFrame, Generic[S1]):
         self,
         buf: FilePath | WriteBuffer[str],
         na_rep: _str = ...,
-        formatters=...,
         float_format=...,
```
Dr-Irv, line 456: We should fix this to be `Callable[[float], str]` (in both overloads)

Dr-Irv (changes requested):
Let's fix `float_format` here as well.

Dr-Irv (approved):
thanks @twoertwein

Issue #718, opened 2023-05-29
CLEAN:  Investigate whether `TimestampSeries`, `TimedeltaSeries`, etc. can be removed


Dr-Irv:
See comment #2 in  https://github.com/microsoft/pyright/issues/5178#issuecomment-1566190533

> Some of the overloads for pd.Series.__new__ return types other than pd.Series (e.g. TimestampSeries, PeriodSeries). These other types appear nowhere within the pandas library. I presume they were added to the stubs as sort of a hack to expose certain methods that apply only for certain specializations of Series. There are better ways of handling this that more closely model the library's runtime behavior. One of the big problems with introducing these "made-up" subclasses is that Series cannot be properly subclassed. For example, if you create a subclass of Series (let's call it MySeries) but then pass arguments to the constructor that invoke an overload that returns a TimestampSeries, you will now have a TimestampSeries rather than a MySeries. If I were maintaining these stubs, I'd look at eliminating these made-up classes and folding the functionality into the main Series class. You can use a specialized type annotation for self to limit the behavior to a specific specialization of Series. For example, def __add__(self: Series[Timedelta], ....

NOTE:  Those extra types were intrroduced in March, 2022, to make it so that `mypy` would work, because using something like `def __add__(self: Series[Timedelta]) ...` did not work with the version of mypy at that time.  But mypy has evolved, so we should see if they can be removed.

Dr-Irv:
See https://github.com/microsoft/pyright/issues/5216 and https://github.com/python/mypy/issues/15452

loicdiridollou:
We made some more advances with `UnknownSeries`, maybe we should close that issue since we are moving in the opposite direction at the moment, what do you think @Dr-Irv ?

Dr-Irv:
closing via #1273 and #1274

PR #722, opened 2023-05-31
Fix Series.__new__

### LINKED ISSUE #717: CLEAN:  Fix overloads for `Series.__new__()` to be consistent in argument ordering.
See comment 1 at https://github.com/microsoft/pyright/issues/5178#issuecomment-1566190533

> The pd.Series.__new__ method has some pretty glaring bugs. Some of the overrides swap the index and dtype parameters. I presume this was done to accommodate the case where the dtype argument is specified by keyword. This can be handled properly by retaining the correct parameter order and adding a keyword-only separator (*). Let me know if you need more details. I definitely recommend addressing this bug because it will lead to incorrect type evaluations.

twoertwein:
- [x] Closes #717 (Replace xxxx with the Github issue number)

I only added `*` for one overload as the implementation does not requires keyword-only arguments.

I also removed `fastpath` as it is not documented.


pandas-stubs/core/series.pyi
```diff
@@ -208,55 +209,52 @@ class Series(IndexOpsMixin, NDFrame, Generic[S1]):
     _ListLike: TypeAlias = ArrayLike | dict[_str, np.ndarray] | list | tuple | Index
     __hash__: ClassVar[None]
 
+    # TODO: can __new__ be converted to __init__? Pandas implements __init__
```
twoertwein, line 212: Can probably be done if #718 is possible

pandas-stubs/core/series.pyi
```diff
@@ -208,55 +209,52 @@ class Series(IndexOpsMixin, NDFrame, Generic[S1]):
...
     @overload
     def __new__(
         cls,
         data: DatetimeIndex | Sequence[Timestamp | np.datetime64 | datetime],
         index: Axes | None = ...,
-        dtype=...,
+        dtype: Dtype = ...,
```
Dr-Irv, line 218: If `data` represents times, then the values of `Dtype` are restricted to be `TimestampDtypeArg`

pandas-stubs/core/series.pyi
```diff
@@ -208,55 +209,52 @@ class Series(IndexOpsMixin, NDFrame, Generic[S1]):
...
     ) -> TimestampSeries: ...
     @overload
-    def __new__(
+    def __new__(  # type: ignore[misc]
         cls,
         data: _ListLike,
-        dtype: Literal["datetime64[ns]"],
         index: Axes | None = ...,
+        *,
+        dtype: Literal["datetime64[ns]"],
```
Dr-Irv, line 228: Probably should be `TimestampDtypeArg`

pandas-stubs/core/series.pyi
```diff
@@ -208,55 +209,52 @@ class Series(IndexOpsMixin, NDFrame, Generic[S1]):
...
         copy: bool = ...,
-        fastpath: bool = ...,
     ) -> TimestampSeries: ...
     @overload
     def __new__(
         cls,
         data: PeriodIndex,
         index: Axes | None = ...,
-        dtype=...,
+        dtype: Dtype = ...,
```
Dr-Irv, line 237: Seems like we need to create a `PeriodDtypeArg` in `_typing.pyi` with the valid values here

pandas-stubs/core/series.pyi
```diff
@@ -208,55 +209,52 @@ class Series(IndexOpsMixin, NDFrame, Generic[S1]):
...
         copy: bool = ...,
-        fastpath: bool = ...,
     ) -> PeriodSeries: ...
     @overload
     def __new__(
         cls,
         data: TimedeltaIndex | Sequence[Timedelta | np.timedelta64 | timedelta],
         index: Axes | None = ...,
-        dtype=...,
+        dtype: Dtype = ...,
```
Dr-Irv, line 246: Use `TimedeltaDtypeArg`

pandas-stubs/core/series.pyi
```diff
@@ -265,10 +263,9 @@ class Series(IndexOpsMixin, NDFrame, Generic[S1]):
         | Interval[float]
         | Sequence[Interval[float]],
         index: Axes | None = ...,
-        dtype=...,
+        dtype: Dtype = ...,
```
Dr-Irv, line 266:
```suggestion
        dtype: Literal["Interval"]= ...,
```

pandas-stubs/core/series.pyi
```diff
@@ -277,10 +274,9 @@ class Series(IndexOpsMixin, NDFrame, Generic[S1]):
         | Interval[Timestamp]
         | Sequence[Interval[Timestamp]],
         index: Axes | None = ...,
-        dtype=...,
+        dtype: Dtype = ...,
```
Dr-Irv, line 277:
```suggestion
        dtype: Literal["interval"]= ...,
```

pandas-stubs/core/series.pyi
```diff
@@ -289,10 +285,9 @@ class Series(IndexOpsMixin, NDFrame, Generic[S1]):
         | Interval[Timedelta]
         | Sequence[Interval[Timedelta]],
         index: Axes | None = ...,
-        dtype=...,
+        dtype: Dtype = ...,
```
Dr-Irv, line 288:
```suggestion
        dtype: Literal["interval"]= ...,
```

Dr-Irv (changes requested):
For `astype`, we created valid argument types for various types.  It may not be fully inclusive, but we can do some restrictions in various places.

pandas-stubs/core/series.pyi
```diff
@@ -208,92 +210,55 @@ class Series(IndexOpsMixin, NDFrame, Generic[S1]):
...
-        copy: bool = ...,
-        fastpath: bool = ...,
-    ) -> IntervalSeries[Timestamp]: ...
-    @overload
-    def __new__(
-        cls,
-        data: IntervalIndex[Interval[Timedelta]]
-        | Interval[Timedelta]
-        | Sequence[Interval[Timedelta]],
+        data: IntervalIndex[Interval[_OrderableT]]
```
twoertwein, line 254: Needed to switch to `bound` to simplify these overloads. All the other changes are related to this change.

pandas-stubs/core/series.pyi
twoertwein: I think all the other overloads would also need that: one overload with a specific `data` and one for `_ListLike` but a specific `dtype`?

Dr-Irv: My concern there would be that it would be too wide.  In other words, there would be contents of lists that wouldn't work with certain dtypes.  I think the one is in there for `TimestampSeries` because it handles any kind of list.

Dr-Irv (changes requested):
All looks good.  Can you merge to `main` to resolve conflicts?

pandas-stubs/core/series.pyi
Dr-Irv: > Probably should be `TimestampDtypeArg`

I think this is the only change left to make.

twoertwein: changed

Dr-Irv (approved):
thanks @twoertwein

PR #724, opened 2023-06-06
Argument 1 to "apply" of "Series" has incompatible type "Callable[[Any], Union[int, NAType]]"

### LINKED ISSUE #681: Argument 1 to "apply" of "Series" has incompatible type "Callable[[Any], Union[int, NAType]]"
**To Reproduce**
```python
import pandas as pd

# these values are allowed
allowed = ["A", "B", "C"]

df = pd.DataFrame()

# we add a few that are not allowed
df["Column"] = ["A", "B", "C", "D"]

# replace values by index, and NA if not allowed
df["Label"] = df.Column.apply(lambda x: allowed.index(x) if x in allowed else pd.NA)

print(df.Label)
```

`mypy 1.2.0 (compiled: yes)`

```
bug.py:12: error: Argument 1 to "apply" of "Series" has incompatible type "Callable[[Any], Union[int, NAType]]"; expected "Callable[..., Union[str, bytes, date, datetime, timedelta, datetime64, timedelta64, bool, int, float, Timestamp, Timedelta, integer[Any], floating[Any], complex, Sequence[Any], Set[Any], Mapping[Any, Any]]]"  [arg-type]
bug.py:12: error: Incompatible return value type (got "Union[int, NAType]", expected "Union[str, bytes, date, datetime, timedelta, datetime64, timedelta64, bool, int, float, Timestamp, Timedelta, integer[Any], floating[Any], complex, Sequence[Any], Set[Any], Mapping[Any, Any]]")  [return-value]
```

 - OS: Windows 11
 - OS Version: 22H2
 - python version: 3.10.11
 - version of type checker: 1.2.0

Related: #262?

ABizzinotto:
- [x] Closes #681 (Replace xxxx with the Github issue number)

### DIFF
--- a/pandas-stubs/core/series.pyi
+++ b/pandas-stubs/core/series.pyi
@@ -851,7 +851,7 @@ class Series(IndexOpsMixin, NDFrame, Generic[S1]):
     @overload
     def apply(
         self,
-        func: Callable[..., Scalar | Sequence | set | Mapping | None],
+        func: Callable[..., Scalar | Sequence | set | Mapping | NAType | None],
         convertDType: _bool = ...,
         args: tuple = ...,
         **kwds,
--- a/tests/test_series.py
+++ b/tests/test_series.py
@@ -35,6 +35,7 @@ from typing_extensions import (
 )
 import xarray as xr
 
+from pandas._libs.missing import NAType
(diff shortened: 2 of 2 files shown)


Dr-Irv (approved):
thanks @ABizzinotto

PR #728, opened 2023-06-12
period_range and Period should accept date, datetime and pd.Timestamp

### LINKED ISSUE #675: `period_range` and `Period` should accept `date`, `datetime` and `pd.Timestamp`
`pd.period_range` and `pd.Period` currently only accept `str` and `pd.Period` as arguments, but the implementation also accepts `datetime`, `date` and `pd.Timestamp` as can be seen here in the cython implementation: https://github.com/pandas-dev/pandas/blob/v2.0.1/pandas/_libs/tslibs/period.pyx#L2626-L2643

It technically also allows integers/floats. `period_range` internally uses `Period` to convert the `start` and `end` arguments, so the type annotations can be used interchangeably between those two.

**To Reproduce**
```python
import pandas as pd

pd.Period(pd.Timestamp('2020-01-01'), freq='Q')  # mypy: arg-type — Argument 1 to "Period" has incompatible type "Timestamp"
pd.period_range(pd.Timestamp('2020-01-01), '2020-06-01', freq='Q')  # mypy: arg-type — Argument 1 to "period_range" has incompatible type "Timestamp"
```

 - Python 3.10
 - mypy 1.0.1
 - pandas-stubs 2.0.1.230501

ABizzinotto:
- [x] Closes #675 (Replace xxxx with the Github issue number)


ABizzinotto:
@Dr-Irv looks like this was indeed a documentation issue, I submitted https://github.com/pandas-dev/pandas/pull/53611 and https://github.com/pandas-dev/pandas/pull/53632 to address the issue reported there (https://github.com/pandas-dev/pandas/issues/53038). The first one has been merged, but the solution was incomplete to I submitted the second one (pending review).

pandas-stubs/_libs/tslibs/period.pyi
```diff
@@ -5,6 +5,7 @@ from typing import (
 )
 
 import numpy as np
+import pandas
```
Dr-Irv, line 8: This is not needed and is causing CI failures.

pandas-stubs/core/series.pyi
```diff
@@ -851,7 +851,7 @@ class Series(IndexOpsMixin, NDFrame, Generic[S1]):
     @overload
     def apply(
         self,
-        func: Callable[..., Scalar | Sequence | set | Mapping | None],
+        func: Callable[..., Scalar | Sequence | set | Mapping | NAType | None],
```
Dr-Irv, line 854: I think this change came up because you haven't merged with `main`.  Can you make sure you have done that?

tests/test_series.py
```diff
@@ -470,6 +471,8 @@ def get_depth(url: str) -> int:
     ss = s.astype(str)
     check(assert_type(ss.apply(get_depth), pd.Series), pd.Series, np.int64)
 
+    check(assert_type(s.apply(lambda x: pd.NA), pd.Series), pd.Series, NAType)
```
Dr-Irv, line 474: same comment as above - this change is already in `main`

tests/test_timefuncs.py
```diff
@@ -838,6 +838,27 @@ def test_period_range() -> None:
...
+        ),
+        pd.PeriodIndex,
+    )
+    check(
+        assert_type(
+            pd.period_range(dt.date(2001, 1, 1), end="2002-01-01", freq="Q"),
+            pd.PeriodIndex,
+        ),
+        pd.PeriodIndex,
+    )
```
Dr-Irv, line 861: For these tests, you should also change the `end` argument to be the corresponding type.

Dr-Irv (changes requested):

tests/test_timefuncs.py
ABizzinotto: Done

Dr-Irv (approved):
thanks @ABizzinotto

PR #739, opened 2023-07-02
Add overload for DataFrameGroupBy.groupby("size") return Series

### LINKED ISSUE #736: DataFrameGroupBy.aggregate can return Series with argument "size", typed as only returning DataFrame
**Describe the bug**
See title, and following line.
https://github.com/pandas-dev/pandas-stubs/blob/6fe90bbb37128a79068ba41c8ebaaf269ddeab90/pandas-stubs/core/groupby/generic.pyi#L162

**To Reproduce**

MRE:

```py
import pandas as pd

df = pd.DataFrame(
    [[pd.NA], [pd.NA], [pd.NA]],
    index=[0, 1, 1],
)

gdf = df.groupby(level=0)

s = gdf.aggregate("size")

reveal_type(s)  # DataFrame
type(s)  # Series
```

Type checker = mypy 1.4.1

When I use `s > 1` for indexing in `loc` I get:

```py
Invalid index type "tuple[DataFrame, str]" for "_LocIndexerFrame"; expected type "Series[bool] | ndarray[Any, dtype[bool_]] | list[bool] | str | str_ | tuple[Index | Series[bool] | ndarray[Any, dtype[bool_]] | list[bool] | str | bytes | date | datetime | timedelta | datetime64 | timedelta64 | bool | int | float | Timestamp | Timedelta | complex | list[Any] | slice | tuple[str | bytes | date | datetime | timedelta | datetime64 | timedelta64 | bool | int | float | Timestamp | Timedelta | complex, ...], ...] | list[<nothing>]"  [index]
```

 - MacOS 13.4.1 (22F82)
 - Python 3.10.11
 - Mypy 1.4.1
 - pandas-stubs-2.0.2.230605

ljmc-github:
Fairly straightforward, I had to add ignores for mypy and pyright since `"size"` is included in `AggFuncTypeFrame`'s `str`. 

I made as few changes as possible leaving `agg = aggregate` which seems to work, although I saw the `SeriesGroupBy.agg` is a redeclaration rather than assignment, I am happy to switch if necessary.

https://github.com/pandas-dev/pandas-stubs/blob/6fe90bbb37128a79068ba41c8ebaaf269ddeab90/pandas-stubs/core/groupby/generic.pyi#L53-L60

- [x] Closes #736


tests/test_frame.py
```diff
@@ -880,6 +880,10 @@ def test_types_groupby() -> None:
     s1: pd.Series = df.set_index("col1")["col2"]
     s2: pd.Series = s1.groupby("col1").transform("sum")
 
+    # GH 736
+    check(assert_type(df1.groupby(by="col1").aggregate("size"), pd.Series), pd.Series)
+    check(assert_type(df1.groupby(by="col1").agg("size"), pd.Series), pd.Series)
+
```
Dr-Irv, line 886: Can you move these tests to the function `test_types_groupby_agg()` ?

pandas-stubs/core/groupby/generic.pyi
```diff
@@ -159,6 +159,10 @@ class DataFrameGroupBy(GroupBy, Generic[ByT]):
     def apply(  # pyright: ignore[reportOverlappingOverload]
         self, func: Callable[[Iterable], float], *args, **kwargs
     ) -> DataFrame: ...
+    # error: overload 1 overlaps overload 2 as "size" is in AggFuncTypeFrame
```
Dr-Irv, line 162: Better to say `# error: overload 1 overlaps overload 2 because of different return types`

Dr-Irv (changes requested):
Thanks for the PR.  You wrote:
> although I saw the SeriesGroupBy.agg is a redeclaration rather than assignment, I am happy to switch if necessary.

Actually, this was an oversight from a previous PR 10 months ago.  I think I'd rather use assignment in `SeriesGroupBy`, so can you make that change there so we are consistent?  You don't have to add a test for that.

tests/test_frame.py
ljmc-github: I have, what about the other aggregates/transforms in `test_types_groupby` ?

https://github.com/pandas-dev/pandas-stubs/blob/6fe90bbb37128a79068ba41c8ebaaf269ddeab90/tests/test_frame.py#L870-L881

Dr-Irv: If you want, you could add extra tests in another PR, but let's leave this PR to focus on just the one issue.  Thanks.

Dr-Irv (approved):
thanks @ljmc-github

PR #746, opened 2023-07-18
Add missing __hash__ stubs

TheTripleV:
These missing `__hash__` methods are in the pandas stubs but are missing here. The missing `__hash__`es incorrectly marked these classes as being unhashable.

https://github.com/microsoft/pyright/issues/5513#issuecomment-1638382323


pandas-stubs/_libs/tslibs/timedeltas.pyi
```diff
@@ -386,6 +386,7 @@ class Timedelta(timedelta):
     ) -> npt.NDArray[np.bool_]: ...
     @overload
     def __gt__(self, other: TimedeltaSeries | Series[pd.Timedelta]) -> Series[bool]: ...
+    def __hash__(self) -> int: ...
```
twoertwein, line 389: Can you please explain why this is needed: Timedelta inherits from `datetime.timedelta` which declares to have `__hash__` see [typeshed](https://github.com/python/typeshed/blob/4088b8425fa71457b9b13166e23ee3dee88a5704/stdlib/datetime.pyi#L227)

TheTripleV: According to the stubs, `Timedelta` does not inherit `__hash__` from `datetime.timedelta` because `Timedelta` defines `__eq__` (https://github.com/pandas-dev/pandas-stubs/pull/746/files#diff-0321b1cf2f2c79264daae754a98c26bcac7040336168706d7f12408919c01ad4L313).

Classes that define `__eq__` do not inherit `__hash__`. [Python Docs](https://docs.python.org/3/reference/datamodel.html#object.__hash__:~:text=if%20it%20defines%20__eq__()%20but%20not%20__hash__()%2C%20its%20instances%20will%20not%20be%20usable%20as%20items%20in%20hashable%20collections)

twoertwein: okay, I understand it now, if `__eq__` is overwritten, `__hash__` is set to `None` unless it is explicitly declared.

can you please add two quick tests for this?

Dr-Irv: @twoertwein this is in the pandas repo, so I see no reason not to accept this.

Dr-Irv (approved):
will merge this pending answer from @twoertwein

pandas-stubs/_libs/tslibs/timedeltas.pyi
Dr-Irv: > I'm for accepting, but typically we like to have tests?

Yes, but in this case, I'm OK to skip the tests.

PR #756, opened 2023-07-21
added pyarrow/numpy dtype literals and allowed `str` | `DtypeObj` as input for `Series.astype`

### LINKED ISSUE #733: error: Argument 1 to "astype" of "DataFrame" has incompatible type "Literal['int32[pyarrow]']";
```python
import pandas as pd

df = pd.DataFrame([1,23])
df = df.astype("int32[pyarrow]")
```

- pandas                        2.0.2
- pandas-stubs                  2.0.2.230605

**Additional context**

### LINKED ISSUE #747: error: Argument 1 to "astype" of "Series" has incompatible type "dtype[generic] | ExtensionDtype"; expected "type[object] | ExtensionDtype"  [arg-type]
**Describe the bug**

`Series.dtype` is incompatible with `Series.astype`

**To Reproduce**

```python
import pandas as pd

s = pd.Series([1, 2, 3])
s = s.astype(s.dtype)  # ✘ [arg-type]
```

 - OS: Ubuntu 22.04
 - python version 3.10
 - mypy 1.4.1
- pandas                        2.0.3
- pandas-stubs                  2.0.2.230605

randolf-scholz:
- [x] Closes #733
- [x] Closes #747

- pandas nullable `UInt` data types were missing
- numpy type code literals
- numpy alternative literals (half, short, double, etc.)
- pyarrow literals
- added `DtypeObj | _str` to final `astype` overload (resulting in `Series[Any]`)
- added few tests
- fixed random bug: `def __getattr__(self, name: str) -> S1` accidentally using `str` instead of `_str`.

### DIFF
--- a/pandas-stubs/_typing.pyi
+++ b/pandas-stubs/_typing.pyi
@@ -78,6 +78,17 @@ NpDtype: TypeAlias = str | np.dtype[np.generic] | type[str | complex | bool | ob
 Dtype: TypeAlias = ExtensionDtype | NpDtype
 DtypeArg: TypeAlias = Dtype | Mapping[Any, Dtype]
 DtypeBackend: TypeAlias = Literal["pyarrow", "numpy_nullable"]
+
+# NOTE: we want to catch all the possible dtypes from np.sctypeDict
+# timedelta64
+# M
+# m8
+# M8
+# object_
+# object0
+# m
+# datetime64
+
 BooleanDtypeArg: TypeAlias = (
     # Builtin bool type and its string alias
     type[bool]  # noqa: Y030
@@ -88,7 +99,7 @@ BooleanDtypeArg: TypeAlias = (
     # Numpy bool type
     # https://numpy.org/doc/stable/reference/arrays.scalars.html#numpy.bool_
     | type[np.bool_]
-    | Literal["?", "bool8", "bool_"]
+    | Literal["?", "b1", "bool8", "bool_"]
     # PyArrow boolean type and its string alias
     | Literal["bool[pyarrow]", "boolean[pyarrow]"]
 )
@@ -102,73 +113,56 @@ IntDtypeArg: TypeAlias = (
     | pd.Int32Dtype
     | pd.Int64Dtype
     | Literal["Int8", "Int16", "Int32", "Int64"]
+    # Numpy signed integer types and their string aliases
+    # https://numpy.org/doc/stable/reference/arrays.scalars.html#numpy.byte
+    | type[np.byte]
+    | Literal["b", "i1", "int8", "byte"]
+    # https://numpy.org/doc/stable/reference/arrays.scalars.html#numpy.short
+    | type[np.short]
+    | Literal["h", "i2", "int16", "short"]
+    # https://numpy.org/doc/stable/reference/arrays.scalars.html#numpy.intc
+    | type[np.intc]
+    | Literal["i", "i4", "int32", "intc"]
+    # https://numpy.org/doc/stable/reference/arrays.scalars.html#numpy.int_
+    | type[np.int_]
+    | Literal["l", "i8", "int64", "int0", "int_", "long"]
+    # https://numpy.org/doc/stable/reference/arrays.scalars.html#numpy.longlong
+    | type[np.longlong]
+    | Literal["q", "longlong"]  # NOTE: int128 not assigned
+    # https://numpy.org/doc/stable/reference/arrays.scalars.html#numpy.intp
+    | type[np.intp]  # signed pointer (=`intptr_t`, platform dependent)
+    | Literal["p", "intp"]
+    # PyArrow integer types and their string aliases
+    | Literal["int8[pyarrow]", "int16[pyarrow]", "int32[pyarrow]", "int64[pyarrow]"]
+)
+UIntDtypeArg: TypeAlias = (
     # Pandas nullable unsigned integer types and their string aliases
-    | pd.UInt8Dtype
+    pd.UInt8Dtype  # noqa: Y030
     | pd.UInt16Dtype
     | pd.UInt32Dtype
     | pd.UInt64Dtype
     | Literal["UInt8", "UInt16", "UInt32", "UInt64"]
-    # Numpy signed integer types and their string aliases
-    # https://numpy.org/doc/stable/reference/arrays.scalars.html#numpy.byte
-    | type[np.byte]
-    | Literal["b", "int8", "byte"]
-    # https://numpy.org/doc/stable/reference/arrays.scalars.html#numpy.short
-    | type[np.short]
-    | Literal["h", "int16", "short"]
-    # https://numpy.org/doc/stable/reference/arrays.scalars.html#numpy.intc
-    | type[np.intc]
-    | Literal["i", "int32", "intc"]
-    # https://numpy.org/doc/stable/reference/arrays.scalars.html#numpy.int_
-    | type[np.int_]
-    | Literal["l", "int64", "int_", "intp", "long"]
-    # https://numpy.org/doc/stable/reference/arrays.scalars.html#numpy.longlong
-    | type[np.longlong]
-    | Literal["q", "longlong"]  # NOTE: int128 not assigned
     # Numpy unsigned integer types and their string aliases
     # https://numpy.org/doc/stable/reference/arrays.scalars.html#numpy.ubyte
     | type[np.ubyte]
-    | Literal["B", "uint8", "ubyte"]
+    | Literal["B", "u1", "uint8", "ubyte"]
     # https://numpy.org/doc/stable/reference/arrays.scalars.html#numpy.ushort
     | type[np.ushort]
-    | Literal["H", "uint16", "ushort"]
+    | Literal["H", "u2", "uint16", "ushort"]
     # https://numpy.org/doc/stable/reference/arrays.scalars.html#numpy.uintc
     | type[np.uintc]
-    | Literal["I", "uint32", "uintc"]
+    | Literal["I", "u4", "uint32", "uintc"]
     # https://numpy.org/doc/stable/reference/arrays.scalars.html#numpy.uint
     | type[np.uint]
-    | Literal["L", "uint64", "uint", "uintp"]
+    | Literal["L", "u8", "uint", "ulong", "uint64", "uint0"]
     # https://numpy.org/doc/stable/reference/arrays.scalars.html#numpy.ulonglong
     | type[np.ulonglong]
     | Literal["Q", "ulonglong"]  # NOTE: uint128 not assigned
-    # PyArrow integer types and their string aliases
-    | Literal["int8[pyarrow]", "int16[pyarrow]", "int32[pyarrow]", "int64[pyarrow]"]
+    # https://numpy.org/doc/stable/reference/arrays.scalars.html#numpy.uintp
+    | type[np.uintp]  # unsigned pointer (=`uintptr_t`, platform dependent)
+    | Literal["P", "uintp"]
     # PyArrow unsigned integer types and their string aliases
     | Literal["uint8[pyarrow]", "uint16[pyarrow]", "uint32[pyarrow]", "uint64[pyarrow]"]
-)
-StrDtypeArg: TypeAlias = (
-    # Builtin str type and its string alias
-    type[str]  # noqa: Y030
-    | Literal["str"]
-    # Pandas nullable string type and its string alias
-    | pd.StringDtype
-    | Literal["string"]
-    # Numpy string type and its string alias
-    # https://numpy.org/doc/stable/reference/arrays.scalars.html#numpy.str_
-    | type[np.str_]
-    | Literal["U", "str_", "unicode"]
-    # PyArrow string type and its string alias
-    | Literal["string[pyarrow]"]
-)
-BytesDtypeArg: TypeAlias = (
-    # Builtin bytes type and its string alias
-    type[bytes]  # noqa: Y030
-    | Literal["bytes"]
-    # Numpy bytes type and its string alias
-    # https://numpy.org/doc/stable/reference/arrays.scalars.html#numpy.bytes_
-    | type[np.bytes_]
-    | Literal["S", "bytes_", "string_"]
-    # PyArrow binary type and its string alias
-    | Literal["binary[pyarrow]"]
 )
 FloatDtypeArg: TypeAlias = (
     # Builtin float type and its string alias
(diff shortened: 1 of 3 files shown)


pandas-stubs/_typing.pyi
```diff
@@ -192,11 +263,10 @@ AstypeArg: TypeAlias = (
     | TimedeltaDtypeArg
     | TimestampDtypeArg
     | CategoryDtypeArg
-    | ExtensionDtype
+    | DtypeObj
     | type[object]
+    | str
```
Dr-Irv, line 268: I don't think we should accept a general string here.  The whole idea of the other aliases is to constrain which strings are acceptable.

Dr-Irv (changes requested):
For all the possible arguments you added for `astype()`, can you add an appropriate test in `test_series.test_updated_astype()` ?  We were really careful there to make sure the tests matched all the things as possible arguments.

pandas-stubs/_typing.pyi
randolf-scholz: `str` must be acceptable, as e.g. a string might be provided dynamically. Same is if you accept generic dtype object, which one must if `s.astye(s.dtype)` should work without raising a warning.

The correct way to handle this is to be careful with overload order, going from special to generic.
All of the existing tests still work with this change.

Dr-Irv: It makes the type check too wide.  It allows `pd.DataFrame().astype("foobar")` to pass.  I'd rather have the stubs catch the most used cases (not dynamic strings).  Having said that, if we want `s.astype(s.dtype)` to work, maybe we should just say that `s.dtype()` returns `AsTypeArg`

Dr-Irv:
@randolf-scholz  tests are failing

pandas-stubs/_typing.pyi
gandhis1: If there was a way to specify a non-literal string, that would be ideal.  If it sees a literal, the literal should be verified. I am not aware of a way to achieve that with current typing features.

gandhis1: Another note - I think in subjective cases like this, you have to ask, what is the most common use case. I would contend dynamic/runtime based `astype` calls are atypical. The majority of the time, it's a static transformation, and the developer is aware of the source dtype and target dtype. In that case, they would hard-code the destination type.

randolf-scholz: @Dr-Irv `pd.DataFrame().astype("foobar")` is a `ValueError`, type hints are for avoiding `TypeError`.

randolf-scholz: But I get that catching typos would be very useful.

randolf-scholz:
- added tests for all the types
- added `ObjectDtypeArg`

```python
ObjectDtypeArg: TypeAlias = (
    # Builtin object type and its string alias
    type[object]  # noqa: Y030
    | Literal["object"]
    | type[np.object_]
    | Literal["O"]  # NOTE: "object_" not assigned
)
```

pandas-stubs/_typing.pyi
randolf-scholz: @gandhis1 As an example for dynamic astype(string): I have some data-pipelines for reading csv-files where I store config files with the target schema (e.g. column dtypes).

randolf-scholz: @gandhis1 @Dr-Irv In python 3.11 [`LiteralString`](https://docs.python.org/3/library/typing.html#typing.LiteralString) was added, I wonder id this can be used to raise a typing error if none of the defined literals is matched? This would give best of both worlds.

randolf-scholz: I asked this here: https://github.com/python/typing/discussions/1434

randolf-scholz: Hm, doesn't seem possible currently. Possibly in the future if both `Intersection` and `Not` are supported: https://github.com/python/typing/issues/801#issuecomment-1646171898

Dr-Irv: > @gandhis1 As an example for dynamic astype(string): I have some data-pipelines for reading csv-files where I store config files with the target schema (e.g. column dtypes).

That's an interesting example, but that's a case where the argument you are using is dynamic, and you're trying to check it in a static typing context.  If you config files had an incorrect string `"foobar"`not covered by the stubs, your code would fail at runtime.  

IMHO, the goal of the static type checks is to catch things before runtime.  So we have two alternatives:

1. Do not include `str` (as it is right now in the stubs) as a valid argument.  Disadvantage is that you can't use a dynamic string as an argument as in your use case.  Advantage is it covers only the valid strings that are statically declared.
2. Include `str` as a valid argument.  Disadvantage is that we don't help people who use `.astype("ing")` instead of `.astype("int")` prior to run time.  Advantage is your use case.

I would argue that people use `.astype("some-string")` more than have a dynamic string as the argument.  So that's why I would prefer (1) above.

Dr-Irv (changes requested):
Tests are still failing.

I also think we should add the following test:
```python
        s.astype("foobar")  # type: ignore[call-arg] # pyright: ignore[reportGeneralTypeIssues]
```
This will check that we are not allowing arbitrary strings.

randolf-scholz:
@Dr-Irv I removed the `str` overload for now, tests should pass. But I think this still should be discussed, maybe if LiteralString is supported by typecheckers one can do some overload trickery.

Apart from that, since I added `ObjectDtypeArg` one might consider adding `object` to `S1`. (and I'd suggest renaming `S1` to something more descriptive like `SeriesKind`).

Also as a side questions: why do `TimestampSeries` and `TimedeltaSeries`, etc. exist? Why don't  `Series[Timestamp]` / `Series[Timestamp]` work?

Dr-Irv:
> @Dr-Irv I removed the `str` overload for now, and added the test you suggested. But I think this still should be discussed, possibly when `LiteralString` is supported by typecheckers one can do some overload trickery to get both dynamical string and typo-checking to work.

I'm still seeing `AsTypeArg` including `str`, so will ask for that to be removed.
> 
> Apart from that, since I added `ObjectDtypeArg` one might consider adding `object` to `S1`. 

Let's handle in a separate PR.   I think if we do that there is a possibility of other errors getting in.  

> (and I'd suggest renaming `S1` to something more descriptive like `SeriesKind`).

That's a historical artifact.  Open to a PR that would change that.

> 
> Also as a side questions: why do `TimestampSeries` and `TimedeltaSeries`, etc. exist?  Why don't `Series[Timestamp]` / `Series[Timestamp]` work?

This is because `mypy` can't handle the overloads correctly for the arithmetic operators on `Series[Timestamp]` and `Series[Timedelta]` .    See #718 and links to discussion in `pyright` there.

Dr-Irv (changes requested):
Tests failing on Windows because of the `int32` type.   Still need to remove `str` from `AsTypeArg`

tests/test_series.py
```diff
@@ -1726,48 +1746,114 @@ def test_updated_astype() -> None:
...
+    check(assert_type(s.astype("b"), "pd.Series[int]"), pd.Series, np.byte)
+    # int16
+    check(assert_type(s.astype(np.short), "pd.Series[int]"), pd.Series, np.short)
+    check(assert_type(s.astype("short"), "pd.Series[int]"), pd.Series, np.short)
+    check(assert_type(s.astype("int16"), "pd.Series[int]"), pd.Series, np.short)
+    check(assert_type(s.astype("h"), "pd.Series[int]"), pd.Series, np.short)
+    # int32
+    check(assert_type(s.astype(np.intc), "pd.Series[int]"), pd.Series, np.intc)
+    check(assert_type(s.astype("intc"), "pd.Series[int]"), pd.Series, np.intc)
+    check(assert_type(s.astype("int32"), "pd.Series[int]"), pd.Series, np.intc)
```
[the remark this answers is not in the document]
randolf-scholz: Hm, I'm confused now: Reading https://numpy.org/doc/stable/reference/arrays.scalars.html it sounds like `intc` should be the platform independent type. https://numpy.org/doc/stable/reference/arrays.scalars.html#sized-aliases says

> Along with their (mostly) C-derived names, the integer, float, and complex data-types are also available using a bit-width convention so that an array of the right size can always be ensured.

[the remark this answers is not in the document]
Dr-Irv:
```python
>>> import pandas as pd
>>> import numpy as np
>>> s = pd.Series([1,2,3])
>>> s.dtype
dtype('int64')
>>> s32 = s.astype("int32")
>>> s32.dtype
dtype('int32')
>>> type(s32[0])
<class 'numpy.int32'>
>>> isinstance(s32[0], np.intc)
False
```
The above is on Windows.  When you use `s.astype("int32")`, then the resulting `int` is not an instance of `np.intc`.  On Linux and Mac, it is.  Not sure I understand why there is a difference, but it is what it is!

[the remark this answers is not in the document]
randolf-scholz: I refactored the tests to use `pytest.mark.parametrize`, because I expect more windows tests to fail, and need to find out which. Also added "<M8[...]"  and "<m8[...]" type-codes for timestamp/timedelta.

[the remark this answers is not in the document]
randolf-scholz: What happens on windows if you try `astype(np.intc)` ?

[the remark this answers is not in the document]
Dr-Irv: > What happens on windows if you try `astype(np.intc)` ?
```python
>>> sc = s.astype(np.intc)
>>> sc.dtype
dtype('int32')
>>> type(sc[0])
<class 'numpy.intc'>
```

So I guess on Windows `s.astype("int32")` and `s.astype(np.intc)` store objects of different types inside the series.

Dr-Irv:
It looks like the Windows tests are failing, with the new parameterize stuff, you can get a complete list of what is failing.  You may need to test the O/S to get the values appropriate for Windows.

randolf-scholz:
There is another thing that came up: pandas supports numpys extended type codes:
https://numpy.org/doc/stable/reference/arrays.dtypes.html#string-dtype-note

There are hundreds of possible combinations, and for some data types unlimited number of literal strings that are allowed. For example `s.astype("S25")` implicitly creates a `bytes` datatype with 25-bytes. any positive number of integer bytes is valid.

The question is: how many of these should be hard-coded as literals, to avoid false-postives as currently `str` will not be allowed.

Dr-Irv:
> There is another thing that came up: pandas supports numpys extended type codes: https://numpy.org/doc/stable/reference/arrays.dtypes.html#string-dtype-note
> 
> There are hundreds of possible combinations, and for some data types unlimited number of literal strings that are allowed. For example `s.astype("S25")` implicitly creates a `bytes` datatype with 25-bytes. any positive number of integer bytes is valid.
> 
> The question is: how many of these should be hard-coded as literals, to avoid false-postives as currently `str` will not be allowed.

I say none.  The goal of the pandas stubs is to cover the most common use cases.  We try to strike a balance between being too narrow and too wide.  IMHO, having `str` as a valid type for `astype()` is too wide and reduces the usefulness of the stubs in the most common use cases. 

If someone actually uses these more esoteric strings with pandas, then they can open an issue and we can add them one by one.

randolf-scholz:
I found `np.sctypeDict` which seems like a good list of aliases to support. I added a test to check that all these aliases are tested. The windows stuff is hopefully fixed as well.

tests/test_series.py
```diff
@@ -61,14 +60,28 @@
...
+    )
     from pandas._typing import np_ndarray_int  # noqa: F401
 
 
 def test_types_init() -> None:
     pd.Series(1)
     pd.Series((1, 2, 3))
     pd.Series(np.array([1, 2, 3]))
-    pd.Series(data=[1, 2, 3, 4], name="series")
+    pd.Series(data=[1, 2, 3, 4], name="pd.Series")
```
Dr-Irv, line 84: revert this change

tests/test_series.py
```diff
@@ -122,7 +135,7 @@ def test_types_select() -> None:
     s = pd.Series(data={"row1": 1, "row2": 2})
     with pytest_warns_bounded(
         FutureWarning,
-        "Series.__getitem__ treating keys as positions is deprecated",
+        "pd.Series.__getitem__ treating keys as positions is deprecated",
```
Dr-Irv, line 138: revert this change

tests/test_series.py
```diff
@@ -255,7 +268,7 @@ def test_types_fillna() -> None:
     check(assert_type(s.fillna(0, axis="index"), pd.Series), pd.Series)
     with pytest_warns_bounded(
         FutureWarning,
-        "Series.fillna with 'method' is deprecated",
+        "pd.Series.fillna with 'method' is deprecated",
```
Dr-Irv, line 271: revert this change

tests/test_series.py
```diff
@@ -1693,532 +1706,851 @@ def test_change_to_dict_return_type() -> None:
...
+def test_astype_int(cast_arg: IntDtypeArg, target_type: type) -> None:
+    s = pd.Series([1, 2, 3])
 
-    # Test incorrect Literal
-    if TYPE_CHECKING_INVALID_USAGE:
-        s.astype("foobar")  # type: ignore[call-overload] # pyright: ignore[reportGeneralTypeIssues]
+    if platform.system() == "Windows":
+        # Different behavior for uint32, uint64 and uintp on Windows
+        if cast_arg in ["uint32", "uint64", "uintp"]:
+            pass
```
Dr-Irv, line 2115: I just tired this on Windows and Linux and got the same result.  So remove this

tests/test_series.py
```diff
@@ -1693,532 +1706,851 @@ def test_change_to_dict_return_type() -> None:
...
-    check(assert_type(s.astype("string_"), "pd.Series[bytes]"), pd.Series, bytes)
-    check(assert_type(s.astype("S"), "pd.Series[bytes]"), pd.Series, bytes)
-    # pyarrow bytes types
-    check(
-        assert_type(s.astype("binary[pyarrow]"), "pd.Series[bytes]"), pd.Series, bytes
-    )
+    if platform.system() != "Windows" and cast_arg == "float96":
+        with pytest.raises(TypeError):
+            s.astype(cast_arg)
+        pytest.skip("Unix does not support float96")
```
Dr-Irv, line 2240: Neither does Windows.  So let's just remove this one from the typing.

tests/test_series.py
```diff
@@ -1693,532 +1706,851 @@ def test_change_to_dict_return_type() -> None:
...
+    if platform.system() == "Windows" and cast_arg == "float128":
+        with pytest.raises(TypeError):
+            s.astype(cast_arg)
+        pytest.skip("Windows does not support float128")
```
Dr-Irv, line 2244: that is correct.

Dr-Irv (changes requested):
some things were changed inadvertently.  There are some tests you are skipping that don't need to be skipped on Windows.  Still have some failures in CI for Windows.  

There are cases where the results are different for Windows and Linux, so I think you will have to just do a different check call for those few cases based on platform

tests/test_series.py
```diff
@@ -1870,7 +1870,7 @@ def test_change_to_dict_return_type() -> None:
     ("longfloat", np.longdouble),
     ("g", np.longdouble),
     ("f16", np.longdouble),
-    ("float96", np.longdouble),  # NOTE: WINDOWS ONLY
+    # ("float96", np.longdouble),  # NOTE: WINDOWS ONLY
```
Dr-Irv: Maybe delete this line - or at least change the comment to indicate why you commented it out?

randolf-scholz: It's changed to NOTE: unsupported in the last commit

Dr-Irv: oh I see you did this in a later commit

tests/test_series.py
```diff
@@ -1906,7 +1906,7 @@ def test_change_to_dict_return_type() -> None:
     ("longcomplex", np.clongdouble),
     ("G", np.clongdouble),
     ("c32", np.clongdouble),
-    ("complex192", np.clongdouble),  # NOTE: WINDOWS ONLY
+    # ("complex192", np.clongdouble),  # NOTE: WINDOWS ONLY
```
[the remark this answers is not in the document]
Dr-Irv: oh I see you did this in a later commit

randolf-scholz:
Tests pass now

Dr-Irv (approved):
Thanks @randolf-scholz  .  Nice PR

PR #757, opened 2023-07-24
Fix SQL parameter type

Tolker-KU:
Align parameter type to SQL functions: https://github.com/pandas-dev/pandas/blob/main/pandas/io/sql.py#L370

### DIFF
--- a/pandas-stubs/io/sql.pyi
+++ b/pandas-stubs/io/sql.pyi
@@ -2,6 +2,7 @@ from collections.abc import (
     Callable,
     Generator,
     Iterable,
+    Mapping,
 )
 import sqlite3
 from typing import (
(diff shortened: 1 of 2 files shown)


Dr-Irv (changes requested):
Can you add a test that illustrates why this is needed, i.e., why the current types for the `params` argument don't work?

Tolker-KU:
Hi!

Thanks for getting back quickly! I'm not quite sure how to test this change. Eventually we are calling into SQLAlchemy when using these methods - and SQLAlchemy types the parameters as `Any`. This is the reasoning behind the change

Dr-Irv:
> Hi!
> 
> Thanks for getting back quickly! I'm not quite sure how to test this change. Eventually we are calling into SQLAlchemy when using these methods - and SQLAlchemy types the parameters as `Any`. This is the reasoning behind the change

If you look at `tests/test_io.py`, and search for `read_sql_query()`, you will want to add a test similar to one of those, using example code that currently fails with the stubs, but would work with your change.

Dr-Irv (changes requested):
In thinking about this, I think the `Any` type is too wide.  For example, you couldn't pass an instance of any class as a parameter to an SQL call, it would have to be some kind of scalar.  So can you change the places where you used `Any` in this PR to use `Scalar` instead, which you can import from `pandas._typing` ?

Tolker-KU:
Sounds like a good solution to me. I've made the change

Dr-Irv (approved):
Thanks @Tolker-KU

PR #759, opened 2023-08-02
make sub of timedelta work

### LINKED ISSUE #758: No overload variant of "__sub__" of "Series" matches argument type "Timedelta"
**Describe the bug**
After upgrading to pandas 2.0 started getting errors as described in the title

**To Reproduce**
```python
import pandas as pd

df = pd.read_csv("afile.csv", parse_dates=["foo"])
df["foo"] = df["foo"] - pd.Timedelta("1D")
```

Using mypy:

```sh
pandas-stubs-mre.py:4: error: No overload variant of "__sub__" of "Series" matches argument type "Timedelta"  [operator]
pandas-stubs-mre.py:4: note: Possible overload variants:
pandas-stubs-mre.py:4: note:     def __sub__(self, Timestamp | datetime | TimestampSeries, /) -> TimedeltaSeries
pandas-stubs-mre.py:4: note:     def __sub__(self, complex | ExtensionArray | ndarray[Any, Any] | dict[str, ndarray[Any, Any]] | list[Any] | tuple[Any, ...] | Index | Series[Any], /) -> Series[Any]
```

 - OS: Discovered on both Windows / Linux
 - python version: 3.11.0
 - version of type checker: mypy 1.4.1 (compiled: yes)
 - version of installed `pandas-stubs`: 2.0.2.230605

**Additional context**
The issue does not happen with older versions of pandas-stubs - 1.5.3.230321 issues no complaint. I haven't done a full bisect but guess this may go back to https://github.com/pandas-dev/pandas-stubs/pull/616 ?

Dr-Irv:
- [x] Closes #758 
   - `test_timefuncs.py:test_timedelta64_and_arithmatic_operator()` - added tests there

### DIFF
--- a/tests/test_indexes.py
+++ b/tests/test_indexes.py
@@ -25,6 +25,8 @@ from tests import (
 
 if TYPE_CHECKING:
     MYPY_CHECKING: bool = True
+    # See test_sorted_and_list() where mypy and pyright do different
+    # inference on sorted(pd.Index)
     if MYPY_CHECKING:
         from typing import Any
 
@@ -728,6 +730,8 @@ def test_interval_index_tuples():
 def test_sorted_and_list() -> None:
     # GH 497
     i1 = pd.Index([3, 2, 1])
+    # mypy infers sorted(i1) as list[Any], while pyright infers sorted(i1) as
+    # list[SupportsRichComparison]
     check(
         assert_type(
             sorted(i1),
--- a/tests/test_indexes.py
+++ b/tests/test_indexes.py
@@ -25,14 +25,15 @@ from tests import (
 
 if TYPE_CHECKING:
     MYPY_CHECKING: bool = True
-    if MYPY_CHECKING:
+    if not MYPY_CHECKING:
+        from _typeshed import SupportsRichComparison
+    else:
         from typing import Any
 
         from typing_extensions import TypeAlias
 
         SupportsRichComparison: TypeAlias = Any
-    else:
-        pass
+
     from pandas.core.indexes.base import (
         _ComplexIndexType,
         _FloatIndexType,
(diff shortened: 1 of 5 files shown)


twoertwein (commented):
Looks good to me but there might be a numpy-pyright error

Dr-Irv:
> Looks good to me but there might be a numpy-pyright error

Actually, it was an issue with pyright 1.1.320.  Needed to test differently on mypy vs. pyright, which I figured out how to do.  Take another look

tests/test_indexes.py
```diff
@@ -24,6 +24,15 @@
...
 if TYPE_CHECKING:
+    MYPY_CHECKING: bool = True
+    if MYPY_CHECKING:
+        from typing import Any
+
+        from typing_extensions import TypeAlias
+
+        SupportsRichComparison: TypeAlias = Any
+    else:
+        from _typeshed import SupportsRichComparison  # noqa: F401
```
twoertwein, line 35: Do you mind referencing an issue or describing the discrepancy? Is it just that mypy has an older version of typestubs?

Dr-Irv: Added comments in https://github.com/pandas-dev/pandas-stubs/pull/759/commits/5310eb682334877ae00a06b80082fc80f7467aee

twoertwein:
Thanks @Dr-Irv !

tests/test_indexes.py
Dr-Irv: > Does this also work for `poe mypy --mypy_nightly`?

Looks like it did!

We may want to use this pattern (`MYPY_CHECKING`) elsewhere when `mypy` and `pyright` make different inferences.  Can't put it in the stubs themselves - that would require users to define `MYPY_CHECKING` as `false` if using pyright, but we can use it in our tests.

PR #760, opened 2023-08-03
Make IndexOpsMixin (and Index) generic

### LINKED ISSUE #340: Generic type for Indexes
Hi,

We have just moved over to pandas-stubs from data-science-types.
Overall it is vastly superior so I am enjoying being able to remove countless type-ignores that we needed due to the incompleteness of the data-science-types stubs.

However, one thing that is not ideal is that we did have a lot of Indexes with types specified, i.e., `Index[int]` as data-science-types have their Index class as Generic: `class Index(Generic[_T]):`
Now with pandas-stubs we are having to remove those index types which isn't the end of the world, but they are nice to have.
If it isn't too hard to get that working, I would definitely appreciate having generic types for Indexes

### LINKED ISSUE #502: Bug with iterating df.columns
**Describe the bug**

```py
import pandas as pd

t: pd.DataFrame
s: list[str] = list(t.columns)
```

```
t.py:4: error: Argument 1 to "list" has incompatible type "Index"; expected
"Iterable[str]"
    s: list[str] = list(t.columns)
                        ^~~~~~~~~
t.py:4: note: Following member(s) of "Index" have conflicts:
t.py:4: note:     Expected:
t.py:4: note:         def __iter__(self) -> Iterator[str]
t.py:4: note:     Got:
t.py:4: note:         def __iter__(self) -> Iterator[Union[Union[str, bytes, date, datetime, timedelta, bool, int, float, complex, Timesta
mp, Timedelta], Tuple[Hashable, ...]]]
```

Variant:

```py
import pandas as pd

t: pd.DataFrame
s: str
for s in t.columns:
    pass
```

```
t.py:5: error: Incompatible types in assignment (expression has type
"Union[Union[str, bytes, date, datetime, timedelta, bool, int, float, complex, Timestamp, Timedelta], Tuple[Hashable, ...]]",
variable has type "str")
    for s in t.columns:
    ^
```

**To Reproduce**
See above

- Version `1.5.2.230105`
- MyPy 0.991

### LINKED ISSUE #723: Create `__iter__()` for `DatetimeIndex`
I also ran in to this using [`date_range`](https://pandas.pydata.org/docs/reference/api/pandas.date_range.html).

Workaround is good for now:
```py
for ts in cast(Iterator[datetime], date_range(start='1/1/2023', end='1/08/2023', freq='6H', tz=timezone.utc)):
   ....
```

_Originally posted by @davetapley in https://github.com/pandas-dev/pandas-stubs/issues/502#issuecomment-1571003583_

Should create `DatetimeIndex.__iter__()` returning `Timestamp`

### LINKED ISSUE #732: FEAT:  Add InterpolateOptions to `interpolate()` methods for `method` argument.
See PR https://github.com/pandas-dev/pandas/pull/53659 where `InterpolateOptions` is introduced in `pandas._typing.py` for type checking the valid arguments to the `interpolate()` method.

We should carry that over to `pandas-stubs`

### LINKED ISSUE #744: Index.intersection does not return correct sub type
**Describe the bug**
When I form the `intersection` of two `pd.DatetimeIndex`, the return type is `Index`. This leads to problems with index entries: mypy complains about `intersection[0].date()`.

**To Reproduce**
1. Minimal example
```py
import pandas as pd
intersection = pd.DatetimeIndex().intersection(pd.DatetimeIndex)

reveal_type(intersection)  # Index
reveal_type(intersection[0])  # "str | bytes | date | datetime | timedelta | datetime64 | timedelta64 | bool | int | float | Timestamp | Timedelta | complex"
```

2. mypy

3. error message

```py
agri_data/utils.py:96: error: Item "str" of "str | bytes | date | datetime | timedelta | datetime64 | timedelta64 | bool | int | float | Timestamp | Timedelta | complex" has no attribute "date"  [union-attr]
agri_data/utils.py:96: error: Item "bytes" of "str | bytes | date | datetime | timedelta | datetime64 | timedelta64 | bool | int | float | Timestamp | Timedelta | complex" has no attribute "date"  [union-attr]
agri_data/utils.py:96: error: Item "date" of "str | bytes | date | datetime | timedelta | datetime64 | timedelta64 | bool | int | float | Timestamp | Timedelta | complex" has no attribute "date"  [union-attr]
agri_data/utils.py:96: error: Item "timedelta" of "str | bytes | date | datetime | timedelta | datetime64 | timedelta64 | bool | int | float | Timestamp | Timedelta | complex" has no attribute "date"  [union-attr]
agri_data/utils.py:96: error: Item "datetime64" of "str | bytes | date | datetime | timedelta | datetime64 | timedelta64 | bool | int | float | Timestamp | Timedelta | complex" has no attribute "date"  [union-attr]
agri_data/utils.py:96: error: Item "timedelta64" of "str | bytes | date | datetime | timedelta | datetime64 | timedelta64 | bool | int | float | Timestamp | Timedelta | complex" has no attribute "date"  [union-attr]
agri_data/utils.py:96: error: Item "bool" of "str | bytes | date | datetime | timedelta | datetime64 | timedelta64 | bool | int | float | Timestamp | Timedelta | complex" has no attribute "date"  [union-attr]
agri_data/utils.py:96: error: Item "int" of "str | bytes | date | datetime | timedelta | datetime64 | timedelta64 | bool | int | float | Timestamp | Timedelta | complex" has no attribute "date"  [union-attr]
agri_data/utils.py:96: error: Item "float" of "str | bytes | date | datetime | timedelta | datetime64 | timedelta64 | bool | int | float | Timestamp | Timedelta | complex" has no attribute "date"  [union-attr]
agri_data/utils.py:96: error: Item "Timedelta" of "str | bytes | date | datetime | timedelta | datetime64 | timedelta64 | bool | int | float | Timestamp | Timedelta | complex" has no attribute "date"  [union-attr]
agri_data/utils.py:96: error: Item "complex" of "str | bytes | date | datetime | timedelta | datetime64 | timedelta64 | bool | int | float | Timestamp | Timedelta | complex" has no attribute "date"  [union-attr]
```

4. Expected type: `reveal_type(intersection[0])  # Timestamp`

 - OS: Ubuntu
 - OS Version: bullseye
 - python version: 3.11.1
 - version of type checker: 1.2.0
 - version of installed `pandas-stubs`: 2.0.2.230605

**Additional context**
Possibly related: #723

twoertwein:
- [x] Closes #723,  closes #744, closes #502, and closes #340

Mostly working, but still a few failing tests and I haven't added new tests.

xref https://github.com/microsoft/pyright/issues/5642


tests/test_scalars.py
```diff
@@ -1745,7 +1754,7 @@ def test_period_add_subtract() -> None:
     as_dt_td = dt.timedelta(days=1)
     as_np_td = np.timedelta64(1, "D")
     as_np_i64 = np.int64(1)
-    as_int = int(1)
+    as_int: int = 1
```
twoertwein, line 1757: from ruff, I added the explicit `: int` but it should not be needed

twoertwein:
Most of the the remaining errors caused by `Index[S1]` where `S1` is  `Any`/`Unknown`, e.g., `df.columns`. Even though `S1` is bound, the type system/checkers widen the type to `Any`/`Unknown`.

Dr-Irv:
I think this may close #502 and #340 .  Can you check if that should be added to the list?

pandas-stubs/core/frame.pyi
```diff
@@ -86,6 +86,7 @@ from pandas._typing import (
     IndexingInt,
     IndexLabel,
     IndexType,
+    InterpolateOptions,
```
Dr-Irv, line 89: Maybe the `InterpolateOptions` stuff should go in a separate PR

pandas-stubs/core/indexes/base.pyi
```diff
@@ -113,7 +107,7 @@ class Index(IndexOpsMixin, PandasObject):
         **kwargs,
     ) -> Self: ...
```
Dr-Irv, line 108: Should probably add specific `__new__()` to return `DatetimeIndex`, `PeriodIndex`, `TimedeltaIndex`  .  In some sense, the pattern of `__new__()` here should mirror what is done for `Series`

pandas-stubs/core/indexes/base.pyi
```diff
@@ -195,15 +187,15 @@ class Index(IndexOpsMixin, PandasObject):
...
+    def __neg__(self) -> Self: ...
     def __nonzero__(self) -> None: ...
     __bool__ = ...
     def union(self, other: list[HashableT] | Index, sort=...) -> Index: ...
-    def intersection(self, other: list[T1] | Index, sort: bool = ...) -> Index: ...
-    def difference(self, other: list | Index, sort: bool | None = None) -> Index: ...
+    def intersection(self, other: list[S1] | Self, sort: bool = ...) -> Self: ...
+    def difference(self, other: list | Index, sort: bool | None = None) -> Self: ...
     def symmetric_difference(
-        self, other: list[T1] | Index, result_name=..., sort=...
```
Dr-Irv, line 205: FYI - the `T1` references are from the original Microsoft stubs where `Index` was generic, but I couldn't ever get that to pass `mypy` (about 18 months ago).  So I just removed it, but probably missed the various references to `T1`

tests/test_indexes.py
```diff
@@ -25,27 +25,6 @@
...
 if TYPE_CHECKING:
     MYPY_CHECKING: bool = True
-    # See test_sorted_and_list() where mypy and pyright do different
-    # inference on sorted(pd.Index)
-    if MYPY_CHECKING:
-        from typing import Any
-
-        from typing_extensions import TypeAlias
-
-        SupportsRichComparison: TypeAlias = Any
```
Dr-Irv, line 35: probably still need this in there??

Dr-Irv (commented):
Noticed some things to consider.

Am wondering how `MultiIndex` will work as a subclass of `Index[S1]`  .  Is everything good because we just say `class MultiIndex(Index)`, so the `S1` is ignored?  But will some of the methods of `Index` that refer to `S1` still work right if the index is a `MultiIndex` ?

pandas-stubs/core/indexes/base.pyi
twoertwein: That would be needed when people use `pd.Index(<list of datetime>)`, or? If they use `pd.DatetimeIndex(...)` it will already work (but not check the type of the iterable)

tests/test_indexes.py
twoertwein: Not at the moment, that was only needed because the item type of list was Any/Unknown which created a discrepancy between mypy and pyright.

twoertwein:
> Am wondering how `MultiIndex` will work as a subclass of `Index[S1]` . Is everything good because we just say `class MultiIndex(Index)`, so the `S1` is ignored? But will some of the methods of `Index` that refer to `S1` still work right if the index is a `MultiIndex` ?

I'm not sure, same applies to `CategoricalIndex` (`pd.Categorical` is not in `S1`, but probably should be there).

pandas-stubs/core/indexes/base.pyi
twoertwein: I think I shouldn't use S1 here - would probably prevent comparing int with float. Should maybe use Scalar/Any (or wait for someone raising an issue).

pandas-stubs/core/frame.pyi
twoertwein: I agree - I naively assumed that adding `__iter__` would be simply :)

Dr-Irv:
@twoertwein  I'm headed out of town for 3 days, so won't look at this until Monday.

pandas-stubs/core/accessor.pyi
```diff
@@ -1,8 +1,5 @@
 from typing import Any
 
-class DirNamesMixin:
-    def __dir__(self): ...
```
twoertwein, line 4: PandasObject and this class were causing pyright's MRO issue. I believe these two classes are not public and they provide no typing benefit (their methods are already provided by object).

pandas-stubs/core/computation/ops.pyi
```diff
@@ -34,7 +34,6 @@ class Term:
     def ndim(self) -> int: ...
 
 class Constant(Term):
-    def __init__(self, value, env, side=..., encoding=...) -> None: ...
```
twoertwein, line 37: Enabled the check to sync `__init__` and `__new__` in the hope of understanding some of pyright's behavior.

pandas-stubs/core/indexes/base.pyi
```diff
@@ -52,37 +60,50 @@ class InvalidIndexError(Exception): ...
...
-    def __getitem__(self, idx: int) -> S1: ...
-
-class Index(IndexOpsMixin, PandasObject):
+class Index(IndexOpsMixin[S1]):
     __hash__: ClassVar[None]  # type: ignore[assignment]
+    # overloads with additional dtypes
+    @overload
+    def __new__(  # type: ignore[misc]
+        cls,
+        data: Iterable[int | np.integer],
```
twoertwein, line 69: One might assume this would work with np.ndarray - it doesn't: https://github.com/numpy/numpy/issues/24335

tests/test_indexes.py
```diff
@@ -709,11 +708,8 @@ def test_interval_index_tuples():
 def test_sorted_and_list() -> None:
     # GH 497
     i1 = pd.Index([3, 2, 1])
-    check(
-        assert_type(sorted(i1), list[int]),
-        list,
-    )
-    check(assert_type(list(i1), list[int]), list)
+    check(assert_type(sorted(i1), "list[int]"), list, int)
+    check(assert_type(list(i1), "list[int]"), list, int)
```
Dr-Irv, line 712: same comment as before - don't see why the quotes are needed

Dr-Irv (commented):
This is looking pretty good, IMHO

tests/test_frame.py
```diff
@@ -1761,7 +1761,8 @@ def test_getmultiindex_columns() -> None:
         [(i, s) for i in [1] for s in df.columns.get_level_values(1)]
     ]
     res4: pd.DataFrame = df[[df.columns[0]]]
-    check(assert_type(df[df.columns[0]], pd.Series), pd.Series)
+    column: Scalar = df.columns[0]
+    check(assert_type(df[column], pd.Series), pd.Series)
```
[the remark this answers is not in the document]
twoertwein: I'm not sure if this can be fixed - maybe we can do some gymnastic with the order of overloads. `df.columns` is `Index[Any]` and would therefore probably match the first overload.

pandas-stubs/core/indexes/timedeltas.pyi
```diff
@@ -44,6 +44,18 @@ class TimedeltaIndex(DatetimeTimedeltaMixin[Timedelta], TimedeltaIndexProperties
         copy: bool = ...,
         name: str = ...,
     ) -> None: ...
+    def __new__(
```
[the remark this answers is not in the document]
twoertwein: If we don't declare `__init__`, then pyright compares the `__init__` from the parent classes and sees that it is not compatible with `__new__`. Removing `__init__` from `Properties` would allow us to remove `__init__` from `TimedeltaIndex`.

tests/test_indexes.py
```diff
@@ -70,17 +69,17 @@ def test_multiindex_get_level_values() -> None:
 
 def test_index_tolist() -> None:
     i1 = pd.Index([1, 2, 3])
-    check(assert_type(i1.tolist(), list[int]), list, int)
-    check(assert_type(i1.to_list(), list[int]), list, int)
+    check(assert_type(i1.tolist(), "list[int]"), list, int)
+    check(assert_type(i1.to_list(), "list[int]"), list, int)
```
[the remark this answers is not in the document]
twoertwein: Seems that Pyright's behavior changed?

pandas-stubs/core/indexes/base.pyi
twoertwein: Added overloads for ndarray

tests/test_frame.py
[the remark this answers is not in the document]
twoertwein: Given that both mypy and pyright do not limit a bounded TypeVar to its bound when it is unknown/any, we have only two options:

- make people cast the Index to the appropriate type
- return Scalar
- return Scalar now; in the future, make pd.DataFrame generic in terms of the Index; return S1

I have no particular preference, except that making DataFrame generic might not happen anytime soon

btw. I will offline from Friday-Sunday, you are welcome to push changes to this PR or also merge it. Since this is a large PR and the tests might not cover everything, I would prefer if you or others could run this version of pandas-stubs on your internal codebase before the next release to catch potential regressions.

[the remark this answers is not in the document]
twoertwein: unrelated: seems that the newest version of numexp broke CI/pandas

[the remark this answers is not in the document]
Dr-Irv: > Given that both mypy and pyright do not limit a bounded TypeVar to its bound when it is unknown/any, we have only two options:
> 
> * make people cast the Index to the appropriate type
> * return Scalar
> * return Scalar now; in the future, make pd.DataFrame generic in terms of the Index; return S1
> 
> I have no particular preference, except that making DataFrame generic might not happen anytime soon

Would changing `DataFrame.columns()` to return `Index[Scalar]` work?

> 
> btw. I will offline from Friday-Sunday, you are welcome to push changes to this PR or also merge it. Since this is a large PR and the tests might not cover everything, I would prefer if you or others could run this version of pandas-stubs on your internal codebase before the next release to catch potential regressions.

I will see if I can give this a try.

tests/test_indexes.py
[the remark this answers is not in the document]
Dr-Irv: So it was a 3.8 failure only?  surprising that it didn't pick this up before.

tests/test_scalars.py
```diff
@@ -745,8 +746,12 @@ def test_timedelta_mul_div() -> None:
...
         mp_series_int // td  # type: ignore[operator] # pyright: ignore[reportGeneralTypeIssues]
         md_series_float // td  # type: ignore[operator] # pyright: ignore[reportGeneralTypeIssues]
-        md_int64_index // td  # type: ignore[operator] # pyright: ignore[reportGeneralTypeIssues]
-        md_float_index // td  # type: ignore[operator] # pyright: ignore[reportGeneralTypeIssues]
+        assert_type(
+            md_int64_index // td, Never  # pyright: ignore[reportGeneralTypeIssues]
+        )
+        assert_type(
+            md_float_index // td, Never  # pyright: ignore[reportGeneralTypeIssues]
+        )
```
Dr-Irv, line 754: Concerned about the above change.  Previously, `mypy` was seeing that this is an invalid operation.  If you make the above change, mypy is no longer detecting that.  That's not a good thing.

tests/test_scalars.py
```diff
@@ -779,8 +784,12 @@ def test_timedelta_mul_div() -> None:
...
         mp_series_int / td  # type: ignore[operator] # pyright: ignore[reportGeneralTypeIssues]
         md_series_float / td  # type: ignore[operator] # pyright: ignore[reportGeneralTypeIssues]
-        md_int64_index / td  # type: ignore[operator] # pyright: ignore[reportGeneralTypeIssues]
-        md_float_index / td  # type: ignore[operator] # pyright: ignore[reportGeneralTypeIssues]
+        assert_type(
+            md_int64_index / td, Never  # pyright: ignore[reportGeneralTypeIssues]
+        )
+        assert_type(
+            md_float_index / td, Never  # pyright: ignore[reportGeneralTypeIssues]
+        )
```
Dr-Irv, line 792: Same as above

Dr-Irv (commented):
Can you test if using `DataFrame.columns()` returning `Index[Scalar]` would solve the problem noted in another comment so that the expression `df[df.columns[0]]` would work without having to do a `cast` or creating a temporary variable?

tests/test_scalars.py
twoertwein: I agree, I don't know how to fix it. I think the reason why it worked previously was a behavior of Never when it is being used on input arguments: it is interpreted by mypy/pyright to indicate that that call is invalid. Now, with the generic class, we need to have annotations on the input arguments (we don't have the luxury of using Never there anymore). We can still return Never (but that has a slightly different semantic meaning).

tests/test_indexes.py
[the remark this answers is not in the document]
twoertwein: Yes (but it might be a pyright bug):
https://github.com/pandas-dev/pandas-stubs/actions/runs/5797814916/job/15714189351

twoertwein:
> Can you test if using `DataFrame.columns()` returning `Index[Scalar]` would solve the problem noted in another comment so that the expression `df[df.columns[0]]` would work without having to do a `cast` or creating a temporary variable?

It sounds as if it should work but it creates different problems:
- Scalar is a Union: using `df.columns[0]` for comparisons or other operations would fail (we have a few tests for that - they work because it is currently Any)
- Scalar is actually not compatible with S1 (I believe np.tiemdelta64 and np.datetime64 are missing in S1)
- it makes `columns: Index[str] = df.columns` not possible anymore #502

I try to see how far I get on my local branch but I'm not too optimistic that there is an ideal solution for this problem (except for DataFrame being generic in index and columns)

tests/test_indexes.py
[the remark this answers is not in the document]
Dr-Irv: I actually think this is correct.  It's basically telling us that the code wouldn't run with python 3.8.  So making this change is OK.  I think I saw this happen in another PR that I did.

twoertwein:
> I will see if I can find some time to test that one issue related to why `mypy` isn't picking up the incorrect arithmetic operation

I believe you have an open issue at mypy related to that :) In theory, mypy should still warn about unreachable code after those lines and prevent people from using the return value.

pandas-stubs/core/indexes/base.pyi
```diff
@@ -66,9 +67,9 @@ class Index(IndexOpsMixin[S1]):
     @overload
     def __new__(  # type: ignore[misc]
         cls,
-        data: Iterable[int | np.integer],
+        data: Sequence[int | np.integer] | IndexOpsMixin[int] | npt.NDArray[np.integer],
```
Dr-Irv, line 70: Can use `np_ndarray_anyint` instead of the `npt` type here

pandas-stubs/core/indexes/base.pyi
```diff
@@ -88,12 +89,11 @@ class Index(IndexOpsMixin[S1]):
     @overload
     def __new__(  # type: ignore[misc]
         cls,
-        data: Iterable[float | np.float32 | np.float64],
+        data: Sequence[float | np.floating]
+        | IndexOpsMixin[float]
+        | npt.NDArray[np.floating],
```
Dr-Irv, line 94: Maybe we should introduce a `np_ndarray_anyfloat` in `_typing` and use that here for consistency.

pandas-stubs/core/indexes/base.pyi
```diff
@@ -116,9 +113,11 @@ class Index(IndexOpsMixin[S1]):
     @overload
     def __new__(  # type: ignore[misc]
         cls,
-        data: Iterable[complex],
+        data: Sequence[complex | np.complexfloating]
+        | IndexOpsMixin[complex]
+        | npt.NDArray[np.complexfloating],
```
Dr-Irv, line 118: Same as above - let's be consistent with the type alias usage

Dr-Irv (changes requested):
Suggestion to avoid using `npt` in the constructors

tests/test_scalars.py
Dr-Irv: I found a fix.  I created a PR against your branch at https://github.com/twoertwein/pandas-stubs/pull/3

Key was to remove `__floordiv__()` from `OpsMixin` and let any subclass of `OpsMixin` define only the valid values.

Dr-Irv: So the mypy folks say it's not a bug, and pyright says that its implementation was incorrect. So I think the only way to do this is with what I did - you can't use `Never` or `NoReturn` to "override" the implementation in a subclass that matches `Any`

tests/test_frame.py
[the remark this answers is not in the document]
Dr-Irv: > Since this is a large PR and the tests might not cover everything, I would prefer if you or others could run this version of pandas-stubs on your internal codebase before the next release to catch potential regressions.

I tried the version I placed in your repo with the PR on two large code bases and no new errors appeared, so I think once you merge that in, I can approve and merge in this PR.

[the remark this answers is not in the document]
twoertwein: Thank you for testing it and finding a way to let mypy (and pyright) clearly indicate the unintended calls!

Dr-Irv (approved):
Thanks @twoertwein .  Great contribution

PR #766, opened 2023-08-17
Infer dtype of Series in more cases

### LINKED ISSUE #710: Create pd.Series initializer overload to propagate the generic type of an input container
> > How can the type of `iter(something)` be different from `something.__iter__` ? Is there some double check I should do ?
> 
> I think the issue is that when you are doing `s.groupby(series_groupby)`, `s` is `Series[Any]` because there was no `dtype` information supplied. If you had created `s=pd.Series([1, 1, 2], dtype=int)`, then you'd get the results you expect.

Maybe this is deserving of a separate issue, but shouldn't we be able to create an initializer overload that takes a `list[T]` and returns a `pd.Series[T]`?

_Originally posted by @gandhis1 in https://github.com/pandas-dev/pandas-stubs/issues/708#issuecomment-1562040106_

twoertwein:
- [ ] Closes #710 (Replace xxxx with the Github issue number)

### DIFF
--- a/pandas-stubs/core/frame.pyi
+++ b/pandas-stubs/core/frame.pyi
@@ -555,7 +555,7 @@ class DataFrame(NDFrame, OpsMixin):
     @overload
     def __getitem__(  # type: ignore[misc]
         self,
-        key: Series[_bool]
+        key: Series
         | DataFrame
         | Index
         | np_ndarray_str
--- a/pandas-stubs/core/series.pyi
+++ b/pandas-stubs/core/series.pyi
@@ -219,7 +219,7 @@ class Series(IndexOpsMixin[S1], NDFrame):
     __hash__: ClassVar[None]
 
     @overload
-    def __new__(
+    def __new__(  # type: ignore[misc]
         cls,
         data: DatetimeIndex | Sequence[np.datetime64 | datetime],
         index: Axes | None = ...,
@@ -229,7 +229,7 @@ class Series(IndexOpsMixin[S1], NDFrame):
         copy: bool = ...,
     ) -> TimestampSeries: ...
     @overload
-    def __new__(
+    def __new__(  # type: ignore[misc]
         cls,
         data: _ListLike,
         index: Axes | None = ...,
@@ -239,7 +239,7 @@ class Series(IndexOpsMixin[S1], NDFrame):
         copy: bool = ...,
     ) -> TimestampSeries: ...
     @overload
-    def __new__(
+    def __new__(  # type: ignore[misc]
         cls,
         data: PeriodIndex,
         index: Axes | None = ...,
@@ -249,7 +249,7 @@ class Series(IndexOpsMixin[S1], NDFrame):
         copy: bool = ...,
     ) -> PeriodSeries: ...
     @overload
-    def __new__(
+    def __new__(  # type: ignore[misc]
         cls,
         data: TimedeltaIndex | Sequence[np.timedelta64 | timedelta],
         index: Axes | None = ...,
@@ -283,7 +283,7 @@ class Series(IndexOpsMixin[S1], NDFrame):
     @overload
     def __new__(
         cls,
-        data: _ListLike[S1] | dict[int, S1] | dict[_str, S1] = ...,
+        data: _ListLike[S1] | dict[int, S1] | dict[_str, S1],
         index: Axes | None = ...,
         *,
         dtype: Dtype = ...,
@@ -685,7 +685,7 @@ class Series(IndexOpsMixin[S1], NDFrame):
     def diff(self, periods: int = ...) -> Series[S1]: ...
     def autocorr(self, lag: int = ...) -> float: ...
     @overload
-    def dot(self, other: Series[S1]) -> Scalar: ...  # pyright: ignore[reportOverlappingOverload]
+    def dot(self, other: Series[S1]) -> Scalar: ...  # type: ignore[misc] # pyright: ignore[reportOverlappingOverload]
     @overload
     def dot(self, other: DataFrame) -> Series[S1]: ...
     @overload
@@ -693,7 +693,7 @@ class Series(IndexOpsMixin[S1], NDFrame):
     def __matmul__(self, other): ...
     def __rmatmul__(self, other): ...
     @overload
-    def searchsorted(
+    def searchsorted(  # type: ignore[misc]
         self,
         value: _ListLike,
         side: Literal["left", "right"] = ...,
--- a/pandas-stubs/core/series.pyi
+++ b/pandas-stubs/core/series.pyi
@@ -221,7 +221,10 @@ class Series(IndexOpsMixin[S1], NDFrame):
     @overload
     def __new__(  # type: ignore[misc]
         cls,
-        data: DatetimeIndex | Sequence[np.datetime64 | datetime],
+        data: DatetimeIndex
+        | Sequence[np.datetime64 | datetime]
+        | np.datetime64
+        | datetime,
         index: Axes | None = ...,
         *,
         dtype: TimestampDtypeArg = ...,
@@ -251,7 +254,10 @@ class Series(IndexOpsMixin[S1], NDFrame):
     @overload
     def __new__(  # type: ignore[misc]
         cls,
-        data: TimedeltaIndex | Sequence[np.timedelta64 | timedelta],
+        data: TimedeltaIndex
+        | Sequence[np.timedelta64 | timedelta]
+        | np.timedelta64
+        | timedelta,
         index: Axes | None = ...,
         *,
         dtype: TimedeltaDtypeArg = ...,
@@ -273,7 +279,7 @@ class Series(IndexOpsMixin[S1], NDFrame):
     @overload
     def __new__(
         cls,
-        data: object,
+        data: Scalar | _ListLike | dict[int, Any] | dict[_str, Any] | None,
         index: Axes | None = ...,
         *,
         dtype: type[S1],
@@ -283,7 +289,7 @@ class Series(IndexOpsMixin[S1], NDFrame):
     @overload
     def __new__(
         cls,
-        data: _ListLike[S1] | dict[int, S1] | dict[_str, S1],
+        data: S1 | _ListLike[S1] | dict[int, S1] | dict[_str, S1],
         index: Axes | None = ...,
         *,
         dtype: Dtype = ...,
@@ -293,7 +299,7 @@ class Series(IndexOpsMixin[S1], NDFrame):
     @overload
     def __new__(
         cls,
-        data: object = ...,
+        data: Scalar | _ListLike | dict[int, Any] | dict[_str, Any] | None = ...,
         index: Axes | None = ...,
         *,
         dtype: Dtype = ...,
@@ -685,11 +691,13 @@ class Series(IndexOpsMixin[S1], NDFrame):
     def diff(self, periods: int = ...) -> Series[S1]: ...
     def autocorr(self, lag: int = ...) -> float: ...
     @overload
-    def dot(self, other: Series[S1]) -> Scalar: ...  # type: ignore[misc] # pyright: ignore[reportOverlappingOverload]
+    def dot(self, other: Series[S1]) -> Scalar: ...
     @overload
     def dot(self, other: DataFrame) -> Series[S1]: ...
     @overload
-    def dot(self, other: _ListLike) -> np.ndarray: ...
+    def dot(
+        self, other: ArrayLike | dict[_str, np.ndarray] | Sequence[S1] | Index[S1]
+    ) -> np.ndarray: ...
     def __matmul__(self, other): ...
     def __rmatmul__(self, other): ...
     @overload
@@ -843,7 +851,7 @@ class Series(IndexOpsMixin[S1], NDFrame):
         axis: AxisIndex = ...,
         *args,
         **kwargs,
-    ) -> Series[S1]: ...
+    ) -> Self: ...
     agg = aggregate
     @overload
     def transform(
@@ -1440,14 +1448,10 @@ class Series(IndexOpsMixin[S1], NDFrame):
     # just failed to generate these so I couldn't match
     # them up.
     @overload
-    def __add__(self, other: TimestampSeries) -> TimestampSeries: ...
-    @overload
-    def __add__(self, other: DatetimeIndex) -> TimestampSeries: ...
-    @overload
-    def __add__(self, other: Timestamp) -> TimestampSeries: ...
+    def __add__(self, other: S1 | Self) -> Self: ...
     @overload
     def __add__(
-        self, other: num | _str | Timedelta | _ListLike | Series[S1] | np.timedelta64
+        self, other: num | _str | Timedelta | _ListLike | Series | np.timedelta64
     ) -> Series: ...
     # ignore needed for mypy as we want different results based on the arguments
     @overload  # type: ignore[override]
@@ -1488,7 +1492,10 @@ class Series(IndexOpsMixin[S1], NDFrame):
     ) -> Series[bool]: ...
     @overload
     def __or__(self, other: int | np_ndarray_anyint | Series[int]) -> Series[int]: ...
-    def __radd__(self, other: num | _str | _ListLike | Series[S1]) -> Series[S1]: ...
+    @overload
+    def __radd__(self, other: S1 | Series[S1]) -> Self: ...
+    @overload
+    def __radd__(self, other: num | _str | _ListLike | Series) -> Series: ...
     # ignore needed for mypy as we want different results based on the arguments
     @overload  # type: ignore[override]
     def __rand__(  # type: ignore[misc]
@@ -1973,6 +1980,13 @@ class Series(IndexOpsMixin[S1], NDFrame):
     ) -> Self: ...
     def set_axis(self, labels, *, axis: Axis = ..., copy: _bool = ...) -> Self: ...
     def __iter__(self) -> Iterator[S1]: ...
+    def xs(
+        self,
+        key: Hashable,
+        axis: AxisIndex = ...,
+        level: Level | None = ...,
+        drop_level: _bool = ...,
+    ) -> Self: ...
 
 class TimestampSeries(Series[Timestamp]):
     # ignore needed because of mypy
@@ -2099,13 +2113,6 @@ class TimedeltaSeries(Series[Timedelta]):
         numeric_only: _bool = ...,
         **kwargs,
     ) -> Timedelta: ...
-    def xs(
-        self,
-        key: Hashable,
-        axis: AxisIndex = ...,
-        level: Level | None = ...,
-        drop_level: _bool = ...,
-    ) -> Self: ...
 
 class PeriodSeries(Series[Period]):
     # ignore needed because of mypy
--- a/pandas-stubs/core/series.pyi
+++ b/pandas-stubs/core/series.pyi
@@ -837,6 +837,14 @@ class Series(IndexOpsMixin[S1], NDFrame):
     ) -> DataFrame: ...
     def map(self, arg, na_action: Literal["ignore"] | None = ...) -> Series[S1]: ...
     @overload
+    def aggregate(  # type: ignore[misc]
+        self: Series[int],
+        func: Literal["mean"],
+        axis: AxisIndex = ...,
+        *args,
+        **kwargs,
+    ) -> float: ...
+    @overload
     def aggregate(
         self,
         func: AggFuncTypeBase,
--- a/pandas-stubs/core/series.pyi
+++ b/pandas-stubs/core/series.pyi
@@ -859,7 +859,7 @@ class Series(IndexOpsMixin[S1], NDFrame):
         axis: AxisIndex = ...,
         *args,
         **kwargs,
-    ) -> Self: ...
+    ) -> Series: ...
     agg = aggregate
     @overload
     def transform(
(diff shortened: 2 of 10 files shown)


pandas-stubs/core/frame.pyi
```diff
@@ -555,7 +555,7 @@ class DataFrame(NDFrame, OpsMixin):
     @overload
     def __getitem__(  # type: ignore[misc]
         self,
-        key: Series[_bool]
+        key: Series
```
twoertwein, line 558: One DataFrame test does `df[Series[str]]`

pandas-stubs/core/series.pyi
```diff
@@ -232,25 +235,27 @@ class Series(IndexOpsMixin[S1], NDFrame):
         index: Axes | None = ...,
         *,
         dtype: TimestampDtypeArg,
-        name: Hashable | None = ...,
+        name: Hashable = ...,
         copy: bool = ...,
     ) -> TimestampSeries: ...
```
twoertwein, line 240: Mypy says that the first two overloads overlap with incompatible return types even though they return exactly the same type! (The other overlapping errors might be legit - I have not yet ignored them)

pandas-stubs/core/series.pyi
```diff
@@ -211,18 +211,21 @@ class _LocIndexerSeries(_LocIndexer, Generic[S1]):
...
     __hash__: ClassVar[None]
 
-    # TODO: can __new__ be converted to __init__? Pandas implements __init__
     @overload
     def __new__(
         cls,
-        data: DatetimeIndex | Sequence[Timestamp | np.datetime64 | datetime],
+        data: DatetimeIndex | Sequence[np.datetime64 | datetime],
         index: Axes | None = ...,
+        *,
```
twoertwein, line 226: Consistently use `*,` after `index` and simplify some unions.

pandas-stubs/core/series.pyi
```diff
@@ -260,35 +265,39 @@ class Series(IndexOpsMixin[S1], NDFrame):
...
-        name: Hashable | None = ...,
+        name: Hashable = ...,
         copy: bool = ...,
     ) -> IntervalSeries[_OrderableT]: ...
     @overload
     def __new__(
         cls,
-        data: object | _ListLike | Series[S1] | dict[int, S1] | dict[_str, S1] | None,
-        dtype: type[S1],
+        data: object,
```
twoertwein, line 276: If we have object here, we need nothing else.

tests/test_series.py
```diff
@@ -242,8 +246,8 @@ def test_types_drop_multilevel() -> None:
 
 
 def test_types_dropna() -> None:
-    s = pd.Series([1, np.nan, np.nan])
-    check(assert_type(s.dropna(), pd.Series), pd.Series)
+    s = pd.Series([1.0, np.nan])
```
twoertwein, line 249: pyright would otherwise infer it as `Series[int | float]` while mypy infers `Series[float]`

tests/test_series.py
```diff
@@ -842,7 +872,7 @@ def test_update() -> None:
     s1.update(pd.Series([0, 2, 12]))
     # Series.update() accepting objects that can be coerced to a Series was added in 1.1.0 https://pandas.pydata.org/docs/whatsnew/v1.1.0.html
     s1.update([1, 2, -4, 3])
-    s1.update([1, "b", "c", "d"])
+    s1.update([1, "b", "c", "d"])  # type: ignore[list-item] # pyright: ignore[reportGeneralTypeIssues]
```
twoertwein, line 875: I'm not sure what to do here - update is inplace, so I think we have to reject this type-unsafe usage.

twoertwein:
I don't know why both mypy and pyright think that `Series[int].xs` is `int`!

pandas-stubs/core/frame.pyi
Dr-Irv: So shouldn't we just do `Series[_bool] | Series[str]` to not widen it too much?

pandas-stubs/core/series.pyi
Dr-Irv: Should we consider removing `object` and use `_ListLike | Series[S1] | dict[int, S1] | dict[_str, S1] | None` ?

pandas-stubs/core/series.pyi
```diff
@@ -260,35 +265,39 @@ class Series(IndexOpsMixin[S1], NDFrame):
...
+        *,
+        dtype: type[S1],
+        name: Hashable = ...,
         copy: bool = ...,
     ) -> Self: ...
     @overload
     def __new__(
         cls,
-        data: Series[S1] | dict[int, S1] | dict[_str, S1] = ...,
+        data: _ListLike[S1] | dict[int, S1] | dict[_str, S1],
```
Dr-Irv, line 286: Is `_ListLike` considered generic?

pandas-stubs/core/series.pyi
```diff
@@ -260,35 +265,39 @@ class Series(IndexOpsMixin[S1], NDFrame):
...
         dtype: Dtype = ...,
-        name: Hashable | None = ...,
+        name: Hashable = ...,
         copy: bool = ...,
     ) -> Self: ...
     @overload
     def __new__(
         cls,
-        data: object | _ListLike | None = ...,
+        data: object = ...,
```
Dr-Irv, line 296: Maybe we should use `_ListLike | None` and remove `object`?

pandas-stubs/core/series.pyi
```diff
@@ -676,15 +685,15 @@ class Series(IndexOpsMixin[S1], NDFrame):
     def diff(self, periods: int = ...) -> Series[S1]: ...
     def autocorr(self, lag: int = ...) -> float: ...
     @overload
-    def dot(self, other: Series[S1]) -> Scalar: ...
+    def dot(self, other: Series[S1]) -> Scalar: ...  # type: ignore[misc] # pyright: ignore[reportOverlappingOverload]
     @overload
     def dot(self, other: DataFrame) -> Series[S1]: ...
     @overload
     def dot(self, other: _ListLike) -> np.ndarray: ...
```
Dr-Irv, line 692: Maybe use something besides `_ListLike` here to avoid the overlapping overload a few lines earlier?

pandas-stubs/core/series.pyi
```diff
@@ -1344,7 +1353,7 @@ class Series(IndexOpsMixin[S1], NDFrame):
         na_option: Literal["keep", "top", "bottom"] = ...,
         ascending: _bool = ...,
         pct: _bool = ...,
-    ) -> Series: ...
+    ) -> Series[float]: ...
```
Dr-Irv, line 1356: I think `rank` returns `Series[int]`

pandas-stubs/core/series.pyi
```diff
@@ -2098,7 +2105,7 @@ class TimedeltaSeries(Series[Timedelta]):
         axis: AxisIndex = ...,
         level: Level | None = ...,
         drop_level: _bool = ...,
-    ) -> Series: ...
+    ) -> Self: ...
```
Dr-Irv, line 2108: Maybe `xs` should only be in the parent class for `Series` ?

tests/test_series.py
Dr-Irv: OK.

tests/test_series.py
```diff
@@ -263,28 +267,42 @@ class MyEnum(Enum):
 
 
 def test_types_fillna() -> None:
-    s = pd.Series([1, np.nan, np.nan, 3])
-    check(assert_type(s.fillna(0), pd.Series), pd.Series)
-    check(assert_type(s.fillna(0, axis="index"), pd.Series), pd.Series)
+    s = pd.Series([1.0, np.nan])
```
Dr-Irv, line 270: why not just use `s = pd.Series([1.0, np.nan, np.nan, 3.0])` to make it closer to the original test?

tests/test_series.py
```diff
@@ -263,28 +267,42 @@ class MyEnum(Enum):
...
     # GH 263
-    check(assert_type(s.fillna(pd.NA), pd.Series), pd.Series)
+    check(assert_type(s.fillna(pd.NA), "pd.Series[float]"), pd.Series, float)
 
 
 def test_types_sort_index() -> None:
-    s = pd.Series([1, 2, 3], index=[2, 3, 1])
-    check(assert_type(s.sort_index(), pd.Series), pd.Series)
-    check(assert_type(s.sort_index(ascending=False), pd.Series), pd.Series)
+    s = pd.Series([1, 2], index=[2, 3])
```
Dr-Irv, line 295: why change this?

tests/test_series.py
```diff
@@ -294,25 +312,37 @@ def test_types_sort_index_with_key() -> None:
 
 
 def test_types_sort_values() -> None:
-    s = pd.Series([4, 2, 1, 3])
-    check(assert_type(s.sort_values(), pd.Series), pd.Series)
+    s = pd.Series([4, 2])
```
Dr-Irv, line 315: why change this?

tests/test_series.py
```diff
@@ -294,25 +312,37 @@ def test_types_sort_index_with_key() -> None:
...
+        assert_type(s.sort_values(ignore_index=True), "pd.Series[int]"),
+        pd.Series,
+        np.intp,
+    )
 
 
 # This was added in 1.1.0 https://pandas.pydata.org/docs/whatsnew/v1.1.0.html
 def test_types_sort_values_with_key() -> None:
-    s = pd.Series([1, 2, 3], index=[2, 3, 1])
+    s = pd.Series([1, 2], index=[2, 3])
```
Dr-Irv, line 340: why change this?

tests/test_series.py
```diff
@@ -294,25 +312,37 @@ def test_types_sort_index_with_key() -> None:
...
     res: pd.Series = s.sort_values(key=lambda k: -k)
 
 
 def test_types_shift() -> None:
-    s = pd.Series([1, 2, 3])
+    s = pd.Series([1, 2])
```
Dr-Irv, line 345: why change this?

tests/test_series.py
```diff
@@ -359,11 +389,11 @@ def test_types_sum() -> None:
     # 2. Runtime return types of `series.sum(min_count=...)` are NOT
     #    tested (because of potential `nan`s).
 
-    s0 = assert_type(pd.Series([1, 2, 3, np.nan]), "pd.Series")
-    check(assert_type(s0.sum(), "Any"), np.float64)
-    check(assert_type(s0.sum(skipna=False), "Any"), np.float64)
-    check(assert_type(s0.sum(numeric_only=False), "Any"), np.float64)
-    assert_type(s0.sum(min_count=4), "Any")
+    s0 = assert_type(pd.Series([1.0, np.nan]), "pd.Series[float]")
```
Dr-Irv, line 392: Just use original `Series`, with same values as float

tests/test_series.py
Dr-Irv: It's valid usage.  Change the `def update` to use `Scalar` instead of `S1`

tests/test_series.py
```diff
@@ -878,40 +908,58 @@ def test_types_between() -> None:
...
-    check(assert_type(s.agg("min"), Any), np.int64)
-    check(assert_type(s.agg(["min", "max"]), pd.Series), pd.Series)
-    check(assert_type(s.agg({"a": "min"}), pd.Series), pd.Series)
-    check(assert_type(s.agg("mean", axis=0), Any), np.float64)
+    check(assert_type(s.agg("min"), int), np.int64)
+    check(assert_type(s.agg(["min", "max"]), "pd.Series[int]"), pd.Series, np.intp)
+    check(assert_type(s.agg({"a": "min"}), "pd.Series[int]"), pd.Series, np.intp)
+    check(
+        assert_type(  # type: ignore[assert-type]
+            s.agg("mean", axis=0), float  # pyright: ignore[reportGeneralTypeIssues]
```
Dr-Irv, line 916: that's not a good change

tests/test_series.py
```diff
@@ -878,40 +908,58 @@ def test_types_between() -> None:
...
-        check(assert_type(s.agg(x=max, y="min", z=np.mean), pd.Series), pd.Series)
+        check(assert_type(s.agg(min), int), np.int64)
+        check(assert_type(s.agg([min, max]), "pd.Series[int]"), pd.Series, np.intp)
+        check(assert_type(s.agg({0: min}), "pd.Series[int]"), pd.Series, np.intp)
+        check(
+            assert_type(  # type: ignore[assert-type]
+                s.agg(
+                    x=max, y="min", z=np.mean
+                ),  # pyright: ignore[reportGeneralTypeIssues]
+                "pd.Series[float]",
```
Dr-Irv, line 933: the ignores here aren't a good change

tests/test_series.py
```diff
@@ -1143,18 +1191,18 @@ def test_reset_index() -> None:
...
-    check(assert_type(r6, pd.Series), pd.Series)
+    check(assert_type(r6, "pd.Series[int]"), pd.Series, np.intp)
     assert assert_type(s.reset_index(inplace=True, drop=True), None) is None
 
 
 def test_series_add_str() -> None:
     s = pd.Series(["abc", "def"])
-    check(assert_type(s + "x", pd.Series), pd.Series)
-    check(assert_type("x" + s, pd.Series), pd.Series)
+    check(assert_type(s + "x", pd.Series), pd.Series, str)
```
Dr-Irv, line 1204: can't you assert against `pd.Series[str]` here?

tests/test_series.py
```diff
@@ -1165,14 +1213,20 @@ def test_series_dtype() -> None:
...
+        np.intp,
+    )
     assert assert_type(s.replace(1, 2, inplace=True), None) is None
 
 
 def test_cat_accessor() -> None:
     # GH 43
-    s = pd.Series(pd.Categorical(["a", "b", "a"], categories=["a", "b"]))
+    s: pd.Series[str] = pd.Series(
+        pd.Categorical(["a", "b", "a"], categories=["a", "b"])
```
Dr-Irv, line 1228: Maybe we should have a `CategoricalSeries` ?

tests/test_series.py
```diff
@@ -1679,8 +1733,8 @@ def test_AnyArrayLike_and_clip() -> None:
     ser = pd.Series([1, 2, 3])
     s1 = ser.clip(lower=ser)
     s2 = ser.clip(upper=ser)
-    check(assert_type(s1, pd.Series), pd.Series)
-    check(assert_type(s2, pd.Series), pd.Series)
+    check(assert_type(s1, "pd.Series[int]"), pd.Series)
+    check(assert_type(s2, "pd.Series[int]"), pd.Series)
```
Dr-Irv, line 1737: could add `np.intp` to the checks here

tests/test_series.py
```diff
@@ -2675,17 +2741,17 @@ def test_types_mask() -> None:
     s = pd.Series([1, 2, 3, 4, 5])
 
     # Test case with a boolean condition and a scalar value
-    check(assert_type(s.mask(s > 3, 10), pd.Series), pd.Series, np.integer)
+    check(assert_type(s.mask(s > 3, 10), "pd.Series[int]"), pd.Series, np.integer)
```
Dr-Irv, line 2744: We're inconsistent using `np.intp` versus `np.integer`, so we should use one or the other throughout (probably the latter)

Dr-Irv (commented):
I saw this is in draft form, but I think my comments would need to be addressed anyway

pandas-stubs/core/frame.pyi
twoertwein: I think it can be any Series:

```py
>>> pd.DataFrame({1: [0]})[pd.Series([1])]
   1
0  0
>>> pd.DataFrame({1.0: [0]})[pd.Series([1.0])]
   1.0
0    0
>>> x = object()
>>> pd.DataFrame({x: [0]})[pd.Series([x])]
   <object object at 0x7fa66230ff30>
0                                  0
```

pandas-stubs/core/series.pyi
twoertwein: This PR made it generic:

```
_ListLike: TypeAlias = (
    ArrayLike | dict[_str, np.ndarray] | Sequence[S1] | IndexOpsMixin[S1]
)
```

pandas-stubs/core/series.pyi
twoertwein: One would assume that, but that isn't the case:

```py
>>> pd.Series([1, 2, 3, 4]).rank()
0    1.0
1    2.0
2    3.0
3    4.0
dtype: float64
```

I will add a test case for that

tests/test_series.py
twoertwein: I was shortening some of the Series as their length is irrelevant from a typing perspective. Can revert it back.

tests/test_series.py
twoertwein: This means that `pd.Series([0, 2, 12]).update([1, "b", "c", "d"])` is of type `Series[int]`.


If we want to catch bugs, I would rather reject it and have people be aware of this type-unsafe usage. But I'm also happy to allow it (most stubs are far more lenient and avoid falsely flagging an error).

tests/test_series.py
twoertwein: I think we can either have `s.agg("min")` or `s.agg("mean")` to be correct or go back to `Any`. Adding literal overloads for all possible function names just because of a small difference between int and float sounds like a pain.

tests/test_series.py
twoertwein: I think we might need to reconsider S1 at some point (not in this PR). I would slightly prefer `S1 = TypeVar("S1")` as any type can be in a Series.

tests/test_series.py
twoertwein: I would prefer the narrower `np.intp` and only use `np.integer` when pandas inconsistently returns np.int32/64 on windows vs linux. I'm also fine with going for `np.integer` as this is just a runtime check.

pandas-stubs/core/series.pyi
twoertwein: also had to add `Scalar` (or `S1`) as `pd.Series(1)` is valid and tested.

pandas-stubs/core/frame.pyi
Dr-Irv: Sure, but do we want to encourage that?  I'd rather limit the possibilities than allow anything, because it is implying that we are allowing anything to be a column name.

tests/test_series.py
Dr-Irv: I see your point.  Then let's then keep the test, but surround it with a `TYPE_CHECKING_INVALID_USAGE`

pandas-stubs/core/series.pyi
Dr-Irv: Actually, it is `Series[float]` .  If there is a tie, it returns the average of the ranks.

tests/test_series.py
Dr-Irv: I think we should go back to `Any` if a string function name is passed.

tests/test_series.py
Dr-Irv: Yes, but I'd rather encourage "typical" types to be in a `Series`, and let people do `cast` or `ignore` if they are putting other types in a `Series`.

tests/test_series.py
Dr-Irv: But if we start using `np.integer`, and make that our going forward practice, we won't have to worry about the platform inconsistencies (as you had to do in the other PR for generic `Index`)

tests/test_series.py
Dr-Irv: We could do a `CategoricalSeries` as a future PR

Dr-Irv (changes requested):
It all looks good.  Some open comments to address.  I think this is the list:

1. Where you changed the construction of the `Series` to shorten them, I'd rather leave them as is to have less changes to the tests.
2. For `Series.update()`, need to surround that `ignore` with a `TYPE_CHECKING_INVALID_USAGE` pattern
3. Use `np.integer` rather  than `np.intp`
4. For strings passed into `Series.agg()`, maybe we just need to have an overload for `np.mean` and `"mean"` to say they return `Series[float]` (as opposed to what I wrote elsewhere about returning `Any`) and then  you can remove the `ignore` in the tests.

pandas-stubs/core/frame.pyi
twoertwein: > implying that we are allowing anything to be a column name.

I think that is already the case `pd.DataFrame(..., columns: Axes)` and `Axes: TypeAlias = AnyArrayLike | list | dict | range | tuple`. The only potential harm I see is that some `__getitem__` overloads might become ambiguous but that doesn't seem the be the case.

Dr-Irv: OK - let's go with your change with an untyped `Series`

twoertwein:
I believe I fixed everything but agg/aggregate.  will look into whether we can do something like:

- special case "mean"
- any string return S1
- for callables: return the actual return type

pandas-stubs/core/series.pyi
```diff
@@ -834,7 +859,7 @@ class Series(IndexOpsMixin[S1], NDFrame):
         axis: AxisIndex = ...,
         *args,
         **kwargs,
-    ) -> Series[S1]: ...
+    ) -> Self: ...
```
Dr-Irv, line 862: Maybe make this `Series` since we really don't know the type of the `Series` - could be mix of int and float, or one, or the other

Dr-Irv (commented):
Pretty close.  Just that one case where there is a mix of funcs on the `Series.agg()` call.  Maybe the solution is that if a list is passed, the result is `Series` instead of `Series[S1]` .  See comment

pandas-stubs/core/series.pyi
twoertwein: Thank you, that made it work!

Dr-Irv (approved):
thanks @twoertwein

PR #772, opened 2023-09-11
Fix flavor param with incorrect type hint in read_html

matheusfelipeog:
## refs:
- https://github.com/pandas-dev/pandas/issues/55059
- https://github.com/pandas-dev/pandas/pull/55076


twoertwein:
Please add a test or this change.

twoertwein:
A test similar to https://github.com/pandas-dev/pandas-stubs/blob/6bb12157a810cc26e034e012bff6a0ec2da191d0/tests/test_io.py#L1240 Pandas-stubs writes test to verify that type checkers pick up the proposed changes (and to prevent regressions).

pandas-stubs/io/html.pyi
```diff
@@ -28,7 +28,7 @@ def read_html(
     io: FilePath | ReadBuffer[str],
     *,
     match: str | Pattern = ...,
-    flavor: str | None = ...,
+    flavor: str | Sequence[str] | None = ...,
```
Dr-Irv, line 31: I don't see in the docs how a "sequence" of `str` is supported (although I think it is supported in the code).

So you should also do a PR in pandas to fix the docs.  Once that is accepted, then we can modify the stubs accordingly to support a `Sequence`.

Secondly, a better improvement here would be to only indicate the valid values rather than any string, i.e. `Literal["lxml", "html5lib", "bs4"] | None`  and if a sequence is supported, you could do the following

```python
HTMLFlavors: TypeAlias = Literal["lxml", "html5lib", "bs4"] | None
```
and then `flavor` is `HTMLFlavors | Sequence[HTMLFlavors] | None`

matheusfelipeog: Sorry for the delay in responding 😅

> I don't see in the docs how a "sequence" of `str` is supported (although I think it is supported in the code).

Yes, apparently it is accepted. I mentioned this in this issue: https://github.com/pandas-dev/pandas/issues/55059

> Secondly, a better improvement here would be to only indicate the valid values rather than any string, i.e. `Literal["lxml", "html5lib", "bs4"] | None` and if a sequence is supported, you could do the following
> 
> ```python
> HTMLFlavors: TypeAlias = Literal["lxml", "html5lib", "bs4"] | None
> ```
> 
> and then `flavor` is `HTMLFlavors | Sequence[HTMLFlavors] | None`

Personally, this is much better. Thanks.

I will work on adding this type to pandas instead of a generic str sequence and submit a PR for review.

Then I modify the stubs accordingly as mentioned.

matheusfelipeog:
Done, @twoertwein and @Dr-Irv.

The PR https://github.com/pandas-dev/pandas/pull/55529 adding the HTMLFlavors type has been approved in the pandas repository. As discussed, I've updated it here to align with it.

Dr-Irv (approved):
thanks @matheusfelipeog

PR #775, opened 2023-09-18
Added `Timestamp` to `Series#__getitem__`

### LINKED ISSUE #773: `Series#__getitem__` doesn't accept `pd.Timestamp`
**Describe the bug**

`Series#__getitem__` doesn't accept `pd.Timestamp`

**To Reproduce**

```py

import pandas as pd

index = pd.date_range("2018-01-01", periods=5, freq="D")
series = pd.Series([1, 2, 3, 4, 5], index=index)
value = series[index[-1]]  # error
print(value)
```

`mypy` output:

```
test.py:7: error: No overload variant of "__getitem__" of "Series" matches argument type "Timestamp"  [call-overload]
test.py:7: note: Possible overload variants:
test.py:7: note:     def __getitem__(self, list[str] | Index[Any] | Series[Any] | slice | Series[bool] | ndarray[Any, dtype[bool_]] | list[bool] | tuple[Any | slice, ...], /) -> Series[Any]
test.py:7: note:     def __getitem__(self, int | str, /) -> Any
```

Until it's fixed in `pandas-stubs` this can be used as a quick-fix:

```diff
-value = series[index[-1]]
+value = series.loc[index[-1]]
```

 - OS: macOS Ventura 13.5.2 (22G91)
 - Python: 3.10.12
 - Mypy: 1.5.1 (compiled: yes)
 - `pandas-stubs`: 2.0.3.230814

GarrisonD:
Closes #773

### DIFF
--- a/tests/test_series.py
+++ b/tests/test_series.py
@@ -1002,7 +1002,7 @@ def test_types_getitem() -> None:
 def test_types_getitem_by_timestamp() -> None:
     index = pd.date_range("2018-01-01", periods=2, freq="D")
     series = pd.Series(range(2), index=index)
-    check(assert_type(series[index[-1]], int), int)
+    check(assert_type(series[index[-1]], int), np.integer)
 
 
 def test_types_eq() -> None:


tests/test_series.py
```diff
@@ -998,6 +998,11 @@ def test_types_getitem() -> None:
     value: int = s2[0]
     s3: pd.Series = s[:2]
 
+def test_types_getitem_by_timestamp() -> None:
+    index = pd.date_range("2018-01-01", periods=2, freq="D")
+    series = pd.Series(range(2), index=index)
+    value: int = series[index[-1]]
```
Dr-Irv, line 1004: Can you replace this with `check(assert_type(series[index[-1]], int), int)`

Dr-Irv (changes requested):
It also appears our CI is failing for unrelated reasons, so I need to investigate that and fix it.  Probably a versioning issue.

tests/test_series.py
GarrisonD: The type check of the returned value was a nice to have. Should I just replace it with `series[index[-1]]` since it is not the subject of the test?

Dr-Irv: With new tests, we are always using `check` and `assert_type`, so I'm suggesting to replace `value: int = series[index[-1]]` with `check(assert_type(series[index[-1]], int), int)`

Dr-Irv:
We need to get #776 approved and merged, then you remerge with main, then I can run the tests.

Dr-Irv:
@GarrisonD can you merge with `main` and push, then I will fire off the CI, as the issue with that has been fixed.

pandas-stubs/core/series.pyi
```diff
@@ -360,7 +360,7 @@ class Series(IndexOpsMixin[S1], NDFrame):
         | tuple[Hashable | slice, ...],
     ) -> Self: ...
     @overload
-    def __getitem__(self, idx: int | _str) -> S1: ...
+    def __getitem__(self, idx: int | _str | Timestamp) -> S1: ...
```
twoertwein, line 363: Is the underlying issue here that Series is not generic in terms of its index's dtype?

As Series is not generic in its index, I would be inclined to widen it to `idx: Any` - can easily come up with other false positives otherwise.

GarrisonD:
@Dr-Irv Sure, will do tomorrow.

pandas-stubs/core/series.pyi
Dr-Irv: I think `Any` would be too wide.   Maybe we should use `Scalar` ?  Thoughts @twoertwein ?

twoertwein: `Scalar` sounds like a good idea - that probably also avoids overlapping overloads.

Dr-Irv: @GarrisonD based on the discussion above, can you replace `int | _str` with `Scalar`, merge with `main` and then we can run the test.

GarrisonD:
Can I move this check to https://github.com/pandas-dev/pandas-stubs/pull/775/files#diff-ea813cbfeb870b2b17f19af553004745162233003506303a2a577e86da7a5836R994 and in my test just have `series[index[-1]]` that will be checked by `mypy`/`pyright`?

tests/test_series.py
```diff
@@ -999,6 +999,12 @@ def test_types_getitem() -> None:
     s3: pd.Series = s[:2]
 
 
+def test_types_getitem_by_timestamp() -> None:
+    index = pd.date_range("2018-01-01", periods=2, freq="D")
+    series = pd.Series(range(2), index=index)
+    check(assert_type(series[index[-1]], int), np.integer)
```
GarrisonD, line 1005: I know, it was me initially who added a check for the return type but I see it useless now.

`mypy`/`pyright` will check if it's legal:

```suggestion
    series[index[-1]]
```

twoertwein: I wouldn't mind keeping the return check, even if it isn't necessary for your change

Dr-Irv: I agree.  We're hoping to get all the tests to use that pattern.

Dr-Irv:
> Can I move this check to https://github.com/pandas-dev/pandas-stubs/pull/775/files#diff-ea813cbfeb870b2b17f19af553004745162233003506303a2a577e86da7a5836R994 and in my test just have `series[index[-1]]` that will be checked by `mypy`/`pyright`?

Merging your test function with the one above would have been fine, but we do want to try using `check(assert_type(...` in new tests.  I'm going to approve/merge this in now because it passed our tests.

Dr-Irv (approved):
thanks @GarrisonD

PR #776, opened 2023-09-18
fix matplotlib imports, pd.util

Dr-Irv:
CI was failing due to matplotlib changes.  This fixes those issues.


pandas-stubs/plotting/_core.pyi
```diff
@@ -239,7 +239,7 @@ class PlotAccessor:
         **kwargs: Any,
     ) -> pd.Series: ...
     @overload
-    def line(
+    def line(  # type: ignore[misc]
```
twoertwein, line 242: Just for me to understand why these changes are needed: mypy&pyright cannot resolve Axis, so it is treated as Any/Unknown, therefore the overloads overlap. Is that correct?

tests/test_pandas.py
```diff
@@ -13,6 +13,7 @@
 import pandas as pd
 from pandas import Grouper
 from pandas.api.extensions import ExtensionArray
+import pandas.util as pdutil
```
twoertwein, line 16: Maybe add a `# TODO: github.com/pandas-dev/pandas/issues/55023` I assume this would fix the workaround?

tests/test_plotting.py
```diff
@@ -1,6 +1,8 @@
 import io
 from typing import Any
 
+from matplotlib.axes import Axes
+from matplotlib.figure import Figure
```
twoertwein, line 5: I assume this just hides the explicit re-export issue (I'm surprised that mypy/pyright are fooled that easily).

twoertwein (approved):
Except for adding a comment about `pd.util` it looks good to me.

(Should probably open an issue/pr at matplotlib so that Axis/Figure are ex-exported)

pandas-stubs/plotting/_core.pyi
Dr-Irv: I think this is because `matplotlib` changed the way that `Axes` is declared.

tests/test_plotting.py
Dr-Irv: No, I think `matplotlib` got updated, and now `plt.Axes` and `plt.Figure` aren't valid any more in code from a type checking perspective.

tests/test_pandas.py
Dr-Irv: I think so.  In next commit.

pandas-stubs/plotting/_core.pyi
twoertwein: I looked at their code: maybe the `from ... import *` or the many class decorators throw mypy/pyright off.

twoertwein:
Thanks @Dr-Irv !

Dr-Irv:
Create the matplotlib issue at https://github.com/matplotlib/matplotlib/issues/26812

pandas-stubs/plotting/_core.pyi
ksunden: I don't think this has to do with `matplotlib`, but is rather a change in `mypy` (Saw similar things in `xarray`)

Here, I think the problem may actually be that you have a default value indicated for both the `True` and `False` case of the subplots kwarg, when I think that should only apply in one case.

twoertwein: @ksunden I think you are right - one of the overloads should require a user-provided input!

Dr-Irv: > @ksunden I think you are right - one of the overloads should require a user-provided input!

Created #777

PR #779, opened 2023-09-19
fix: chained_assignment could be warn, raise or None

leviplj:
According to panda's code, the option mode.chained_assignment could be `warn`, `raise` or None. 

https://github.com/pandas-dev/pandas/blob/1.5.x/pandas/core/config_init.py#L563-L575


Dr-Irv (approved):
Thanks @leviplj .  We usually require tests for any changes to the stubs, but since we had no tests for config options, I'm not going to require adding one.

PR #781, opened 2023-09-23
Fix nightly and support 2.1.1

Dr-Irv:
- Updated some PYI files to disallow future deprecated features. 
- Updated tests to work with 2.1 and with nightly


pandas-stubs/_libs/tslibs/offsets.pyi
```diff
@@ -125,7 +128,13 @@ class BusinessMixin(SingleConstructorOffset):
         self, n: int = ..., normalize: bool = ..., offset: timedelta = ...
     ) -> None: ...
 
-class BusinessDay(BusinessMixin): ...
+class _NotForPeriodBusinessMixin(_NotForPeriodSingleConstructorOffset):
```
twoertwein, line 131: Those two private helper classes seem to be only used by BusinessDay. Would it make sense to add their methods to BusinessDay instead?

pandas-stubs/_libs/tslibs/offsets.pyi
```diff
@@ -125,7 +128,13 @@ class BusinessMixin(SingleConstructorOffset):
...
+    def __init__(
+        self, n: int = ..., normalize: bool = ..., offset: timedelta = ...
+    ) -> None: ...
+
+class _NotForPeriodDtype: ...
```
twoertwein, line 136: Seems unused

pyproject.toml
```diff
@@ -30,16 +30,17 @@ packages = [
 "Documentation" = "https://pandas.pydata.org/pandas-docs/stable"
 
 [tool.poetry.dependencies]
-python = ">=3.9"
+python = ">=3.9 <3.13"
```
twoertwein, line 33: I assume that is needed because one of the dependencies added this constraint? Might be nicer to move this constraint to that specific dependency. Something like `{ version = ">=x.y.z" , python = "<4"} `

xref #625

pyproject.toml
```diff
@@ -30,16 +30,17 @@ packages = [
...
 
 [tool.poetry.dev-dependencies]
 mypy = "1.5.1"
-pandas = "2.0.3"
+pandas = "2.1.1"
 pyarrow = ">=10.0.1"
 pytest = ">=7.1.2"
-pyright = ">= 1.1.327"
+# pyright 1.1.327 has bug fixed in 1.1.328
+pyright = "1.1.326"
```
twoertwein, line 43: Which bug is affecting pandas-stubs?

tests/test_frame.py
```diff
@@ -180,9 +179,8 @@ class MyEnum(Enum):
 
 
 def test_slice_setitem() -> None:
-    # Due to the bug in pandas 1.2.3(https://github.com/pandas-dev/pandas/issues/40440), this is in separate test case
     df = pd.DataFrame(data={"col1": [1, 2], "col2": [3, 4], 5: [6, 7]})
-    df[1:] = ["a", "b", "c"]
+    df[1:] = [10, 11, 12]
```
twoertwein, line 183: Is this because pandas does no longer upcast the dtype?

tests/test_frame.py
```diff
@@ -2207,53 +2206,61 @@ def test_groupby_apply() -> None:
...
+        pd.Series,
     )
 
     def sample_to_df(x: pd.DataFrame) -> pd.DataFrame:
         return x.sample()
 
     check(
         assert_type(
-            df.groupby("col1", group_keys=False).apply(sample_to_df), pd.DataFrame
+            df.groupby("col1", group_keys=False)["col2"].apply(sample_to_df), pd.Series
```
twoertwein, line 2237: Is it possible to reflect this change in the stubs and test it as an invalid usage?

pandas-stubs/_libs/tslibs/offsets.pyi
Dr-Irv: Good point.  I got rid of the two classes, but made a comment in there that indicates why it is done this way.

pandas-stubs/_libs/tslibs/offsets.pyi
Dr-Irv: removed

pyproject.toml
Dr-Irv: It now pulls in numpy 1.26.0 and scipy 1.11.2, which have the python < 3.13 restriction, so I put that in.

pyproject.toml
Dr-Irv: Not clear, but 1.1.328 is installed in VS Code pre-release and within VS Code, I get no complaints, but when I install 1.1.327, it doesn't like the use of this pattern in `core/generic.pyi`:
```python
def somefunc(self: NDFrameT, someargs) -> NDFrameT: ...
```
complaining like this:
```
core\generic.pyi:60:15 - error: Type of parameter "self" must be a supertype of its class "NDFrame" (reportGeneralTypeIssues)
```

1.1.328 is in VSCode, but not on pypi yet

tests/test_frame.py
Dr-Irv: Yes, and since the goal of the test was to test the slice, I figured I would just change the values that way.

Otherwise in 2.1, we get a `FutureWarning`

tests/test_frame.py
Dr-Irv: I think I was too aggressive on this change with respect to supporting nightly.  I've put the code back the way it was and surrounded it with `pytest_warns_bounded`

Dr-Irv:
@twoertwein addressed all your comments in https://github.com/pandas-dev/pandas-stubs/pull/781/commits/3274b29470501acabca37c57ffd03de31eba2b9a

pyproject.toml
twoertwein: Would Self fix the issue?

twoertwein:
Thank you @Dr-Irv !

pyproject.toml
Dr-Irv: If 1.1.328 also fails once released, then I will try that

PR #782, opened 2023-09-23
Fix a few issues found in my team's code

### LINKED ISSUE #641: For `read_excel()`, change result of `dict` type based on type of the list or `None`
With `read_excel()`, if `sheet_name=None`, or  is `list[str]` then we know that the returned keys of the dictionary will be strings, i.e., the result is `dict[str, pd.DataFrame]`

The result could only be `dict[int, pd.DataFrame]` if  `list[int]` is passed. 

So we should split the first overload into two - returning different results based on the type passed to `sheet_name`.

### LINKED ISSUE #743: `DataFrame.loc.__setitem__()` and `Series.loc.__setitem__()` need to support `pd.NA` and `pd.NaT` as possible values
**Describe the bug**
For assigning a value using `.loc` and `.iloc`, we should allow `pd.NA` and `pd.NaT` to be the value.

**To Reproduce**
```python
import pandas as pd
df = pd.DataFrame({"x":[1,2,3]}, index=pd.Index(["a", "b", "c"])).convert_dtypes()
print("orig ", df)

ind = pd.Index(["a", "c"])

df.loc[ind, :] = pd.NA

print("new ", df)
```

With pyright, message is:
```text
setind.py
setind.py:7:1 - error: Argument of type "tuple[Index, slice]" cannot be assigned to parameter "idx" of type "tuple[_IndexSliceTuple[Unknown], HashableT@__setitem__]" in function "__setitem__"
    "tuple[Index, slice]" is incompatible with "tuple[_IndexSliceTuple[Unknown], HashableT@__setitem__]"
        "Index" is incompatible with "_IndexSliceTuple[Unknown]" (reportGeneralTypeIssues)
setind.py:7:1 - error: Argument of type "NAType" cannot be assigned to parameter "value" of type "Scalar | ArrayLike | Series[Unknown] | list[Unknown] | None" in function "__setitem__"
    Type "NAType" cannot be assigned to type "Scalar | ArrayLike | Series[Unknown] | list[Unknown] | None"
      "NAType" is incompatible with "str"
      "NAType" is incompatible with "bytes"
      "NAType" is incompatible with "date"
      "NAType" is incompatible with "datetime"
      "NAType" is incompatible with "timedelta"
      "NAType" is incompatible with "datetime64"
      "NAType" is incompatible with "timedelta64" (reportGeneralTypeIssues)
```

 - OS: Windows 10 
 - OS Version 2H22
 - python version  3.10
 - version of type checker   pyright 1.1.316
 - version of installed `pandas-stubs`    "2.0.2.230605

**Additional context**

Need to change the `__setitem__()` declarations in `frame.pyi` and `series.pyi` for `_LocIndexerFrame`, `_iLocIndexerFrame`, `_LocIndexerSeries`, `_iLocIndexerSeries` that whenever they accept `Scalar`, they also accept `NAType` and `NaTType`

### LINKED ISSUE #767: `Series.loc.__setitem__()` with a `MultiIndex`, using `pd.IndexSlice` with an Index not picked up - OK for DataFrame
**Describe the bug**
Not allowing valid indexing with MultiIndex backing a Series.

**To Reproduce**
```python
import pandas as pd

df = (
    pd.DataFrame({"x": [1, 2, 3, 4]})
    .assign(y=lambda df: df["x"] * 10, z=lambda df: df["x"] * 100)
    .set_index(["x", "y"])
)

ind = pd.Index([2, 3])

df.loc[pd.IndexSlice[ind, :], "z"] = 25

s = df["z"]

s.loc[pd.IndexSlice[ind, :]] = 30

print(s)
```
produces
```
setitemindex.py:15:1 - error: Argument of type "tuple[Index[int], slice]" cannot be assigned to parameter "idx" of type "list[int] | list[str] | list[str | int]" in function "__setitem__"
    Type "tuple[Index[int], slice]" cannot be assigned to type "list[int] | list[str] | list[str | int]"
      "tuple[Index[int], slice]" is incompatible with "list[int]"
      "tuple[Index[int], slice]" is incompatible with "list[str]"
      "tuple[Index[int], slice]" is incompatible with "list[str | int]" (reportGeneralTypeIssues)
```

Note that it works for a `DataFrame`, but not a `Series`, so have to copy some of the code from `frame.pyi` to `series.pyi`

 - OS Windows 10
 - OS Version  22H1
 - python version   3.10
 - version of type checker   pyright 3.23
 - version of installed `pandas-stubs`   :  2.0.3.230814

**Additional context**

Dr-Irv:
- [x] Closes #767
- [x] Closes #743
- [x] Closes #641

- [x] Tests added: 
  - `test_frame.test_frame_setitem_na`
  - `test_series.test_series_setitem_multiindex`
  - `test_series.test_series_setitem_na`

Modified tests for `read_excel` to differentiate the cases of different arguments for `sheetname`


pandas-stubs/io/excel/_base.pyi
```diff
@@ -31,6 +31,86 @@ from pandas._typing import (
...
+    | Sequence[Sequence[str] | Sequence[int]]
+    | dict[str, Sequence[int] | list[str]] = ...,
+    date_format: dict[Hashable, str] | str | None = ...,
+    thousands: str | None = ...,
+    decimal: str = ...,
+    comment: str | None = ...,
+    skipfooter: int = ...,
+    storage_options: StorageOptions = ...,
+    dtype_backend: DtypeBackend | NoDefault = ...,
+) -> dict[str, DataFrame]: ...
```
twoertwein, line 113: Can the first two overloads be combined using a _StrIntT TypeVar? (I think we might even have such a typevar in _typing)

pandas-stubs/core/frame.pyi
```diff
@@ -152,7 +153,7 @@ class _iLocIndexerFrame(_iLocIndexer):
         | tuple[IndexType, int]
         | tuple[IndexType, IndexType]
         | tuple[int, IndexType],
-        value: Scalar | Series | DataFrame | np.ndarray | None,
+        value: Scalar | Series | DataFrame | np.ndarray | NAType | NaTType | None,
```
twoertwein, line 156: I wouldn't mind having `value: object` as we do not keep track of the type anyway.

pandas-stubs/io/excel/_base.pyi
twoertwein: pandas does that: https://github.com/pandas-dev/pandas/blob/9e1096e8373bc99675fd1b3490cfb7cf26041395/pandas/io/excel/_base.py#L426

pandas-stubs/core/frame.pyi
Dr-Irv: I'm not sure we should encourage putting arbitrary types inside of a `DataFrame` .  So I want to leave this as is.

pandas-stubs/io/excel/_base.pyi
Dr-Irv: I've put that in.  Still need 4 overloads because `sheetname = None` gives a `dict[str, pd.DataFrame]` result.

Dr-Irv:
@twoertwein  ready for another review

pandas-stubs/io/excel/_base.pyi
```diff
@@ -112,7 +111,8 @@ def read_excel(  # type:ignore[misc]
     dtype_backend: DtypeBackend | NoDefault = ...,
 ) -> dict[str, DataFrame]: ...
 @overload
-def read_excel(
+# mypy says this won't be matched
+def read_excel(  # type: ignore[misc]
```
twoertwein, line 115: Could maybe use bound=Int|Str for IntStrT to potentially get rid of this last overload?

twoertwein: the bound option would make the 3rd overload obsolete for pyright (it passes all tests) but mypy still needs the 3rd overload.

twoertwein:
Thanks @Dr-Irv !

pandas-stubs/core/frame.pyi
twoertwein: I'm not sure where pandas-stubs should be on the spectrum of what is documented to the subset of what is also recommended.

In terms of deprecations (basically a recommendation), I'm all for staying closer on the recommendation side, even though the deprecated behavior is still possible. When it comes to what to put in a Series (and especially in a dtype-unaware DataFrame), I feel more comfortable being on the what is documented side. I guess I prefer annotations that have a chance of being ported to pandas (I would probably object to restricting the type of value in pandas).

Dr-Irv: > I'm not sure where pandas-stubs should be on the spectrum of what is documented to the subset of what is also recommended.

I call this the "Goldilocks problem". We don't want the types to be too wide, or too narrow, but to be "just right".  The debate is what is "just right"
> 
> In terms of deprecations (basically a recommendation), I'm all for staying closer on the recommendation side, even though the deprecated behavior is still possible. When it comes to what to put in a Series (and especially in a dtype-unaware DataFrame), I feel more comfortable being on the what is documented side. I guess I prefer annotations that have a chance of being ported to pandas (I would probably object to restricting the type of value in pandas).

I agree that within pandas, what should be in there is what works, which probably means wide types in many places.  We'll have to see how this evolves over time.  My concern with having wide types is that it doesn't help users catch certain bugs via type checking.

PR #787, opened 2023-10-01
CI: use compiled nightly mypy

twoertwein:
See https://github.com/microsoft/pyright/issues/6044#issuecomment-1741429961


Dr-Irv (approved):
thanks @twoertwein

twoertwein:
Not bad, from 3min46 to 1min17 :)

twoertwein:
I'm unsure whether we can fix that issue by re-writing the overloads. In the worst case, we could ignore overload-specific ignores on the nightly run (we already ignore unused ignore comments).

PR #788, opened 2023-10-04
'Fix' mypy-nightly

twoertwein:
When specifying the generic type of IntervalSeries, the current mypy also sees it as an error.

### DIFF
--- a/pandas-stubs/_typing.pyi
+++ b/pandas-stubs/_typing.pyi
@@ -517,7 +517,7 @@ S1 = TypeVar(
     | datetime.datetime  # includes pd.Timestamp
     | datetime.timedelta  # includes pd.Timedelta
     | Period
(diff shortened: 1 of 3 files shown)


pandas-stubs/_typing.pyi
```diff
@@ -517,7 +517,7 @@ S1 = TypeVar(
     | datetime.datetime  # includes pd.Timestamp
     | datetime.timedelta  # includes pd.Timedelta
     | Period
-    | Interval[int | float | Timestamp | Timedelta]
```
twoertwein, line 520: mypy can't match _OrderableT with a Union of _OrderableT's bound

Dr-Irv (approved):
thanks @twoertwein .  I wouldn't have come up with that solution!

PR #792, opened 2023-10-06
TYP: NDFrame.__round__

### LINKED ISSUE #791: Argument `value` to `insert` of `DataFrame` has incompatible type `NDFrame`
I am having an issue with this. Here is my example:

```python
import pandas as pd
d = {'col1': [1, 2], 'col2': [3, 4]}
df = pd.DataFrame(data=d)
percent = round(df["col1"] / (df["col2"]) * 100,2)
df.insert(
            loc=1,
            column= "percent",
            value=percent
        )
```

This throws the following output in mypy:

```bash
error: Argument "value" to "insert" of "DataFrame" has incompatible type "NDFrame"; expected "str | bytes | date | datetime | timedelta | datetime64 | timedelta64 | bool | int | float | Timestamp | Timedelta | complex | Sequence[Any] | ndarray[Any, Any] | Series[Any] | Index[Any] | None"  [arg-type]
```

The code works just fine, the problem is specifically with mypy and the stubs.

I was able to make a workaround using numpy transformations and operations, but I preferred to stick with the pandas only code I show at the beginning and ended up ignoring the line.

twoertwein:
- [x] Closes #791 (Replace xxxx with the Github issue number)


twoertwein:
There are probably a few more cases where we could use Self or remove the method from NDFrame when DataFrame/Series both explicitly declare the method (typically with more precise dtype-aware version in Series).

pandas-stubs/core/generic.pyi
```diff
@@ -74,19 +74,16 @@ class NDFrame(indexing.IndexingMixin):
     def ndim(self) -> int: ...
     @property
     def size(self) -> int: ...
-    def swapaxes(
```
twoertwein, line 77: Is deprecated since 2.1

Dr-Irv (approved):
thanks @twoertwein .  Ideally, we'd have tests for the other things you changed, but that's a lot of work, so I'm OK with just putting these kinds of improvements in there.

PR #801, opened 2023-10-20
Add types for pandas.DataFrame.map

andrewpmk:
Added types for pandas.DataFrame.map (renamed from applymap in pandas 2.1.0 and later https://pandas.pydata.org/docs/dev/reference/api/pandas.DataFrame.map.html, applymap has been deprecated)

- [ ] Closes #774 and #800

PR #802, opened 2023-10-21
reportIncompatibleMethodOverride

twoertwein:
Compatibility with new pyright. Mostly:

- added ignores when mypy had ignores
- removed methods in child classes when they seem to be covered by the parent class (those might need double-checking)


Dr-Irv (changes requested):
Thanks for this PR.  The pyright changes were as a result of https://github.com/microsoft/pyright/issues/6158 addressed in PR https://github.com/microsoft/pyright/pull/6159



In #801 I asked to have pyright pinned to 1.1.331, but with your fix, we can remove that pinning, so can you now update to `>= 1.1.332` in `pyproject.toml` ?

Dr-Irv (approved):
Thanks @twoertwein .  This was a big change that I wasn't looking forward to investigating

PR #806, opened 2023-11-02
Update required python version to >= 3.9 in docs

matheusfelipeog:
ref: https://github.com/pandas-dev/pandas-stubs/blob/9aac8e31ba69eb4c0583e55dd2198755fb031620/pyproject.toml#L33

If you attempt to use version 3.8:
```zsh
❯ poetry env use 3.8
The currently activated Python version 3.8.18 is not supported by the project (>=3.9).
Trying to find and use a compatible version.
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -101,7 +101,7 @@ pip install pandas-stubs
 
 ## Installation from sources
 
-- Make sure you have `python >= 3.8` installed.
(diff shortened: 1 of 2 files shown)


Dr-Irv (approved):
thanks @matheusfelipeog

PR #807, opened 2023-11-02
Use the new poetry dev dependency group

matheusfelipeog:
The `*.dev-dependencies` section and the `--dev` flag of the `poetry add` command are deprecated. You will receive if you run:

```zsh
❯ poetry add --dev foo-pkg
The --dev option is deprecated, use the `--group dev` notation instead.
```

Instead, use groups (`*.group.dev.dependencies`):

```zsh
❯ poetry add foo-pkg --group dev
```

---

refs:
- https://python-poetry.org/docs/cli/#options-4
- https://python-poetry.org/docs/managing-dependencies/#dependency-groups

### DIFF
--- a/README.md
+++ b/README.md
@@ -111,7 +111,7 @@ conda install poetry
 
 ```sh
 # or PyPI
-pip install poetry
+pip install 'poetry>=1.2'
 ```
 
 - Install the project dependencies
--- a/docs/setup.md
+++ b/docs/setup.md
@@ -1,7 +1,7 @@
 ## Set Up Environment
 
 - Make sure you have `python >= 3.9` installed.
-- Install poetry: `pip install poetry`
+- Install poetry: `pip install 'poetry>=1.2'`
 - Install the project dependencies: `poetry update`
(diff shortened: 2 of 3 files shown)


Dr-Irv (changes requested):
Since `poetry` changed things, it would be best if you updated line 4 of `docs/setup.md` to indicate the minimal version of `poetry` required to use the group syntax.  That will help anybody (like myself) who has an older version installed, since I don't think there is a way to specify a poetry version in the `pyproject.toml` file (but maybe there is??)

matheusfelipeog:
@Dr-Irv You're right, it's better to specify the version. I saw there is a discussion about this in https://github.com/python-poetry/poetry/issues/3316, but there isn't currently a way to define it within the `pyproject.toml` file.

For now, we can specify it in the `docs/setup.md` file, as you suggested.

matheusfelipeog:
Done 😄

Dr-Irv (approved):
Thanks @matheusfelipeog

PR #815, opened 2023-11-08
Add more valid options for compression= in to_parquet()

gwerbin:
This matches the description in the API documentation for v2.1: https://pandas.pydata.org/pandas-docs/version/2.1/reference/api/pandas.DataFrame.to_parquet.html


Dr-Irv (approved):
Thanks @gwerbin .  Usually, we ask for tests, but in this case, I don't think it's worth it.

PR #816, opened 2023-11-08
Fix: dataframe.loc int label could be warn

dangjinghao:
According to [the pandas document](https://pandas.pydata.org/docs/reference/api/pandas.DataFrame.loc.html)
> A single label, e.g. 5 or 'a', (note that 5 is interpreted as a label of the index, and never as an integer position along the index).

dataframe.loc can accept int label, but for the previous stubs,`df.loc[1]` or similar codes would be warn.


Dr-Irv (approved):
Thanks @dangjinghao

PR #819, opened 2023-11-24
fix pyright and pytest --nightly

### LINKED ISSUE #803: Nightly failing as of 10/24/23
See  https://github.com/pandas-dev/pandas-stubs/actions/runs/6627496119/job/18002538307#step:4:623

twoertwein:
- [ ] Closes #803 (Replace xxxx with the Github issue number)

There are still many FutureWarnings triggered by `poe pytest --nightly` but they do not seem to fail pytest.


Dr-Irv (commented):
I tried this, and while we now don't fail, shouldn't we be trying to catch the `FutureWarning`  cases in the tests?  There were some `pytest_warns_bounded()` that you removed, but if you changed it to `FutureWarning`, it would have been correct.

The other advantage of trying to catch the warnings now is that when it is invalid due to a future release (e.g., 2.2 or 3.0), we know we have to change the tests to indicate this is invalid usage, and also change the stubs to disallow the usage.  So by having the `pytest_warns_bounded` in there, those changes are easier to spot.

twoertwein:
> I tried this, and while we now don't fail, shouldn't we be trying to catch the `FutureWarning` cases in the tests?

That would be good! Or should we already remove this functionality (assuming the new behavior is already possible)?


> There were some `pytest_warns_bounded()` that you removed, but if you changed it to `FutureWarning`, it would have been correct.

I didn't realize - will fix that.

Dr-Irv:
> > I tried this, and while we now don't fail, shouldn't we be trying to catch the `FutureWarning` cases in the tests?
> 
> That would be good! Or should we already remove this functionality (assuming the new behavior is already possible)?
> 
No, because for the current release, the "future behavior" is allowed without deprecation.  Assuming the next release is version 2.2, when that's released, we then remove the deprecated behavior from the stubs.

twoertwein:
How about warnings that are caught with `lower="2.0.99"` - they should move to the invalide section and the correspondig annotations should be removed, or?

Dr-Irv:
> How about warnings that are caught with `lower="2.0.99"` - they should move to the invalide section and the correspondig annotations should be removed, or?

Yes, that is what should happen.

Dr-Irv (approved):
Thanks @twoertwein

twoertwein:
> Yes, that is what should happen.

I will try to look into that in the coming weeks - I have been busy with my thesis and job search.

PR #824, opened 2023-12-10
Fix nightly CI again

twoertwein:
and remove some old warnings.


tests/test_frame.py
```diff
@@ -52,6 +49,10 @@
 DF = pd.DataFrame(data={"col1": [1, 2], "col2": [3, 4]})
 
 
+def getSeriesData() -> dict[str, pd.Series[float]]:
+    return {key: pd.Series([1.0, 2.0]) for key in ("A", "B", "C", "D")}
+
```
Dr-Irv, line 54: This got removed in the pandas dev branch, but I'd like to put back in here the version that was there:
```python
_N = 30
_K = 4

def getCols(k) -> str:
    return string.ascii_uppercase[:k]


def getSeriesData() -> dict[str, Series]:
    index = makeStringIndex(_N)
    return {
        c: pd.Series(np.random.default_rng(i).standard_normal(_N), index=index)
        for i, c in enumerate(getCols(_K))
    }
```

Dr-Irv (changes requested):

tests/test_frame.py
twoertwein: I added the function as it was in pandas but I inlined unused kwargs in two of the helper functions (`makeStringIndex` and `rands_array`).

Dr-Irv (approved):
thanks @twoertwein

jonashaag:
This removed `is_categorical_dtype` :(

twoertwein:
> This removed `is_categorical_dtype` :(

Pandas 2.1 recommends `isinstance(obj.dtype, pd.CategoricalDtype)` https://pandas.pydata.org/docs/dev/whatsnew/v2.1.0.html

PR #837, opened 2023-12-22
Allow sequence in groupby level

### LINKED ISSUE #836: Level parameter of DataFrame.groupby should allow Sequences
The level parameter of DataFrame.groupby does not except sequences, when it should. ([Corresponding Doc](https://pandas.pydata.org/docs/reference/api/pandas.DataFrame.groupby.html))
This is the same as https://github.com/pandas-dev/pandas/issues/47548

**To Reproduce**
From the linked Issue above
```python
import pandas as pd
df = pd.DataFrame([[1, 2, 3, 4], [5, 6, 7, 8]], columns=['a', 'b', 'c', 'd']).set_index(['a', 'b', 'c'])

print(df.groupby(level=[0, 1]).sum())
```
**My System**
 - Windows 10
 - Python 3.11
 - MyPy 1.5.1
 - `pandas-stubs` 2.1.4.231218

**Potential Fix**
The issue linked above was fixed by changing the type of `level` from `Level | None` to `IndexLabel | None`. (https://github.com/pandas-dev/pandas/pull/47560)
I suppose, the same Fix should be applied to the groupby-Overloads in pandas-stubs/core/frame.pyi.
I am happy to create a pull request, if that helps.
But i would appreciate it, if someone more involved with this library could confirm that the proposed fix makes sense.

jens-diewald:
- [x] Closes #836
- [x] Tests added:
  - One for series and one for DataFrame. The test case is pretty much the one from the Issue, adapted to the style of the surrounding code and I added a third row, so something actually would be grouped.
  - The template comment tells me to add `assert_type()` to assert the type of any return value. I do not really see value in checking any return type here, though?


Dr-Irv:
> * The template comment tells me to add `assert_type()` to assert the type of any return value. I do not really see value in checking any return type here, though?

We are doing this for consistency purposes.  There are still tests that have to adopt that convention (they are close to 2 years old), but we are using the convention on any new tests that are added.  So if you could add the `assert_type(check(...))` pattern into the tests, we'd maintain that consistency.

Dr-Irv (changes requested):
This all looks good.  As mentioned elsewhere, convert the tests to use the `check(assert_type(...))` pattern, and it should be good to go.

jens-diewald:
Is it fine like that?
Or should is split up the command and test the return type of groupby?

tests/test_frame.py
```diff
@@ -1082,7 +1082,10 @@ def test_types_groupby_level() -> None:
         "col4": [1, 2, 3],
     }
     df = pd.DataFrame(data=data).set_index(["col1", "col2", "col3"])
-    df.groupby(level=["col1", "col2"]).sum()
+    check(
+        assert_type(df.groupby(level=["col1", "col2"]).sum(), "pd.DataFrame"),
```
Dr-Irv, line 1086:
```suggestion
        assert_type(df.groupby(level=["col1", "col2"]).sum(), pd.DataFrame),
```
You don't need the quotes here because the type is known at runtime.

tests/test_series.py
```diff
@@ -445,7 +445,11 @@ def test_types_groupby_level() -> None:
         [(0, 0, 1), (0, 1, 2), (0, 0, 3)], names=["col1", "col2", "col3"]
     )
     s = pd.Series([1, 2, 3], index=index)
-    s.groupby(level=["col1", "col2"]).sum()
+    check(
+        assert_type(s.groupby(level=["col1", "col2"]).sum(), "pd.Series[int]"),
```
Dr-Irv, line 449: You do need the quotes here around `pd.Series[int]`, because the generic version is not known at runtime.

Dr-Irv (changes requested):
One small change in the tests.  I just ran them - failure is due to an odd MacOS issue that is intermittent, so once you make the change, I will run again, and if it passes, approve/merge the PR. 

Thanks!

tests/test_frame.py
jens-diewald: Makes sense, I did the change.
Thanks for the feedback.

Dr-Irv (approved):
thanks @jens-diewald

PR #839, opened 2023-12-23
CI: test on 3.12

### LINKED ISSUE #813: python 3.12 and pyright 1.1.334 error:  Method "fromtimestamp" overrides class "datetime" in an incompatible manner
**Describe the bug**
I'm Follwing the setup.md to set up an development envirment. When I  execute  `poe test_all` .Pyright test cannot be finished and prints this error message.

**To Reproduce**
unit test in `tests/`
3. Indicate which type checker you are using (`mypy` or  `pyright`).
pyright
5. Show the error message received from that type checker while checking your example.
```
prompt> poe test_all                                                                         (pandas_stubs_test) 
Poe => test_all

Success: no issues found in 224 source files

End: 'Run mypy on 'tests' (using the local stubs) and on the local stubs', runtime: 27.658 seconds.


Beginning: 'Run pyright on 'tests' (using the local stubs) and on the local stubs'

/pathto/pandas_stubs_test/pandas_stubs_v2/pandas-stubs/_libs/tslibs/timestamps.pyi
  /pathto/pandas_stubs_test/pandas_stubs_v2/pandas-stubs/_libs/tslibs/timestamps.pyi:103:9 - error: Method "fromtimestamp" overrides class "datetime" in an incompatible manner
    Parameter 2 name mismatch: base parameter is named "timestamp", override parameter is named "t" (reportIncompatibleMethodOverride)
1 error, 0 warnings, 0 informations 

Step: 'Run pyright on 'tests' (using the local stubs) and on the local stubs' failed!
```

 - OS: archlinux
 - OS Version `kernel version: 6.5.8`
 - python version `3.12.0`
 - version of type checker `pyright 1.1.334`
 - version of installed `pandas-stubs` `2.1.1.230928`

**Additional context**
I didn't modify anything in this project.

twoertwein:
- [ ] Closes #813 (Replace xxxx with the Github issue number)

`poe pytest` (circular import error) fails for me but `pytest tests` works?! Let's see what happens on the CI.


twoertwein:
Unrelated: could replace isort and black with ruff (pandas might replace black with ruff https://github.com/pandas-dev/pandas/pull/56554)

tests/__init__.py
```diff
@@ -7,23 +7,29 @@
...
 
-import pandas as pd
-from pandas.util.version import Version
 import pytest
 
-from pandas._typing import T
+if sys.version_info < (3, 12):
+    import pandas as pd
+else:
+    with pytest.warns(DeprecationWarning, match="datetime.datetime.utcfromtimestamp"):
```
Dr-Irv, line 22: I don't understand why this is needed. Why does pytest see the deprecation warning, but when we import pandas into a regular python interpreter, the warning isn't shown?

Dr-Irv (commented):
When I look into the Actions on github, it shows that the CI is failing.  

But I also don't understand why the jobs aren't showing up in the Conversation tab.

tests/__init__.py
twoertwein: pandas triggers a warning when importing on 3.12 - with `-Werror` - it errors (while pandas is not fully initialized).

Dr-Irv: But why does pytest pick up the warning, but not from the python interpreter?

If I run `pytest` with `-W error`, I get the full stack trace showing that it is an import of `dateutil` from `timestamps.pyx` .  I just don't understand why a regular import of pandas isn't showing the warning.  What is pytest doing differently?

twoertwein: It does, the warning is just not printed by default:

```sh
$ python -Wdefault -c "import pandas"
/var/home/twoertwein/.cache/pypoetry/virtualenvs/pandas-stubs-KrNUNgd3-py3.12/lib/python3.12/site-packages/dateutil/tz/tz.py:37: DeprecationWarning: datetime.datetime.utcfromtimestamp() is deprecated and scheduled for removal in a future version. Use timezone-aware objects to represent datetimes in UTC: datetime.datetime.fromtimestamp(timestamp, datetime.UTC).
  EPOCH = datetime.datetime.utcfromtimestamp(0)
```


I will squash and force push to hopefully trigger the CI.

twoertwein: just the random mac failure :)

Dr-Irv: Looks like they fixed it but haven't released:  https://github.com/dateutil/dateutil/pull/1285

twoertwein: Happy to wait until the next release then

Dr-Irv: > just the random mac failure :)

I reran by hand and it worked.

tests/test_frame.py
```diff
@@ -1821,24 +1822,30 @@ def test_frame_getitem_isin() -> None:
...
-        check(assert_type(pd.read_excel(path), pd.DataFrame), pd.DataFrame)
-    with ensure_clean() as path:
-        df.to_excel(path, engine="openpyxl", columns=["col1"])
-        check(assert_type(pd.read_excel(path), pd.DataFrame), pd.DataFrame)
+    with pytest_warns_bounded(
+        DeprecationWarning,
+        match="datetime.datetime.utcnow",
+        lower="3.11.99",
+        version_str=platform.python_version(),
+    ):
```
Dr-Irv, line 1830: Given that this pattern is appearing all over the place, and hopefully we can one day remove it, it might be better to just create a special item in `tests/__init__.py` that returns  
```python
 pytest_warns_bounded(
        DeprecationWarning,
        match="datetime.datetime.utcnow",
        lower="3.11.99",
        version_str=platform.python_version(),
    )
```
Then we know there is one place that is catching the same warning.

Dr-Irv (commented):
One suggestion to help with maintenance.  Convert out of draft if you're happy with it.

tests/__init__.py
Dr-Irv: > Happy to wait until the next release then

They haven't released for over 2 years (last release was July, 2021), so I don't think we should wait.  If they ever do release, we can then revert that part of the change.

tests/test_frame.py
twoertwein: I found a nicer solution :) `pytest_warns_bounded` is used in only one place now, so I don't think creating a new function for it would make sense anymore.

Dr-Irv (approved):
Thanks @twoertwein .  I like the idea of using `pyproject.toml` and `[tool.pytest.ini_options]` to handle the `dateutil` issue.

PR #840, opened 2023-12-23
pin tables to 3.9.1

Dr-Irv:
I am guessing that the MacOS failures in testing are due to the version of pytables.  Let's see if this fixes it.


Dr-Irv:
So this worked and there were no crashes.  But with tables 3.8.2, we sometimes get crashes on MacOS, and other times we don't, if I keep re-running the job, it might eventually succeed.

@twoertwein So should we pin this for now to (hopefully) avoid the CI failures that are intermittent?

twoertwein:
Sounds okay to me

twoertwein:
Thanks @Dr-Irv !

PR #842, opened 2023-12-24
itertuples

### LINKED ISSUE #834: With the most recent pandas-stubs, mypy now complains about every call to itertuples()
With the most recent pandas-stubs, mypy now complains about every call to itertuples()
Is there a way that I'm supposed to tell the new release what my column names are?

Sample 
```python
import pandas as pd                                                                                 
                                                                                                    
df = pd.DataFrame({"a": [1, 2, 3, 4], "b": [5, 6, 7, 8]})                                           
                                                                                                    
for item in df.itertuples():                                                                        
    a = item.a                                                                                      
    b = item.b     
```

```
- hook id: mypy
- exit code: 1

foo.py:6: error: "tuple[Any, ...]" has no attribute "a"  [attr-defined]
foo.py:7: error: "tuple[Any, ...]" has no attribute "b"  [attr-defined]
```

_Originally posted by @j-carson in https://github.com/pandas-dev/pandas-stubs/discussions/832_

twoertwein:
- [ ] Closes #834 (Replace xxxx with the Github issue number)


pandas-stubs/core/frame.pyi
```diff
@@ -250,7 +250,7 @@ class DataFrame(NDFrame, OpsMixin):
     def iterrows(self) -> Iterable[tuple[Hashable, Series]]: ...
     def itertuples(
         self, index: _bool = ..., name: _str | None = ...
-    ) -> Iterable[tuple[Any, ...]]: ...
+    ) -> Iterable[Any]: ...
```
Dr-Irv, line 253: I'm wondering if we could do something like this:
```python
class _PandasNamedTuple(tuple[Any,...]):
    def __getattr__(self, field: str) -> Any: ...
```
Then have `itertuples()` return `Iterable[_PandasNamedTuple]`

Then at least you know you're getting a tuple.

twoertwein: Nice solution!

tests/test_frame.py
```diff
@@ -20,6 +20,7 @@
     Any,
     Callable,
     Generic,
+    TypeAlias,
```
Dr-Irv, line 23: Needs to come from `typing_extensions`.  Failed under python 3.9

Dr-Irv (changes requested):

Dr-Irv (changes requested):
Can you look at `test_types_itertuples()` in `test_frame.py` and see if that should be modified?  Maybe introduce `check(assert_type` in there just to be sure.

pandas-stubs/core/frame.pyi
```diff
@@ -2279,3 +2279,6 @@ class DataFrame(NDFrame, OpsMixin):
     ) -> Self: ...
     def __truediv__(self, other: float | DataFrame | Series | Sequence) -> Self: ...
     def __rtruediv__(self, other: float | DataFrame | Series | Sequence) -> Self: ...
+
+class _PandasNamedTuple(tuple[Any, ...]):
+    def __getattr__(self, field: str) -> Any: ...
```
Dr-Irv, line 2284: I'm thinking we should return `Scalar` here, because we also return `Scalar` when someone does `df.loc[3, "a"]`

While it's true that some non-scalar value could be an individual element of a `DataFrame`, I've taken the philosophy of limiting the types to what is "normal" usage, and if you put a funky type in a `DataFrame` or `Series`, then you can do a `cast` to fix it.   I've done that in some of our application code when we have lists or other objects inside a `Series` or `DataFrame`.

Dr-Irv (approved):
thanks @twoertwein

PR #843, opened 2023-12-24
Series(Mapping)

### LINKED ISSUE #831: dict argument to `pd.Series` constructor should allow `Timestamp` and `tuple` as keys
Additional migration issues for the `pd.Series` constructor involve passing a dictionary.  It fails with a Timestamp or multi-index key whereas before it did not:

```
sr1 = pd.Series(
        {
            pd.Timestamp(2023, 1, 2): "b",
        }
    )

    sr2 = pd.Series(
        {
            ("a", "b"): "c",
        }
    )
```

```
test.py:172: error: Need type annotation for "sr1"  [var-annotated]
test.py:174: error: Dict entry 0 has incompatible type "Timestamp": "str"; expected "str": "ndarray[Any, Any]"  [dict-item]
test.py:178: error: Need type annotation for "sr2"  [var-annotated]
test.py:180: error: Dict entry 0 has incompatible type "Tuple[str, str]": "str"; expected "str": "ndarray[Any, Any]"  [dict-item]
```

_Originally posted by @amgcc in https://github.com/pandas-dev/pandas-stubs/issues/826#issuecomment-1858252851_

Might be best to allow `Hashable` as the `dict` key.

twoertwein:
- [ ] Closes #831 (Replace xxxx with the Github issue number)


pandas-stubs/core/series.pyi
```diff
@@ -294,7 +297,7 @@ class Series(IndexOpsMixin[S1], NDFrame):
     @overload
     def __new__(
         cls,
-        data: S1 | _ListLike[S1] | dict[int, S1] | dict[_str, S1],
+        data: S1 | Mapping[Any, S1] | _ListLike[S1],
```
twoertwein, line 300: I'm not happy with `_ListLike` but a `Mapping` is also `Iterable`.

pandas-stubs/core/series.pyi
```diff
@@ -239,7 +240,7 @@ class Series(IndexOpsMixin[S1], NDFrame):
     @overload
     def __new__(  # type: ignore[overload-overlap]
         cls,
-        data: _ListLike,
+        data: Iterable,
```
twoertwein, line 243: Wouldn't mind widening `data` to `Any` when `dtype` is specified (there are combinations that will error but that is already the case)

pandas-stubs/core/series.pyi
```diff
@@ -228,6 +228,7 @@ class Series(IndexOpsMixin[S1], NDFrame):
         cls,
         data: DatetimeIndex
         | Sequence[np.datetime64 | datetime]
+        | Mapping[Any, np.datetime64 | datetime]
```
Dr-Irv, line 231: Shouldn't this be `Hashable` instead of `Any` ?

pandas-stubs/core/series.pyi
Dr-Irv: Wouldn't we be able to say the kinds of iterables that are allowed?  E.g., if the specified dtype is `TimestampDtypeArg`, then we know only certain kinds of iterables are allowed.

Dr-Irv (changes requested):
I think we should consider using `Mapping[Hashable, _appropriate_type_here]` instead of `Mapping[Any, _approprirate_type_here]`, since we know that the keys of the dict will end up as column names.

pandas-stubs/core/series.pyi
twoertwein: I would prefer not doing that for now: there are so many coerciable types, that I assume it might create false negatives.

Dr-Irv: OK, that makes sense.

pandas-stubs/core/series.pyi
```diff
@@ -284,7 +287,7 @@ class Series(IndexOpsMixin[S1], NDFrame):
     @overload
     def __new__(
         cls,
-        data: Scalar | _ListLike | dict[int, Any] | dict[_str, Any] | None,
+        data: Scalar | Iterable | None,
```
Dr-Irv, line 290: I don't think `Iterable` is correct here, because we don't (or shouldn't) accept a `set`.

pandas-stubs/core/series.pyi
```diff
@@ -304,7 +307,7 @@ class Series(IndexOpsMixin[S1], NDFrame):
     @overload
     def __new__(
         cls,
-        data: Scalar | _ListLike | dict[int, Any] | dict[_str, Any] | None = ...,
+        data: Scalar | Iterable | None = ...,
```
Dr-Irv, line 310: same comment as above about `Iterable`

Dr-Irv (changes requested):

pandas-stubs/core/series.pyi
twoertwein: Would it be correct to have one Never-overload for `set` and then simply use `data: Any` + specified `dtype`?

Dr-Irv: > Would it be correct to have one Never-overload for `set` and then simply use `data: Any` + specified `dtype`?

I don't know.  I'd rather keep things narrow by just using `Scalar | _ListLike | dict[Hashable, Any] | None` .  Or maybe we can use `Mapping[Hashable, Scalar]`  instead of `dict[Hashable, Any]` .

My feeling is that we should make small increments when we change the stubs so as to not widen things too much.

twoertwein: > My feeling is that we should make small increments when we change the stubs so as to not widen things too much.

[sorry, I edited your comment by accident - wanted to quote reply]

I agree about that! But I also feel like we are making ourselves too much work by trying to be as restrictive as possible.

Dr-Irv: > I agree about that! But I also feel like we are making ourselves too much work by trying to be as restrictive as possible.

Possibly.  But I'd rather let the community report issues, and we really haven't had too many issues where we made things too restrictive.  

See what I wrote here 18 months ago about this topic:  https://github.com/pandas-dev/pandas-stubs/blob/main/docs/philosophy.md#narrow-vs-wide-arguments

Dr-Irv (approved):
Thanks @twoertwein

Issue #820, opened 2023-11-25
ENH: Forward operators to underlying types


twoertwein:
**Describe the bug**
Series is a container that forwards operators (+, -, ...) to the underlying types.

Some of the operators of Series could be better, for example, `__add__` which:

- returns `Series[Unknown]` instead of `Series[return type]`
- or uses non-exiting helper classes to restrict the types

Instead of returning too wide types, enumerating all typing combinations, or creating non-existing helper classes, it might(?) be possible to abuse Protocols to forward operators to the underlying types.

**To Reproduce**
The following code works with both mypy and pyright

```py
from __future__ import annotations
from typing import Protocol, TypeVar, overload, Generic

_T = TypeVar("_T", covariant=True)
_Other = TypeVar("_Other", contravariant=True)
_Result = TypeVar("_Result", covariant=True)

class add(Protocol[_Other, _Result]):
    def __add__(self, other: _Other, /) -> _Result:
        ...

class radd(Protocol[_Other, _Result]):
    def __radd__(self, other: _Other, /) -> _Result:
        ...

class Series(Generic[_T]):  # simulate Series with one element
    def __init__(self, x: _T) -> None:
        self.x = x

    # Series + Series
    @overload
    def __add__(
        self: Series[add[_Other, _Result]], other: Series[_Other], /
    ) -> Series[_Result]:
        ...

    @overload
    def __add__(self, other: Series[radd[_T, _Result]], /) -> Series[_Result]:
        ...

    # Series + scalar
    @overload
    def __add__(
        self: Series[add[_Other, _Result]], other: _Other, /
    ) -> Series[_Result]:
        ...

    @overload
    def __add__(self, other: radd[_T, _Result], /) -> Series[_Result]:
        ...

    def __add__(self, other):
        if isinstance(other, Series):
            return Series(self.x + other.x)
        return Series(self.x + other)

def test_int(x: Series[int], y: int):
    reveal_type(x + y)  # Series[int] (pandas-stubs: Series[int])

def test_float(x: Series[int], y: float):
    reveal_type(x + y)  # Series[float] (pandas-stubs: Series[Unknown])

def test_series_int(x: Series[int], y: Series[int]):
    reveal_type(x + y)  # Series[int] (pandas-stubs: Series[int])

def test_series_float(x: Series[int], y: Series[float]):
    reveal_type(x + y)  # Series[float] (pandas-stubs: Series[Unknown])

def test_error(x: Series[int], y: str):
[...]

PR #844, opened 2023-12-24
OffsetSeries inherits from Series[BaseOffset]

twoertwein:
I honestly would prefer to remove `OffsetSeries` and replace it with `Series[Any]` (personally, I'm not keen on maintaining operators for Series as there is no generic solution #820). If we want to keep this class, it would make sense to let it inherit from `Series[BaseOffset]` instead of `Series[Any/Unknown]`

### DIFF
--- a/pandas-stubs/_typing.pyi
+++ b/pandas-stubs/_typing.pyi
@@ -32,6 +32,7 @@ from typing_extensions import TypeAlias
 
 from pandas._libs.interval import Interval
 from pandas._libs.tslibs import (
+    BaseOffset,
     Period,
     Timedelta,
     Timestamp,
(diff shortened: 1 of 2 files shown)


Dr-Irv (approved):
thanks @twoertwein

Dr-Irv:
> I honestly would prefer to remove `OffsetSeries` and replace it with `Series[Any]` (personally, I'm not keen on maintaining operators for Series as there is no generic solution #820). If we want to keep this class, it would make sense to let it inherit from `Series[BaseOffset]` instead of `Series[Any/Unknown]`

that makes sense.  Good improvement.

twoertwein:
This could potentially hit people who rely on methods provided only by sub-types of BaseOffset.

PR #845, opened 2023-12-26
Upgrade pyright to 1.1.343

paw-lu:

Pin pyright to 1.1.343. Unblocks #823 by including change introduced in:

- https://github.com/microsoft/pyright/pull/6797


Dr-Irv (approved):
thanks @paw-lu

PR #848, opened 2024-01-06
Rework groupby and resample core modules

### LINKED ISSUE #737: `DataFrame.Groupby.Rolling` is missing proper typing
**Describe the bug**

In this line:

https://github.com/pandas-dev/pandas-stubs/blob/6fe90bbb37128a79068ba41c8ebaaf269ddeab90/pandas-stubs/core/groupby/groupby.pyi#L64

we are not returning a `Rolling` object with the proper subtype of `Series` or `DataFrame`

**To Reproduce**
```python
from typing_extensions import reveal_type
import pandas as pd

df = pd.DataFrame(
    {"A": [1, 1, 2, 2], "B": [1, 2, 3, 4], "C": [0.362, 0.227, 1.267, -0.562]}
)

d1 = df.groupby("A").rolling(2).sum()

s1 = df.groupby("A")["B"].rolling(2).sum()

s2 = df.groupby("A").rolling(2)["B"].sum()

reveal_type(d1)

reveal_type(s1)

reveal_type(s2)
```
at runtime produces
```text
Runtime type is 'DataFrame'
Runtime type is 'Series'
Runtime type is 'Series'
```
but pyright produces
```text
  grbroll.py:14:13 - information: Type of "d1" is "Unknown"
  grbroll.py:16:13 - information: Type of "s1" is "Unknown"
  grbroll.py:18:13 - information: Type of "s2" is "Unknown"
```

 - OS:  Windows 10
 - python version:  3.9
 - version of type checker:  pyright 1.1.313
 - version of installed `pandas-stubs`  :  2.0.2.230605

**Additional context**

This will need some care to make sure that the types are done correctly.  May have to move the `def rolling` from `pandas-stubs/core/groupby/groupby.pyi` to `pandas-stubs/core/groupby/generic.pyi` in `DataFrameGroupby` and `SeriesGroupBy`.

### LINKED ISSUE #810: `core.groupby.generic.DataFrameGroupBy` missing keyword-only arguments `engine` and `engine_kwargs`
**Describe the bug**
`core.groupby.generic.DataFrameGroupBy` missing keyword-only arguments

**To Reproduce**

Current:
```py
DataFrameGroupBy().aggregate("size", "some", "args", engine=0, engine_kwargs="not valid", other_kwarg="")
# no error despite `engine` as `engine_kwargs` not being valid types for those kw-only args
```

Expected:
`mypy`
```
error: No overload variant of "aggregate" of "DataFrameGroupBy" matches argument types "str", "str", "str", "int", "int", "str"  [call-overload]
note:     def aggregate(self, arg: Literal['size']) -> Series[Any]
note:     def aggregate(self, arg: Union[Union[Callable[..., Any], str, ufunc], list[Union[Callable[..., Any], str, ufunc]], Mapping[Any, Union[Union[Callable[..., Any], str, ufunc], list[Union[Callable[..., 
Any], str, ufunc]]]]] = ..., *args: Any, engine: Optional[str] = ..., engine_kwargs: Optional[dict[str, bool]] = ..., **kwargs: Any) -> DataFrame
```
`pyright`
```
error: Argument of type "Literal[0]" cannot be assigned to parameter "engine" of type "str | None" in function "aggregate"
    Type "Literal[0]" cannot be assigned to type "str | None"
      "Literal[0]" is incompatible with "str"
      Type cannot be assigned to type "None" (reportGeneralTypeIssues)
error: Argument of type "Literal[0]" cannot be assigned to parameter "engine_kwargs" of type "dict[str, bool] | None" in function "aggregate"
    Type "Literal[0]" cannot be assigned to type "dict[str, bool] | None"
      "Literal[0]" is incompatible with "dict[str, bool]"
      Type cannot be assigned to type "None" (reportGeneralTypeIssues)
```

 - OS: Windows
 - OS Version	10.0.19045 Build 19045
 - python version: 3.9.13
 - version of type checker:
   - mypy 1.6.1 (compiled: yes)
   - pyright@1.1.334
 - version of installed `pandas-stubs`: 2.1.1.230928

**Additional context**
Noticed while drilling down seaborn types to complete some method signatures in typeshed

### LINKED ISSUE #825: Return type from `pandas.core.groupby.generic.DataFrameGroupBy.resample`
**Describe the bug**
The type hint for `pandas.core.groupby.generic.DataFrameGroupBy.resample` says it returns a `pandas.core.groupby.grouper.Grouper`; however in at least some situations this function can return a `pandas.core.resample.DatetimeIndexResamplerGroupby`, which is not a Grouper.

**To Reproduce**
This results in a false positive on MyPy 1.7.1:
```
df.set_index("date").groupby("intensity").resample("90D")["duration"]
```

The error message is
```
error: Value of type "Grouper" is not indexable  [index]
```

Python 3.11
Mypy 1.7.1
pandas-stubs 2.1.1.230928

hamdanal:
- [x] Closes #825, closes #810, closes #737, improves the situation for #456

Firstly, my apologies for the lengthy PR.
I intended to fix `df.groupby().resample()` but it was very hard to do so given how much was missing in that area. I then went down the rabbit hole of running stubtest and fixing all allow list entries related to groupby and its methods. Hopefully the extensive testing I added and the happy stubtest will make it easier to review.


hamdanal:
Just a note on stubtest:

Given that stubtest does not run as part of the CI, it generated 4073 errors on the main branch. With the changes in this PR, I got the number down to 3396 errors, that almost 700 less errors. With some tweaking organization, I got the number of allow list entries down to a manageable 840 entries I would happily upload this allow list file and add stubtest to CI if you are willing. It will help the project stay up-to-date with upstream pandas and will prevent regressions in already fixed code.

pandas-stubs/_typing.pyi
```diff
@@ -48,6 +48,8 @@ from pandas.core.dtypes.dtypes import (
 
 from pandas.io.formats.format import EngFormatter
 
+Incomplete: TypeAlias = Any
```
twoertwein, line 51: I assume that is inspired by typeshed? We haven't used that yet. Open to using it (is `_typeshed.Incomplete` possible?)

pandas-stubs/core/apply.pyi
```diff
@@ -0,0 +1,242 @@
...
+)
+from pandas.util._decorators import cache_readonly
+
+_AggObjT = TypeVar("_AggObjT", bound=NDFrame | GroupBy | BaseWindow | Resampler)
+_AggGroupByObjT = TypeVar("_AggGroupByObjT", bound=GroupBy | BaseWindow | Resampler)
+_AggResamplerWindowObjT = TypeVar(
+    "_AggResamplerWindowObjT", bound=BaseWindow | Resampler
+)
+
+ResType = dict[int, Any]  # noqa: PYI026
```
twoertwein, line 46: Is that part of the public API? If not, it can/should be removed.

twoertwein: Same also for all methods/attributes for all the classes.

If you cannot find a method/class here https://pandas.pydata.org/docs/reference/index.html, it is considered private. We still keep some private parent classes that are used by multiple public child classes.

It is already confusing what is private in pandas (most people expect that only _* are private), we do not want to add more confusion by highlighting private API in pandas-stubs. (That creates some inherent conflict with stubtest)

pandas-stubs/core/generic.pyi
```diff
@@ -432,6 +439,21 @@ class NDFrame(indexing.IndexingMixin):
         end_time,
         axis=...,
     ) -> Self: ...
+    @final
+    def resample(
+        self,
+        rule: Frequency,
+        axis: Axis | NoDefault = ...,
+        closed: Literal["right", "left"] | None = None,
```
twoertwein, line 447: So far, we have tried to not have the default values in the stubs. (I believe pyright/mypy(?) have an option to create stubs with/without default values as people requested that having default values in stubs can be helpful.)

pandas-stubs/core/groupby/generic.pyi
```diff
@@ -1,150 +1,191 @@
...
+    TakeIndexer,
+    WindowingEngine,
+    WindowingEngineKwargs,
 )
 
+from pandas.plotting import boxplot_frame_groupby
+
 AggScalar: TypeAlias = str | Callable[..., Any]
-ScalarResult = ...
+ScalarResult = TypeVar("ScalarResult")  # noqa: PYI001
```
twoertwein, line 55: Should probably be removed and the usage by replaced by `pandas._typing.T`

pandas-stubs/core/groupby/generic.pyi
```diff
@@ -1,150 +1,191 @@
...
+        engine: WindowingEngine = None,
+        engine_kwargs: WindowingEngineKwargs = None,
+        **kwargs,
+    ) -> Series: ...
+    def filter(
+        self, func: Callable | str, dropna: bool = True, *args, **kwargs
+    ) -> Series: ...
+    def nunique(self, dropna: bool = True) -> Series[int]: ...
+    # describe delegates to super() method but here it has keyword-only parameters
+    def describe(  # type: ignore[override] # pyright: ignore[reportIncompatibleMethodOverride]
```
twoertwein, line 94: Feel free to open  PRs at pandas for those inconsistencies (preferably smaller PRs: typically easier to review/merge).

pandas-stubs/core/groupby/grouper.pyi
```diff
@@ -7,67 +14,115 @@ from pandas import (
...
+        self,
+        key: KeysArgType | None = None,
+        level: Level | ListLikeHashable[Level] | None = None,
+        freq: Frequency | None = None,
+        axis: Axis | NoDefault = ...,
+        sort: bool = False,
+        dropna: bool = True,
     ) -> None: ...
     @property
+    @deprecated("Grouper.ax is deprecated. Use Resampler.ax instead.")
```
twoertwein, line 66: To make maintaining pandas-stubs easier, we opted to remove methods/arguments that trigger a warning on the current pandas. Unless more people work on pandas-stubs, I would be inclined to keep it simpler by removing these deprecated methods.

(I assume stubtests doesn't like this simplification.)

twoertwein (commented):
Thank you very much for this giant PR!

Please wait for feedback from @Dr-Irv before making further changes.

pandas-stubs/_typing.pyi
hamdanal: Yes and yes.

> I assume that is inspired by typeshed?

At typeshed, `Incomplete` is used as a placeholder for missing type annotations. Sometimes it is needed, for example to annotate module-level or class-level variables, and sometimes it is used for documentation purposes. It also helps find where missing annotations are as you can literally grep for "Incomplete" to find them.
Although this is equivalent to `Any` from a type checker's perspective, they signal different intention from the project's point of view.

> is _typeshed.Incomplete possible?

It can be used if you prefer. I defined it in _typing because I noticed you define all your reusable typing symbols there.

pandas-stubs/core/apply.pyi
hamdanal: Yeah I tried to fix everything stubtest complained about. We can add these symbols to the allowlist. I'll wait for the review from Dr-Irv and do a pass to handle those after if needed.

pandas-stubs/core/generic.pyi
hamdanal: The same thing was done at typeshed until a year ago where the [view has shifted towards keeping the default values](https://github.com/python/typeshed/issues/8988). Since then, types checkers and linters like flake8-pyi have been updated to not flag simple default values in stubs and stubtest was updated to include a check for default values so they do not go out of sync. I also recently [modified stubgen](https://github.com/python/mypy/commit/e69c5cde8643e04a54a644cc27814ab98181541d) to preserve default values in generated stubs unconditionally; this change was included in the latest release of mypy.

Default values, although not very useful for command line type checkers, proved useful for other applications like IDEs that use the stubs like VSCode and Pycharm and thus are now encouraged.

pandas-stubs/core/groupby/grouper.pyi
hamdanal: > (I assume stubtests doesn't like this simplification.)

Yes but stubtest can be silenced about a missing symbol once we established the symbol is not to be included in the stubs. There will always be an allowlist of things we intentionally want to deviate from runtime, stubtest can still help with unintentional deviations though.

pandas-stubs/core/groupby/ops.pyi
```diff
@@ -8,91 +17,141 @@ from pandas import (
...
-    def slabels(self): ...
-    def __iter__(self): ...
-
-class SeriesSplitter(DataSplitter): ...
-class FrameSplitter(DataSplitter): ...
+    def __iter__(self) -> Iterator[NDFrameT]: ...
 
-def get_splitter(data: DataFrame | Series, *args, **kwargs) -> DataSplitter: ...
+class SeriesSplitter(DataSplitter[Series]): ...
+class FrameSplitter(DataSplitter[DataFrame]): ...
```
Dr-Irv, line 157: delete - not documented

pandas-stubs/core/resample.pyi
```diff
@@ -51,138 +70,238 @@ _SeriesGroupByFuncArgs: TypeAlias = (
     _SeriesGroupByFuncTypes | Mapping[Hashable, _SeriesGroupByFunc | str]
 )
 
-class Resampler(BaseGroupBy, Generic[NDFrameT]):
+class Resampler(BaseGroupBy[NDFrameT]):
+    grouper: BinGrouper  # pyright: ignore[reportIncompatibleVariableOverride]  # variance incompatibility
+    binner: DatetimeIndex | TimedeltaIndex | PeriodIndex
+    exclusions: frozenset[Hashable]
+    ax: Index
+    def __init__(
```
Dr-Irv, line 78: Class can't be instantiated, so delete `__init__()`

Dr-Irv (changes requested):
Thanks for this big contribution.  It's a hard one to review, so I may not have caught everything.

Some meta-level-notes:
1. The initial version of these stubs was produced by using `stubgen` (before I was involved in the project), probably using pandas 1.1.  We've been slowly removing things that are not documented.  So if you use `stubgen` to generate things from the pandas source, we end up having to do those removals again.
2. Using the pandas source to generate stubs has risks because the types within the pandas source code are meant to make sure that the types are consistent within pandas for pandas developers, but not necessarily are they appropriate for users. So we use the stubs as a way of providing type checking for the most common usage patterns of pandas.  That also means we try to only include classes, methods and functions  in the stubs that are documented, as well as parameters that are documented.  One advantage of this is that it has helped us identify incorrect documentation in pandas and to have discussions on whether we want to accept certain types in various methods.  So the rule of thumb is that the stubs are provided for what is documented, and we tend to be as narrow as we can to help users in the best way possible.
3. If we need some helper classes, we make them private with a preceding underscore.  We're probably not consistent, but that's an overall goal.  One thing I noticed is that there are classes that can be returned by pandas (and are documented as such), but users should never instantiate - so we should not include `__init__()` in the stubs for those classes.
4. You included `NoDefault` in many places -t that is used in the pandas source to tell users there is no default value, or that if you don't supply a value, there is an assumed default within pandas.  So users should never pass `NoDefault`, so we should not include it in the stubs.
5. I think the setting of default values of the parameters should be moved to a separate PR so as to minimize the changes coming from this PR.
6. I don't think we should be using `@final` or any of the pandas decorators in the stubs.  The latter is not documented.  I don't see the value for including `@final` in the stubs that are meant for user code (but I'm open to an alternative point of view).

One other request:
7. Can we add a test for the issue reported in #810  (testing the `engine` keyword)?  If you just do it for the specific example provided there, that's fine - no need to create a test for all methods that accept `engine` .

pandas-stubs/_typing.pyi
twoertwein: I'm fine either way: I believe typeshed is already included with type checkers, so it wouldn't be a dependency

pandas-stubs/core/base.pyi
```diff
@@ -19,13 +24,21 @@ from pandas._typing import (
...
 
 class NoNewAttributesMixin:
-    def __setattr__(self, key, value) -> None: ...
+    def __setattr__(self, key: str, value: Any) -> None: ...
 
 class SelectionMixin(Generic[NDFrameT]):
+    obj: NDFrameT
+    exclusions: frozenset[Hashable]
+    @final
+    @cache_readonly
```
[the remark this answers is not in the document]
twoertwein: I would be inclined to keep final and cache_readonly:

- `final`: could help downstream developers who extend pandas classes to know not to overwrite a method
- `cache_readonly`: we have to use at least `property` to indicate the correct type

pandas-stubs/core/groupby/grouper.pyi
```diff
@@ -7,67 +14,115 @@ from pandas import (
...
+        cls,
+        key: KeysArgType | None = None,
+        level: Level | ListLikeHashable[Level] | None = None,
+        axis: Axis | NoDefault = ...,
+        sort: bool = False,
+        dropna: bool = True,
+    ) -> Self: ...
+    @overload
+    def __new__(cls, *args, freq: Frequency, **kwargs) -> TimeGrouper: ...
     def __init__(
```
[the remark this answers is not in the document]
hamdanal: I suggest keeping `__init__` for now. `__new__` that returns different types from `__new__` than the class itself is known to be a source of compatibility issues between type checkers. Until this behavior is standardized, we can keep `__init__` because all type checkers are happy with it.

pandas-stubs/plotting/_core.pyi
```diff
@@ -83,6 +84,54 @@ def boxplot(
     return_type: Literal["both"],
     **kwargs,
 ) -> _BoxPlotT: ...
+@overload
+def boxplot_frame_groupby(
```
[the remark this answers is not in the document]
hamdanal: This function is public. It is re-exported (included in `__all__`) in the `pandas.plotting` module which is documented as _"Plotting public API"_. The function is also used as alias for a public method on the dataframe groupby class.

pandas-stubs/core/groupby/ops.pyi
```diff
@@ -8,91 +17,141 @@ from pandas import (
...
+        obj: Series,
+        func: Callable[[Series], object],
+        preserve_dtype: bool = False,
+    ) -> ArrayLike: ...
+    @final
+    def apply_groupwise(
+        self, f: Callable[[NDFrameT], T], data: NDFrameT, axis: AxisInt = 0
+    ) -> tuple[list[T], bool]: ...
 
 class BinGrouper(BaseGrouper):
```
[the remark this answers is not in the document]
hamdanal: The class is used as a type elsewhere. I removed it `__init__` instead.

pandas-stubs/core/window/ewm.pyi
```diff
@@ -1,62 +1,113 @@
...
+        pairwise: bool | None = None,
+        numeric_only: bool = False,
+    ) -> NDFrameT: ...
+
+class ExponentialMovingWindowGroupby(
+    BaseWindowGroupby[NDFrameT], ExponentialMovingWindow[NDFrameT]
+):
+    def __init__(self, obj, *args, _grouper=None, **kwargs) -> None: ...
+
+class OnlineExponentialMovingWindow(ExponentialMovingWindow[NDFrameT]):
```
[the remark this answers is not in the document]
hamdanal: This is used as a return type for a public method. I removed its `__init__`

pandas-stubs/core/groupby/grouper.pyi
```diff
@@ -7,67 +14,115 @@ from pandas import (
...
+    @deprecated("Grouper.grouper is deprecated. Use GroupBy.grouper instead.")
+    def grouper(self): ...
+    @property
+    @deprecated("Grouper.groups is deprecated. Use GroupBy.groups instead.")
     def groups(self): ...
+    @final
+    def __repr__(self) -> str: ...  # noqa: PYI029 __repr__ here is final
 
+@final
 class Grouping:
```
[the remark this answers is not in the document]
hamdanal: This class is used in a public method. I removed its `__init__` instead.

pandas-stubs/core/groupby/groupby.pyi
```diff
@@ -1,115 +1,436 @@
...
+        alpha: float | None = None,
+        min_periods: int | None = 0,
+        adjust: bool = True,
+        ignore_na: bool = False,
+        axis: Axis = 0,
+        times: str | np.ndarray | Series | np.timedelta64 | None = None,
+        method: CalculationMethod = "single",
+        *,
+        selection: IndexLabel | None = None,
+    ) -> ExponentialMovingWindowGroupby[NDFrameT]: ...
```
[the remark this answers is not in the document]
hamdanal: These are definitely documented somewhere. I remember using them before. I'll try to dig up where they are documented.

pandas-stubs/core/base.pyi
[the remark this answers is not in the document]
hamdanal: What @twoertwein said.
These are not added only to make stubtest happy.

`final` is actually a more strict interface. It tells the users that they are not supposed to override a function or a class otherwise they are on their own if something breaks.

`cache_readonly` is currently an alias to `property` so removing it has the same effect as removing a `property` decorator. I think using `cache_readonly` has two more advantages over using `property` apart from consistency with runtime. 1) its name is descriptive; it cahces the result and it is not writable, and 2) if/when type checkers start to recognize custom property definitions, this can be changed in one place instead of having to go through all the code base to change `property` to `cache_readonly` (both are actually used by pandas).

pandas-stubs/core/groupby/indexing.pyi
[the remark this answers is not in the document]
hamdanal: Agree. But they are used in public methods. Removed their `__init__` instead.

pandas-stubs/core/groupby/generic.pyi
```diff
@@ -1,150 +1,191 @@
...
-        ylabelsize: int | None = ...,
-        yrot: float | None = ...,
-        figsize: tuple[float, float] | None = ...,
-        bins: int | Sequence = ...,
-        backend: str | None = ...,
-        legend: bool = ...,
+        by: IndexLabel | None = None,
+        ax: PlotAxes | None = None,
+        grid: bool = True,
+        xlabelsize: float | None = None,
```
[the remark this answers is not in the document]
hamdanal: Oh , the upstream docs of this, and other label size usages, will need to be updated to accept `float | str | None` (See the [matplotlib docs](https://matplotlib.org/stable/api/_as_gen/matplotlib.axes.Axes.tick_params.html)). I typed them as `int | None` here for now because they need to be fixed everywhere in pandas-stubs, not just this file. We can do that in a separate PR.

pandas-stubs/core/groupby/groupby.pyi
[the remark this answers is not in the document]
hamdanal: Found the doc page: https://pandas.pydata.org/docs/user_guide/window.html#overview. You can see in the table that these methods support chained groupby operations. There is even an example in this page:
```python
In [6]: df = pd.DataFrame({'A': ['a', 'b', 'a', 'b', 'a'], 'B': range(5)})

In [7]: df.groupby('A').expanding().sum()
Out[7]: 
       B
A       
a 0  0.0
  2  2.0
  4  6.0
b 1  1.0
  3  4.0
```

pandas-stubs/core/generic.pyi
```diff
@@ -432,6 +439,21 @@ class NDFrame(indexing.IndexingMixin):
...
+        label: Literal["right", "left"] | None = None,
+        convention: TimestampConvention = "start",
+        kind: Literal["period", "timestamp", "timedelta"] | None = None,
+        on: Level | None = None,
+        level: Level | None = None,
+        origin: TimeGrouperOrigin | TimestampConvertibleTypes = "start_day",
```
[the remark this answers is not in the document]
hamdanal: Mmmm. I am not sure about this one. In other places in the code where this function calls, it is deliberately treated as "`Timestamp` convertible". There is also an explicit
```
try:
    self.origin = Timestamp(origin)
except (ValueError, TypeError) as err:
    raise
```
which raises on invalid inputs with the following message:
`ValueError: 'origin' should be equal to 'epoch', 'start', 'start_day', 'end', 'end_day' or should be a Timestamp convertible type.`
Notice it says "or should be a Timestamp convertible" which includes other strings like "2000-01-01". I think the docs need to be changed here to say "_..._ If string, must be _Timestamp convertible or_ one of the following: _..._"

pandas-stubs/core/groupby/groupby.pyi
```diff
@@ -1,115 +1,436 @@
...
+        origin: TimeGrouperOrigin | TimestampConvertibleTypes = "start_day",
+        offset: TimedeltaConvertibleTypes | None = None,
+        group_keys: bool = False,
+        **kwargs,
+    ) -> _ResamplerGroupBy[NDFrameT]: ...
+    @final
+    def rolling(
+        self,
+        # Arguments must be kept roughly inline with pandas.core.window.RollingGroupby
+        window: int | dt.timedelta | str | BaseOffset | BaseIndexer | None = None,
```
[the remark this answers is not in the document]
hamdanal: I am afraid this will not work. There is an isinstance check at runtime with `dt.timedelta` (which also includes pandas Timedelta) that would fail if passed `np.timedelta64` for example (I tested and it does fail).

pandas-stubs/core/groupby/generic.pyi
```diff
@@ -341,5 +270,82 @@ class DataFrameGroupBy(GroupBy, Generic[ByT]):
...
+        ax: PlotAxes | None = None,
+        sharex: bool = False,
+        sharey: bool = False,
+        figsize: tuple[float, float] | None = None,
+        layout: tuple[int, int] | None = None,
+        bins: int | Sequence[int] = 10,
+        backend: str | None = None,
+        legend: bool = False,
+        **kwargs,
+    ) -> Series: ...  # Series[Axes] but this is not allowed
```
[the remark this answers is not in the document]
hamdanal: Could you clarify what you mean here please? What is `AxesSubplot`? This method returns a series of matplotlib `Axes` as evidenced by the tests I added in `test_plotting.py`.

pandas-stubs/core/generic.pyi
```diff
@@ -432,6 +439,21 @@ class NDFrame(indexing.IndexingMixin):
         end_time,
         axis=...,
     ) -> Self: ...
+    @final
```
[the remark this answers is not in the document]
hamdanal: See the answer above

hamdanal:
I think I responded to / made the required changes for all the comments above except for this one about `NoDefault`

> I don't think we should be adding NoDefault as a possible argument in the stubs, because a user really shouldn't pass that value. It's just used internally by the pandas code to say that if a value wasn't passed, we assume a specific default.

I am afraid that we have to disagree here. I believe that the benefit of running stubtest on public methods that use `NoDefault` is far more important than hiding the symbol from the user. This is demonstrated in a couple of places in this PR where stubtest helped us detect wrong function signature (example there is no `squeeze` parameter to `DataFrame.groupy`). Unfortunately, because the default value at runtime is an object of type `NoDefault`, stubtest will complain that the type of the argument does not cover the default value if this type is to be omitted and we will have to add the whole method to the allowlist. 
Additionally, the symbol `NoDefault` is imported from an underscored module, this is enough indication to the user that it is private. It also has a descriptive name of what it does -- it is not very inviting for users to use it. I believe if you compare the pros and cons, the symbol should be used where needed.

I could've imported it as `_NoDefault` (or any underscored name you like) and used this "more private" name instead but I didn't because I saw it used without the underscore in other places already. If you prefer the underscored name, maybe we can change all its uses in a separate PR?

pandas-stubs/core/frame.pyi
```diff
@@ -1006,104 +1006,96 @@ class DataFrame(NDFrame, OpsMixin):
     def groupby(
         self,
         by: Scalar,
-        axis: AxisIndex = ...,
+        axis: AxisIndex | NoDefault = ...,
```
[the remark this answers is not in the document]
twoertwein: no_default/NoDefault seems to be public'ish: https://github.com/pandas-dev/pandas/blob/main/doc/source/reference/extensions.rst

Dr-Irv:
> I could've imported it as `_NoDefault` (or any underscored name you like) and used this "more private" name instead but I didn't because I saw it used without the underscore in other places already. If you prefer the underscored name, maybe we can change all its uses in a separate PR?

I did a search, and the only place it is used is in the parameter `dtype_backend` in various I/O routines.

I see your point about the value of having `stubtest` pick up things that are incorrect in the stubs vs. the source.  So let's fix in another PR.  Created an issue https://github.com/pandas-dev/pandas-stubs/issues/851   for this.

pandas-stubs/core/groupby/grouper.pyi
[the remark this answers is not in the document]
Dr-Irv: We use that "feature" of different `__new__()` methods to allow different types to be returned in many places in `pandas-stubs`.  So since you know that the argument `freq` being present changes what type of object is returned, we should take advantage of that.  I'd prefer just having `__new__()` in this case.

pandas-stubs/plotting/_core.pyi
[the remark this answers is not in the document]
Dr-Irv: Our philosophy is to not publish things that are not documented.  While it is used as an alias for the public method, the `groupby` stub has `boxplot` defined, so we should keep it defined there.

pandas-stubs/core/base.pyi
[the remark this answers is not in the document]
Dr-Irv: OK, thanks for the explanation

pandas-stubs/core/groupby/generic.pyi
```diff
@@ -77,74 +113,78 @@ class SeriesGroupBy(GroupBy, Generic[S1, ByT]):
...
-    def idxmin(self, axis: Axis = ..., skipna: bool = ...) -> Series: ...
-    def __iter__(self) -> Iterator[tuple[ByT, Series[S1]]]: ...
-    def diff(self, periods: int = ..., axis: Axis = ...) -> Series: ...
+    ) -> Series: ...  # Series[Axes] but this is not allowed
+    @property
+    def dtype(self) -> Series: ...
+    def unique(self) -> Series: ...
+    # Overrides that provide more precise return types over the GroupBy class
+    @final  # type: ignore[misc]
+    def __iter__(self) -> Iterator[tuple[ByT, Series[S1]]]: ...  # pyright: ignore
```
Dr-Irv, line 183: should put the specific `pyright` thing we are ignoring here.

pandas-stubs/core/groupby/generic.pyi
```diff
@@ -159,143 +199,94 @@ class DataFrameGroupBy(GroupBy, Generic[ByT]):
...
-        limit: int | None = ...,
-        downcast: dict | None = ...,
+        axis: Axis | None | NoDefault = ...,
+        skipna: bool = ...,
+        numeric_only: bool = ...,
     ) -> DataFrame: ...
-    def first(self, **kwargs) -> DataFrame: ...
-    def head(self, n: int = ...) -> DataFrame: ...
-    def hist(
+    boxplot = boxplot_frame_groupby
```
Dr-Irv, line 253: As mentioned elsewhere, let's keep the definition of `boxplot()` here.

pandas-stubs/core/groupby/generic.pyi
```diff
@@ -305,41 +296,57 @@ class DataFrameGroupBy(GroupBy, Generic[ByT]):
...
+        axis: Axis | NoDefault = ...,
+        drop: bool = ...,
+        method: CorrelationMethod = ...,
+        numeric_only: bool = ...,
+    ) -> DataFrame: ...
     def __getattr__(self, name: str) -> SeriesGroupBy[Any, ByT]: ...
-    def __iter__(self) -> Iterator[tuple[ByT, DataFrame]]: ...
+    # Overrides that provide more precise return types over the GroupBy class
+    @final  # type: ignore[misc]
+    def __iter__(self) -> Iterator[tuple[ByT, DataFrame]]: ...  # pyright: ignore
```
Dr-Irv, line 352: add specific ignore info here

pandas-stubs/core/groupby/ops.pyi
```diff
@@ -1,98 +1,103 @@
...
-    def __iter__(self): ...
-
-class SeriesSplitter(DataSplitter): ...
-class FrameSplitter(DataSplitter): ...
+class BinGrouper(BaseGrouper):
+    bins: npt.NDArray[np.int64]
+    binlabels: Index
+    indexer: npt.NDArray[np.intp]
+    @cache_readonly
+    def indices(self) -> dict[Incomplete, list[int]]: ...  # type: ignore[override] # pyright: ignore
```
Dr-Irv, line 96: add specific pyright `ignore` info here

pandas-stubs/core/resample.pyi
```diff
@@ -181,8 +140,68 @@ class Resampler(BaseGroupBy, Generic[NDFrameT]):
...
+    def quantile(
+        self,
+        q: float | list[float] | npt.NDArray[np.float_] | Series[float] = ...,
+        **kwargs,
+    ) -> NDFrameT: ...
+
+# We lie about inheriting from Resampler because at runtime inherits all Resampler
+# attributes via setattr
+class _GroupByMixin(Resampler[NDFrameT]):
+    k: str | list[str] | None
```
Dr-Irv, line 173: seems odd that we would export/document something called `k`

pandas-stubs/core/window/__init__.pyi
```diff
@@ -8,3 +11,13 @@ from pandas.core.window.rolling import (
...
+
+__all__ = [
+    "Expanding",
+    "ExpandingGroupby",
+    "ExponentialMovingWindow",
+    "ExponentialMovingWindowGroupby",
+    "Rolling",
+    "RollingGroupby",
+    "Window",
+]
```
Dr-Irv, line 23: Since none of these should be instantiated by a user, do we still need to have `__all__` here even though they are returned by various pandas methods?

Dr-Irv (changes requested):
This is pretty close.  Found a few small things, and we still have some open discussions.

pandas-stubs/core/generic.pyi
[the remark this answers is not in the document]
Dr-Irv: Created https://github.com/pandas-dev/pandas/issues/56877

pandas-stubs/core/groupby/generic.pyi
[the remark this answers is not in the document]
Dr-Irv: created an issue https://github.com/pandas-dev/pandas-stubs/issues/852 and https://github.com/pandas-dev/pandas/issues/56876#issuecomment-1891041139

pandas-stubs/plotting/_core.pyi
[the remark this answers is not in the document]
hamdanal: Thank you for the second pass of review; I'll address your comments soon.

-------------

While I understand the philosophy of not including everything upstream in the stubs, I am having a hard time understanding the rational behind this specific case. The function is explicitly public, included in `__all__` of a module documented as "Plotting public API", mentioned explicitly in this module's docstring:
> backends are expected to implement the next functions:
> ...
> - boxplot_frame and boxplot_frame_groupby

and doesn't require any added maintenance in the stub since it has to be defined anyway for its use as a method of dataframe groupby. Could you please explain why you see this as a negative.

pandas-stubs/core/window/__init__.pyi
hamdanal: Will remove

pandas-stubs/core/resample.pyi
hamdanal: Good call, this should've been `key`, I'll fix it.

pandas-stubs/core/groupby/generic.pyi
hamdanal: For some reason this pyright error and the one on the last line of the same file so not have a specific error code! I cannot add a specific pyright ignore rule here! This is what pyright emits:
`pandas-stubs/core/groupby/generic.pyi:352:9 - error: Method "__iter__" cannot override final method defined in class "BaseGroupBy"`

hamdanal: Created a pyright issue: https://github.com/microsoft/pyright/issues/6985

pandas-stubs/plotting/_core.pyi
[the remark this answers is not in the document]
twoertwein: It probably is in `__all__` so that other code from within pandas can import it but it isn't documented publicly https://pandas.pydata.org/docs/search.html?q=boxplot_frame_groupby I think pandas symbols are considered public only if they are publicly documented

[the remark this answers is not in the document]
Dr-Irv: > I think pandas symbols are considered public only if they are publicly documented

This is the rule we've used with pandas-stubs.  We don't want users to call `boxplot_frame_groupby()`, so a user with a call to that function should see it as an error, as it would be unsupported.

pandas-stubs/core/groupby/generic.pyi
Dr-Irv: Looks like he fixed it and a new release of pyright will be available, so I recommend updating the pyright version in `pyproject.toml` and then using the correct error code.

twoertwein (approved):
I didn't cross-check it with the documentation/pandas code but the changes overall look amazing!

Dr-Irv:
@hamdanal let me know if you want me to do another review, or if you have more commits coming

hamdanal:
> @hamdanal let me know if you want me to do another review, or if you have more commits coming

I addressed your previous comments. Feel free to do another review.

pyproject.toml
```diff
@@ -191,6 +191,7 @@ reportUnknownParameterType = false
 reportUnknownVariableType = false
 reportUnusedVariable = false
 reportPrivateUsage = false
+reportSelfClsParameterName = false
```
Dr-Irv, line 194: Why did you add this?

Dr-Irv (changes requested):
Everything looks fine codewise, but there is a change to `pyproject.toml` that I don't think should be there.

pyproject.toml
hamdanal: Oops I am sorry I changed it without leaving a comment, I forgot I made this change.

This is a lint rule, rather than a type check, that doesn’t apply to stubs. The name of the variable is out of control of the stub, it is defined by the implementation  (see for example [typeshed](https://github.com/python/typeshed/blob/8b9b0f4d031fe614a0211e1faadf57b3340e643e/pyrightconfig.json#L43-L44)). If, for instance, the implementation defines a method like:
```python
class C:
    @classmethod
    def method(klass): …
```
and the stub defines it as:
```python
class C:
    @classmethod
    def method(cls): …
```
just to silence the linter, then if someone calls `C.method(cls=…)` they will get a TypeError.

I started getting errors about this when I added the boxplot method.

Anyhow, if you don’t like this change please feel free to change it back. I won’t have access to computer for some time.

hamdanal:
I reverted the pyproject.toml change and fixed some new warnings in the tests. I also pinned pyright since it changed some of its error codes and started failing in places unrelated to this PR.

pandas-stubs/core/groupby/generic.pyi
```diff
@@ -252,7 +252,7 @@ class DataFrameGroupBy(GroupBy[DataFrame], Generic[ByT]):
     ) -> DataFrame: ...
     @overload
     def boxplot(
-        grouped,
+        grouped,  # pyright: ignore[reportSelfClsParameterName]
```
Dr-Irv, line 255: Why not just use `self` here (and in the other overloads)?

pandas-stubs/core/groupby/generic.pyi
```diff
@@ -159,143 +199,141 @@ class DataFrameGroupBy(GroupBy, Generic[ByT]):
...
-        subplots: bool = ...,
-        column: str | Sequence | None = ...,
-        fontsize: float | str = ...,
+        axis: Axis | None | NoDefault = ...,
+        skipna: bool = ...,
+        numeric_only: bool = ...,
+    ) -> DataFrame: ...
+    @overload
+    def boxplot(
+        grouped,  # pyright: ignore[reportSelfClsParameterName]
```
hamdanal, line 255:
```suggestion
        self,
```

pandas-stubs/core/groupby/generic.pyi
```diff
@@ -159,143 +199,141 @@ class DataFrameGroupBy(GroupBy, Generic[ByT]):
...
-    def head(self, n: int = ...) -> DataFrame: ...
-    def hist(
-        self,
-        data: DataFrame,
-        column: str | Sequence | None = ...,
-        by=...,
+    ) -> Series: ...  # Series[PlotAxes] but this is not allowed
+    @overload
+    def boxplot(
+        grouped,  # pyright: ignore[reportSelfClsParameterName]
```
hamdanal, line 271:
```suggestion
        self,
```

pandas-stubs/core/groupby/generic.pyi
```diff
@@ -159,143 +199,141 @@ class DataFrameGroupBy(GroupBy, Generic[ByT]):
...
+        figsize: tuple[float, float] | None = ...,
+        layout: tuple[int, int] | None = ...,
         sharex: bool = ...,
         sharey: bool = ...,
+        backend: str | None = ...,
+        **kwargs,
+    ) -> PlotAxes: ...
+    @overload
+    def boxplot(
+        grouped,  # pyright: ignore[reportSelfClsParameterName]
```
hamdanal, line 287:
```suggestion
        self,
```

Dr-Irv (approved):
Thanks @hamdanal !  Lots of good work here

PR #858, opened 2024-02-09
Fix pd.concat to accept None values as input.

### LINKED ISSUE #857: pd.concat with None
**Describe the bug**
`concat` with `None` values are valid use but pandas-stubs rejects.

**To Reproduce**
```
def x() -> pd.DataFrame:
    df1: pd.DataFrame | None = None
    df2 = pd.DataFrame()
    return pd.concat([df1, df2])
```
2. Using mypy
3. Error
`List item 0 has incompatible type "DataFrame | None"; expected "DataFrame"  [list-item]`

 - OS: Fedora 
 - OS Version 7.8
 - python version 3.11.7
 - version of type checker 1.8.0n
 - version of installed `pandas-stubs` 2.1.1

amgcc:
- [x] Closes #857

### DIFF
--- a/pandas-stubs/core/reshape/concat.pyi
+++ b/pandas-stubs/core/reshape/concat.pyi
@@ -12,8 +12,10 @@ from pandas import (
     DataFrame,
     Series,
 )
+from typing_extensions import Never
 
 from pandas._typing import (
+    Axis,
     AxisColumn,
     AxisIndex,
     HashableT1,
@@ -24,9 +26,23 @@ from pandas._typing import (
 
 @overload
 def concat(
-    objs: Iterable[DataFrame | None] | Mapping[HashableT1, DataFrame],
+    objs: Iterable[None] | Mapping[HashableT1, None],
     *,
-    axis: AxisIndex = ...,
+    axis: Axis = ...,
+    join: Literal["inner", "outer"] = ...,
+    ignore_index: bool = ...,
+    keys: Iterable[HashableT2] = ...,
+    levels: Sequence[list[HashableT3] | tuple[HashableT3, ...]] = ...,
+    names: list[HashableT4] = ...,
+    verify_integrity: bool = ...,
+    sort: bool = ...,
+    copy: bool = ...,
+) -> Never: ...
+@overload
+def concat(  # type: ignore[overload-overlap] # pyright: ignore[reportOverlappingOverload]
+    objs: Iterable[DataFrame | None] | Mapping[HashableT1, DataFrame | None],
+    *,
+    axis: Axis = ...,
     join: Literal["inner", "outer"] = ...,
     ignore_index: bool = ...,
     keys: Iterable[HashableT2] = ...,
@@ -38,7 +54,7 @@ def concat(
 ) -> DataFrame: ...
 @overload
 def concat(
-    objs: Iterable[Series | None] | Mapping[HashableT1, Series],
+    objs: Iterable[Series | None] | Mapping[HashableT1, Series | None],
     *,
     axis: AxisIndex = ...,
     join: Literal["inner", "outer"] = ...,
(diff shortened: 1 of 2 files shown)


Dr-Irv (changes requested):
Please add a test in `test_pandas.py`.  Follow the pattern in there of using `check()` and `assert_type()`.

Dr-Irv:
You have some failures in the tests.  See https://github.com/pandas-dev/pandas-stubs/actions/runs/7850999903/job/21487238109?pr=858

Be sure to set up a local environment and run tests locally.  See https://github.com/pandas-dev/pandas-stubs/blob/main/docs/setup.md and https://github.com/pandas-dev/pandas-stubs/blob/main/docs/tests.md

twoertwein:
Sorry, had to force push - merging upstream messed things somehow up.

pandas-stubs/core/reshape/concat.pyi
```diff
@@ -24,9 +26,23 @@ from pandas._typing import (
...
+    ignore_index: bool = ...,
+    keys: Iterable[HashableT2] = ...,
+    levels: Sequence[list[HashableT3] | tuple[HashableT3, ...]] = ...,
+    names: list[HashableT4] = ...,
+    verify_integrity: bool = ...,
+    sort: bool = ...,
+    copy: bool = ...,
+) -> Never: ...
+@overload
+def concat(  # type: ignore[overload-overlap] # pyright: ignore[reportOverlappingOverload]
```
twoertwein, line 42: overlaps because of `Iterable[None]`

tests/test_pandas.py
```diff
@@ -51,17 +54,25 @@ def test_types_to_datetime() -> None:
...
-    check(assert_type(pd.concat([df1, df2]), pd.DataFrame), pd.DataFrame)
+    check(assert_type(pd.concat({"a": None, "b": series}), pd.Series), pd.Series)
+    check(assert_type(pd.concat({"a": None, "b": df}), pd.DataFrame), pd.DataFrame)
+    check(
+        assert_type(pd.concat({"a": None, "b": series, "c": df}, axis=1), pd.DataFrame),
+        pd.DataFrame,
+    )
 
-    check(assert_type(pd.concat([s1, df1, df2]), pd.DataFrame), pd.DataFrame)
+    if TYPE_CHECKING_INVALID_USAGE:
```
amgcc, line 73: @twoertwein can you explain what this block means and why this is needed?  Trying to understand what was missing from my tests.

Dr-Irv: This is testing that the type checkers see the code as invalid.  Although I don't think the test is constructed correctly....

tests/test_pandas.py
```diff
@@ -49,6 +52,29 @@ def test_types_to_datetime() -> None:
...
+    check(assert_type(pd.concat({"a": None, "b": series}), pd.Series), pd.Series)
+    check(assert_type(pd.concat({"a": None, "b": df}), pd.DataFrame), pd.DataFrame)
+    check(
+        assert_type(pd.concat({"a": None, "b": series, "c": df}, axis=1), pd.DataFrame),
+        pd.DataFrame,
+    )
+
+    if TYPE_CHECKING_INVALID_USAGE:
+        assert_type(pd.concat({"a": None}), Never)
+        assert_type(pd.concat([None]), Never)
```
Dr-Irv, line 75: I think these tests should be of the form:
```python
    pd.concat({"a": None})  # type: ignore[some_mypy_error] # pyright: ignore[some_pyright_error]
    pd.concat([None])  # type: ignore[some_mypy_error] # pyright: ignore[some_pyright_error]
```
so we are checking that the type checkers see that invalid code as an error.

Dr-Irv (changes requested):

tests/test_pandas.py
twoertwein: I believe I tried that first - the issue is that the second call will not be checked as it cannot be reached (by the type checkers). Would need to split it into two functions.

Dr-Irv: That makes sense.  I think for consistency's sake, we should do that, although I could be convinced otherwise.  If we want to use your pattern here, then add a comment to indicate why we can't just check for a specific type checker error based on your comment here.

twoertwein: I think we need the `assert_type` only when testing at least two invalid function calls where the first one "returns"  `NoReturn/Never`. I think we also have many cases were we do not explicitly return NoReturn/Never. In that case, we might not need the `assert_type`.

I would be inclined to use the `assert_type` only when we have to.

Dr-Irv (approved):
thanks @amgcc and @twoertwein

PR #861, opened 2024-02-11
Fix calling df.agg with scalar dict values

### LINKED ISSUE #846: `DataFrame.agg()` can return `Series` or `DataFrame` with `dict` argument
**Describe the bug**

There is no overload for `agg` returning a scalar, which it can per [the docs](https://pandas.pydata.org/docs/reference/api/pandas.DataFrame.agg.html) and:

**To Reproduce**

```py
df = pd.DataFrame([[1, 2, 3],
                   [4, 5, 6],
                   [7, 8, 9],
                   [np.nan, np.nan, np.nan]],
                  columns=['A', 'B', 'C'])



# Expression of type "DataFrame" cannot be assigned to declared type "float64"
#  "DataFrame" is incompatible with "float64"Pylance[reportGeneralTypeIssues](https://github.com/microsoft/pyright/blob/main/docs/configuration.md#reportGeneralTypeIssues)
A_sum: np.float64 = df.agg({'A': 'sum'})


# But:
print(A_sum)

# A    12.0
# dtype: float64
```

 - Linux
 - Ubuntu 20.04.6 LTS
 - python 3.11.4
 - pyright
 - `pandas-stubs == 2.1.4.231227

**Additional context**

hamdanal:
- [x] Closes #846

As promised in https://github.com/pandas-dev/pandas-stubs/issues/846#issuecomment-1880005640

The overlap in overloads is necessary as a dict with only scalar values returns a series but a dict with lists *and optionally scalar values* returns a dataframe. So both overloads have to accept scalar values but that doesn't seem to be a problem as both mypy and pyright are able to infer the good overload.

### DIFF
--- a/pandas-stubs/core/frame.pyi
+++ b/pandas-stubs/core/frame.pyi
@@ -61,6 +61,7 @@ from pandas._typing import (
     S1,
     AggFuncTypeBase,
     AggFuncTypeDictFrame,
+    AggFuncTypeDictSeries,
     AggFuncTypeFrame,
     AnyArrayLike,
     ArrayLike,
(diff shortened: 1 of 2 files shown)


Dr-Irv (approved):
thanks @hamdanal

PR #864, opened 2024-02-12
Add `unit` property and `as_unit` method to `Timestamp`, `Timedelta` and `NaTType`

### LINKED ISSUE #797: "Timestamp" has no attribute "as_unit" [attr-defined]
I think we are missing the `.as_unit` method added to pd.Timestamp in 2.0

### LINKED ISSUE #859: Timestamp.as_unit
**Describe the bug**
Method `as_unit` not defined on `Timestamp`

**To Reproduce**
`pd.Timestamp(2023, 1, 1).as_unit("ns")`
mypy
error: "Timestamp" has no attribute "as_unit"  [attr-defined]

 - OS: Linux
 - OS Fedora 7
 - python version: 3.11
 - version of type checker: 1.8
 - version of installed `pandas-stubs`=2.1.4

skatsuta:
- [x] Closes #797
- [x] Closes #859 

Add `unit` property and `as_unit` method to `Timestamp`, `Timedelta` and `NaTType`.

Adding them to `DatetimeIndex`, `TimedeltaIndex` and `Series.dt` will be done in #865.


pandas-stubs/_libs/tslibs/nattype.pyi
```diff
@@ -121,3 +125,8 @@ class NaTType:
     __le__: _NatComparison
     __gt__: _NatComparison
     __ge__: _NatComparison
+    @property
+    def unit(self) -> str: ...
+    def as_unit(
+        self, unit: Literal["s", "ms", "us", "ns"], round_ok: bool = ...
+    ) -> Self: ...
```
Dr-Irv, line 132: Can you create a new type in `_typing.pyi` called `TimeUnit` that is `Literal["s", "ms", "us", "ns"]` and use that as the return type of `unit` and as the argument to `as_unit()` ?
Also replace wherever we have in the stubs that `Literal` and use `TimeUnit` instead.

Note that `TimeUnit` is now in `pandas/_typing.py` .

tests/test_scalars.py
```diff
@@ -517,6 +517,7 @@ def test_timedelta_properties_methods() -> None:
     check(assert_type(td.value, int), int)
     check(assert_type(td.resolution_string, str), str)
     check(assert_type(td.components, Components), Components)
+    check(assert_type(td.unit, str), str)
```
Dr-Irv, line 520: Since `TimeUnit` is in `pandas`, you could import that and use it here in the `assert_type` test.

Dr-Irv (changes requested):
Suggest using `TimeUnit` that is now in pandas.

Dr-Irv:
Closing in favor of #864

skatsuta (commented):
@Dr-Irv Thanks for your suggestion.

Defined the `TimeUnit` type alias and used it throughout the stubs.

Also rebased to resolve conflicts with the main branch, so apologize if additional changes are not clear.

Will update #865 to use `TimeUnit` as well once this PR gets merged.

Dr-Irv (approved):
thanks @skatsuta

PR #867, opened 2024-02-12
Fix `Series#get` and `DataFrame#get`

### LINKED ISSUE #808: pd.Series.get causes typing error in second argument (default value)
**Describe the bug**
`pd.Series.get()` seems to ignore the Series' type and causes an unexpected mypy error in the default argument.

**To Reproduce**
Minimal runnable `pandas` example:
```python
from typing import cast
import pandas as pd

s = pd.Series([1, 3, 3, 7, "l33t"], index=["a", "b", "c", "d"])

i = s.get("b", -1)

j = cast(pd.Series[int], s).get("b", -1)
```

I am using `mypy`, and the error message I receive is
`Argument 2 to "get" of "NDFrame" has incompatible type "int"; expected "ExtensionDtype | str | dtype[generic] | type[str] | type[complex] | type[bool] | type[object] | None"  [arg-type]mypy(error)`

image

I can understand why the `i` line can cause problems (as the Series' dtype is not uniform), but at least the `j` line should work fine in my opinion, and I have no idea how to fix this other than via `# type: ignore`.

 - OS: Windows 10
 - Python 3.10.12
 - `mypy=1.6.1` (shipped with VS Code extension; env contains `1.5.1`)
 - `pandas-stubs==2.1.1.230928` (via `conda-forge`)

skatsuta:
- [x] Closes #808

Fix `Series#get` and `DataFrame#get` to resolve #808.


pandas-stubs/core/frame.pyi
```diff
@@ -1683,7 +1686,14 @@ class DataFrame(NDFrame, OpsMixin):
     # def from_dict
     # def from_records
     def ge(self, other, axis: Axis = ..., level: Level | None = ...) -> DataFrame: ...
-    # def get
+    @overload
+    def get(self, key: Hashable, default: None = ...) -> Series | None: ...
+    @overload
+    def get(self, key: Hashable, default: _T = ...) -> Series | _T: ...
```
Dr-Irv, line 1692:
```suggestion
    def get(self, key: Hashable, default: _T ) -> Series | _T: ...
```
I think the `...` should be removed because you would only return `_T` if a specific non-None value was specified.

pandas-stubs/core/frame.pyi
```diff
@@ -1683,7 +1686,14 @@ class DataFrame(NDFrame, OpsMixin):
...
+    @overload
+    def get(self, key: list[Hashable], default: None = ...) -> DataFrame | None: ...
+    @overload
+    def get(self, key: list[Hashable], default: _T = ...) -> DataFrame | _T: ...
```
Dr-Irv, line 1696: same comment as above about `_T`

pandas-stubs/core/frame.pyi
```diff
@@ -127,6 +128,8 @@ from pandas.plotting import PlotAccessor
 _str = str
 _bool = bool
 
+_T = TypeVar("_T")
```
Dr-Irv, line 131: You can import `T` from `_typing` here and use that instead.

pandas-stubs/core/series.pyi
```diff
@@ -160,6 +161,8 @@ from pandas.plotting import PlotAccessor
 _bool = bool
 _str = str
 
+_T = TypeVar("_T")
```
Dr-Irv, line 164: You can import `T` from `_typing` here and use that instead.

pandas-stubs/core/series.pyi
```diff
@@ -371,6 +374,10 @@ class Series(IndexOpsMixin[S1], NDFrame):
     @overload
     def __getitem__(self, idx: Scalar) -> S1: ...
     def __setitem__(self, key, value) -> None: ...
+    @overload
+    def get(self, key: Hashable, default: None = ...) -> S1 | None: ...
+    @overload
+    def get(self, key: Hashable, default: S1 | _T = ...) -> S1 | _T: ...
```
Dr-Irv, line 380: Remove the `...` for `default` .  Also, split this into 2 overloads, one for `S1` and another for `_T` (which should change to `T`) .  That might require an `ignore` because of overlaps.

tests/test_frame.py
```diff
@@ -3107,3 +3107,34 @@ def test_itertuples() -> None:
...
+    df = pd.DataFrame({"a": [1, 2, 3], "b": [4, 5, 6], "c": [7, 8, 9]})
+
+    # Get single column
+    check(assert_type(df.get("a"), Union[pd.Series, None]), pd.Series, np.int64)
+    check(assert_type(df.get("z"), Union[pd.Series, None]), type(None))
+    check(
+        assert_type(df.get("a", default=None), Union[pd.Series, None]),
+        pd.Series,
+        np.int64,
+    )
```
Dr-Irv, line 3122: Repeat this test with `df.get("z", default=None)` that would return `None`

Dr-Irv (changes requested):

pandas-stubs/core/frame.pyi
```diff
@@ -114,6 +114,7 @@ from pandas._typing import (
     StorageOptions,
     StrLike,
     Suffixes,
+    T as _T,
```
skatsuta, line 117: Use alias to avoid conflict with `DataFrame#T` method.

skatsuta (commented):
@Dr-Irv Thanks for your suggestions.

Added the following changes:
* Used `_typing.T` instead of locally defined type vars
* Removed `= ...` from the overloads for cases where the `default` parameter is specified
* Split `Series#get(..., default: S1) -> S1` and `Series#get(..., default: T) -> S1 | T`
* Added test cases for when the `default` parameter is `None` and no key is found

Dr-Irv (approved):
thanks @skatsuta

PR #869, opened 2024-02-13
CI: fix nightly

twoertwein:


tests/test_io.py
```diff
@@ -427,18 +428,21 @@ def test_hdf_series():
...
-            assert_type(read_spss(str(path), usecols=["VAR00002"]), DataFrame),
-            DataFrame,
-        )
+    if PD_LTE_22:
+        path = Path(CWD, "data", "labelled-num.sav")
+        with pytest_warns_bounded(
+            FutureWarning,
+            "ChainedAssignmentError: behaviour will change",
+            lower="2.1.99",
+        ):
```
Dr-Irv, line 437: I think we need to do this test on nightly, but make the error message dependent on the version.  For version 2.2, the message is `"ChainedAssignmentError: behaviour will change"` .  For nightly, it is `"ChainedAssignmentError: A value is trying to be set on"`

So let's keep the test, and just make the message that is checked based on `PD_LTE_22`

There is an issue about this:  https://github.com/pandas-dev/pandas/issues/51315

Dr-Irv (changes requested):
one issue with respect to `read_spss()` test, otherwise ok

tests/test_io.py
twoertwein: Changed! There were a few new errors on nightly (finally removed support for `"ignore"`)

Dr-Irv (approved):
thanks @twoertwein

PR #875, opened 2024-02-21
Added types to df.info()

askmyhat:


pandas-stubs/core/frame.pyi
```diff
@@ -570,7 +570,7 @@ class DataFrame(NDFrame, OpsMixin):
         storage_options: StorageOptions = ...,
     ) -> str: ...
     def info(
-        self, verbose=..., buf=..., max_cols=..., memory_usage=..., null_counts=...
+        self, verbose=..., buf=..., max_cols=..., memory_usage=..., show_counts=...
```
twoertwein, line 573: I believe you can also add the types `show_counts: bool | None = ...`

twoertwein (commented):
Can you please add a test case, something like:
```py
check(assert_type(df.info(show_counts=True), None), None)
check(assert_type(df.info(show_counts=None), None), None)
```

pandas-stubs/core/frame.pyi
```diff
@@ -570,7 +570,12 @@ class DataFrame(NDFrame, OpsMixin):
         storage_options: StorageOptions = ...,
     ) -> str: ...
     def info(
-        self, verbose=..., buf=..., max_cols=..., memory_usage=..., null_counts=...
+        self,
+        verbose: bool | None = ...,
+        buf: WriteBuffer[str] = ...,
+        max_cols: int | None = ...,
+        memory_usage: bool | str | None = ...,
```
Dr-Irv, line 577: The `str` value can only be `Literal["deep"]` according to the docs.

tests/test_frame.py
```diff
@@ -3147,3 +3147,22 @@ def test_get() -> None:
         pd.DataFrame,
     )
     check(assert_type(df.get(["z"], default=1), Union[pd.DataFrame, int]), int)
+
+
+def test_info() -> None:
+    df = pd.DataFrame()
+
+    check(assert_type(df.info(verbose=True), None), type(None))
+    check(assert_type(df.info(verbose=None), None), type(None))
```
Dr-Irv, line 3156: For any of these tests that are testing values of  `True` and `None`, should also add a test testing a value of `False`

Dr-Irv (changes requested):
Some small changes

Dr-Irv (approved):
thanks @askmyhat

PR #880, opened 2024-02-26
fix concat for axis=0

### LINKED ISSUE #879: Issue with `concat` and passing a `Union` due to presence of checking for `Iterable[None]`
You are right, the example doesn't work, my bad. This example does work though, and the same question still applies.

```python
data: pd.DataFrame | pd.Series = pd.Series()

pd.concat([pd.DataFrame(), data])
```

returns

```
error: List item 0 has incompatible type "DataFrame"; expected "None"  [list-item]
error: List item 1 has incompatible type "DataFrame | Series[Any]"; expected "None"  [list-item]
```

_Originally posted by @tvdboom in https://github.com/pandas-dev/pandas-stubs/discussions/878#discussioncomment-8594364_

twoertwein:
closes #879

Works on pyright - mypy behaves weirdly (the 3rd overload is clearly the first matching overload). @Dr-Irv, feel free to change this PR - not sure whether I will have time to debug the mypy issue before the weekend.

### DIFF
--- a/tests/test_pandas.py
+++ b/tests/test_pandas.py
@@ -62,11 +62,16 @@ def test_types_concat_none() -> None:
     check(
         assert_type(pd.concat([None, series, df], axis=1), pd.DataFrame), pd.DataFrame
     )
+    check(assert_type(pd.concat([None, series, df]), pd.DataFrame), pd.DataFrame)
 
     check(assert_type(pd.concat({"a": None, "b": series}), pd.Series), pd.Series)
     check(assert_type(pd.concat({"a": None, "b": df}), pd.DataFrame), pd.DataFrame)
     check(
         assert_type(pd.concat({"a": None, "b": series, "c": df}, axis=1), pd.DataFrame),
+        pd.DataFrame,
+    )
+    check(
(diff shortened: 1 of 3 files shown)


tests/test_pandas.py
```diff
@@ -77,8 +82,8 @@ def test_types_concat_none() -> None:
 
 
 def test_types_concat() -> None:
-    s: pd.Series = pd.Series([0, 1, -10])
-    s2: pd.Series = pd.Series([7, -5, 10])
```
twoertwein, line 81: These two annotations messed mypy in the previous commit. up

Dr-Irv (approved):
thanks @twoertwein

tvdboom:
thanks for the fix. @Dr-Irv any idea when the next release will be?

Dr-Irv:
> thanks for the fix. @Dr-Irv any idea when the next release will be?

I usually wait for a number of PR's to come through before doing a release. We've only had 3 since the last release.

Dr-Irv:
> thanks for the fix. @Dr-Irv any idea when the next release will be?

@tvdboom just released v2.2.1.240316

PR #890, opened 2024-03-15
fix concat, update mypy, pyright

### LINKED ISSUE #888: `concat` of types `Iterable[Any]` cause Pyright to think subsequent code is unreachable because of recent stubs change
**Describe the bug**
In the following code the last line is wrongly considered to be unreachable by Pyright:
```python
from typing import Any

import pandas as pd


def generate_series() -> Any:
    return pd.Series([1, 2, 3])


df = pd.concat({k: generate_series() for k in range(10)}, axis=1)

print("Unreachable?!")
```
This is caused by the recent change https://github.com/pandas-dev/pandas-stubs/pull/858 adding a stub returning `Never`. This is presumably a misuse of `Never` or the stub as a whole should be moved to the very end so that `concat` on types `Iterable[Any]` can first match a stub with valid return.

**To Reproduce**
See https://github.com/microsoft/pylance-release/issues/5631 and https://github.com/microsoft/pylance-release/issues/5630 for issues reported on the Pylance repo.

 - OS: Linux
 - OS Version 5.15.0-46-generic
 - Python 3.11.6
 - Pylance v2024.3.1
 - version of installed `pandas-stubs`: the one shipped with above Pylance version

Dr-Irv:
- [x] Closes #888 

Updated mypy, pyright versions as well.


Dr-Irv:
@twoertwein I figured out a way that we can make things work by adding additional overloads that don't include `None` at the top, and leaving in the `Never` so we get the best of both worlds.

twoertwein (approved):

twoertwein:
Thank you @Dr-Irv ! It's great that you found a way to get it to work (even if it creates a few near-duplicate overloads).

PR #894, opened 2024-03-22
Fixed TYP of names argument of concat method and included test

### LINKED ISSUE #883: pd.concat with names=None
**Describe the bug**
`pd.concat` should accept `names=None` as it is the default value.

**To Reproduce**

```
def maybe_concat(
    dd: dict[str, pd.DataFrame], names: list[str] | None = None
) -> pd.DataFrame:
    """Concatenate a dict of dataframes if available."""
    return pd.concat(dd, names=names) if dd else pd.DataFrame()
```

**mypy**
```
error: Argument "names" to "concat" has incompatible type "list[str] | None"; expected "list[str]"  [arg-type]
```

 - OS: [e.g. Windows, Linux, MacOS] **Linux**
 - OS Version [e.g. 22] **Fedora 7**
 - python version **3.11.8**
 - version of type checker **1.8.0**
 - version of installed `pandas-stubs` **2.1.4**

Aftabby:
The TYP of the `names` argument in the `pd.concat()` method, was updated with additional `| None` values. 
And 5 tests were added.

- [ ] Closes #883 
- [ ] Tests added: 
`
check(
        assert_type(
            pd.concat([s, s2], keys=["first", "second"], names=None),
            pd.Series,
        ),
        pd.Series,
    )
`
`
    check(
        assert_type(
            pd.concat([df, df2], keys=["first", "second"], names=None),
            pd.DataFrame,
        ),
        pd.DataFrame,
    )
`
`
    check(
        assert_type(
            pd.concat([df, df2], keys=["df1", "df2"], names=None), pd.DataFrame
        ),
        pd.DataFrame,
    )
`
`
    check(
        assert_type(
            pd.concat([df, df2], keys=["df1", "df2"], names=None),
            pd.DataFrame,
        ),
        pd.DataFrame,
    )
`
`
    check(
        assert_type(
            pd.concat(
                [df, df2, df, df2],
                keys=[("foo", "one"), ("foo", "two"), ("baz", "one"), ("baz", "two")],
                levels=levels,
                names=None,
            ),
            pd.DataFrame,
        ),
        pd.DataFrame,
    )
`


tests/test_pandas.py
```diff
@@ -189,13 +203,26 @@ def test_concat_args() -> None:
...
         ),
         pd.DataFrame,
     )
+    check(
+        assert_type(
+            pd.concat([df, df2], keys=["df1", "df2"], names=None),
+            pd.DataFrame,
+        ),
+        pd.DataFrame,
+    )
```
Dr-Irv, line 225: Unless I'm missing something, I think this test is the same as the test on lines 206-211, so you can delete it.

Dr-Irv (changes requested):
Looks good.  I think you have a duplicate test that you can delete.

Dr-Irv (approved):
thanks @Aftabby

PR #897, opened 2024-03-29
896 dataframe from dict columns dtype args

mutricyl:
- [x] Closes #896 
- [x] Tests added within `test_frame.test_types_from_dict`
- [ ] ⚠️ b4c85be1df85ab02396d19610ca16286089a2c11 is not in the scope of issue #896 . It was only commited to pass the tests.

PR #898, opened 2024-03-31
Add calamine Excel reader

davetapley:
Since https://github.com/pandas-dev/pandas/pull/54998

No existing tests covering `engine`, I can add if needed?

### DIFF
--- a/pandas-stubs/io/excel/_base.pyi
+++ b/pandas-stubs/io/excel/_base.pyi
@@ -23,6 +23,8 @@ from pandas._libs.lib import NoDefault
 from pandas._typing import (
     Dtype,
     DtypeBackend,
+    ExcelReadEngine,
+    ExcelWriteEngine,
     FilePath,
     IntStrT,
     ListLikeHashable,
@@ -50,7 +52,7 @@ def read_excel(
     index_col: int | Sequence[int] | None = ...,
     usecols: str | UsecolsArgType = ...,
     dtype: str | Dtype | Mapping[str, str | Dtype] | None = ...,
-    engine: Literal["xlrd", "openpyxl", "odf", "pyxlsb"] | None = ...,
+    engine: ExcelReadEngine | None = ...,
     converters: Mapping[int | str, Callable[[object], object]] | None = ...,
     true_values: Iterable[Hashable] | None = ...,
     false_values: Iterable[Hashable] | None = ...,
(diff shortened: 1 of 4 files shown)


Dr-Irv:
Thanks for the PR.  Would like opinion of @twoertwein to determine if we should add tests for `engine` keyword.  On the one hand, it would be nice to have for completeness.  On the other hand, we'd have to add a build dependency for `calamine`, which might be fine to do assuming all the installation issues work out for developers and in CI.

pandas-stubs/io/excel/_base.pyi
```diff
@@ -9,6 +9,7 @@ from types import TracebackType
 from typing import (
     Any,
     Literal,
+    TypeAlias,
```
Dr-Irv, line 12: You'll need to import `TypeAlias` from `typing_extensions` so we can support earlier versions of python.

Dr-Irv (changes requested):

Dr-Irv:
Also, CI is failing because a change was made in pyright 1.1.356, so you can lock down pyright to 1.1.354 in this PR in `pyproject.toml`.  There's another PR out there that will allow us to upgrade pyright to 1.1.356

twoertwein:
>  we'd have to add a build dependency for `calamine`, which might be fine to do assuming all the installation issues work out for developers and in CI.

I'm not familiar with how python-calamine is packaged. Assuming its pypi package contains everything to test it, I don't see a reason not to make it a development dependency. If we have to install packages that are not on Pypi, I would be inclined to have tests inside a TYPE_CHECKING block instead.

Dr-Irv:
> > we'd have to add a build dependency for `calamine`, which might be fine to do assuming all the installation issues work out for developers and in CI.
> 
> I'm not familiar with how python-calamine is packaged. Assuming its pypi package contains everything to test it, I don't see a reason not to make it a development dependency. If we have to install packages that are not on Pypi, I would be inclined to have tests inside a TYPE_CHECKING block instead.

It is on pypi and conda-forge, so I think we'd be OK.  I was concerned that it is built on top of rust, and I don't know if that means a whole lot of other stuff gets installed as a result.

@davetapley Open to have you add tests to this PR

pyproject.toml
```diff
@@ -39,7 +39,7 @@ mypy = "1.9.0"
 pandas = "2.2.1"
 pyarrow = ">=10.0.1"
 pytest = ">=7.1.2"
-pyright = ">=1.1.354"
+pyright = "=1.1.354"
```
Dr-Irv, line 42: #897 fixed this issue, so you can use `pyright = ">=1.1.357"` and CI should pass

Dr-Irv (changes requested):
Can you merge with `upstream/main` - CI is failing because of that.  And update the pyright version as indicated.

pandas-stubs/io/excel/_base.pyi
```diff
@@ -32,6 +35,8 @@ from pandas._typing import (
     WriteExcelBuffer,
 )
 
+ReadEngine: TypeAlias = Literal["xlrd", "openpyxl", "odf", "pyxlsb", "calamine"]
```
twoertwein, line 38: Could go in `pandas._typing`.

twoertwein (approved):

PR #900, opened 2024-04-02
887 select_dtypes stubs fixing

### LINKED ISSUE #887: In `DataFrame.select_dtypes` type values like `np.number` or `np.datetime64` should be allowed
**Describe the bug**
The documentation of `DataFrame.select_dtypes` clearly states that it is allowed to use values like `np.number` or `np.datetime64` for the `include` and `exclude` parameters, but the current stubs don't allow it, they enforce `include: _str | list[_str] | None` and `exclude: _str | list[_str] | None`. This leads to typecheckers like `mypy` reporting false positives on such calls.

**To Reproduce**
If I run mypy on the following short snippet I get an error:

```python
import numpy as np
import pandas as pd

df = pd.DataFrame()
df.select_dtypes(np.number)
```

the error is:

```
min_example.py:5: error: Argument 1 to "select_dtypes" of "DataFrame" has incompatible type "type[number[Any]]"; expected "str | list[str] | None"  [arg-type]
```

 - OS: Linux
 - OS Version: Arch Linux
 - python version: 3.11
 - version of type checker: mypy 1.8.0
 - version of installed `pandas-stubs`: 2.2.0.240218

mutricyl:
- [ ] Closes #887

I tried include with overloads limitations stated in `select_dtypes` documentation: 

> ValueError
> - If both of include and exclude are empty
> - If include and exclude have overlapping elements
> - If any kind of string dtype is passed in.

I managed to cover somehow the first tick covering when neither of include or exclude is provided but I am not sure that it is worth the introduced complexity in stubs.

### DIFF
--- a/pandas-stubs/core/frame.pyi
+++ b/pandas-stubs/core/frame.pyi
@@ -51,6 +51,7 @@ from pandas.core.window.rolling import (
     Window,
 )
 from typing_extensions import (
+    Never,
     Self,
     TypeAlias,
 )
@@ -115,6 +116,7 @@ from pandas._typing import (
     SortKind,
     StataDateFormat,
     StorageOptions,
+    StrDtypeArg,
     StrLike,
     Suffixes,
     T as _T,
@@ -624,6 +626,18 @@ class DataFrame(NDFrame, OpsMixin):
         ]
     )
     AstypeArgExtList: TypeAlias = AstypeArgExt | list[AstypeArgExt]
+    @overload
+    def select_dtypes(
+        self, include: StrDtypeArg, exclude: AstypeArgExtList | None = ...
+    ) -> Never: ...
+    @overload
+    def select_dtypes(
+        self, include: AstypeArgExtList | None, exclude: StrDtypeArg
+    ) -> Never: ...
+    @overload
+    def select_dtypes(self, exclude: StrDtypeArg) -> Never: ...
+    @overload
+    def select_dtypes(self, include: list[Never], exclude: list[Never]) -> Never: ...
     @overload
     def select_dtypes(
         self,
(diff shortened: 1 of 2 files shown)


pandas-stubs/core/frame.pyi
```diff
@@ -608,10 +611,35 @@ class DataFrame(NDFrame, OpsMixin):
...
+            "number",
+            "datetime64",
+            "datetime",
+            "timedelta",
+            "timedelta64",
+            "datetimetz",
+            "datetime64[ns]",
+        ]
+    )
+    AstypeArgExtList: TypeAlias = AstypeArgExt | list[AstypeArgExt]
```
Dr-Irv, line 626: I think if you added the following here as the first 2 overloads, you could catch the case of the string arguments not being allowed:
```python
@overload
def select_dtypes(self, include: StrDtypeArg, exclude: AstypeArgExtList | None = ...) -> Never: ...
@overload
def select_dtypes(self, include: AstypeArgExtList | None = ..., exclude: StrDtypeArg) -> Never: ...
```
And if you added this, I think that you'd catch the case of 2 empty lists:
```python
@overload
def select_dtypes(self, include: list[Never], exclude: list[Never]) -> Never: ...
```
You import `Never` from `typing_extensions`

There's a possibility that either mypy and/or pyright will complain about overlapping overloads, but you can add `#type: ignore` and `# pyright: ignore` as needed in the stubs to ignore that complaint.

Dr-Irv (changes requested):
Overall, this looks good.  I have ideas for you to try to catch the case of passing `str` or `str` as the dtype, as well as the 2 empty list case,  so see if that works.

I don't think that overlapping lists can be detected with static typing.

pandas-stubs/core/frame.pyi
mutricyl: That is implemented. `Never` is actually stronger than I expected since pylance read the stubs and visually show that the program will stop there.

mutricyl:
@Dr-Irv your comments are implemented, you can review this PR one more time.
Regards

PR #903, opened 2024-04-06
BUG/TST: fixed series apply frozenset error and created test case

### LINKED ISSUE #871: Series apply frozenset raises error, frozenset not in enumerated Callable output types
**Describe the bug**
Series.apply has two overloads, the first one enumerates the possible return types of a Callable, this is missing frozenset (probably others as well).

https://github.com/pandas-dev/pandas-stubs/blob/23b35e48fb6b159d9ddab7f007ede26ae8a05564/pandas-stubs/core/series.pyi#L930-L937

**To Reproduce**

```python
import pandas as pd

s = pd.Series(["A", "B", "AB"])

s.apply(tuple)
s.apply(list)
s.apply(set)
s.apply(frozenset)
```

`mypy`

```none
series-apply.py:8: error: Argument 1 to "apply" of "Series" has incompatible type "type[frozenset[Any]]"; expected "Callable[..., str | bytes | date | datetime | timedelta | datetime64 | timedelta64 | bool | int | float | Timestamp | Timedelta | complex | Sequence[Any] | set[Any] | Mapping[Any, Any] | NAType | None]"  [arg-type]
```

 - OS: MacOS
 - OS Version 14.1.2 (23B92)
 - python version 3.11.4 (pyenv managed)
 - version of type checker `mypy==1.8.0`
 - version of installed `pandas-stubs==2.1.4.231227`

teshao:
- [x] Closes #871
- [x] Tests Added and Passed: test_frozenset.py 
- [x] All [code checks passed]

When trying to use series.Apply() on "frozenset", it raises the following error:

error: Argument 1 to "apply" of "Series" has incompatible type "type[frozenset[Any]]"; expected "Callable[..., str | bytes | date | datetime | timedelta | datetime64 | timedelta64 | bool | int | float | Timestamp | Timedelta | complex | Sequence[Any] | set[Any] | Mapping[Any, Any] | NAType | None]"  [arg-type]

Fixed this by adding "frozenset" to the enumerated Callable output types.
Created a test case by calling series.Apply(frozenset) and running "poe mypy".

### DIFF
--- a/tests/test_frozenset.py
+++ b/tests/test_frozenset.py
@@ -1,8 +1,0 @@
-import pandas as pd
-
-s = pd.Series(["A", "B", "AB"])
-
-s.apply(tuple)
-s.apply(list)
-s.apply(set)
-s.apply(frozenset)
--- a/tests/test_frozenset.py
+++ b/tests/test_frozenset.py
@@ -1,0 +1,8 @@
+import pandas as pd
+
+s = pd.Series(["A", "B", "AB"])
+
+s.apply(tuple)
+s.apply(list)
+s.apply(set)
+s.apply(frozenset)
--- a/tests/test_series.py
+++ b/tests/test_series.py
@@ -3069,3 +3069,11 @@ def test_pipe() -> None:
             ),
             1,
         )
+
+def test_series_apply() -> None:
+    s = pd.Series(["A", "B", "AB"])
+    
+    s.apply(tuple)
+    s.apply(list)
+    s.apply(set)
+    s.apply(frozenset)
--- a/tests/test_series.py
+++ b/tests/test_series.py
@@ -3072,7 +3072,11 @@ def test_pipe() -> None:
 
 def test_series_apply() -> None:
     s = pd.Series(["A", "B", "AB"])
-    s.apply(tuple)
-    s.apply(list)
-    s.apply(set)
-    s.apply(frozenset)
+
+    check(assert_type(s.apply(tuple), pd.Series), pd.Series, tuple)
+
+    check(assert_type(s.apply(list), pd.Series), pd.Series, list)
+
+    check(assert_type(s.apply(set), pd.Series), pd.Series, set)
+    
+    check(assert_type(s.apply(frozenset), pd.Series), pd.Series, frozenset)
--- a/tests/test_series.py
+++ b/tests/test_series.py
@@ -3073,11 +3073,7 @@ def test_pipe() -> None:
 
 def test_series_apply() -> None:
     s = pd.Series(["A", "B", "AB"])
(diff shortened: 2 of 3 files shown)


tests/test_frozenset.py
```diff
@@ -0,0 +1,8 @@
+import pandas as pd
+
+s = pd.Series(["A", "B", "AB"])
+
+s.apply(tuple)
+s.apply(list)
+s.apply(set)
+s.apply(frozenset)
```
twoertwein, line 8: This should probably be a function in test_series.py

teshao: Removed the extra test file and put it as a function in test_series.py!

twoertwein: I know you are not changing the output, but we try to use `check` and `assert_type` in all new tests
`check(assert_type(s.apply(frozenset), "pd.Series[dtype]" , pd.Series)`

teshao: Just fixed the tests.

tests/test_series.py
```diff
@@ -3069,3 +3069,15 @@ def first_arg_not_series(argument_1: int, ser: pd.Series) -> pd.Series:
             ),
             1,
         )
+
+
+def test_series_apply() -> None:
+    s = pd.Series(["A", "B", "AB"])
+    s.apply(tuple)
+    check(assert_type(s, "pd.Series[str]"), pd.Series, tuple)
```
twoertwein, line 3077: That is testing `s`, but it should be testing the return value of `s.apply`.

teshao: That makes sense but I can't seem to get it to work. 

I am getting errors when I do this:
check(assert_type(s.apply(tuple), "pd.Series[tuple]"), pd.Series) 
The error I'm getting is: 
image

Any tips?

twoertwein: I think you might have to use `"pd.Series[Any]"`

teshao: Fixed it!

twoertwein: `black` and then it should be ready

teshao: When I run "pre-commit run --all-files" locally, black passes. Can't seem to find the issue.

twoertwein: Strange, there could be a discrepancy between the black in pyproject and .pre-commit but when running pre-commit it should always use the one from .pre-commit (which should be the same version that the CI uses).

teshao: My local was automatically fixed when running pre-commit and I didn't notice that it had changed the file to fix the black error.

twoertwein (approved):

Dr-Irv (approved):
thanks @teshao

PR #905, opened 2024-04-10
Add stub for Styler.map_index

cquick01:

    Couldn't find an issue, but it might belong in #624. This is for this method https://github.com/pandas-dev/pandas/blob/5232bee8df3b57766c44b62152aa3fdd24e40ada/pandas/io/formats/style.py#L2018

    The function signature in code is 

    ```python
       def map_index(
            self,
            func: Callable,
            axis: AxisInt | str = 0,
            level: Level | list[Level] | None = None,
            **kwargs,
        ) -> Styler:
    ```

    but this looks to be the same as how `Styler.apply_index` is defined in code, but the stub is a little different so I just copied that instead

- [x] Tests added: Adds `test_styler.test_map_index`, similarly adapted from `test_apply_index`


pandas-stubs/io/formats/style.pyi
```diff
@@ -218,6 +218,13 @@ class Styler(StylerRenderer):
         level: Level | list[Level] | None = ...,
         **kwargs: Any,
     ) -> Styler: ...
+    def map_index(
+        self,
+        func: Callable[[Series], npt.NDArray[np.str_] | list[str] | Series[str]],
```
Dr-Irv, line 223: Based on the docs, I think this should be `Callable[[Scalar], str]`

tests/test_styler.py
```diff
@@ -63,6 +63,18 @@ def f1(s: Series) -> Series[str]:
...
+def test_map_index() -> None:
+    def f(s: Series) -> npt.NDArray[np.str_]:
+        return np.asarray(s, dtype=np.str_)
+
+    check(assert_type(DF.style.map_index(f), Styler), Styler)
+
+    def f1(s: Series) -> Series[str]:
+        return Series(s, dtype=str)
+
+    check(assert_type(DF.style.map_index(f1), Styler), Styler)
```
Dr-Irv, line 75: The tests should be modified to something similar to what you see in the docs for `map_index()`

Dr-Irv (changes requested):

pandas-stubs/io/formats/style.pyi
cquick01: Thanks, I updated to `Callable[[Scalar], str | None]` to match the docs

tests/test_styler.py
cquick01: Updated to show the first example from the docs. The second one showed the following error

```
    def f1(s: Scalar):
        return "background-color: yellow;" if "x" in s else None
                                              ~~~~~~~~
Unsupported right operand type for in ("str | bytes | date | datetime | timedelta | datetime64 | timedelta64 | bool | int | float | Timestamp | Timedelta | complex")
```

PR #914, opened 2024-04-28
adding str to read_excel index_col, resolves #809

### LINKED ISSUE #809: read_excel/index_col: string not accepted
Either the documentation or the type hint for index_col in pandas.read_excel is wrong.

The documentation says a string may be used.

The type hint forbids that

Documentation:
index_col     int, **str**, list of int, default None

Type hints:
`index_col: int | Sequence[int] | None = ...,`

mutricyl:
- [ ] Closes #809


Dr-Irv (approved):
Thanks @mutricyl

PR #916, opened 2024-04-29
copy paste from pandas readme, resolves #909

### LINKED ISSUE #909: README.md still suggests to join a gitter room
I would expect that to be updated to the pandas slack instead

mutricyl:
- [ ] Closes #909


Dr-Irv (approved):
thanks @mutricyl

PR #918, opened 2024-05-09
Add callable condition type in Series.mask, DataFrame.mask

### LINKED ISSUE #917: Passing a callable to DataFrame.mask's 'cond' argument results in arg-type error (mypy)
**Describe the bug**

The `cond` argument of the `DataFrame.mask` method currently [doesn't expect a callable](https://github.com/pandas-dev/pandas-stubs/blob/e63d1cbf5087cd7cd934db1c84184cb2caacbac9/pandas-stubs/core/frame.pyi#L1836), even though it should, according to the method's [documentation page](https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.DataFrame.mask.html):

> If cond is callable, it is computed on the Series/DataFrame and should return boolean Series/DataFrame or array. The callable must not change input Series/DataFrame (though pandas doesn’t check it).

**To Reproduce**

1. Minimally runnable `pandas` example that is not properly checked by the stubs:

```python
import pandas as pd
df = pd.DataFrame({"a": [1]}).mask(cond=lambda x: x.notna(), other=2)
print(df)
#    a
# 0  2
```

2. Type checker: `mypy`

```
(.venv) ➜  python -m mypy test.py
test.py:2: error: Argument "cond" to "mask" of "DataFrame" has incompatible type "Callable[[Any], Any]"; expected "Series[Any] | DataFrame | ndarray[Any, Any]"  [arg-type]
```

 - OS: MacOS
 - OS Version: Sonoma 14.2.1
 - python version: Python 3.10.14
 - version of type checker: 1.10.0
 - version of installed `pandas-stubs`: 2.2.1.240316

**Additional context**

I've also tried using `typing.cast` to cast the lambda into `Callable[[DataFrame], DataFrame]`, but it doesn't work, as expected. Happy to provide more details if needed.

georgesittas:
- [x] Closes #917

I played around a bit and found that `Series.mask()` actually had the same issue as `DataFrame.mask()`:

```python
import pandas as pd
s = pd.Series([1]).mask(cond=lambda x: x % 2 == 1, other=2)
print(s)
# 0    2
# dtype: int64
```

```
(.venv) ➜  test python -m mypy test.py
test.py:2: error: Argument "cond" to "mask" of "Series" has incompatible type "Callable[[Any], Any]"; expected "Series[bool] | ndarray[Any, dtype[bool_]] | list[bool]"  [arg-type]
```

I tried to fix both in this PR. The new `cond` types match those used in [`Series.where`](https://github.com/pandas-dev/pandas-stubs/blob/e63d1cbf5087cd7cd934db1c84184cb2caacbac9/pandas-stubs/core/series.pyi#L1420-L1426) and [`DataFrame.where`](https://github.com/pandas-dev/pandas-stubs/blob/e63d1cbf5087cd7cd934db1c84184cb2caacbac9/pandas-stubs/core/frame.pyi#L2309-L2315). Let me know if it's desirable to factor the `cond` type out by moving it to `_typing.pyi`.

### DIFF
--- a/pandas-stubs/core/frame.pyi
+++ b/pandas-stubs/core/frame.pyi
@@ -1833,7 +1833,13 @@ class DataFrame(NDFrame, OpsMixin):
     def lt(self, other, axis: Axis = ..., level: Level | None = ...) -> DataFrame: ...
     def mask(
         self,
-        cond: Series | DataFrame | np.ndarray,
+        cond: (
+            Series
+            | DataFrame
+            | np.ndarray
+            | Callable[[DataFrame], DataFrame]
+            | Callable[[Any], _bool]
+        ),
         other=...,
         *,
         inplace: _bool = ...,
--- a/pandas-stubs/core/groupby/generic.pyi
+++ b/pandas-stubs/core/groupby/generic.pyi
@@ -199,7 +199,7 @@ class DataFrameGroupBy(GroupBy[DataFrame], Generic[ByT]):
         **kwargs,
     ) -> DataFrame: ...
     @overload
-    def apply(  # pyright: ignore[reportOverlappingOverload,reportIncompatibleMethodOverride]
+    def apply(  # pyright: ignore[reportOverlappingOverload]
         self,
         func: Callable[[Iterable], float],
         *args,
--- a/pandas-stubs/core/series.pyi
+++ b/pandas-stubs/core/series.pyi
@@ -1433,7 +1433,13 @@ class Series(IndexOpsMixin[S1], NDFrame):
     ) -> Series[S1]: ...
     def mask(
         self,
-        cond: MaskType,
+        cond: (
+            Series[S1]
+            | Series[_bool]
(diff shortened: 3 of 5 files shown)


Dr-Irv:
> CI failed with:
> 
> > error: Unnecessary "# pyright: ignore" rule: "reportIncompatibleMethodOverride"
> 
> I saw this one locally as well, but wasn't sure if it was due to my configuration. Should we get rid of that rule from the `ignore` hint? The test suite ran fine for me locally after removing it, but haven't investigated what it does / implications of removing it.

I am guessing that pyright has a new version that says that the rule is unnecessary, so you can remove it as part of this PR.  It won't have an effect - we put those in there when we disagree with the type checker, so in this case, they fixed a bug making the ignore statement unnecessary.

georgesittas:
Got it - done.

Dr-Irv (approved):
Thanks @georgesittas

PR #921, opened 2024-05-12
Fix `Styler.apply()` to accept keyword arguments

### LINKED ISSUE #919: No overload variant of "apply" of "Styler" matches argument types "Callable[[Series[int], str], list[str]]", "str", "int"
**Describe the bug**
Following the example in [pandas website](https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.io.formats.style.Styler.apply.html), I defined a function to highlight the maximum of each row. However, `mypy` does not recognize the call signature.

**To Reproduce**
1. Minimal working example
```
from typing import List

import numpy as np
import pandas as pd


def highlight_max(x: 'pd.Series[int]', color: str) -> List[str]:
    """Highlight max."""
    return [f"color: {color}" if val == x.max() else "" for val in x]


df = pd.DataFrame(np.random.randint(0, high=100, size=(10, 3)))
styler = df.style.apply(highlight_max, color='red', axis=1)
styler.to_html("output.html")
```
2. I'm using `mypy`
In terminal
```
$ mypy --explicit-package-bases --strict .
proj.py:14: error: No overload variant of "apply" of "Styler" matches argument types "Callable[[Series[int], str], list[str]]", "str", "int"  [call-overload]
proj.py:14: note: Possible overload variants:
proj.py:14: note:     def apply(self, func: Callable[[Series[Any]], Union[list[Any], Series[Any]]], axis: Union[Literal['index', 0], Literal['columns', 1]] = ..., subset: Optional[Union[_IndexSlice, slice, tuple[slice, ...], list[Any], Index[Any]]] = ..., **kwargs: Any) -> Styler
proj.py:14: note:     def apply(self, func: Callable[[DataFrame], Union[ndarray[Any, dtype[Any]], DataFrame]], axis: None, subset: Optional[Union[_IndexSlice, slice, tuple[slice, ...], list[Any], Index[Any]]] = ..., **kwargs: Any) -> Styler
```

 - OS: Mac OS
 - OS Version 14.4.1 
 - python version: Python 3.9.6
 - version of type checker: mypy 1.8.0 (compiled: yes)
 - version of installed `pandas-stubs`: from `pip list`: `pandas-stubs                      2.2.1.240316`

Dr-Irv:
- [x] Closes #919
- [x] Tests added: 
    - `test_styler.py:test_apply()` has a new test

### DIFF
--- a/pandas-stubs/io/formats/style.pyi
+++ b/pandas-stubs/io/formats/style.pyi
@@ -42,7 +42,14 @@ from pandas.io.formats.style_render import (
 )
 
 class SeriesFunc(Protocol):
-    def __call__(self, series: Series, *args: Any, **kwargs: Any) -> list | Series: ...
+    def __call__(
+        self, series: Series, /, *args: Any, **kwargs: Any
+    ) -> list | Series: ...
+
+class DataFrameFunc(Protocol):
+    def __call__(
+        self, series: DataFrame, /, *args: Any, **kwargs: Any
+    ) -> npt.NDArray | DataFrame: ...
 
 class Styler(StylerRenderer):
     def __init__(
@@ -202,7 +209,7 @@ class Styler(StylerRenderer):
     @overload
     def apply(
         self,
-        func: SeriesFunc,
+        func: SeriesFunc | Callable[[Series], list | Series],
         axis: Axis = ...,
         subset: Subset | None = ...,
         **kwargs: Any,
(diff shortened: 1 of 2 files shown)


pandas-stubs/io/formats/style.pyi
```diff
@@ -40,6 +41,16 @@ from pandas.io.formats.style_render import (
     Subset,
 )
 
+class SeriesFunc(Protocol):
```
twoertwein, line 44: I would prefer renaming it to `_SeriesFunc` or put it in `_typing` so that users don't think they can import it.

twoertwein (approved):

PR #922, opened 2024-05-13
import from_dataframe

Dr-Irv:
For some reason, `mypy` has trouble figuring out the type of `pd.api.interchange.from_dataframe`, but if you import it, the error goes away.


twoertwein (approved):

twoertwein:
Thank you @Dr-Irv !

PR #923, opened 2024-05-14
add reduce/accumulate methods stubs + minimal tests, resolves #850

### LINKED ISSUE #850: Implement ExtensionArray _accumulate and _reduce
**Describe the bug**
The stubs for ExtensionArray (in `pandas-stubs/core/arrays/base.pyi`) does not provide type signatures for _accumulate and _reduce.  To properly add typing information to the Pint-Pandas project, these need to be defined.

**To Reproduce**
1. Minimal Runnable Example:
```
import numpy as np
import pandas as pd
from pandas.arrays import IntegerArray
from pandas.api.extensions import ExtensionArray

_data: ExtensionArray = IntegerArray(values=np.array([1, 2, 3], dtype=int), mask=np.array([True, True, True], dtype=bool))
if isinstance(_data, ExtensionArray):
    reveal_type(_data)
    reveal_type(_data._accumulate)
    reveal_type(_data._reduce)
```
2.  Using `mypy`
3.  Show the error message received from that type checker while checking your example.
```
(pint-dev) % pre-commit run mypy --files foo.py
- hook id: mypy
- duration: 1.41s
- exit code: 1

foo.py:9: note: Revealed type is "pandas.core.arrays.base.ExtensionArray"
foo.py:10: error: "ExtensionArray" has no attribute "_accumulate"  [attr-defined]
foo.py:10: note: Revealed type is "Any"
foo.py:11: error: "ExtensionArray" has no attribute "_reduce"  [attr-defined]
foo.py:11: note: Revealed type is "Any"
```

Note that running the script in python works, because it uses actual Pandas code, not Pandas-Stubs:
```
(pint-dev) % python foo.py
Runtime type is 'IntegerArray'
Runtime type is 'method'
Runtime type is 'method'
```

 - OS: Mac OS
 - OS Version  14.1.2
 - python 3.11.4
 - mypy 1.8.0
 - version of installed `pandas-stubs`: 2.1.4.231227

**Additional context**

mutricyl:
- [x] Closes #850 
   - `test_ExtensionArray_reduce_accumulate` added in `test_extension.py`


Dr-Irv (approved):
Thanks @mutricyl

Issue #925, opened 2024-05-15
Stack method - Unexpected keyword argument


rinzool:
**Describe the bug**
Method `.stack` of dataframe is not fully typed, it misses argument `sort` and `future_stack`

**To Reproduce**
1. Provide a minimal runnable `pandas` example that is not properly checked by the stubs.
```python
import pandas as pd
df = pd.DataFrame([1, 2, 3])
df.stack(future_stack=True)
df.stack(sort=True)
```
2. Indicate which type checker you are using (`mypy` or  `pyright`).
I am using mypy (1.10.0)
3. Show the error message received from that type checker while checking your example.
```
ex.py:3: error: Unexpected keyword argument "future_stack" for "stack" of "DataFrame"  [call-arg]
ex.py:4: error: Unexpected keyword argument "sort" for "stack" of "DataFrame"  [call-arg]
Found 2 errors in 1 file (checked 1 source file)
```

**Please complete the following information:**
 - OS: ArchLinux
 - python version: 3.10.14
 - version of type checker: 1.10.0
 - version of installed `pandas-stubs` 2.2.2.240514

**Additional context**
I'll be glad to open a PR to add those two arguments with some tests if that's ok for you, but I'm not entirely sure about what is the philosophy about pandas-stubs here. Since they are new arguments in 2.1, could it be bad to add those arguments in pandas stubs for someone who sill uses 2.0?

Dr-Irv:
> I'll be glad to open a PR to add those two arguments with some tests if that's ok for you, but I'm not entirely sure about what is the philosophy about pandas-stubs here. Since they are new arguments in 2.1, could it be bad to add those arguments in pandas stubs for someone who sill uses 2.0?

Not a problem, because we only support the latest version of `pandas` in the stubs.  Also, the arguments are optional, so anyone using an earlier version of pandas wouldn't be using them anyway.

Thanks for the report.  Given that `future_stack` is incompatible with `sort` and `dropna`, you'd need overloads to show the possible combinations of the arguments.

PR with test welcome

PR #926, opened 2024-05-15
fix(frame): add arguments for stack method

rinzool:
- [X] Closes https://github.com/pandas-dev/pandas-stubs/issues/925 


tests/test_frame.py
```diff
@@ -2256,6 +2256,20 @@ def test_frame_stack() -> None:
             ),
             pd.Series,
         )
+        check(
+            assert_type(
+                df_multi_level_cols2.stack(0, future_stack=False),
+                Union[pd.DataFrame, "pd.Series[Any]"],
+            ),
+            pd.DataFrame,
+        )
```
Dr-Irv, line 2265: Can you surround this with `if PD_LTE_22:`  as the `future_stack` parameter is being removed in the future, and it will allow the nightly tests that get run upon a merge to succeed?

Dr-Irv (changes requested):

tests/test_frame.py
rinzool: Sure, I just added the condition 👍

tests/test_frame.py
```diff
@@ -2256,6 +2256,21 @@ def test_frame_stack() -> None:
             ),
             pd.Series,
         )
+        check(
+            assert_type(
+                df_multi_level_cols2.stack(0, dropna=True, sort=True),
+                Union[pd.DataFrame, "pd.Series[Any]"],
+            ),
+            pd.DataFrame,
+        )
```
Dr-Irv, line 2265: Looks like you have to include the `if PD_LTE_22` on this test as well, as the `dropna` argument will also change behavior in 3.0, so have one `if` statement surround both of the new tests you added.

To see this, do `poe pytest --nightly` on your current branch, and when you make the fix the problem will go away.

rinzool: Oh ok nice I didn't notice the `--nightly` option to test with pandas 3! It seems to pass now

Dr-Irv (approved):
Thanks @rinzool

PR #930, opened 2024-05-26
pandas 2.0 removals for NDFrame and subclasses

hamdanal:
- [x] Work on #624
- [ ] Tests added

Reference: https://pandas.pydata.org/docs/dev/whatsnew/v2.0.0.html#removal-of-prior-version-deprecations-changes


tests/test_plotting.py
```diff
@@ -486,22 +486,23 @@ def test_plot_scatter(close_figures) -> None:
...
-            IRIS_DF.plot.scatter(x="SepalLength", y="SepalWidth", subplots=True),
-            npt.NDArray[np.object_],
-        ),
-        np.ndarray,
-    )
-    check(
-        assert_type(
-            IRIS_DF.plot(
-                x="SepalLength", y="SepalWidth", subplots=True, kind="scatter"
+    with pytest.warns(UserWarning):
```
Dr-Irv, line 489: I don't see why this is necessary.

Dr-Irv (changes requested):
All looks good - don't understand the change in the tests.

tests/test_plotting.py
hamdanal: See the failing CI tests https://github.com/pandas-dev/pandas-stubs/actions/runs/9244454746/job/25429762574. A new version of matplotlib is causing new test failures with the warning.

Dr-Irv:
So the issue with needing `pytest.warns()` is that `matplotlib` got updated to version 3.9.0, which causes the warning.  See https://github.com/pandas-dev/pandas/issues/58851 for a bug report that I filed as a result.

Rather than add the `pytest.warns()` here and in your other PR's, can you change `pyproject.toml` to reflect a restriction on `matplotlib`:

```
matplotlib = "3.8.4"
```

And change the `pyright` version to be `>= 1.1.364` since that was used in your tests and why you had to make that change for `aggregate()`

You'll have to do this in all 4 PR's

Dr-Irv (approved):
Thanks @hamdanal

PR #943, opened 2024-06-25
Support numpy 2.0 in tests

Dr-Irv:
CI is now pulling in numpy 2.0, so this removes the numpy deprecated `np.float_` and changes it to `np.double`, which works with older and newer versions of  numpy.

In addition, there are a number of other changes where `numpy` has removed certain aliases for dtypes, so removed those from the tests.

Set up the CI to use numpy 1.x with python <=3.11, and numpy 2.0 with python=3.12 so both versions of numpy are tested.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -8,22 +8,20 @@ readme = "README.md"
 homepage = "https://pandas.pydata.org"
 repository = "https://github.com/pandas-dev/pandas-stubs"
 classifiers = [
-    "Development Status :: 5 - Production/Stable",
-    "Environment :: Console",
-    "Intended Audience :: Science/Research",
-    "License :: OSI Approved :: BSD License",
-    "Operating System :: OS Independent",
-    "Programming Language :: Python",
-    "Programming Language :: Python :: 3",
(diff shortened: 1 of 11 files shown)


pyproject.toml
```diff
@@ -114,18 +116,38 @@ script = "scripts.test:test(dist=True, type_checker='pyright')"
...
+  "FUTURE",
+  "STDLIB",
+  "THIRDPARTY",
+  "PRE_LIBS",
+  "PRE_CORE",
+  "DTYPES",
+  "FIRSTPARTY",
+  "POST_CORE",
+  "LOCALFOLDER",
+]
```
twoertwein, line 150: Do you use a formatting tool for these changes?

twoertwein (approved):

pyproject.toml
Dr-Irv: I used VSCode to do the editing, and I have a setting to tell it to auto-format files after saving, so I'm guessing that's what happened.

Will you merge?

twoertwein:
Thank you @Dr-Irv !

Issue #944, opened 2024-06-27
pd.notna does not work with pyright strict mode.


JanEricNitschke:
As discussed [here](https://github.com/microsoft/pyright/discussions/8250), `pd.notna` (and also `pd.isna`) do not work with pyright strict mode.

This is due to the missing type parameters for `Series`, `list`, `Index` and `ArrayLike` in these two overloads

```
@overload
def notna(obj: Series) -> Series[bool]: ...
@overload
def notna(obj: Index | list | ArrayLike) -> npt.NDArray[np.bool_]: ...
```

The fix would be to add the required type parameters (or Any) to them. However i am not sure if/what bounds are needed for these type parameters.

JanEricNitschke:
Closed by: https://github.com/pandas-dev/pandas-stubs/pull/945

PR #942, opened 2024-06-24
Fix signature of Series.map()

JanEricNitschke:
- [x] Closes https://github.com/pandas-dev/pandas-stubs/issues/941
- [x] Tests added: Please use `assert_type()` to assert the type of any return value

Only added the proposed changes so far. Will try to work on tests as soon as possible.

Does not seem to break any existing tests so far at least.

(I think the failures are all due to numpy 2.0 and separate from my changes)

PR #945, opened 2024-06-28
Add type parameters to generics in isna and notna

JanEricNitschke:
- [x] Closes https://github.com/pandas-dev/pandas-stubs/issues/944

This is to fix issues with pyright strict mode which complains about partially unknown types.

I verified that [this](https://github.com/pandas-dev/pandas-stubs/blob/main/tests/test_pandas.py#L319) test fails in strict mode without these changes and passes with them.

There are still some points i am unsure about:
 - Whether to use `Any` or an (unbounded?) `TypeVar` (or maybe i should use `Incomplete`?)
 - I had to split the `ArrayLike` annotation into its components to add the type parameters to the ndarray. While i did also change it in the `pandas-stubs/_typing.pyi`, it was still importing them from the actual pandas.

### DIFF
--- a/pandas-stubs/core/dtypes/missing.pyi
+++ b/pandas-stubs/core/dtypes/missing.pyi
@@ -10,12 +10,12 @@ from pandas import (
     Index,
     Series,
 )
-from pandas.core.arrays.base import ExtensionArray
 from typing_extensions import TypeGuard
 
 from pandas._libs.missing import NAType
 from pandas._libs.tslibs import NaTType
 from pandas._typing import (
+    ArrayLike,
     Scalar,
     ScalarT,
 )
@@ -28,9 +28,7 @@ def isna(obj: DataFrame) -> DataFrame: ...
 @overload
 def isna(obj: Series[Any]) -> Series[bool]: ...
 @overload
-def isna(
-    obj: Index[Any] | list[Any] | ExtensionArray | np.ndarray[Any, Any]
-) -> npt.NDArray[np.bool_]: ...
+def isna(obj: Index[Any] | list[Any] | ArrayLike) -> npt.NDArray[np.bool_]: ...
 @overload
 def isna(
     obj: Scalar | NaTType | NAType | None,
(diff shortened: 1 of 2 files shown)


JanEricNitschke:
Closes https://github.com/pandas-dev/pandas-stubs/issues/944

JanEricNitschke:
I [checked](https://github.com/microsoft/pyright/discussions/8257#discussioncomment-9904375) that the `Any` is probably the way to go unless we already know some bounds that need to be respected here.

Only open question would then be the handling of the `ArrayLike`. The best approach would be if we could modify (which i did) and pull in the version from the stubs. But i dont know how, do i explicitely import from pandas-stubs? But it seems to work with importing from pandas in the [previous pr](https://github.com/pandas-dev/pandas-stubs/pull/942/files#diff-6f90f15b437974706a3c0a75636efd6d163a42d4a786bc5513a5069b1040f031R96).

pandas-stubs/core/dtypes/missing.pyi
```diff
@@ -23,9 +26,11 @@ isneginf_scalar = ...
 @overload
 def isna(obj: DataFrame) -> DataFrame: ...
 @overload
-def isna(obj: Series) -> Series[bool]: ...
+def isna(obj: Series[Any]) -> Series[bool]: ...
 @overload
-def isna(obj: Index | list | ArrayLike) -> npt.NDArray[np.bool_]: ...
+def isna(
+    obj: Index[Any] | list[Any] | ExtensionArray | np.ndarray[Any, Any]
```
Dr-Irv, line 32: Given that you redefined `ArrayLike` in `_typing.pyi`, couldn't you just use `ArrayLike` here?

pandas-stubs/core/dtypes/missing.pyi
```diff
@@ -36,9 +41,11 @@ isnull = isna
 @overload
 def notna(obj: DataFrame) -> DataFrame: ...
 @overload
-def notna(obj: Series) -> Series[bool]: ...
+def notna(obj: Series[Any]) -> Series[bool]: ...
 @overload
-def notna(obj: Index | list | ArrayLike) -> npt.NDArray[np.bool_]: ...
+def notna(
+    obj: Index[Any] | list[Any] | ExtensionArray | np.ndarray[Any, Any]
```
Dr-Irv, line 47: Same comment as above - use `ArrayLike`

Dr-Irv (changes requested):

pandas-stubs/core/dtypes/missing.pyi
JanEricNitschke: Done

pandas-stubs/core/dtypes/missing.pyi
JanEricNitschke: Done

JanEricNitschke:
Just double checked and you are right. I was getting some weird messages from vscode there. I guess it was an issue with that.

I ran pyright over just that one test and it did pick up the ArrayLike from the stubs _typing.

Dr-Irv (approved):
thanks @JanEricNitschke

Issue #948, opened 2024-06-28
`Series.astype` is partially unknown.


JanEricNitschke:
Same as the other ones.

Specifically these two overloads have the problem:

```
    @overload
    def astype(
        self,
        dtype: CategoryDtypeArg,
        copy: _bool = ...,
        errors: IgnoreRaise = ...,
    ) -> Series: ...
    @overload
    def astype(
        self,
        dtype: ObjectDtypeArg | VoidDtypeArg | ExtensionDtype | DtypeObj,
        copy: _bool = ...,
        errors: IgnoreRaise = ...,
    ) -> Series: ...
```

Is there anything more concrete that can be done here besides `Series[Any]`?

Possibly a type parameter with the respective bound and return a `Series[T]`?

JanEricNitschke:
According to the [tests](https://github.com/pandas-dev/pandas-stubs/blob/main/tests/test_series.py#L2172) it should probably be `Series[object]` (thats what the runtime checks do) or `Series[Any]` thats what they [type checking tests](https://github.com/pandas-dev/pandas-stubs/blob/main/tests/test_series.py#L2593) do.

Dr-Irv:
The results that are not `Any` have to be part of `S1`.  I think if you use `Series[CategoricalDtype]` in the first one, and `Series[Any]` in the second, it should be OK.

PR #949, opened 2024-06-28
Add type parameter to remaining Series.astype overloads.

JanEricNitschke:

Closes https://github.com/pandas-dev/pandas-stubs/issues/948

Verified that some of the test statements [here](https://github.com/pandas-dev/pandas-stubs/blob/main/tests/test_series.py#L2191) pass pyright strict with this change that do not do so without.

Tested using `object` instead of `Any` as that is the runtime behaviour, but that did not work due to the parameter bounds.

Other option (maybe better?) would be to explicitely return as series matching the specified parameter? But i think that does not work either because the [bounds of the type parameter for series](https://github.com/pandas-dev/pandas-stubs/blob/main/pandas-stubs/_typing.pyi#L533) are more restrictive than the input arguments for `astype`.


pandas-stubs/core/series.pyi
```diff
@@ -1320,14 +1320,14 @@ class Series(IndexOpsMixin[S1], NDFrame):
         dtype: CategoryDtypeArg,
         copy: _bool = ...,
         errors: IgnoreRaise = ...,
-    ) -> Series: ...
+    ) -> Series[Any]: ...
```
Dr-Irv, line 1323:
```suggestion
    ) -> Series[CategoricalDtype]: ...
```
If this change is failing a test, then modify the test.

pandas-stubs/core/series.pyi
```diff
@@ -1320,14 +1320,14 @@ class Series(IndexOpsMixin[S1], NDFrame):
...
     @overload
     def astype(
         self,
         dtype: ObjectDtypeArg | VoidDtypeArg | ExtensionDtype | DtypeObj,
         copy: _bool = ...,
         errors: IgnoreRaise = ...,
-    ) -> Series: ...
+    ) -> Series[Any]: ...
```
Dr-Irv, line 1330: This is correct.  Any generic parameter to `Series` needs to be in the `TypeVar` `S1` that is declared in `_typing.pyi`

Dr-Irv (changes requested):

pandas-stubs/core/series.pyi
JanEricNitschke: Made that change and made one of the test that previously expected the `Series[Any]` fit with this.

Also double checked that pyright strict that not throw any new warnings due to this.

pandas-stubs/core/series.pyi
```diff
@@ -1320,14 +1320,14 @@ class Series(IndexOpsMixin[S1], NDFrame):
         dtype: CategoryDtypeArg,
         copy: _bool = ...,
         errors: IgnoreRaise = ...,
-    ) -> Series: ...
+    ) -> Series[CategoryDtypeArg]: ...
```
Dr-Irv, line 1323: This needs to be `CategoryDtype` as it is value in `S1`

tests/test_series.py
```diff
@@ -2581,8 +2581,8 @@ def test_astype_categorical(cast_arg: CategoryDtypeArg, target_type: type) -> No
 
     if TYPE_CHECKING:
         # pandas category
-        assert_type(s.astype(pd.CategoricalDtype()), "pd.Series[Any]")
-        assert_type(s.astype("category"), "pd.Series[Any]")
+        assert_type(s.astype(pd.CategoricalDtype()), "pd.Series[CategoryDtypeArg]")
+        assert_type(s.astype("category"), "pd.Series[CategoryDtypeArg]")
```
Dr-Irv, line 2585: Should be `CategoryDtype`

Dr-Irv (changes requested):

pandas-stubs/core/series.pyi
JanEricNitschke: Made the change

tests/test_series.py
JanEricNitschke: Here too

Dr-Irv (approved):
thanks @JanEricNitschke

Issue #950, opened 2024-06-29
`df.drop` signature does not capture all constraints


JanEricNitschke:
I was looking at `df.drop` to make it work with pyright strict mode.

The type hints are these:

```
    @overload
    def drop(
        self,
        labels: Hashable | Sequence[Hashable] | Index = ...,
        *,
        axis: Axis = ...,
        index: Hashable | Sequence[Hashable] | Index = ...,
        columns: Hashable | Sequence[Hashable] | Index = ...,
        level: Level | None = ...,
        inplace: Literal[True],
        errors: IgnoreRaise = ...,
    ) -> None: ...
    @overload
    def drop(
        self,
        labels: Hashable | Sequence[Hashable] | Index = ...,
        *,
        axis: Axis = ...,
        index: Hashable | Sequence[Hashable] | Index = ...,
        columns: Hashable | Sequence[Hashable] | Index = ...,
        level: Level | None = ...,
        inplace: Literal[False] = ...,
        errors: IgnoreRaise = ...,
    ) -> DataFrame: ...
    @overload
    def drop(
        self,
        labels: Hashable | Sequence[Hashable] | Index = ...,
        *,
        axis: Axis = ...,
        index: Hashable | Sequence[Hashable] | Index = ...,
        columns: Hashable | Sequence[Hashable] | Index = ...,
        level: Level | None = ...,
        inplace: bool = ...,
        errors: IgnoreRaise = ...,
    ) -> DataFrame | None: ...
```

However, if "labels" is specified, neither index, nor columns should be.

Additionally, one of "labels", "index", "columns" has to be specified.

Dr-Irv:
Seems like this isn't documented, although it does make sense.

PR #951, opened 2024-06-29
Expand `drop()` overloads to encode more argument restriction and add type parameters to `df.drop()` signature.

JanEricNitschke:

This is again to increase compatibility with pyright strict mode. I verified that the df.drop [tests](https://github.com/pandas-dev/pandas-stubs/blob/main/tests/test_frame.py#L329) do not pass pyright strict without this change but do with it.

Also (maybe) closes https://github.com/pandas-dev/pandas-stubs/issues/950.

This is done by adding a bunch of overloads that specify the allowed combinations in more detail.

For that i added a test that fails without this change and passes with it,

However i was not able to specify the full functionality using the overloads. Specifically, explicit "None" arguments are allowed by the signature, but cause errors at runtime. This is mostly because `None` is Hasheable.
This could be improved substantially in correctness and conciseness of the hints if it were possible to type hint the general signature using `Non-None-Hasheable | None` and then split these up in the overloads.


JanEricNitschke:
One more thing that could be done would be to add overloads for the cases where the `None`s are explicitely specified and mark that as deprecated and `Never` return. But i dont know how clear that would be.

pandas-stubs/core/generic.pyi
```diff
@@ -305,10 +305,22 @@ class NDFrame(indexing.IndexingMixin):
     @overload
     def drop(
```
Dr-Irv, line 306: I think the signatures for `DataFrame.drop()` and `Series.drop()` are all the same, so that means we can delete the definitions of `drop()` from `frame.pyi` and `series.pyi` and then put all of your changes in `generic.pyi` , and return `Self` in there.

I believe that the use of `HashableT1`, `HashableT2`, etc . in `series.pyi` for `drop()` were unnecessary, so let's use this opportunity to consolidate.

Dr-Irv (changes requested):

pandas-stubs/core/generic.pyi
JanEricNitschke: Currently generic has `Hashable | Sequence[Hashable]` while frame has `Hashable | Sequence[Hashable] | Index[Any]` and series has `Hashable | list[HashableT2] | Index`.

So yeah, i think we can just use `Hashable | Sequence[Hashable] | Index[Any]` in generic and remove the overwrites in frame and series.

JanEricNitschke: added the `Index[Any]` to the signature in generic and removed the ones in series and frame.

Dr-Irv (approved):
Thanks @JanEricNitschke

Issue #952, opened 2024-06-29
Type of "apply" is partially unknown


JanEricNitschke:
I am again interested in getting some pyright strict warnings fixed that i get due to a lack of type parameters. This time for `df.apply`. However this already has a lot of [overloads](https://github.com/pandas-dev/pandas-stubs/blob/main/pandas-stubs/core/frame.pyi#L1249) and i do not yet fully understand all of them.

Firstly i do not quite understand this one:

```
   @overload
    def apply(
        self,
        f: Callable[..., Mapping],
        axis: AxisIndex = ...,
        raw: _bool = ...,
        result_type: None = ...,
        args=...,
        **kwargs,
    ) -> Series: ...
```

that would be something like 

```
def returns_dict(x: pd.Series) -> dict[str, int]:
    return {"col4": 7, "col5": 8}

df = pd.DataFrame(data={"col1": [1, 2], "col2": [3, 4], "col3": [5, 6]})
print(df.apply(returns_dict))
```

Which just gives me an object series.

```
col1    {'col4': 7, 'col5': 8}
col2    {'col4': 7, 'col5': 8}
col3    {'col4': 7, 'col5': 8}
```

is that really the intended thing here? In that case that one could just go from taking a `Mapping -> Series` to `Mapping[Any, Any] -> Series[Any]`.

All the ones that just return a dataframe should probably also have `Series[Any]` (or the equivalent for other  generics).

For all overloads that take a 

```
ListLikeExceptSeriesAndStr = TypeVar(
    "ListLikeExceptSeriesAndStr", MutableSequence, np.ndarray, tuple, Index
)
```

and return a series i think one could probably try to parametrize that further and have the series again be `Series[S1]`. But bounds in a typevar cant be generic themselves.

Dr-Irv:
We can only use `Series[S1]` as a return type if we know that the input type is also `Series[S1]`.  My hypothesis here is that wherever we just use `Series` in the stubs without specifying the associated type, we could just replace that with `Series[Any]`, and then you'd get strict mode to work better.

JanEricNitschke:
Yeah, replacing every `Unknown` that pyright reports with `Any` would fix most issues with strict mode. I am currently just doing that bit by bit as i am encountering them in my own project and also checking if there are any spots missed where there could be a more constrained one.

Dr-Irv:
[...]

PR #953, opened 2024-06-29
Add explicit type parameters to arguments of df.apply

JanEricNitschke:

Adresses: https://github.com/pandas-dev/pandas-stubs/issues/952

Currently just added the explicit `Any` annotations everywhere, moved `ListLikeExceptSeriesAndStr` from a TypeVar to a Union and verified that more tests pass pyright strict mode.

However i still have a confusion about the overloads with `Mapping`. Isnt that just any generic object?

### DIFF
--- a/pandas-stubs/_typing.pyi
+++ b/pandas-stubs/_typing.pyi
@@ -490,8 +490,8 @@ Axis: TypeAlias = AxisIndex | AxisColumn
 DtypeNp = TypeVar("DtypeNp", bound=np.dtype[np.generic])
 KeysArgType: TypeAlias = Any
 ListLike = TypeVar("ListLike", Sequence, np.ndarray, Series, Index)
-ListLikeExceptSeriesAndStr = TypeVar(
-    "ListLikeExceptSeriesAndStr", MutableSequence, np.ndarray, tuple, Index
+ListLikeExceptSeriesAndStr: TypeAlias = (
+    MutableSequence[Any] | np.ndarray[Any, Any] | tuple[Any, ...] | Index[Any]
 )
 ListLikeU: TypeAlias = Sequence | np.ndarray | Series | Index
 ListLikeHashable: TypeAlias = (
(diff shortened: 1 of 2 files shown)


Dr-Irv (approved):
thanks @JanEricNitschke

Dr-Irv:
> However i still have a confusion about the overloads with `Mapping`. Isnt that just any generic object?

I don't understand your question here.

JanEricNitschke:
Here for example:

```
    @overload
    def apply(
        self,
        f: Callable[..., ListLikeExceptSeriesAndStr | Series[Any] | Mapping[Any, Any]],
        axis: Axis = ...,
        raw: _bool = ...,
        args: Any = ...,
        *,
        result_type: Literal["expand"],
        **kwargs: Any,
```

If i do 

```
def returns_dict(x: pd.Series) -> dict[str, int]:
    return {"col4": 7, "col5": 8}

df = pd.DataFrame(data={"col1": [1, 2], "col2": [3, 4], "col3": [5, 6]})
print(df.apply(returns_dict))

col1    {'col4': 7, 'col5': 8}
col2    {'col4': 7, 'col5': 8}
col3    {'col4': 7, 'col5': 8}
```

I just get a series where each entry is this dict object. The mappings there just made me feel like it should be something like `Mapping[S1, S1] -> Series[S2]` or somethinglike that, but it `Mapping -> Series[Mapping]`

Dr-Irv:
> I just get a series where each entry is this dict object. The mappings there just made me feel like it should be something like `Mapping[S1, S1] -> Series[S2]` or somethinglike that, but it `Mapping -> Series[Mapping]`

You could do `Mapping[Any, S1]` and have it return `Series[S1]`, but you'd need to then  have to separate the overloads for `Mapping` from the existing ones, and also you'd have to add another overload of `Mapping[Any, Any] -> Series[Any]` as well.

The key idea in the stubs is that we define `S1` to be the "regular" types of things that we see in `Series`, and where it is useful to make a differentiation.  While it is possible to put any object into a `Series`, we didn't include `object` in `S1`.  That might be a change to make for cases like these.  You'd possibly end up with overlapping overloads that mypy won't be happy with, but we can just `#type: ignore` statements for that.

PR #954, opened 2024-06-29
 Setup optional CI jobs using pyright strict mode

JanEricNitschke:
Setup a optional CI job that uses full pyright strict mode by taking its settings from `pyrightconfig-strict.json`.

Addresses the other points made here: https://github.com/pandas-dev/pandas-stubs/issues/952

### DIFF
--- a/.github/workflows/optional.yml
+++ b/.github/workflows/optional.yml
@@ -4,7 +4,6 @@ on:
   push:
     branches:
       - main
-      - strict-ci
   workflow_dispatch:
 
 jobs:
--- a/.github/workflows/optional.yml
+++ b/.github/workflows/optional.yml
@@ -4,6 +4,7 @@ on:
   push:
     branches:
       - main
+      - strict-ci
   workflow_dispatch:
 
 jobs:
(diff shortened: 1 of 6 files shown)


JanEricNitschke:
Did a test run here: https://github.com/JanEricNitschke/pandas-stubs/actions/runs/9726191446/job/26844350295

Dr-Irv (approved):
thanks @JanEricNitschke 

It looks like there is a LOT of work to do to make the stubs work in strict mode.

JanEricNitschke:
Yeah, it is mostly missing type hints or type parameters. Would it make sense to enable the pyright_strict test also in PRs, currently i made it like the others where it only runs on commits to main and not on PR creation like the other optionals.

Might make sense to also have it run here to see the number of errors getting lower on changes?

Dr-Irv:
> Yeah, it is mostly missing type hints or type parameters. Would it make sense to enable the pyright_strict test also in PRs, currently i made it like the others where it only runs on commits to main and not on PR creation like the other optionals.
> 
> Might make sense to also have it run here to see the number of errors getting lower on changes?

I'm hesitant to do that, because then every PR will show a failure, which can be confusing to contributors.

PR #955, opened 2024-06-30
Trying to make pd.concat work with pyright strict mode.

JanEricNitschke:
Steps are:
- [x] Add explicit anys
- [ ] Add tests that should work already
- [ ] Add explicit S1 overloads

The first step is done. The second step does not work because of mypy. Those test failures also happen without the explicit `Any`.

### DIFF
--- a/pandas-stubs/core/reshape/concat.pyi
+++ b/pandas-stubs/core/reshape/concat.pyi
@@ -4,6 +4,7 @@ from collections.abc import (
     Sequence,
 )
 from typing import (
+    Any,
     Literal,
     overload,
 )
@@ -39,7 +40,7 @@ def concat(
 ) -> DataFrame: ...
 @overload
 def concat(  # type: ignore[overload-overlap] # pyright: ignore[reportOverlappingOverload]
-    objs: Iterable[Series] | Mapping[HashableT1, Series],
+    objs: Iterable[Series[Any]] | Mapping[HashableT1, Series[Any]],
     *,
     axis: AxisIndex = ...,
     join: Literal["inner", "outer"] = ...,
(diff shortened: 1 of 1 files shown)


Dr-Irv:
> The first step is done. The second step does not work because of mypy. Those test failures also happen without the explicit `Any`.

I made a copy of your PR,  and played around with it.  I think this is a bug in `mypy` for which there is no easy workaround.

First, one change I did is to change the last overload of `__new__()` in `series.pyi` to return `Series[Any]` rather than `Series`.  That will make `mypy` recognize that `series2` declared at https://github.com/JanEricNitschke/pandas-stubs/blob/5a53dfca1cfb04263d136f5f2fda3cafcfa15b5d/tests/test_pandas.py#L59

to be type `Union[list[None | pd.Series[Any]]`.   Then if I remove the last overload for `pd.concat()`, `mypy` will get past the first test that uses `series2`.  But removing that overload causes other tests to fail.

So the issue here is that we have to return different results dependent on whether `DataFrame` is present in the `Iterable` passed to `concat`.   So we need that last overload, but when that overload is present, then `mypy` isn't matching the test `pd.concat([None, series2])` to the correct overload.

This is one of the challenges with our stubs - making them work with `pyright` and `mypy`, when the two type checkers have different behaviors in certain cases, and, in some cases, it is due to `mypy` bugs.

JanEricNitschke:
Thanks for checking this out. Then i guess the best we can do here is just taking the first commit for now and keeping it at that until the behaviour between mypy and pyright is consistent or at least can be made compatible here.

JanEricNitschke:
I removed the second commit that added the tests and just kept it add the basic addition of the `Any` type parameters.

Dr-Irv (approved):
thanks @JanEricNitschke

PR #957, opened 2024-07-01
add Series.case_when

Dr-Irv:
`Series.case_when()` added in pandas 2.2, so this adds it to the stubs.

  - `test_series.py:test_case_when()`


twoertwein (approved):

twoertwein:
Thanks @Dr-Irv !

PR #959, opened 2024-07-12
Switch to_timedelta argument type from list to Sequence

### LINKED ISSUE #956: Switch `to_timedelta` argument type from `list` to `Sequence`
**Describe the bug**
`to_timedelta` uses a list of a union argument type.
This will raise false positive errors when passing a list of a single type because list is invariant.

**To Reproduce**
```python
import pandas as pd

pd.to_timedelta(["1 day"])
```

produces:
> error: Argument 1 to "to_timedelta" has incompatible type "list[str]"; expected "Union[Sequence[Union[float, timedelta]], list[Union[str, float, timedelta]], tuple[Union[str, float, timedelta], ...], range, Union[ExtensionArray, ndarray[Any, Any]], Index[Any]]"  [arg-type]
> note: "List" is invariant -- see https://mypy.readthedocs.io/en/stable/common_issues.html#variance
> note: Consider using "Sequence" instead, which is covariant

 - OS: Win
 - OS Version 10
 - python 3.9.7
 - version of type checker mypy 1.8.0
 - version of installed `pandas-stubs` 2.2.2.240603

loicdiridollou:
- [x] Closes #956

Should fix the issue in #956 with allowing the sequence of string type, I was not sure about the `assert_type` step above, please feel free to guide me on that one.

### DIFF
--- a/pandas-stubs/core/tools/timedeltas.pyi
+++ b/pandas-stubs/core/tools/timedeltas.pyi
@@ -32,7 +32,6 @@ def to_timedelta(
 def to_timedelta(
     arg: (
         Sequence[str | float | timedelta]
-        | list[str | float | timedelta]
         | tuple[str | float | timedelta, ...]
         | range
         | ArrayLike
--- a/pandas-stubs/core/tools/timedeltas.pyi
+++ b/pandas-stubs/core/tools/timedeltas.pyi
@@ -31,7 +31,7 @@ def to_timedelta(
 @overload
 def to_timedelta(
     arg: (
-        Sequence[float | timedelta]
+        Sequence[str | float | timedelta]
         | list[str | float | timedelta]
         | tuple[str | float | timedelta, ...]
         | range
--- a/pandas-stubs/core/tools/timedeltas.pyi
+++ b/pandas-stubs/core/tools/timedeltas.pyi
@@ -32,7 +32,7 @@ def to_timedelta(
 @overload
 def to_timedelta(
     arg: (
-        SequenceNotStr[str]
+        SequenceNotStr
         | Sequence[float | timedelta]
         | tuple[str | float | timedelta, ...]
         | range
--- a/pandas-stubs/core/tools/timedeltas.pyi
+++ b/pandas-stubs/core/tools/timedeltas.pyi
@@ -14,6 +14,7 @@ from pandas._libs.tslibs.timedeltas import TimeDeltaUnitChoices
 from pandas._typing import (
     ArrayLike,
     RaiseCoerce,
+    SequenceNotStr,
 )
 
 @overload
@@ -31,7 +32,8 @@ def to_timedelta(
 @overload
 def to_timedelta(
     arg: (
-        Sequence[str | float | timedelta]
+        SequenceNotStr[str]
+        | Sequence[float | timedelta]
         | tuple[str | float | timedelta, ...]
         | range
         | ArrayLike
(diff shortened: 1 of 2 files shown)


pandas-stubs/core/tools/timedeltas.pyi
```diff
@@ -31,7 +31,7 @@ def to_timedelta(
 @overload
 def to_timedelta(
     arg: (
-        Sequence[float | timedelta]
+        Sequence[str | float | timedelta]
         | list[str | float | timedelta]
```
Dr-Irv, line 35: Combine these to one item of `Sequence[str | float | timedelta]`

tests/test_timefuncs.py
```diff
@@ -140,6 +140,15 @@ def test_timedelta_series_arithmetic() -> None:
     r4: pd.TimedeltaIndex = tds1 / 10.2
 
 
+def test_timedelta_series_string() -> None:
+    tds1: pd.TimedeltaIndex = pd.to_timedelta(["1 day"])
+    td1: pd.Timedelta = pd.Timedelta("2 days")
+    r1: pd.TimedeltaIndex = tds1 + td1
+    r2: pd.TimedeltaIndex = tds1 - td1
+    r3: pd.TimedeltaIndex = tds1 * 4.3
+    r4: pd.TimedeltaIndex = tds1 / 10.2
```
Dr-Irv, line 149: You don't need these tests here.

You do need to do the test as follows:
```python
check(assert_type(pd.to_timedelta(["1 day"]), pd.TimedeltaIndex), pd.TimedeltaIndex)
```

Dr-Irv (changes requested):
Please make changes as indicated, and then I will star the CI process to see if there are any other issues.

pandas-stubs/core/tools/timedeltas.pyi
loicdiridollou: Indeed Sequence is encompassing list so might as well.

tests/test_timefuncs.py
loicdiridollou: Thanks for the feedback and the clarity!

pandas-stubs/core/tools/timedeltas.pyi
```diff
@@ -31,8 +31,7 @@ def to_timedelta(
 @overload
 def to_timedelta(
     arg: (
-        Sequence[float | timedelta]
-        | list[str | float | timedelta]
+        Sequence[str | float | timedelta]
```
Dr-Irv, line 34: The CI is failing because a plain string will match `Sequence[str]` .  So you should change this to 
```suggestion
        SequenceNotStr | Sequence[float | timedelta]
```
You can import `SequenceNotStr` from `_typing`

Also, you should set up your environment and test locally.

Dr-Irv (changes requested):

pandas-stubs/core/tools/timedeltas.pyi
loicdiridollou: Great that you are confirming my initial solution, `'1 day'` and `['1 day']` are both `Sequence`.
I forgot to rerun locally when I pushed the feedback last time, should be good now.
Thanks for all the insights on my first PR here!

loicdiridollou: I had to force the typehint in the test, let me know if this is in the spirit.

Dr-Irv (changes requested):
One small change.  It passed the CI, so it's pretty close.

Dr-Irv (approved):
thanks @loicdiridollou

PR #960, opened 2024-07-13
Fix series.map overloads

JanEricNitschke:

In my last PR that handles `series.map()` i introduced a slight error by adding the `= ...` to the overload with `"ignore"` as this is not the actual default value.

Subsequently invalid function that do not handle `pd.NA` were accepted when the `na_action` was not explicitely specified.

This has been addressed and test cases have been added to check this.

### DIFF
--- a/tests/test_series.py
+++ b/tests/test_series.py
@@ -3284,12 +3284,14 @@ def test_map_na() -> None:
         return x
 
     def bad_callable(x: int) -> int:
-        return x + 1
+        return x << 1
 
-    s.map(
-        bad_callable, na_action=None  # type: ignore[arg-type] # pyright: ignore[reportCallIssue, reportArgumentType]
-    )
-    s.map(bad_callable)  # type: ignore[arg-type] # pyright: ignore[reportArgumentType]
+    with pytest.raises(TypeError):
+        s.map(
+            bad_callable, na_action=None  # type: ignore[arg-type] # pyright: ignore[reportCallIssue, reportArgumentType]
+        )
+    with pytest.raises(TypeError):
+        s.map(bad_callable)  # type: ignore[arg-type] # pyright: ignore[reportArgumentType]
     check(
         assert_type(s.map(bad_callable, na_action="ignore"), "pd.Series[int]"),
         pd.Series,
--- a/tests/test_series.py
+++ b/tests/test_series.py
@@ -3283,21 +3283,6 @@ def test_map_na() -> None:
             return str(x)
         return x
 
-    def bad_callable(x: int) -> int:
-        return x << 1
-
-    with pytest.raises(TypeError):
-        s.map(
-            bad_callable, na_action=None  # type: ignore[arg-type] # pyright: ignore[reportCallIssue, reportArgumentType]
-        )
-    with pytest.raises(TypeError):
-        s.map(bad_callable)  # type: ignore[arg-type] # pyright: ignore[reportArgumentType]
-    check(
-        assert_type(s.map(bad_callable, na_action="ignore"), "pd.Series[int]"),
-        pd.Series,
-        int,
-    )
-
     check(
         assert_type(s.map(callable, na_action=None), "pd.Series[str]"), pd.Series, str
     )
@@ -3306,6 +3291,40 @@ def test_map_na() -> None:
     series = pd.Series(["a", "b", "c"])
     check(assert_type(s.map(series, na_action=None), "pd.Series[str]"), pd.Series, str)
     check(assert_type(s.map(series), "pd.Series[str]"), pd.Series, str)
+
+    s2: pd.Series[float] = pd.Series([1.0, pd.NA, 3.0])
+
+    def callable2(x: float) -> float:
+        return x + 1
+
+    check(
+        assert_type(s2.map(callable2, na_action="ignore"), "pd.Series[float]"),
+        pd.Series,
+        float,
+    )
+    check(
+        assert_type(s2.map(callable2), "pd.Series[float]"),
+        pd.Series,
+        float,
+    )
+    if TYPE_CHECKING_INVALID_USAGE:
+        s2.map(callable2, na_action=None)  # type: ignore[arg-type] # pyright: ignore[reportCallIssue, reportArgumentType]
+
+    s3: pd.Series[str] = pd.Series(["A", pd.NA, "C"])
+
+    def callable3(x: str) -> str:
+        return x.lower()
+
+    check(
+        assert_type(s3.map(callable3, na_action="ignore"), "pd.Series[str]"),
+        pd.Series,
+        str,
+    )
+    if TYPE_CHECKING_INVALID_USAGE:
+        s3.map(
+            callable3, na_action=None  # type: ignore[arg-type] # pyright: ignore[reportCallIssue, reportArgumentType]
+        )
+        s3.map(callable3)  # type: ignore[type-var] # pyright: ignore[reportCallIssue, reportArgumentType]
 
 
 def test_case_when() -> None:
--- a/tests/test_series.py
+++ b/tests/test_series.py
@@ -3276,18 +3276,34 @@ def test_map_na() -> None:
 
     mapping = {1: "a", 2: "b", 3: "c"}
     check(assert_type(s.map(mapping, na_action=None), "pd.Series[str]"), pd.Series, str)
+    check(assert_type(s.map(mapping), "pd.Series[str]"), pd.Series, str)
 
     def callable(x: int | NAType) -> str | NAType:
         if isinstance(x, int):
             return str(x)
         return x
 
+    def bad_callable(x: int) -> int:
+        return x + 1
+
+    s.map(
+        bad_callable, na_action=None  # type: ignore[arg-type] # pyright: ignore[reportCallIssue, reportArgumentType]
+    )
+    s.map(bad_callable)  # type: ignore[arg-type] # pyright: ignore[reportArgumentType]
+    check(
+        assert_type(s.map(bad_callable, na_action="ignore"), "pd.Series[int]"),
+        pd.Series,
+        int,
+    )
+
     check(
         assert_type(s.map(callable, na_action=None), "pd.Series[str]"), pd.Series, str
     )
+    check(assert_type(s.map(callable), "pd.Series[str]"), pd.Series, str)
 
     series = pd.Series(["a", "b", "c"])
     check(assert_type(s.map(series, na_action=None), "pd.Series[str]"), pd.Series, str)
+    check(assert_type(s.map(series), "pd.Series[str]"), pd.Series, str)
 
 
 def test_case_when() -> None:
--- a/pandas-stubs/_typing.pyi
+++ b/pandas-stubs/_typing.pyi
@@ -568,6 +568,8 @@ S2 = TypeVar(
     | BaseOffset,
 )
 
+SN = TypeVar("SN", bound=bool | int | float | complex)
+
 IndexingInt: TypeAlias = (
     int | np.int_ | np.integer | np.unsignedinteger | np.signedinteger | np.int8
 )
--- a/pandas-stubs/core/series.pyi
+++ b/pandas-stubs/core/series.pyi
@@ -917,7 +917,7 @@ class Series(IndexOpsMixin[S1], NDFrame):
     def map(
         self,
         arg: Callable[[S1], S2 | NAType] | Mapping[S1, S2] | Series[S2],
-        na_action: Literal["ignore"] = ...,
+        na_action: Literal["ignore"],
     ) -> Series[S2]: ...
     @overload
     def map(
--- a/pandas-stubs/core/series.pyi
+++ b/pandas-stubs/core/series.pyi
@@ -94,6 +94,7 @@ from pandas._libs.tslibs.nattype import NaTType
 from pandas._typing import (
     S1,
     S2,
+    SN,
     AggFuncTypeBase,
     AggFuncTypeDictFrame,
     AggFuncTypeSeriesToFrame,
@@ -913,6 +914,12 @@ class Series(IndexOpsMixin[S1], NDFrame):
         level: Level = ...,
         fill_value: int | _str | dict | None = ...,
     ) -> DataFrame: ...
+    @overload
+    def map(
+        self,
+        arg: Callable[[SN], S2 | NAType] | Mapping[SN, S2] | Series[S2],
+        na_action: Literal["ignore"] = ...,
+    ) -> Series[S2]: ...
     @overload
     def map(
         self,


tests/test_series.py
```diff
@@ -3276,18 +3276,34 @@ def test_map_na() -> None:
...
             return str(x)
         return x
 
+    def bad_callable(x: int) -> int:
+        return x + 1
+
+    s.map(
+        bad_callable, na_action=None  # type: ignore[arg-type] # pyright: ignore[reportCallIssue, reportArgumentType]
+    )
+    s.map(bad_callable)  # type: ignore[arg-type] # pyright: ignore[reportArgumentType]
```
Dr-Irv, line 3292: The type checkers should accept those patterns, because at least in my tests at runtime, no error is raised.  So why are you considering that a "bad callable"?

Dr-Irv (changes requested):

tests/test_series.py
JanEricNitschke: Ah, may bad. So normally if `na_action=None` then the function has to be able to handle `pd.NA` because it gets passed in. And for my `bad_callable` it does not explicitely handle it, at least according to the type hint.

I just chose a poor operation, to perform in the function, because pd.NA supports addition of an integer.

I have now changed it for a different operation on integers that pd.NA does not incidentally support.

tests/test_series.py
```diff
@@ -3276,18 +3276,36 @@ def test_map_na() -> None:
...
 
+    def bad_callable(x: int) -> int:
+        return x << 1
+
+    with pytest.raises(TypeError):
+        s.map(
+            bad_callable, na_action=None  # type: ignore[arg-type] # pyright: ignore[reportCallIssue, reportArgumentType]
+        )
+    with pytest.raises(TypeError):
+        s.map(bad_callable)  # type: ignore[arg-type] # pyright: ignore[reportArgumentType]
```
Dr-Irv, line 3294: From how we want to test the stubs, this is not a good test, and I think it should be deleted.  What we want in the tests is the following:
1) Tests of valid code that should pass the type checker (that's about 98% of our tests)
2) Tests of invalid code that will raise exceptions that the type checker should catch.  We surround such tests with `if TYPE_CHECKING_INVALID_USAGE` and then put the `ignore` tags on those tests, which also means that the code will not be tested at runtime.

In your case, `bad_callable()` has a signature that could be a "good" callable, and the type checker can't tell the difference.  So  I don't think the 2 tests you have here should be included.  The test below is fine.

Dr-Irv (changes requested):

tests/test_series.py
JanEricNitschke: I think the essence of these two tests is important. A callable that does not explicitly allow `pd.NA` as an argument should not be valid to be passed to `Series.map` if `na_action=None` (either explicitly or as a default value). The default value case was not detected before the changes that i made in this PR and caused a runtime error in a project of mine. So i do think it makes sense to explicitly test this case.

I do agree that they should be moved into a `TPYE_CHECHKING_INVALID_USAGE` block and have the pytest.raises removed.

JanEricNitschke: But then i would argue that the hints are just poor and that is purely on the user.

Take for example:
```
def add(a: float, b: float) -> float:
    return a+b

add(1.0, 2.0)
add("1", "2")
```
While both calls do not cause issues at run time, the second should definitely cause an error from the type checker. Because the type checker only validates the signature, because that is what external users interact with. The implementation could change at any point and cause runtime errors in the future if you are using the function in a way that does not match the signature.

even more explicitly
```
float_callable(pd.NA)
```
causes a typechecking failure by itself.

Dr-Irv: > But then i would argue that the hints are just poor and that is purely on the user (author of the function).

I agree.  But we also make choices in the pandas stubs that cover the more common usage of pandas.  Handling the case of `np.nan` versus `pd.NA` is a long topic of discussion in the pandas team, and we hope to resolve it some day.  But right now, the majority of pandas users use `np.nan` to represent missing values, so we can't have stubs that prevent their code from working.  The changes you are proposing cause existing (and valid) pandas code to fail.

JanEricNitschke: But that point is only incidentally related to this, right? Because the current behavior is that only `Callable[[S1], S2 | NAType]` is allowed if the `na_type=None` (explicitely) and `Callable[[S1 | NAType], S2 | NAType]` for `na_type="ignore` (explicitely).

This PR mainly changes the case that not both cases are also accepted as the default value (which is only `None`).

If `np.nan` is commonly used as a nan value that the signature could be changed to `Callable[[S1], S2 | NAType | Literal[np.nan]]` and  `Callable[[S1 | NAType | Literal[np.nan]], S2 | NAType | Literal[np.nan]]` (probably also for the Mapping variants key and value types).

Because in the live version `s.map(bad_callable, na_type=None)` will give a type checking error.

Dr-Irv: You could try adding the `Literal[np.nan]` in there, but I would want to make sure the following is accepted from a type checking perspective;
```python
s = pd.Series([1.1, 2.2, 3.3])
def good_callable(x: float) -> float:
    return x+ 1.3
s.map(good_callable, na_action=None)
s.map(good_callable)
s.map(good_callable, na_action="ignore")
```
If you can add that as a test, then I'd be OK with the change.

JanEricNitschke: I dont think the first should every typecheck (and it does not do so currently), because with the explicit `na_action=None` you are specifically saying that the function should be able to handle None-values, which `good_callable` does not do explicitly. And i feel that if you explicitly set this you probably do want to be warned about potential issues with your code (i library you are using could change the implementation of a function in a way that is non-breaking according to the signature and you could get runtime errors)

The third would work explicitly, while the second works currently but would not after this PR. I do concede that this case would probably cause warnings that currently work fine because the none values generally handle the common operations performed on numbers (+,-,*,/) without crashing (although importantly not all of them).

Another option here would be to try and carve out exceptions for int and float (maybe also complex and bool?) functions to not require the explicit none-value-handling because they support most of the operations performed on those, but leave it as is for the other (general case) because there they are likely to cause runtime errors?

Dr-Irv: So if you want to separate the cases of `na_action=None` being explicit which should not be allowed, versus `na_action=None` being implicit (because `None` is the default value), I'd be OK with that.  For me, what's important is that the implicit case should pass type checking, because the default behavior (not specifying `na_action`) is most likely what people use today.

JanEricNitschke: Ok, that is basically exactly what the status is before this PR actually, because it allows the interpretation of `na_action="ignore"` as a default value which is more forgiving. Then i would just add a `TPYE_CHECKING_INVALID_USAGE` for `s.map(bad_callable, na_action=None)` which captures the current behavior.

---

Additionally, with `na_action=None` (especially implicitly), the reason why a lot of current incorrect (pure type theory wise) code doesnt cause runtime errors despite not handling `pd.NA` and `np.nan` explicitly, is because they play nicely with most number interactions by just propagating them. However, this is not the case with most other types. So maybe it would make sense to also fail on the implicit case for functions that take most types, but explicitly allow int/float/complex?

Ill try to see if i can do it and add it to the PR to show what i mean.

tests/test_series.py
```diff
@@ -3285,9 +3286,45 @@ def callable(x: int | NAType) -> str | NAType:
...
     )
+    check(assert_type(s.map(callable), "pd.Series[str]"), pd.Series, str)
 
     series = pd.Series(["a", "b", "c"])
     check(assert_type(s.map(series, na_action=None), "pd.Series[str]"), pd.Series, str)
+    check(assert_type(s.map(series), "pd.Series[str]"), pd.Series, str)
+
+    s2: pd.Series[float] = pd.Series([1.0, pd.NA, 3.0])
+
+    def callable2(x: float) -> float:
```
JanEricNitschke, line 3297: Float (and int, complex, bool) callables pass `na_action="ignore"` and default na_action despite no explicit None handling.

tests/test_series.py
```diff
@@ -3285,9 +3286,45 @@ def callable(x: int | NAType) -> str | NAType:
...
+        assert_type(s2.map(callable2), "pd.Series[float]"),
+        pd.Series,
+        float,
+    )
+    if TYPE_CHECKING_INVALID_USAGE:
+        s2.map(callable2, na_action=None)  # type: ignore[arg-type] # pyright: ignore[reportCallIssue, reportArgumentType]
+
+    s3: pd.Series[str] = pd.Series(["A", pd.NA, "C"])
+
+    def callable3(x: str) -> str:
```
JanEricNitschke, line 3315: Non-numerical functions pass only the explicit `na_action="ignore"` but not the `na_action=None` case (explicit or implicit) because these are likely to cause runtime errors on nan-values.

tests/test_series.py
JanEricNitschke: Added the changes i proposed above and added some comments on the PR to explain them. Let me know what you think.

I actually also wanted to specifically allow mappings to return none-values `Mapping[S1, S2]` -> `Mapping[S1, S2 | NAType]` but this seems to mess with mypys overload resolution again.

Dr-Irv: I'm OK with putting `# type: ignore` statements in the stubs when `mypy` and/or `pyright` complains about things like that.

JanEricNitschke: Yeah, but for mypy it also causes issues at the user side where it just infers things as `Any`.

tests/test_series.py
```diff
@@ -3285,9 +3286,45 @@ def callable(x: int | NAType) -> str | NAType:
     check(
         assert_type(s.map(callable, na_action=None), "pd.Series[str]"), pd.Series, str
     )
+    check(assert_type(s.map(callable), "pd.Series[str]"), pd.Series, str)
 
     series = pd.Series(["a", "b", "c"])
     check(assert_type(s.map(series, na_action=None), "pd.Series[str]"), pd.Series, str)
+    check(assert_type(s.map(series), "pd.Series[str]"), pd.Series, str)
+
+    s2: pd.Series[float] = pd.Series([1.0, pd.NA, 3.0])
```
Dr-Irv, line 3295: We generally don't use the annotations in the tests.  So you can do
```python
s2 = pd.Series([1.0, pd.NA, 3.0], dtype=float)
# OR
s2 = pd.Series([1.0, pd.NA, 3.0]).astype(pd.Float64Dtype())
```
Then the type checkers should recognize `s2` as `Series[float]`

Note that internally, `pd.Series([1.0, pd.NA, 3.0])` has dtype `object`

tests/test_series.py
```diff
@@ -3285,9 +3286,45 @@ def callable(x: int | NAType) -> str | NAType:
...
+    )
+    check(
+        assert_type(s2.map(callable2), "pd.Series[float]"),
+        pd.Series,
+        float,
+    )
+    if TYPE_CHECKING_INVALID_USAGE:
+        s2.map(callable2, na_action=None)  # type: ignore[arg-type] # pyright: ignore[reportCallIssue, reportArgumentType]
+
+    s3: pd.Series[str] = pd.Series(["A", pd.NA, "C"])
```
Dr-Irv, line 3313: Similar comment.  Use `s3 = pd.Series(["A", pd.NA, "C"], dtype=str)`

tests/test_series.py
```diff
@@ -3285,9 +3286,45 @@ def callable(x: int | NAType) -> str | NAType:
...
+    check(
+        assert_type(s3.map(callable3, na_action="ignore"), "pd.Series[str]"),
+        pd.Series,
+        str,
+    )
+    if TYPE_CHECKING_INVALID_USAGE:
+        s3.map(
+            callable3, na_action=None  # type: ignore[arg-type] # pyright: ignore[reportCallIssue, reportArgumentType]
+        )
+        s3.map(callable3)  # type: ignore[type-var] # pyright: ignore[reportCallIssue, reportArgumentType]
```
Dr-Irv, line 3327: I'm a little concerned that this won't pass the type checker.  A user should be able to do:
```python
s = pd.Series({"a", "b", "c"], dtype=str)
s.map(callable3)
``` 
That will execute correctly and should pass the type checker.

Dr-Irv (changes requested):

tests/test_series.py
JanEricNitschke: Unfortunately i dont think the type system supports something like this at the moment.

All that `s.map(callable3)` sees is that the entries of `s` are string or none values and that none-values will be passed into `callable3` as they are. So it has to be able to handle them. The type system does not know if there actually are non-values in there or not, only that there might be. Most of the times the user probably wont know either if the data is loaded from an external file that comes from a sensor or a data repository.

If the user does know more than the type system he could either specify `na_action="ignore"`, which he should do anyway if the function does no handle the case explicitly, or add a type ignore to the line as one normally does when you have more information than can be contained in the type system (or understood by the type checkers).

tests/test_series.py
JanEricNitschke: This first actually gives me an error `TypeError: float() argument must be a string or a real number, not 'NAType'`. While the second version does work, i cant get it to work for the integer case of `s = pd.Series([1, pd.NA, 3])`. While that does work at the construction point and for inferring `s: Series[int]`, the values that are passed into `callable` at `s.map(callable)` are actually floats.

tests/test_series.py
Dr-Irv: I understand your point of view.  But the issue here is that we need to support the "normal" usage of pandas.  The type system can't keep track of what is inside of a `Series` or `DataFrame`.  For example, if you have a `DataFrame` `df` with columns `"a"`, `"b"` then `df["a"]` is `Series[Any]` because we can't track the type inside the `DataFrame`.  

The default of the `na_action` argument in `Series.map()` is `None`, which means that people who know that their series has no non-null elements would now have to handle the case in the declared typing of `callable3()` to handle nulls, but that will likely break existing code.

We've had other cases where we were too restrictive in the stubs, and people complained, so we had to pull back. 

See https://github.com/pandas-dev/pandas-stubs/blob/main/docs/philosophy.md#narrow-vs-wide-arguments for a discussion of this issue.  IMHO, you're proposing that the stubs will be too narrow.

tests/test_series.py
Dr-Irv: For `Series[int]`, you can use `s = pd.Series([1, pd.NA, 2], dtype="Int64")`

Sorry about my error on the `dtype=float`.

JanEricNitschke:
```
# Step 1: Create the Series with pd.NA
s = pd.Series([1, pd.NA, 3], dtype="Int64")

print(s)
print(s.dtype)

# Define the callable function
def callable(x: int | NAType) -> str | NAType:
    print(x, type(x))
    if pd.isna(x):
        return pd.NA
    return str(x)

print(s)
print(s.map(callable, na_action=None))

```

actually still gives me this:

```
$ python testing.py
0       1
1    <NA>
2       3
dtype: Int64
Int64
0       1
1    <NA>
2       3
dtype: Int64
1.0 <class 'float'>
nan <class 'float'>
3.0 <class 'float'>
0     1.0
1    <NA>
2     3.0
dtype: object

```

tests/test_series.py
JanEricNitschke: Yeah, i think im just too much of a typing purist in this case. Because, just because your code doesnt cause runtime errors with the data that you are testing it with, doesnt mean that it isnt unsound from a typing perspective. Normally every interaction should be proveably valid just from the involved signatures. And i am using a type checker to actually ensure this.

But i guess that is not fully possible anyway with the setup of pandas.

tests/test_series.py
Dr-Irv: That's a known pandas issue.  See https://github.com/pandas-dev/pandas/issues/57189

tests/test_series.py
Dr-Irv: Yes, I agree.  We've chosen a middle ground with the stubs that supports the common ways people use pandas, but doesn't necessarily support all possible best typing practices.

Dr-Irv:
@JanEricNitschke can you address my comments at
- https://github.com/pandas-dev/pandas-stubs/pull/960#discussion_r1680033615 
- https://github.com/pandas-dev/pandas-stubs/pull/960#discussion_r1680031096
- https://github.com/pandas-dev/pandas-stubs/pull/960#discussion_r1680031672

I'll then have to take another careful look and see if there is anything else that I'm not comfortable with.

JanEricNitschke:
@Dr-Irv ah, sorry, i had other life stuff come up.

I think it is not easily possible to make the changes i intended while keeping with the general philosophy of not disallowing things with the types that do happen to work at runtime.

So i would probably close this PR  for now and maybe have another look again later when i have more time to see if there is still an option?

Does that work for you?

Dr-Irv:
> So i would probably close this PR for now and maybe have another look again later when i have more time to see if there is still an option?
> 
> Does that work for you?

Sure.  I'll let you do the closing

PR #961, opened 2024-07-13
df.apply should allow pd.NA from Callables

JanEricNitschke:

`df.apply()` should allow Callables to return either an `S1` compatible type or `pd.NA`


Dr-Irv (approved):
Thanks @JanEricNitschke

PR #964, opened 2024-07-28
fix signature of frame.filter

refack:
`items` can be `ListLike`

It is turned into and `Index` which can handle and `ListLike` as data
https://github.com/pandas-dev/pandas/blob/9b375be5aa3610e8a21ef0b5b81e4db04270f3d3/pandas/core/generic.py#L5448

### DIFF
--- a/pandas-stubs/core/frame.pyi
+++ b/pandas-stubs/core/frame.pyi
@@ -1711,7 +1711,7 @@ class DataFrame(NDFrame, OpsMixin):
     ) -> DataFrame: ...
     def filter(
         self,
-        items: list | None = ...,
+        items: ListLike | None = ...,
         like: _str | None = ...,
         regex: _str | None = ...,
         axis: Axis | None = ...,
--- a/tests/test_frame.py
+++ b/tests/test_frame.py
@@ -319,6 +319,8 @@ def test_types_filter() -> None:
     df.filter(items=["col1"])
     df.filter(regex="co.*")
     df.filter(like="1")
+    # [PR 964] Docs state `items` is `list-like`
+    df.filter(items=("col2", "col2", 1, tuple([4])))
 
 
 def test_types_setting() -> None:


Dr-Irv (changes requested):
Can you add a test that fails without your PR and works with your PR?

refack:
> Can you add a test

Allready started (great minds and such)

Dr-Irv:
CI is failing because pyright got upgraded and we have to remove some `#pyright ignore` statements. This is in PR #967 .  Once that is merged in, then you can merge in the new `main` in your branch and push, and I will test the CI again.

Dr-Irv:
> CI is failing because pyright got upgraded and we have to remove some `#pyright ignore` statements. This is in PR #967 . Once that is merged in, then you can merge in the new `main` in your branch and push, and I will test the CI again.

This is now merged, so go ahead and merge with `main` and push

pandas-stubs/core/frame.pyi
```diff
@@ -1711,7 +1711,7 @@ class DataFrame(NDFrame, OpsMixin):
     ) -> DataFrame: ...
     def filter(
         self,
-        items: list | None = ...,
+        items: Iterable | None = ...,
```
Dr-Irv, line 1714: `Iterable` is too wide.   That will accept a plain string.  Can you use `ListLike` here

While the pandas code accepts a set, that's not the intent and why the docs say "list-like".  That would also mean you should remove the test that passes a set.

Dr-Irv (changes requested):

pandas-stubs/core/frame.pyi
refack: I used `ListLike` in the first iteration but then mypy fails for `set` or `generator`

```console
tests\test_frame.py:323: error: Value of type variable "ListLike" of "filter" of "DataFrame" cannot be "set[str]"  [type-var]
tests\test_frame.py:325: error: Value of type variable "ListLike" of "filter" of "DataFrame" cannot be "Generator[str, None, None]"  [type-var]
```

I skimmed the code for something in the middle... Maybe `Iterable[str]`?

refack: Since `items` is converted into an `Index` it's very flexible. And `Index.__new__` has a very complicated signature with multiple overloads.
Your call as to which way to go; stricter and have false negatives, or loose and have false positives.

Dr-Irv: the rule we are using in the stubs is to base things on the docs.  So while `filter` does accept a `set`, it probably shouldn't, given the docs.  So I am for removing the test for sets, and if you want to include a generator, you can use `ListLike | Generator[Hashable, None, None]`

refack: Done

Dr-Irv (approved):
Thanks @refack

refack:
> Thanks @refack

Keep up the great work

refack:
@Dr-Irv apropos, I just stumbled on https://github.com/pandas-dev/pandas/issues/47215
LOL
It's nice to see things stay consistent even after 2 years 👍

PR #966, opened 2024-07-30
GroupBy[Series].count() return type should be Series[int]

### LINKED ISSUE #965: GroupBy[Series].count() should have return type Series[int]
**Describe the bug**

**To Reproduce**

```python
s = pd.Series(["a", "b", "a"], index=[1, 2, 2])
s.groupby(level=0).count()
```

Type checker: pyright (in VS Code)

The expected type is `Series[int]`, but instead pyright shows it as `Series[str]`

 - OS: Ubuntu
 - OS Version: 22.04.4 LTS
 - python version: 3.10.12
 - version of type checker: pyright 1.1.369
 - version of installed `pandas-stubs`: 2.2.2.240603

chrisyeh96:
- [X] Closes #965


pandas-stubs/core/groupby/groupby.pyi
```diff
@@ -176,7 +176,10 @@ class GroupBy(BaseGroupBy[NDFrameT]):
     @overload
     def all(self: GroupBy[DataFrame], skipna: bool = ...) -> DataFrame: ...
     @final
-    def count(self) -> NDFrameT: ...
+    @overload
+    def count(self: GroupBy[Series]) -> Series[int]: ...
+    @overload
+    def count(self: GroupBy[DataFrame]) -> DataFrame: ...
```
Dr-Irv, line 182: I'd prefer removing `def count()` from here, and having declarations in `DataFrameGroupBy` and `SeriesGroupBy` in `core/groupby/generic.pyi`

tests/test_series.py
```diff
@@ -676,6 +676,11 @@ def test_types_groupby_methods() -> None:
     check(assert_type(s.groupby(level=0).idxmax(), pd.Series), pd.Series)
     check(assert_type(s.groupby(level=0).idxmin(), pd.Series), pd.Series)
 
+    s2 = pd.Series(["w", "x", "y", "z"], index=[3, 4, 3, 4], dtype=str)
+    check(
+        assert_type(s2.groupby(level=0).count(), "pd.Series[int]"), pd.Series, np.int_
```
Dr-Irv, line 681: CI is failing on Windows for this.  Use `np.integer` rather than `np.int_`

Dr-Irv (changes requested):

Dr-Irv:
We just uncovered an issue with our CI that was fixed earlier today. Can  you merge with current `main` and push?  Then I will start the CI runs.

chrisyeh96:
I have merged `pandas-dev/main` into my branch

pandas-stubs/core/groupby/groupby.pyi
Dr-Irv: > I'm a new contributor to this repo and eager to learn more about these choices. Is there a specific reason for moving the declarations to `core/groupby/generic.pyi`?
> 
> In the actual pandas repo, the `count()` method is implemented in [`core/groupby/groupby.py`](https://github.com/pandas-dev/pandas/blob/v2.2.2/pandas/core/groupby/groupby.py), not [`core/groupby/generic.py`](https://github.com/pandas-dev/pandas/blob/v2.2.2/pandas/core/groupby/generic.py).
> 
> Thanks in advance for taking the time to explain.

It probably is due to that we had issues in the past with doing overloads where `self` was declared as `Series` vs. `DataFrame`, so putting the stubs in the lower classes was a way to get around that problem.

The type checkers have improved, so it's possible your approach could be used going forward.

Having said that, I see that for `any()` and `all()` in the file you changed, that was the pattern used, so maybe your change was fine the way it was!

As an example, there are a number of methods shared between `Series` and `DataFrame` that are implemented in `NDFrame`, but instead of putting the stub in `NDFrame`, we put them in `Series` and `DataFrame` to clarify that the results are different in the two subclasses.

Dr-Irv:
Ugh, so now CI is failing because when you first tested, it used pyright 1.1.373 and that got locked for further tests.

Can you include this in your PR?  In `pyproject.toml`, change:
```
pyright = ">=1.1.369"
```
to
```
pyright = ">= 1.1.374"
```

chrisyeh96:
I have updated the pyright dependency version in `pyproject.toml` as you suggested.

Dr-Irv (approved):
thanks @chrisyeh96

PR #972, opened 2024-08-06
Use the new TypeIs feature to update the notna/notnull/isna/isnull type guards

### LINKED ISSUE #911: Use the new TypeIs feature to update the notna/notnull/isna/isnull type guards
I originally contributed the annotations for these methods in https://github.com/pandas-dev/pandas-stubs/pull/339. That PR has a bunch of context about how the fact that `TypeGuard` only narrows in the positive case, while the methods should narrow in both the positive and negative case.

`TypeIs` was recently introduced as a PEP and mypy added support for it, I suspect pyright as well: https://mypy-lang.blogspot.com/2024/04/mypy-110-released.html

This would allow narrowing in the negative case and fix the type narrowing semantics of these methods.

loicdiridollou:
- [x] Closes #911 


tests/test_pandas.py
```diff
@@ -366,52 +366,41 @@ def test_isna() -> None:
...
     if not pd.isna(nullable3):
-        # check(assert_type(nullable3, bool), bool)  # TODO: Desired result (see comments above)
-        check(assert_type(nullable3, Union[bool, NAType, None]), bool)
+        check(assert_type(nullable3, bool), bool)
     if pd.isna(nullable3):
-        # assert_type(nullable3, Union[NAType, None])  # TODO: Desired result (see comments above)
-        assert_type(nullable3, Union[NaTType, NAType, None])
+        assert_type(nullable3, Union[NAType, None])
     if not pd.notna(nullable3):
         # assert_type(nullable3, Union[NAType, None])  # TODO: Desired result (see comments above)
```
gandhis1, line 405: This assertion can be un-commented out

tests/test_pandas.py
```diff
@@ -366,52 +366,41 @@ def test_isna() -> None:
...
-        check(assert_type(nullable2, Union[NaTType, NAType, None]), type(None))
+        check(assert_type(nullable2, None), type(None))
     if not pd.notna(nullable2):
-        # check(assert_type(nullable2, None), type(None))  # TODO: Desired result (see comments above)
+        check(assert_type(nullable2, None), type(None))
         # TODO: MyPy and Pyright produce conflicting results:
         # assert_type(nullable2, Union[int, None])  # MyPy result
         # assert_type(
         #     nullable2, Union[int, NaTType, NAType, None]
         # )  # Pyright result
```
gandhis1, line 395: This commented out part may be out-of-date and can be removed?

tests/test_pandas.py
```diff
@@ -366,52 +366,41 @@ def test_isna() -> None:
...
-        # check(assert_type(nullable3, bool), bool)  # TODO: Desired result (see comments above)
-        check(assert_type(nullable3, Union[bool, NAType, None]), bool)
+        check(assert_type(nullable3, bool), bool)
     if pd.isna(nullable3):
-        # assert_type(nullable3, Union[NAType, None])  # TODO: Desired result (see comments above)
-        assert_type(nullable3, Union[NaTType, NAType, None])
+        assert_type(nullable3, Union[NAType, None])
     if not pd.notna(nullable3):
         # assert_type(nullable3, Union[NAType, None])  # TODO: Desired result (see comments above)
         # TODO: MyPy and Pyright produce conflicting results:
```
gandhis1, line 406: This commented out part may be out-of-date and can be removed?

loicdiridollou: New line 405 should be good, working on this, the rest is out-of-date so removing

tests/test_pandas.py
loicdiridollou: Fixing that, good catch!

tests/test_pandas.py
loicdiridollou: Out-of-date correct, removing.

loicdiridollou:
Thanks @gandhis1, not sure though if the CI has triggered on the last commit.

Dr-Irv (approved):
Thanks @loicdiridollou .  Nice change

steve-mavens:
I think this change isn't quite correct: https://github.com/pandas-dev/pandas-stubs/issues/1009. The return value of this function doesn't actually discriminate the type, because some `float` are NaN and others aren't.

PR #975, opened 2024-08-07
Fix nightly CI with split case for pandas (around 3.0.0 version)

### LINKED ISSUE #973: Restore nightly CI build due to new pandas version (3.0.0)
**Describe the bug**
Due to many new deprecations in pandas 3.0.0, the nightly build involving the new version of pandas is failing.

**To Reproduce**
`poetry run poe test_all --nightly`
Also see the Actions tab in the GitHub repo.

 - OS: MacOS
 - OS Version: 14.5
 - python version: 3.12.4
 - version of type checker: 1.1.373
 - version of installed `pandas-stubs`: v2.2.2.240807

loicdiridollou:
- [x] Closes #973


tests/test_utility.py
```diff
@@ -17,20 +17,26 @@
...
-    ):
-        context: AbstractContextManager
-        if PD_LTE_22 and NUMPY20:  # https://github.com/PyTables/PyTables/issues/1172
-            context = pytest.raises(ValueError)
-        else:
-            context = nullcontext()
-        with context:
-            check(assert_type(pd.show_versions(True), None), type(None))
-            check(assert_type(pd.show_versions(False), None), type(None))
+    """Test show_versions method types with split case for pandas and python versions."""
```
Dr-Irv, line 20: If you set `context = nullcontext()` here, then you don't need the `else`, and  you can have the 2 lines for the test as `with context` where `with context` is at the level of `if` and you avoid the repeated test.

tests/test_utility.py
```diff
@@ -17,20 +17,26 @@
...
+            UserWarning,
+            match="Setuptools is replacing distutils",
+            upper="3.11.99",
+            version_str=platform.python_version(),
+        ):
+            context: AbstractContextManager
+            if NUMPY20:  # https://github.com/PyTables/PyTables/issues/1172
+                context = pytest.raises(ValueError)
+            else:
+                context = nullcontext()
```
Dr-Irv, line 33: can remove these two lines based on my suggestion above

tests/test_utility.py
```diff
@@ -17,20 +17,26 @@
...
+            with context:
+                check(assert_type(pd.show_versions(True), None), type(None))
+                check(assert_type(pd.show_versions(False), None), type(None))
+    else:
+        check(assert_type(pd.show_versions(True), None), type(None))
+        check(assert_type(pd.show_versions(False), None), type(None))
```
Dr-Irv, line 39:
```suggestion
    with context:
        check(assert_type(pd.show_versions(True), None), type(None))
        check(assert_type(pd.show_versions(False), None), type(None))
```

you do this and then you've handled all cases.

Dr-Irv (changes requested):

loicdiridollou:
@Dr-Irv I tried to clean up the test but the context is dependent on `pandas<3.0` and `numpy>=2.0` so that makes the logic a little crooked. I improved it but there is no way to have all the `show_versions` at once (since in one case it will raise a `UserWarning` and not in another).
Let me know if you find a better way to write those, I was not overly familiar with the logic at first so did not want to break it all.

tests/test_utility.py
```diff
@@ -17,17 +17,24 @@
...
+                match="Setuptools is replacing distutils",
+                upper="3.11.99",
+                version_str=platform.python_version(),
+            ),
+            context,
+        ):
+            check(assert_type(pd.show_versions(True), None), type(None))
+            check(assert_type(pd.show_versions(False), None), type(None))
+    else:
         with context:
```
Dr-Irv, line 38: Since the `else` will get entered with `context = nullcontext()`, you don't need to do the test with a context manager.  So you can delete this line.

Dr-Irv (changes requested):

tests/test_utility.py
loicdiridollou: Correct that was unnecessary, let me remove!

Dr-Irv (approved):
thanks @loicdiridollou .  Let's see what happens after I merge

loicdiridollou:
> thanks @loicdiridollou . Let's see what happens after I merge

Fingers crossed!

Dr-Irv:
> > thanks @loicdiridollou . Let's see what happens after I merge
> 
> Fingers crossed!

It succeeded.  Thanks for looking into this

PR #977, opened 2024-08-12
Modify optional workflows to be triggered by comments

### LINKED ISSUE #613: Create a way to test nightly locally, or as part of a PR that fixes nightly
Right now, if we get a PR that fixes an issue for the nightly tests, the only way to test it is to approve the PR and merge it, and see what happens as a result of the merge.

We should figure out a way that we can run the nightly tests on demand, either locally, or by doing something in the PR that signals that the nightly tests should be run as part of the PR.

Maybe @twoertwein has an idea here.

loicdiridollou:
- [x] Closes #613  (Replace xxxx with the Github issue number)

This PR tries to modify how we run nightly/strict workflows to avoid having to run them systematically on merge to main but instead by able to do so by commenting on the PR with comments like `/pandas_nightly`, `mypy_nightly` or `/pyright_strict`.

The process would go as such:
- You write some code and open a PR in the repo
- You comment in the PR one of the commands like `/pandas_nightly` as below
image
- From there, the action kicks in and run the tests.
- Once done, the results are published in a check run like below with a link to the action so that it is convenient to access the detailed results.
image


loicdiridollou:
Right now there is a bit of code duplication in the yaml file but open to clean up, I will add a readme file to guide the operations.

Dr-Irv:
I'm going to let @twoertwein review, approve and merge this, but one thing that I see is that you removed the `optional` runs that occur during a merge.  I think those should remain, because otherwise, as someone doing many of the merges and getting those reports, I won't see what's failing on a regular basis and know what needs to be fixed.

loicdiridollou:
> I'm going to let @twoertwein review, approve and merge this, but one thing that I see is that you removed the `optional` runs that occur during a merge. I think those should remain, because otherwise, as someone doing many of the merges and getting those reports, I won't see what's failing on a regular basis and know what needs to be fixed.

I misunderstood then, I thought the idea was to replace as those optionals were put in place as a workaround. I will restore them so that they get triggered at every merge, sorry for that.

.github/workflows/comment_commands.yml
```diff
@@ -0,0 +1,164 @@
...
+        uses: ./.github/setup
+        with:
+          os: ubuntu-latest
+          python-version: "3.11"
+
+      - name: Run pytest (against pandas nightly)
+        id: tests-step
+        run: poetry run poe pytest --nightly
+
+      - name: Get head sha and store value
```
twoertwein, line 28: Can you please add a few comments here in the file and explain what these steps do and why we need them (why do we need to get the hash)?

I'm not very familiar with the github CI scripts.

.github/workflows/comment_commands.yml
```diff
@@ -0,0 +1,164 @@
...
+    types: created
+
+permissions:
+  checks: write
+
+jobs:
+  pandas_nightly:
+    runs-on: ubuntu-latest
+    timeout-minutes: 10
+    if: (github.event.issue.pull_request) && github.event.comment.body == '/pandas_nightly'
```
twoertwein, line 13: Should probably also document those commands

twoertwein: @Dr-Irv Do you prefer to have separate commands (pandas nightly for pytest; mypy nightly; pyright strict) or combine them? Would reduce the common setup steps if we combine them.

.github/workflows/comment_commands.yml
loicdiridollou: Yes all will be documented to reflect what we are doing, agree it is not always super clear.

.github/workflows/comment_commands.yml
loicdiridollou: I was thinking of combining the commands into one since we can only trigger one set of tests by command, just trying to figure out how to make a clean combination.

Dr-Irv: If by "commands", you mean commands that someone could do in Github to cause these to run, I'm indifferent.  If we combine them, then we just have to document that certain failures (`pyright_strict`) are OK, while the others should be fixed sooner than later.

twoertwein: Seems like it might be better if we keep them separated

loicdiridollou: I am going to try and make a mapping so that we can simplify the code, with good docs it will be easier than just a lot of repetition.

loicdiridollou: Sorry @twoertwein for the delay i was moving cross country.
I did concentrate all the actions with one code and a selector as a dictionary that will select the command to run according to the comment, let me know if this is what you intended and I will write the documentation.

twoertwein: Looks great! Can you please mention the three commands here https://github.com/pandas-dev/pandas-stubs/blob/main/docs/tests.md ? It can probably integrated in the optional test section (or be a new section).

.github/workflows/comment_commands.yml
```diff
@@ -0,0 +1,71 @@
...
+  RUN_COMMAND: '{"/pandas_nightly": "pytest --nightly", "/pyright_strict": "pyright_strict", "/mypy_nightly": "mypy --mypy_nightly"}'
+  # store mapping of labels to display in the check runs
+  DISPLAY_COMMAND: '{"/pandas_nightly": "Pandas nightly tests", "/pyright_strict": "Pyright strict tests", "/mypy_nightly": "Mypy nightly tests"}'
+
+jobs:
+  optional_tests:
+    name: "Optional tests run"
+    runs-on: ubuntu-latest
+    timeout-minutes: 10
+    # if more commands are added, they will need to be added here too as we don't have avvecc to env at this stage
```
Dr-Irv, line 20: not sure what you mean by `avvecc`

docs/tests.md
```diff
@@ -20,3 +20,9 @@ The following tests are **optional**. Some of them are run by the CI but it is o
 - Use mypy nightly to validate the annotations: `poe mypy --mypy_nightly`
 - Use pyright in full strict mode: `poe pyright_strict`
 - Run stubtest to compare the installed pandas-stubs against pandas (this will fail): `poe stubtest`. If you have created an allowlist to ignore certain errors: `poe stubtest path_to_the_allow_list`
+
+Among the tests above, the following can be run directly during a PR by commenting in the discussion.
+
+- Run pytest against pandas nightly by commenting `\pandas_nightly`
+- Use mypy nightly to validate the annotations by commenting `\mypy_nightly`
+- Use pyright in full strict mode by commenting `\pyright_strict`
```
Dr-Irv, line 28: Shouldn't these be forward slashes (e.g., `/pyright_string` rather than `\pyright_strict` )?

Dr-Irv (changes requested):

docs/tests.md
loicdiridollou: Forward slashes, my mistake.

.github/workflows/comment_commands.yml
loicdiridollou: I meant access, will fix

twoertwein (approved):

Dr-Irv (approved):
Thanks @loicdiridollou

PR #967, opened 2024-07-31
remove pyright overrides not needed with pyright 1.1.374

Dr-Irv:

PR #979, opened 2024-08-19
Support pyright 376

Dr-Irv:
- #967 created fixes for pyright 1.1.374, but that was a buggy version of pyright, so pyright 1.1.376 has it right, so had to undo that PR
- Upgraded my laptop to Windows 11 and had intermittent issues with the `test_plotting.py` tests, so copied over some code from the pandas repo to handle that


Dr-Irv:
Had to lock numpy to 2.0.1 because with numpy 2.1, we are getting fails now, and as my luck goes, numpy 2.1 was released yesterday.

twoertwein:
Thanks @Dr-Irv !

PR #981, opened 2024-08-20
Support Numpy 2.1

Dr-Irv:
Numpy 2.1 is not available on python 3.9, so dropped 3.9 support.

Built on top of #979


pandas-stubs/core/dtypes/base.pyi
```diff
@@ -16,7 +16,7 @@ class ExtensionDtype:
     @property
     def kind(
         self,
-    ) -> Literal["b", "i", "u", "f", "c", "m", "M", "O", "S", "U", "V"]: ...
+    ) -> Literal["b", "i", "u", "f", "c", "m", "M", "O", "S", "U", "V", "T"]: ...
```
twoertwein, line 19: is that because of the numpy change?

(It would be great if we can reference their annotations, if they expose TypeVariables for that.)

twoertwein:
Needs rebasing

pandas-stubs/core/dtypes/base.pyi
Dr-Irv: Yes, `numpy` added that as one of their literals.

The now have a `_DTypeKind` in their `__init__.pyi` in version 2.1, but it doesn't exist in prior versions, so I didn't want to `import` it. I was concerned about requiring numpy 2.1 in the stubs because they ideally should still work with earlier versions of numpy.

twoertwein (approved):

twoertwein:
Thank you @Dr-Irv !

PR #982, opened 2024-08-21
Add support for arrays in `lower` and `upper` args of `DataFrame.clip`

### LINKED ISSUE #980: `lower` and `upper` from `DataFrame.clip` don't accept array-like
**Describe the bug**
Hi,

There is an inconsistency with Pandas's doc regarding `clip`. This was already fixed for Series in #443, but not for DataFrame

https://github.com/pandas-dev/pandas-stubs/blob/2acc94d60c63484d9b664b8e68aa213a4d39b937/pandas-stubs/core/frame.pyi#L1633-L1641

**To Reproduce**
1. From https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.DataFrame.clip.html
```python
import pandas as pd

data = {"col_0": [9, -3, 0, -1, 5], "col_1": [-2, -7, 6, 8, -5]}
df = pd.DataFrame(data)
t = pd.Series([2, -4, -1, 6, 3])
df.clip(t, t + 4, axis=0)
```
2. Pyright 1.1.373
3. 
```
Argument of type "Series[int]" cannot be assigned to parameter "upper" of type "float | None" in function "clip"
  Type "Series[int]" is incompatible with type "float | None"
    "Series[int]" is incompatible with "float"
    "Series[int]" is incompatible with "None"Pylance[reportArgumentType](https://github.com/microsoft/pyright/blob/main/docs/configuration.md#reportArgumentType)
```

 - OS: Windows 11
 - OS Version 23H2
 - python version 3.12.2
 - version of type checker:  Pyright 1.1.373
 - version of installed `pandas-stubs`: 2.2.2.240807

ldouteau:
- [x] Closes #980 

### DIFF
--- a/tests/test_frame.py
+++ b/tests/test_frame.py
@@ -540,7 +540,7 @@ def test_types_clip(lower, upper, axis) -> None:
         with pytest.raises(ValueError):
             df.clip(lower=lower, upper=upper, axis=axis)
     else:
-        df.clip(lower=lower, upper=upper, axis=axis)
+        assert_type(df.clip(lower=lower, upper=upper, axis=axis), pd.DataFrame)
 
 
 def test_types_abs() -> None:
--- a/tests/test_frame.py
+++ b/tests/test_frame.py
@@ -527,9 +527,20 @@ def test_types_quantile() -> None:
     df.quantile(np.array([0.25, 0.75]))
 
 
-def test_types_clip() -> None:
+@pytest.mark.parametrize("lower", [None, 5, pd.Series([3, 4])])
+@pytest.mark.parametrize("upper", [None, 15, pd.Series([12, 13])])
+@pytest.mark.parametrize("axis", [None, 0, "index"])
+def test_types_clip(lower, upper, axis) -> None:
+    def is_none_or_numeric(val: Any) -> bool:
+        return val is None or isinstance(val, int | float)
+
     df = pd.DataFrame(data={"col1": [20, 12], "col2": [3, 14]})
-    df.clip(lower=5, upper=15)
+    uses_array = not (is_none_or_numeric(lower) and is_none_or_numeric(upper))
+    if uses_array and axis is None:
+        with pytest.raises(ValueError):
+            df.clip(lower=lower, upper=upper, axis=axis)
+    else:
+        df.clip(lower=lower, upper=upper, axis=axis)
 
 
 def test_types_abs() -> None:
--- a/tests/test_frame.py
+++ b/tests/test_frame.py
@@ -540,7 +540,10 @@ def test_types_clip(lower, upper, axis) -> None:
         with pytest.raises(ValueError):
             df.clip(lower=lower, upper=upper, axis=axis)
     else:
-        assert_type(df.clip(lower=lower, upper=upper, axis=axis), pd.DataFrame)
+        check(
+            assert_type(df.clip(lower=lower, upper=upper, axis=axis), pd.DataFrame),
+            pd.DataFrame,
+        )
 
 
 def test_types_abs() -> None:
--- a/pandas-stubs/core/frame.pyi
+++ b/pandas-stubs/core/frame.pyi
@@ -1632,8 +1632,8 @@ class DataFrame(NDFrame, OpsMixin):
     ) -> DataFrame: ...
     def clip(
         self,
-        lower: float | None = ...,
-        upper: float | None = ...,
+        lower: float | AnyArrayLike | None = ...,
+        upper: float | AnyArrayLike | None = ...,
         *,
         axis: Axis | None = ...,
         inplace: _bool = ...,


tests/test_frame.py
```diff
@@ -527,9 +527,20 @@ def test_types_quantile() -> None:
...
+        return val is None or isinstance(val, int | float)
+
     df = pd.DataFrame(data={"col1": [20, 12], "col2": [3, 14]})
-    df.clip(lower=5, upper=15)
+    uses_array = not (is_none_or_numeric(lower) and is_none_or_numeric(upper))
+    if uses_array and axis is None:
+        with pytest.raises(ValueError):
+            df.clip(lower=lower, upper=upper, axis=axis)
+    else:
+        assert_type(df.clip(lower=lower, upper=upper, axis=axis), pd.DataFrame)
```
Dr-Irv, line 543:
```suggestion
        check(assert_type(df.clip(lower=lower, upper=upper, axis=axis), pd.DataFrame), pd.DataFrame)
```
This will add the runtime check.

Dr-Irv (changes requested):

tests/test_frame.py
```diff
@@ -527,9 +527,23 @@ def test_types_quantile() -> None:
...
+def test_types_clip(lower, upper, axis) -> None:
+    def is_none_or_numeric(val: Any) -> bool:
+        return val is None or isinstance(val, int | float)
+
     df = pd.DataFrame(data={"col1": [20, 12], "col2": [3, 14]})
-    df.clip(lower=5, upper=15)
+    uses_array = not (is_none_or_numeric(lower) and is_none_or_numeric(upper))
+    if uses_array and axis is None:
+        with pytest.raises(ValueError):
+            df.clip(lower=lower, upper=upper, axis=axis)
```
Dr-Irv, line 541: It's possible (but not required) that you could also handle this via typing checks by creating overloads that are the acceptable combinations of the arguments `lower`, `upper` and `axis`

Dr-Irv: > It's possible (but not required) that you could also handle this via typing checks by creating overloads that are the acceptable combinations of the arguments `lower`, `upper` and `axis`

I'm OK to approve this as is, but let me know if you want to try to make these changes I suggested as part of this PR.

ldouteau: Just thinking of this implementation. One question though: the case where `upper` and `lower` are both arrays is covered by both overloads. Is that an issue ? I'm not that familiar with overloads in general

```python
    @overload
    def clip(
        self,
        lower: AnyArrayLike = ...,
        upper: float | AnyArrayLike | None = ...,
        *,
        axis: Axis = ...,
        inplace: _bool = ...,
        **kwargs,
    ) -> DataFrame: ...
    @overload
    def clip(
        self,
        lower: float | AnyArrayLike | None = ...,
        upper: AnyArrayLike = ...,
        *,
        axis: Axis = ...,
        inplace: _bool = ...,
        **kwargs,
    ) -> DataFrame: ...
```

Dr-Irv: > I'm not sure how to type the fact that "either one of `lower` or `upper` is ArrayLike". If you have any suggestion to propose, i can add overloads

I don't think that is required.  `df.clip()` works fine, as does `df.clip(3, 4)`.  So maybe your test isn't right?


> 
> Regarding the test i think it should stay as-is, as it tests the pandas version of the function

If you can get the overloads to work, then you'd add tests using `TYPE_CHECKING_INVALID_USAGE` to make sure the type checkers caught the invalid usage.  But I'm now not sure the overloads are needed.

ldouteau: > maybe your test isn't right?
Well the test is not exhaustive: `df.clip([1, 2])` will pass but `df.clip(pd.Series([1, 2])` will fail with a value error.

What's your opinion on this ? Getting dedicated overloads using pd.Series ?

Dr-Irv: > > maybe your test isn't right?
> 
> Well the test is not exhaustive: `df.clip([1, 2])` will pass but `df.clip(pd.Series([1, 2]))` will fail with a value error.
> 
> What's your opinion on this ? Getting dedicated overloads using pd.Series ?

> > maybe your test isn't right?
> 
> Well the test is not exhaustive: `df.clip([1, 2])` will pass but `df.clip(pd.Series([1, 2]))` will fail with a value error.
> 
> What's your opinion on this ? Getting dedicated overloads using pd.Series ?

I think this is pretty complicated, because you need to have overloads for each combination where `axis=None` and using `AnyArrayLike` corresponds to `Never`, but using `AnyArrayLike` requires `axis` to be specified and not be `None`.

I'm going to approve this PR as is - it's an improvement, and if you want to experiment with the overloads, that could be a separate PR.   Created #984 to track that.

Dr-Irv (approved):
thanks @ldouteau

PR #985, opened 2024-08-22
`__arrow_c_stream__` method on DataFrame

kylebarron:
Pandas added support for the [Arrow PyCapsule Interface](https://arrow.apache.org/docs/format/CDataInterface/PyCapsuleInterface.html#protocol-typehints) (the `__arrow_c_stream__` method) in pandas v2.2. https://github.com/pandas-dev/pandas/pull/56587

This method returns a C PyCapsule, so it's not the _types_ of arguments or return value that matters, it's really just the existence of the method, so that when other libraries expect an object defined as 
```py
class ArrowStreamExportable(Protocol):
    def __arrow_c_stream__(
        self,
        requested_schema: object | None = None
    ) -> object:
        ...
```

then a `DataFrame` will satisfy the type checker


Dr-Irv (approved):
Thanks @kylebarron 

I'm not sure we would have ever picked this up, because it isn't documented in the pandas release notes.

PR #990, opened 2024-09-01
fix(typing): #983 return type of StringMethods.match

### LINKED ISSUE #983: df.columns.str.match actually gives npt.NDArray[np.bool_], but mypy thinks it is pd.Index[str]
# Describe the bug
`df.columns.str.match(reg_ex_pattern)` actually gives `npt.NDArray[np.bool_]`, but mypy thinks it's `pd.Index[str]`

# To Reproduce
## Provide a minimal runnable `pandas` example that is not properly checked by the stubs.
```py
from typing import TYPE_CHECKING, cast

import numpy as np
import pandas as pd

if TYPE_CHECKING:
    from numpy import typing as npt


df = pd.DataFrame({"1": [2, 3], "2": 3, "4": 5, "a": 1})
mask = df.columns.str.match(r"\d")
print(mask)  # array([ True,  True,  True, False])
print(type(mask))  # <class 'numpy.ndarray'>
df.loc[:, mask]  # mypy: error: Invalid index type "tuple[slice, Index[str]]" for "_LocIndexerFrame"; expected type "slice | ndarray[Any, dtype[integer[Any]]] | Index[Any] | list[int] | Series[int] | <6 more items>"
df.loc[:, cast("npt.NDArray[np.bool_]", mask)]  # mypy: fine
```
## Indicate which type checker you are using (`mypy` or  `pyright`).
I am using `mypy`.
## Show the error message received from that type checker while checking your example.
```ps
error: Invalid index type "tuple[slice, Index[str]]" for "_LocIndexerFrame"; expected type "slice | ndarray[Any, dtype[integer[Any]]] | Index[Any] | list[int] | Series[int] | <6 more items>"
```

# Please complete the following information:
 - OS: Windows
 - OS Version:
   ```ps
   cmd /c ver  # Microsoft Windows [Version 10.0.19045.4651]
   ```
 - python version
   ```ps
   poetry run python --version  # Python 3.12.5
   ```
 - version of type checker
   ```ps
   poetry run mypy --version  # mypy 1.11.1 (compiled: yes)
   ```
   ```ps
   poetry run pip freeze | findstr pandas-stubs  # pandas-stubs==2.2.2.240603
   ```

# Additional context
Nothing

cmp0xff:
- [ ] Closes #983


cmp0xff:
Hi @Dr-Irv thank you for the comment. In https://github.com/pandas-dev/pandas-stubs/issues/983#issuecomment-2301853575 you mentioned:
> The declaration of `match()` in `core/strings.pyi` is incorrect.

I partially fixed it.
- Now the typing is correct and more accurate for `Series.str.match`.
- However, the return type actually depends on where `str.match` is called.
   - If it is from a `Series`, the return type is `Series[bool]`, as is fixed
   - If it is from an `Index`, the return type is `npt.NDArray[np.bool_]`, as in #983 

How can I capture this expectation?

> But to fix it, the `StringMethods` class will need an additional argument to pass in the expected result of `match()`, similar to what is done with `str.split()`.

I do not get this suggestion. Could you elaborate?
> PR with tests welcome

#990

Dr-Irv:
> How can I capture this expectation?
> 
> > But to fix it, the `StringMethods` class will need an additional argument to pass in the expected result of `match()`, similar to what is done with `str.split()`.
> 
> I do not get this suggestion. Could you elaborate?

In `core/strings.pyi`, we have this:
```python
# The _TS type is what is used for the result of str.split with expand=True
_TS = TypeVar("_TS", DataFrame, MultiIndex)

class StringMethods(NoNewAttributesMixin, Generic[T, _TS]):
```
You need to create an additional `TypeVar`, let's call it `_MR` (for match result).  Then the `Generic` part of `StringMethods` adds `_MR` to the parameters of `Generic`.

Then the `match()` method would return `_MR`.

Then you change the references to `StringMethods` in `core/series.pyi` and `core/indexes/base.pyi` to pass in the correct type of result you want `match` to return as the 3rd argument.

See the pattern used with `split()` - that's how we control the return type of that particular method.

I'll let you address that change before firing off the CI.

cmp0xff:
Hi @Dr-Irv , thanks for the explanations. I have made the changes.

Dr-Irv (changes requested):
In the stubs, instead of using `npt.NDArray[np.bool_]`, can you use `np_ndarray_bool` which you can import from `pandas._typing` .

Dr-Irv (approved):
Thanks @cmp0xff

PR #991, opened 2024-09-03
Validate all values for csv.QUOTE

refack:
4 and 5 added in 3.12 (python/cpython/67230)
ref: https://docs.python.org/3/library/csv.html#csv.QUOTE_NOTNULL
ref: https://github.com/python/cpython/blob/a8bc03696c7c2c03e1d580633151ec7b850366f3/Modules/_csv.c#L88-L106

### DIFF
--- a/pandas-stubs/_typing.pyi
+++ b/pandas-stubs/_typing.pyi
@@ -9,6 +9,7 @@ from collections.abc import (
 )
 import datetime
 from os import PathLike
+import sys
 from typing import (
     Any,
     Literal,
@@ -18,7 +19,6 @@ from typing import (
     TypeVar,
     overload,
 )
-import sys
 
 import numpy as np
 from numpy import typing as npt
--- a/pandas-stubs/_typing.pyi
+++ b/pandas-stubs/_typing.pyi
@@ -738,7 +738,7 @@ JsonSeriesOrient: TypeAlias = Literal["split", "records", "index", "table"]
 TimestampConvention: TypeAlias = Literal["start", "end", "s", "e"]
 
 CSVEngine: TypeAlias = Literal["c", "python", "pyarrow", "python-fwf"]
-CSVQuoting: TypeAlias = Literal[0, 1, 2, 3]
+CSVQuoting: TypeAlias = Literal[0, 1, 2, 3, 4, 5]
 
 HDFCompLib: TypeAlias = Literal["zlib", "lzo", "bzip2", "blosc"]
 ParquetEngine: TypeAlias = Literal["auto", "pyarrow", "fastparquet"]
--- a/pandas-stubs/_typing.pyi
+++ b/pandas-stubs/_typing.pyi
@@ -18,6 +18,7 @@ from typing import (
     TypeVar,
     overload,
 )
+import sys
 
 import numpy as np
 from numpy import typing as npt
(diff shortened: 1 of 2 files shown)


refack:
I didn't consider the fact that this will fail in `python < 3.12` since the functionality is just not there...
I can put this test under a version check condition, but then should there be and `if` in the stub as well, one branch allowing literals up to 3 and the other up to 5?

pandas-stubs/_typing.pyi
```diff
@@ -738,7 +738,7 @@ JsonSeriesOrient: TypeAlias = Literal["split", "records", "index", "table"]
 TimestampConvention: TypeAlias = Literal["start", "end", "s", "e"]
 
 CSVEngine: TypeAlias = Literal["c", "python", "pyarrow", "python-fwf"]
-CSVQuoting: TypeAlias = Literal[0, 1, 2, 3]
+CSVQuoting: TypeAlias = Literal[0, 1, 2, 3, 4, 5]
```
Dr-Irv, line 741: Can you change this to use the references to `csv.MINIMAL`, `csv.QUOTE_ALL`, etc., You can do tests on the python version in a stub, e.g., 
```python
if sys.version_info >= (3, 12):
    CSVQuoting: TypeAlias = Literal[csv.MINIMAL, #rest of values ]
else:
   CSVQuoting: TypeAlias = Literal[csv.MINIMAL, #rest of values]
```

tests/test_frame.py
```diff
@@ -172,6 +172,9 @@ def test_types_to_csv() -> None:
     # Testing support for binary file handles, added in 1.2.0 https://pandas.pydata.org/docs/whatsnew/v1.2.0.html
     df.to_csv(io.BytesIO(), quoting=csv.QUOTE_ALL, encoding="utf-8", compression="gzip")
 
+    with ensure_clean() as path:
```
Dr-Irv, line 175: you'll have to test the python version here as well

Dr-Irv (changes requested):

tests/test_frame.py
refack: done

pandas-stubs/_typing.pyi
Dr-Irv: Oh, so maybe we should use `Union[csv.MINIMAL, csv.QUOTE,  # rest of values]` instead.

refack:
Sorry I'm both dyslexic and ESL 🤣 
image

pandas-stubs/_typing.pyi
```diff
@@ -738,7 +739,20 @@ JsonSeriesOrient: TypeAlias = Literal["split", "records", "index", "table"]
...
+# QUOTE_NONNUMERIC = 2
+# QUOTE_NONE = 3
+# Added in 3.12:
+# QUOTE_STRINGS = 4
+# QUOTE_NOTNULL = 5
+CSVQuotingCompat: TypeAlias = Literal[0, 1, 2, 3]
+if sys.version_info < (3, 12):
+    CSVQuoting: TypeAlias = CSVQuotingCompat
+else:
+    CSVQuoting: TypeAlias = CSVQuotingCompat | Literal[4, 5]
```
Dr-Irv, line 755: Can you try the `Union` idea here, i.e., use `Union[csv.QUOTE_MINIMAL, # other values]` instead of `Literal`

Dr-Irv (changes requested):

pandas-stubs/_typing.pyi
refack: No go.
Both type checker can't resolve the type of the variable at static time, might be an edge case since the Enums are actually defined in the C compiled code, or they might be extra strict with variable values
image

refack:
Can we depend on `typeshed`? They have a stub for the compiled `_csv` code
https://github.com/python/typeshed/pull/12332/files#diff-c58bcb0a60afcd3168bd383c480870da81f23b9cec15e0890cf77d2306323240
(Even they recently moved from `Literal` to `Final`)

Dr-Irv:
> Never mind, the type checkers still don't want to infer the Type at static time, and the only importable Type is just `_QuotingType: TypeAlias = int`
> https://github.com/python/typeshed/blob/e8e9291c76f50c3bcde79e7bb61060f5c24c054e/stdlib/_csv.pyi#L17-L19

Yes, I see that.  I tried some other ideas and they just didn't work.

Dr-Irv (approved):
Thanks @refack , especially for trying the various workarounds.

PR #993, opened 2024-09-03
GH992 Fix pd.DataFrame.pct_change fill_method typing

### LINKED ISSUE #992: DataFrame.pct_change typing for None missing
**Describe the bug**
For the `pd.DataFrame.pct_change` the documentation allows the `None` method but the typing only supports `str`.
This results in an error from pyright to be raised.

**To Reproduce**

```python
df = pd.DataFrame({"x": [1, 2, 2, 3, 3], "y": [10, 20, 30, 40, 50]})
df.pct_change(fill_method=None)
```

This example will yield the error:
```
error: Argument "fill_method" to "pct_change" of "DataFrame" has incompatible type "None"; expected "str"  [arg-type]
```

While the documentation allows for None:
Screenshot 2024-09-03 at 6 04 46 PM

 - OS: MacOs
 - OS Version Sonoma 14
 - python version 3.12.4
 - version of type checker 1.1.378
 - version of installed `pandas-stubs` v2.2.2.240807

**Additional context**

loicdiridollou:
- [x] Closes #992 


pandas-stubs/core/frame.pyi
```diff
@@ -1874,7 +1874,7 @@ class DataFrame(NDFrame, OpsMixin):
     def pct_change(
         self,
         periods: int = ...,
-        fill_method: _str = ...,
+        fill_method: _str | None = ...,
```
Dr-Irv, line 1877: Since the only acceptable argument is now `None` because the string values have been deprecated, can you just change this to 
```suggestion
        fill_method:  None = ...,
```

Dr-Irv (changes requested):

pandas-stubs/core/frame.pyi
loicdiridollou: Wasn't sure how far the support of pandas-stubs would go so I did not want to go too hard but agree with the deprecation!

Dr-Irv (approved):
Thanks @loicdiridollou

PR #995, opened 2024-09-08
Complete signature of value_counts

refack:
For Index, Series, DataFrame

- [X] Closes #445
- [X] Tests added

### DIFF
--- a/tests/test_frame.py
+++ b/tests/test_frame.py
@@ -610,7 +610,12 @@ def test_types_idxmax() -> None:
 # This was added in 1.1.0 https://pandas.pydata.org/docs/whatsnew/v1.1.0.html
 def test_types_value_counts() -> None:
     df = pd.DataFrame(data={"col1": [1, 2], "col2": [1, 4]})
-    df.value_counts()
+    check(assert_type(df.value_counts(), "pd.Series[int]"), pd.Series, np.integer)
+    check(
+        assert_type(df.value_counts(normalize=True), "pd.Series[float]"),
+        pd.Series,
+        float,
+    )
 
 
 def test_types_unique() -> None:
--- a/tests/test_indexes.py
+++ b/tests/test_indexes.py
@@ -1135,3 +1135,13 @@ def test_get_loc() -> None:
         np.ndarray,
         np.bool_,
     )
+
+
+def test_value_counts() -> None:
+    nmi = pd.Index(list("abcb"))
+    check(assert_type(nmi.value_counts(), "pd.Series[int]"), pd.Series, np.integer)
+    check(
+        assert_type(nmi.value_counts(normalize=True), "pd.Series[float]"),
+        pd.Series,
+        float,
+    )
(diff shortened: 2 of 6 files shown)


tests/test_frame.py
```diff
@@ -610,7 +610,10 @@ def test_types_idxmax() -> None:
 # This was added in 1.1.0 https://pandas.pydata.org/docs/whatsnew/v1.1.0.html
 def test_types_value_counts() -> None:
     df = pd.DataFrame(data={"col1": [1, 2], "col2": [1, 4]})
-    df.value_counts()
+    c1: Series[int] = df.value_counts()
+    check(c1, pd.Series, np.integer)
+    c2: Series[float] = df.value_counts(normalize=True)
+    check(c2, pd.Series, float)
```
Dr-Irv, line 616: Can you use the `check(assert_type(...))` pattern here?  We're requiring that on any new tests.

tests/test_indexes.py
```diff
@@ -1135,3 +1135,12 @@ def test_get_loc() -> None:
...
     )
+
+
+def test_value_counts() -> None:
+    non_monotonic_index = pd.Index(list("abcb"))
+    c1: pd.Series[int] = non_monotonic_index.value_counts()
+    check(c1, pd.Series, np.integer)
+
+    c2: pd.Series[float] = non_monotonic_index.value_counts(normalize=True)
+    check(c2, pd.Series, float)
```
Dr-Irv, line 1146: Same comment as above with `check(assert_type(...`

tests/test_series.py
```diff
@@ -556,7 +556,10 @@ def test_types_idxmax() -> None:
 
 def test_types_value_counts() -> None:
     s = pd.Series(["a", "b"])
-    check(assert_type(s.value_counts(), "pd.Series[int]"), pd.Series, np.integer)
+    c1: pd.Series[int] = s.value_counts()
+    check(c1, pd.Series, np.integer)
+    c2: pd.Series[float] = s.value_counts(normalize=True)
+    check(c2, pd.Series, float)
```
Dr-Irv, line 562: And here too with `check(assert_type(...`

Dr-Irv (changes requested):
Changes needed to use the `check(assert_type(...))` pattern in new tests.

tests/test_series.py
refack: Done all three

Dr-Irv (approved):
Thanks @refack

refack:
> We're requiring that on any new tests.

I did that, but those are suboptimal for my setup. Assignment to a type hinted variable is analyzed in PyCharm, while annotating with `assert_type` is transparent since the signature is `assert_type(val: T_, typ: Any, /) -> T_`
Screenshot 2024-09-09 151809

Both forms err for pyright/mypy.
Screenshot 2024-09-09 152247

and since syntactically the type parameter is not in a type-hint positions, we need to do the "string" trick
Screenshot 2024-09-09 153144

Dr-Irv:
> I did that, but those are suboptimal for my setup. Assignment to a type hinted variable is analyzed in PyCharm, while annotating with `assert_type` is transparent since the signature is `assert_type(val: T_, typ: Any, /) -> T_`

That's an issue with the type checker within PyCharm (which we don't support) having a different behavior than `mypy` and `pyright` .  So with `assert_type()`, we have to surround the types that are not available at runtime (e.g., `Series[int]`) in quotes.

PR #997, opened 2024-09-11
Allow tuple-valued params in read_sql[_query]

### LINKED ISSUE #996: read_sql_query should support tuples as `params` dict values
**Describe the bug**
The type for [`pandas.read_sql_query`](https://pandas.pydata.org/docs/reference/api/pandas.read_sql_query.html) is specified by `pandas-stubs` as a dictionary mapping from `str` to `str | bytes | date | timedelta | datetime64 | timedelta64 | int | float | complex`. But for some database drivers like `psycopg2` or `mysql-connector`, it's also possible to pass in tuples for `IN` (`WHERE ... IN (1, 2, 3)`).

**To Reproduce**
A minimal example using Postgres (it looks similar for MySQL except the parameter is written as `:ids` in the query):

1. Launch a Postgres Docker container: `docker run -p 9876:5432 -e POSTGRES_HOST_AUTH_METHOD=trust postgres`
2. Install dependencies and run this script to see that tuple substitutions for `IN` conditions work fine (I included [PEP 723](https://peps.python.org/pep-0723/) script metadata so you can just do `uv run` or `pipx run` on it to avoid needing to install dependencies manually):

```python
# To make this executable directly w/ pipx run or uv run:
# /// script
# dependencies = [
#   "psycopg2==2.9.9",
#   "pandas==2.2.2"
# ]
# ///

import pandas as pd
import psycopg2

conn = psycopg2.connect(
  host="localhost", dbname="postgres", user="postgres", port=9876
)
cur = conn.cursor()
cur.execute("CREATE TABLE test (id serial PRIMARY KEY)")
cur.execute("INSERT INTO test(id) VALUES(1)")
cur.execute("INSERT INTO test(id) VALUES(2)")
cur.execute("INSERT INTO test(id) VALUES(3)")

df = pd.read_sql_query(
  "SELECT * FROM test WHERE id IN %(ids)s",
  conn,
  params={"ids": (1, 2)},
)
print(df)

```
Output is as expected:
```
   id
0   1
1   2
```

3. Run Mypy on it and get this error message:
```
bug.py:24: error: Dict entry 0 has incompatible type "str": "list[int]"; expected "str": "str | bytes | date | timedelta | datetime64 | timedelta64 | int | float | complex"  [dict-item]
```

 - OS: Linux
 - OS Version: Ubuntu 22.04
 - python version: 3.10
 - version of type checker: Mypy 1.11.2
 - version of installed `pandas-stubs`: 2.2.2.240909

**Additional context**
Something like this doesn't work for `sqlite3` connections, because it really does only accept "primitive" values for substitutions. 

But IMO it doesn't make sense to reduce the set of types to the minimal supported subset, because some other weird database could decide not even to support `int` substitutions or something like that.

sh-at-cs:
- [x] Closes #996
- [ ] Tests added: Kind of:
  - As explained in https://github.com/pandas-dev/pandas-stubs/issues/996, the added types are _not_ valid for SQLite databases/drivers, so it's not possible to add a test that actually executes without major changes to the CI setup. But I've added a skipped test instead so that at least the [type checking part](https://github.com/pandas-dev/pandas-stubs/blob/fa7e4446d226ee385412186bc3414a3499f42917/.github/workflows/test.yml#L35-L39) of the CI pipeline can check if the new annotations are at all sane.

### DIFF
--- a/pandas-stubs/io/sql.pyi
+++ b/pandas-stubs/io/sql.pyi
@@ -66,7 +66,14 @@ def read_sql_query(
     con: _SQLConnection,
     index_col: str | list[str] | None = ...,
     coerce_float: bool = ...,
-    params: list[Scalar] | tuple[Scalar, ...] | Mapping[str, Scalar] | None = ...,
+    params: (
+        list[Scalar]
+        | tuple[Scalar, ...]
+        | tuple[tuple[Scalar, ...], ...]
+        | Mapping[str, Scalar]
+        | Mapping[str, tuple[Scalar, ...]]
+        | None
+    ) = ...,
     parse_dates: list[str] | dict[str, str] | dict[str, dict[str, Any]] | None = ...,
     *,
     chunksize: int,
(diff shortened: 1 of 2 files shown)


sh-at-cs:
Sorry, I ran my system black on it which had a different version and hence slightly different formatting. Should be fixed now - at least pre-commit runs without errors.

Dr-Irv (approved):
thanks @sh-at-cs

PR #1003, opened 2024-09-13
Add datetime.date and datetime.datetime specializations

### LINKED ISSUE #755: BDay typing
I was using pylance to check types and the `date + BDay` stubs seems to be incorrect
All 3 (`__rsub__`, `__add__`, `__radd__`) should return `pd.Timestamp`, but instead they are inferred as `datetime.date`:
```python
datetime.date(2023, 1, 1) + BDay(2)
datetime.date(2023, 1, 1) - BDay(2)
BDay(2) + datetime.date(2023, 1, 1)
```

https://github.com/microsoft/pylance-release/issues/4631

alippai:
- [x] Closes #755
- [ ] Tests added:


twoertwein:
Can you please add runtime tests in https://github.com/pandas-dev/pandas-stubs/blob/main/tests/test_timefuncs.py

You will also need to re-order the overloads to have the more specifiy overloads before more general overloads as type checker pick the first fitting overload (see the mypy errors).

alippai:
@twoertwein thanks, I've updated the failing tests and added the cases brought up in the initial issue. Looks like `datetime` is a subclass of `date` so I just removed that, it could fix the overloads.

Dr-Irv (changes requested):
Your tests are still failing.  I think the issue here is that this line is incorrect in `offsets.pyi`:
```python
_DatetimeT = TypeVar("_DatetimeT", bound=date)
```
I think it should be
```python
_DatetimeT = TypeVar("_DatetimeT", bound=datetime)
```
based on what is now in the pandas source.

Can you try that change to see if it will make things work?  Because the error you are getting is that this overload:  `def __add__(self, other: _DatetimeT) -> _DatetimeT: ...` will never get matched, because you put in something explicit for `date`, but `datetime` is a subclass of `date`, so I think changing the `TypeVar` line will fix it.

alippai:
Looks like this is not possible to describe correctly: https://github.com/microsoft/pylance-release/issues/6512

Dr-Irv:
> Looks like this is not possible to describe correctly: [microsoft/pylance-release#6512](https://github.com/microsoft/pylance-release/issues/6512)

I'm OK with putting in the appropriate `#type: ignore` and `#pyright: ignore` statements when the type checkers are reporting overlapping overloads.

alippai:
@Dr-Irv thanks for the guide, I believe the last commit may work as intended and pass the CI

Dr-Irv (approved):
thanks @alippai

PR #1005, opened 2024-09-22
GH677 Allow pd.RangeIndex to be initialized with range

### LINKED ISSUE #677: `pd.RangeIndex` should accept `range` as the first argument
**Describe the bug**
Both according to the docs and the implementation it is valid to pass a python `range` object as the first argument to `pd.RangeIndex`

**To Reproduce**
```python
import pandas as pd

pd.RangeIndex(range(5))  # mypy: arg-type – Argument 1 to "RangeIndex" has incompatible type "range"
```

 - Python 3.10
 - mypy 1.0.1
 - pandas-stubs 2.0.1.230501

loicdiridollou:
- [x] Closes #677 

### DIFF
--- a/pandas-stubs/core/indexes/range.pyi
+++ b/pandas-stubs/core/indexes/range.pyi
@@ -22,15 +22,6 @@ class RangeIndex(Index[int]):
         copy: bool = ...,
         name=...,
     ): ...
-    def __init__(
-        self,
-        start: int | RangeIndex | range = ...,
-        stop: int = ...,
-        step: int = ...,
-        dtype=...,
-        copy: bool = ...,
-        name=...,
-    ) -> None: ...
     @classmethod
     def from_range(cls, data, name=..., dtype=...): ...
     def __reduce__(self): ...
--- a/pandas-stubs/core/indexes/range.pyi
+++ b/pandas-stubs/core/indexes/range.pyi
@@ -15,7 +15,7 @@ from pandas._typing import (
 class RangeIndex(Index[int]):
     def __new__(
         cls,
-        start: int | RangeIndex = ...,
+        start: int | RangeIndex | range = ...,
         stop: int = ...,
         step: int = ...,
         dtype=...,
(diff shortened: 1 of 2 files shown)


pandas-stubs/core/indexes/range.pyi
```diff
@@ -24,7 +24,7 @@ class RangeIndex(Index[int]):
     ): ...
     def __init__(
```
loicdiridollou, line 25: Also quick note here, the pandas implementation does not have a `__init__` method so we may want to remove that from pandas-stubs, @Dr-Irv let me know if you think it is good to drop.

loicdiridollou: https://github.com/pandas-dev/pandas/blob/2419343bfea5dba678146139ca9663d831c47b22/pandas/core/indexes/range.py#L150-L158

loicdiridollou:
\pandas_nightly

pandas-stubs/core/indexes/range.pyi
twoertwein: You can drop __init__.

twoertwein (approved):
Happy to merge or wait if you want to remove `__init__` first

loicdiridollou:
> Happy to merge or wait if you want to remove `__init__` first

I deleted the `__init__` method, just waiting for CI to finish but it is looking ready from a code perspective.

twoertwein:
Thanks @loicdiridollou !

Dr-Irv:
FYI, I think we should have kept `__init__()` and deleted `__new__()`.  In the stubs, we only need to have `__new__()` if we want to have different overloads that produce different results.  Otherwise `__init__()` is sufficient.

Admittedly, we're inconsistent in the stubs about that. There are times we use `__new__()` instead of `__init__()`

twoertwein:
> Admittedly, we're inconsistent in the stubs about that. There are times we use `__new__()` instead of `__init__()`

If we want to become more stubtest compatible (aligning pandas and pandas-stubs), we should use which ever constructor pandas uses, at least when possible.

PR #1010, opened 2024-09-29
GH648 Allowing dateoffset weekday from relativedelta

### LINKED ISSUE #648: `DateOffset` for `weekday` argument requires `weekday` argument from `dateutil`
> `DateOffset`'s docs don't say it allows taking a dateutil `weekday` object, but that's what was documented. I'm a bit out of my depth, I'm not sure how much I should expect the pandas docs to be correct/complete...

The problem is twofold.  One is that the pandas docs aren't documenting that we would accept a `dateutil._common.weekday` type as a value for the `weekday` parameter to `DateOffset`.  The second issue is that the type specification for the `weekday` argument in `pandas-stubs/_libs/tslibs/offsets.pyi` says it is an `int`.  You could fix the latter, but then we have to make `types-python-dateutil` a dependency, which involves modifying `pyproject.toml` and takes some care.

Created a docs issue for pandas at https://github.com/pandas-dev/pandas/issues/52650

_Originally posted by @Dr-Irv in https://github.com/pandas-dev/pandas-stubs/issues/645#issuecomment-1507053774_

loicdiridollou:
- [x] Closes #648 

### DIFF
--- a/tests/test_timefuncs.py
+++ b/tests/test_timefuncs.py
@@ -8,6 +8,15 @@ from typing import (
     cast,
 )
 
+from dateutil.relativedelta import (
+    FR,
+    MO,
+    SA,
+    SU,
+    TH,
+    TU,
+    WE,
+)
 import numpy as np
 from numpy import typing as npt
 import pandas as pd
@@ -19,7 +28,12 @@ from typing_extensions import (
 
 from pandas._libs import NaTType
 from pandas._libs.tslibs import BaseOffset
-from pandas._libs.tslibs.offsets import DateOffset
+from pandas._libs.tslibs.offsets import (
+    DateOffset,
+    FY5253Mixin,
+    WeekOfMonth,
+    WeekOfMonthMixin,
+)
 
 if TYPE_CHECKING:
     from pandas._typing import FulldatetimeDict
@@ -1282,6 +1296,21 @@ def test_weekofmonth_init():
         ),
         pd.offsets.WeekOfMonth,
     )
+
+
+def test_dateoffset_weekday():
+    """Check that you can create a `pd.DateOffset` from weekday of int or dateutil.relativedelta."""
+    # check for int
+    check(
+        assert_type(pd.offsets.DateOffset(weekday=1), pd.offsets.DateOffset),
+        pd.offsets.DateOffset,
+    )
+    # check for relativedelta
+    for weekday in [MO, TU, WE, TH, TH, FR, SA, SU]:
+        check(
+            assert_type(pd.offsets.DateOffset(weekday=weekday), pd.offsets.DateOffset),
+            pd.offsets.DateOffset,
+        )
 
 
 def test_date_range_unit():
--- a/tests/test_timefuncs.py
+++ b/tests/test_timefuncs.py
@@ -16,10 +16,12 @@ from dateutil.relativedelta import (
     TH,
     TU,
     WE,
+    weekday as WeekdayClass,
 )
 import numpy as np
 from numpy import typing as npt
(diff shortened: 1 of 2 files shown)


tests/test_timefuncs.py
```diff
@@ -1284,6 +1295,15 @@ def test_weekofmonth_init():
     )
 
 
+@pytest.mark.parametrize("weekday", [1, MO, TU, WE, TH, TH, FR, SA, SU])
+def test_dateoffset_weekday(weekday: int | WeekdayClass):
+    """Check that you can create a `pd.DateOffset` from weekday of int or relativedelta.weekday."""
+    check(
+        assert_type(pd.offsets.DateOffset(weekday=weekday), pd.offsets.DateOffset),
```
Dr-Irv, line 1302: You can't use `parameterize` to do a type checking test.  

You will need to have explicit lines with `assert_type()` for each value you wish to test.

Dr-Irv (changes requested):

tests/test_timefuncs.py
loicdiridollou: I am not sure I understand what you mean here.
Having parametrize is (I believe) equivalent to splitting the test in individual tests for all the given parameters. So in my mind this is testing the type individually for each value. Maybe the goal is to make it more readable by having one line for each, if so happy to change it, but let me know if I miss anything with `parametrize`.
When I run the tests by hand here is what I see:
```
 pytest tests/test_timefuncs.py::test_dateoffset_weekday -v
========================== test session starts ===========================
platform darwin -- Python 3.12.4, pytest-8.3.2, pluggy-1.5.0 -- /Users/loic.diridollou/Documents/Code/pandas-stubs/venv/bin/python
cachedir: .pytest_cache
rootdir: /Users/loic.diridollou/Documents/Code/pandas-stubs
configfile: pyproject.toml
collected 9 items

tests/test_timefuncs.py::test_dateoffset_weekday[1] PASSED         [ 11%]
tests/test_timefuncs.py::test_dateoffset_weekday[weekday1] PASSED  [ 22%]
tests/test_timefuncs.py::test_dateoffset_weekday[weekday2] PASSED  [ 33%]
tests/test_timefuncs.py::test_dateoffset_weekday[weekday3] PASSED  [ 44%]
tests/test_timefuncs.py::test_dateoffset_weekday[weekday4] PASSED  [ 55%]
tests/test_timefuncs.py::test_dateoffset_weekday[weekday5] PASSED  [ 66%]
tests/test_timefuncs.py::test_dateoffset_weekday[weekday6] PASSED  [ 77%]
tests/test_timefuncs.py::test_dateoffset_weekday[weekday7] PASSED  [ 88%]
tests/test_timefuncs.py::test_dateoffset_weekday[weekday8] PASSED  [100%]

=========================== 9 passed in 0.21s ============================

Dr-Irv: > Having parametrize is (I believe) equivalent to splitting the test in individual tests for all the given parameters. So in my mind this is testing the type individually for each value. Maybe the goal is to make it more readable by having one line for each, if so happy to change it, but let me know if I miss anything with `parametrize`.

The issue is that `parameterize` applies to when `pytest` is run on the file.

But when we test the files with `pyright` and `mypy`, the type checkers don't test that each of those arguments is valid, and that is the main purpose of the tests.

loicdiridollou: Oh I get it now, sorry for missing that. I made all the tests explicit so should be good to go. Let me know if something is missing.

Dr-Irv (approved):
thanks @loicdiridollou

PR #1011, opened 2024-10-04
GH852 Typehint labelsize

### LINKED ISSUE #852: `labelsize` arguments need fixes
Oh , the upstream docs of this, and other label size usages, will need to be updated to accept `float | str | None` (See the [matplotlib docs](https://matplotlib.org/stable/api/_as_gen/matplotlib.axes.Axes.tick_params.html)). I typed them as `int | None` here for now because they need to be fixed everywhere in pandas-stubs, not just this file. We can do that in a separate PR.

_Originally posted by @hamdanal in https://github.com/pandas-dev/pandas-stubs/pull/848#discussion_r1451472602_
            
Probably needs fixes in the stubs and in pandas docs

loicdiridollou:
- [x] Closes #852 


Dr-Irv (approved):
thanks @loicdiridollou

PR #1012, opened 2024-10-08
GH935 Create overload for pd.Index.rename

### LINKED ISSUE #935: Index.rename() should return Self
**Describe the bug**
`Index.rename()` should return `Self`

**To Reproduce**
```python
import pandas as pd

ind = pd.Index([1,2,3], name="foo")
ind2 = ind.rename("goo")
reveal_type(ind2)
```
shows that `ind2` is unknown

 - OS: Windows 10
 - OS Version ??
 - python version  3.9
 - version of type checker  pyright 1.1.362
 - version of installed `pandas-stubs` :  2.2.2.240514

loicdiridollou:
- [x] Closes #935 


tests/test_indexes.py
```diff
@@ -121,11 +121,20 @@ def test_str_match() -> None:
...
     ind = pd.Index([1, 2, 3], name="foo")
     ind2 = ind.rename("goo")
     check(assert_type(ind2, "pd.Index[int]"), pd.Index, np.integer)
 
 
+def test_index_rename_inplace() -> None:
+    """Test that index rename in-place does not return anything (None)."""
+    ind = pd.Index([1, 2, 3], name="foo")
+    ind2 = ind.rename("goo", inplace=True)
+    check(assert_type(ind2, None), type(None))
```
loicdiridollou, line 134: Not totally sure about the structure when there is a None return, will double check with another example from the code.

PR #1013, opened 2024-10-08
Bump to pandas 2.2.3

Dr-Irv:
Test with pandas 2.2.3.  Seems like `show_versions()` got fixed.


twoertwein (approved):

twoertwein:
Thanks @Dr-Irv always good when we can remove workarounds!

PR #1014, opened 2024-10-13
GH203 Split groupby with as_index (temptative)

### LINKED ISSUE #203: BUG: df.groupby(col, as_index=False).value_counts() returns a DataFrame but is annotated as Series

loicdiridollou:
- [x] Closes #203 


loicdiridollou:
I am still having overlap issues in the overload but looking into it, feel free to suggest something else.
Right now just done for the Scalar but can/should be expanded to the other types.

Dr-Irv (changes requested):
Good start, but I think you will have to work on the following:

- Have to follow the suggestion at https://github.com/pandas-dev/pandas-stubs/issues/203#issuecomment-1218439288 , item 2, to add a `Generic` parameter to `DataFrameGroupBy` that essentially represents whether `as_index` is `True` or `False`
- You will have to do the same separation on the overloads for `groupby()` with the other `by` arguments, and add appropriate tests.

loicdiridollou:
Seems like this is going to take a little longer, I think I understand your suggestion in #203, let me try to come up with something.

tests/test_frame.py
```diff
@@ -1048,7 +1066,7 @@ def test_types_groupby() -> None:
 
     df1: pd.DataFrame = df.groupby(by="col1").agg("sum")
     df2: pd.DataFrame = df.groupby(level="ind").aggregate("sum")
-    df3: pd.DataFrame = df.groupby(by="col1", sort=False, as_index=True).transform(
+    df3: pd.Series = df.groupby(by="col1", sort=False, as_index=True).transform(
```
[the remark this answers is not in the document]
loicdiridollou: Agreed, I will tackle those once we agree on the potential solution so I can do one thing at a time not to confuse the amount of code to review too much.

pandas-stubs/core/frame.pyi
```diff
@@ -229,6 +235,11 @@ class _LocIndexerFrame(_LocIndexer):
         value: Scalar | NAType | NaTType | ArrayLike | Series | list | None,
     ) -> None: ...
 
+_TT = TypeVar("TT", bound=Literal[True, False])
+
+class DataFrameGroupByGen(DataFrameGroupBy[ByT], Generic[ByT, _TT]): ...
```
loicdiridollou, line 240: @Dr-Irv these seem to be more what you had originally in mind, please confirm if that is what you were thinking

pandas-stubs/core/groupby/groupby.pyi
```diff
@@ -235,8 +239,10 @@ class GroupBy(BaseGroupBy[NDFrameT]):
     @final
     @overload
     def size(self: GroupBy[Series]) -> Series[int]: ...
-    @overload  # return type depends on `as_index` for dataframe groupby
-    def size(self: GroupBy[DataFrame]) -> DataFrame | Series[int]: ...
+    @overload
+    def size(self: DataFrameGroupByGen[ByT, Literal[True]]) -> Series[int]: ...  # type: ignore[misc]
```
loicdiridollou, line 243: Surprisingly here `mypy` throws the error that the DataFrameGroupByGen is not a supertype of DataFrameGroupByGen, I have ignored the error but open to a better solution.

pandas-stubs/core/frame.pyi
Dr-Irv: No, I was thinking of doing this:
```python
class DataFrameGroupBy(GroupBy[DataFrame], Generic[ByT, _TT]):
```
No need to introduce a new class

pandas-stubs/core/frame.pyi
```diff
@@ -229,6 +235,11 @@ class _LocIndexerFrame(_LocIndexer):
...
+class SeriesGroupByGen(SeriesGroupBy, Generic[_TT, ByT]): ...
```
Dr-Irv, line 241: Same here - you can have more than one parameter to the generic nature.

pandas-stubs/core/groupby/groupby.pyi
Dr-Irv: I think the solution here is to push `def size` down to the `DataFrameGroupBy` and `SeriesGroupBy` classes, and remove it from the `GroupBy` class.

Dr-Irv (changes requested):
Hope my comments help.  Note - I will be not be able to give feedback until after Wed., October 23.

Dr-Irv (changes requested):
I think this is looking good, but CI is failing due to overlapping overload.  You'll have to use `# type: ignore` statements to get around that.

Dr-Irv:
So I am fine with this.  Do you want me to merge it so that `size()` is fixed?  Or do you want to work on other methods where the type of the result depends on `as_index` parameter?

loicdiridollou:
I will open another issue to track down the other methods, would rather merge small changes than one large PR.

Dr-Irv (approved):
thanks @loicdiridollou

PR #1020, opened 2024-10-30
Update enums to reflect typing spec changes

hauntsaninja:
See https://typing.readthedocs.io/en/latest/spec/enums.html#defining-members

https://github.com/python/typing-council/issues/11

The next version of mypy will obey the spec change: https://github.com/python/mypy/pull/18068

pyright already requires this


Dr-Irv:
I think the CI is failing on Windows because of a `pyright` issue that was just fixed.  Can  you update the `pyproject.toml` to require `pyright >= 1.1.387` ?

Dr-Irv (approved):
Thanks @hauntsaninja

PR #1026, opened 2024-11-02
accept float values in pd.Timedelta

### LINKED ISSUE #1015: `pd.Timedelta` should also accept `float` values
The `value` parameter of `Timedelta` currently has the following type-hint:
https://github.com/pandas-dev/pandas-stubs/blob/f71224c6211ff436424be614ba47f494f64c4618/pandas-stubs/_libs/tslibs/timedeltas.pyi#L105

I think it should also accept float, so that type checkers won't complain on following code snippet:
```python
import pandas as pd
pd.Timedelta(3.5, 'hours')
```

```
Argument of type "float" cannot be assigned to parameter "value" of type "str | int | Timedelta | timedelta | timedelta64" in function "__new__"
  Type "float" is not assignable to type "str | int | Timedelta | timedelta | timedelta64"
    "float" is not assignable to "str"
    "float" is not assignable to "int"
    "float" is not assignable to "Timedelta"
    "float" is not assignable to "timedelta"
    "float" is not assignable to "timedelta64"
Pylance[reportArgumentType](https://github.com/microsoft/pyright/blob/main/docs/configuration.md#reportArgumentType)
```

5j9:
- [X] Closes #1015
- [X] Tests added

### DIFF
--- a/pandas-stubs/_libs/tslibs/timedeltas.pyi
+++ b/pandas-stubs/_libs/tslibs/timedeltas.pyi
@@ -102,7 +102,7 @@ class Timedelta(timedelta):
     value: int
     def __new__(
         cls,
-        value: str | int | Timedelta | timedelta | np.timedelta64 | float = ...,
+        value: str | float | Timedelta | timedelta | np.timedelta64 = ...,
         unit: TimeDeltaUnitChoices = ...,
         *,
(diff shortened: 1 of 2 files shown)


Dr-Irv (approved):
Thanks @5j9

PR #1027, opened 2024-11-06
GH1025 Typehint DataFrame.groupby.value_counts method

### LINKED ISSUE #1025: groupby methods can return Series or DataFrame based on `as_index` variable
**Describe the bug**
Depending on the boolean `as_index`, methods on groupby objects will return different type.
This is a follow up for #203,

**To Reproduce**

 - OS Version [e.g. 22]
 - python version
 - version of type checker

**Additional context**

loicdiridollou:
- [x] Closes #1025 

### DIFF
--- a/pandas-stubs/core/groupby/generic.pyi
+++ b/pandas-stubs/core/groupby/generic.pyi
@@ -300,7 +300,6 @@ class DataFrameGroupBy(GroupBy[DataFrame], Generic[ByT, _TT]):
         backend: str | None = ...,
         **kwargs,
     ) -> PlotAxes | Series: ...  # Series[PlotAxes]
-    # def size(self: DataFrameGroupBy[ByT, Literal[True]]) -> Series[int]: ...
     @overload
     def value_counts(
         self: DataFrameGroupBy[ByT, Literal[True]],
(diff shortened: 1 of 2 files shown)


pandas-stubs/core/groupby/generic.pyi
```diff
@@ -300,9 +300,10 @@ class DataFrameGroupBy(GroupBy[DataFrame], Generic[ByT, _TT]):
         backend: str | None = ...,
         **kwargs,
     ) -> PlotAxes | Series: ...  # Series[PlotAxes]
+    # def size(self: DataFrameGroupBy[ByT, Literal[True]]) -> Series[int]: ...
```
Dr-Irv, line 303: Delete the commented out code.

Dr-Irv (changes requested):

pandas-stubs/core/groupby/generic.pyi
loicdiridollou: My bad, I copied out the `size` method to make sure it was done right, should be good now!

Dr-Irv (approved):
Thanks @loicdiridollou

PR #1029, opened 2024-11-10
GH571 Update typehint when creating a Series from an empty list

### LINKED ISSUE #571: Empty Series is considered `TimestampSeries`
**Describe the bug**

Instantiating an empty `Series` object creates a `TimestamptSeries`, should probably be `Series[Any]`.

**To Reproduce**

```python
from pandas import Series
baz = Series([])
reveal_type(baz)  # pandas.core.series.TimestampSeries
```

- pandas                        1.5.3
- pandas-stubs                  1.5.3.230304

loicdiridollou:
- [x] Closes #571

Potential fix for #571 to typehint a `Series` created from an empty list to be typed as `pd.Series[Any]`.
Open for feedback if this is not the optimal type.

### DIFF
--- a/tests/test_frame.py
+++ b/tests/test_frame.py
@@ -7,6 +7,7 @@ from collections.abc import (
     Iterator,
     Mapping,
     MutableMapping,
+    Sequence,
 )
 import csv
 import datetime
@@ -21,6 +22,7 @@ from typing import (
     Any,
     Callable,
     Generic,
+    Never,
     TypedDict,
     TypeVar,
     Union,
@@ -3556,3 +3558,9 @@ def test_series_typed_dict() -> None:
     my_dict = MyDict(a="", b="")
     sr = pd.Series(my_dict)
     check(assert_type(sr, pd.Series), pd.Series)
+
+
+def test_series_empty_dtype() -> None:
+    """Test for the creation of a Series from an empty list GH571."""
+    new_tab: Sequence[Never] = []
+    check(assert_type(pd.Series(new_tab), "pd.Series[Any]"), pd.Series)
--- a/tests/test_frame.py
+++ b/tests/test_frame.py
@@ -22,7 +22,6 @@ from typing import (
     Any,
     Callable,
     Generic,
-    Never,
     TypedDict,
     TypeVar,
     Union,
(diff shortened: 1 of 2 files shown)


tests/test_frame.py
```diff
@@ -3556,3 +3558,9 @@ class MyDict(TypedDict):
     my_dict = MyDict(a="", b="")
     sr = pd.Series(my_dict)
     check(assert_type(sr, pd.Series), pd.Series)
+
+
+def test_series_empty_dtype() -> None:
+    """Test for the creation of a Series from an empty list GH571."""
+    new_tab: Sequence[Never] = []
```
Dr-Irv, line 3565: For the test, it should work without setting a type for `new_tab`.  So if you change this line to `new_tab = []`, does the empty list match `Sequence[Never]` ?

also, double check that passing an empty string will NOT match this, because strings are sequences, and I'm not sure if `""` would be an OK match for `Sequence[Never]`

loicdiridollou: Thanks for the feedback, in order:
- `mypy` raises an error when instantiating an empty sequence so I needed to annotate it
```

tests/test_frame.py:3566: error: Need type annotation for "new_tab" (hint: "new_tab: list[<type>] = ...")  [var-annotated]
Found 1 error in 1 file (checked 226 source files)
```
- I have added a test for the empty string

loicdiridollou: FYI this is the error that you get when trying to typehint `""` as a `Sequence[Never]` so the types are well separated.
```
tests/test_frame.py:3571: error: Incompatible types in assignment (expression has type "str", variable has type "Sequence[Never]")  [assignment]
```
for mypy above and pyright below
```
Diagnostics: 
Type "Literal['']" is not assignable to declared type "Sequence[Never]"  
"Literal['']" is not assignable to "Sequence[Never]"                         
Type parameter "_T_co@Sequence" is covariant, but "str" is not a subtype of "Never"
Type "str" is not assignable to type "Never" [reportAssignmentType]
```

Dr-Irv: OK, I see that you added an explicit test for `pd.Series([])`, so the issue with `new_tab` is nothing to worry about.

Dr-Irv (approved):
thanks @loicdiridollou

PR #1030, opened 2024-11-12
Add `ignore_index` keyword arg in `dropna` and `drop_duplicates` (Part of GH624)

hyperc54:
- [x] Should cover two bullet points of #624 (dropna and drop_duplicates for both Series and DFs) 
- [x] Tests added

- Additionally add overloads to dataframe `drop_duplicates` method
- Additionally convert existing tests of DF `drop_na` method to use `assert_type`.

### DIFF
--- a/pandas-stubs/core/frame.pyi
+++ b/pandas-stubs/core/frame.pyi
@@ -880,6 +880,7 @@ class DataFrame(NDFrame, OpsMixin):
         thresh: int | None = ...,
         subset: ListLikeU | Scalar | None = ...,
         inplace: Literal[True],
+        ignore_index: _bool = ...,
     ) -> None: ...
     @overload
     def dropna(
@@ -890,6 +891,7 @@ class DataFrame(NDFrame, OpsMixin):
         thresh: int | None = ...,
         subset: ListLikeU | Scalar | None = ...,
         inplace: Literal[False] = ...,
+        ignore_index: _bool = ...,
     ) -> DataFrame: ...
     @overload
     def dropna(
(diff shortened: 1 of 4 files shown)


pandas-stubs/core/frame.pyi
```diff
@@ -900,15 +902,35 @@ class DataFrame(NDFrame, OpsMixin):
...
+    @overload
+    def drop_duplicates(
+        self,
+        subset: Hashable | Iterable[Hashable] | None = ...,
+        *,
+        keep: NaPosition | _bool = ...,
+        inplace: Literal[True] = ...,
+        ignore_index: _bool = ...,
+    ) -> None: ...
+    @overload
```
Dr-Irv, line 925: Reverse these two overloads.  Pattern should look like what is in `fillna()`.  First you'd have `inplace: Literal[True]` (without the ellipses), then you'd have `Literal[False] = ...`

pandas-stubs/core/series.pyi
```diff
@@ -726,15 +726,27 @@ class Series(IndexOpsMixin[S1], NDFrame):
...
     ) -> Series[S1]: ...
     @overload
     def drop_duplicates(
-        self, *, keep: NaPosition | Literal[False] = ..., inplace: Literal[True]
+        self,
+        *,
+        keep: NaPosition | Literal[False] = ...,
+        inplace: Literal[True],
+        ignore_index: _bool = ...,
     ) -> None: ...
```
Dr-Irv, line 742: Same comment as within `frame.pyi` - reverse the 2 overloads (looks like they were not using the consistent pattern before)

hyperc54:
Right thanks! definitely missed the subtlety. I applied the suggestions.

Dr-Irv (approved):
thanks @hyperc54

PR #1035, opened 2024-11-16
GH1033 Add overloads of engine for pd.read_json

### LINKED ISSUE #1033: Added new argument engine to read_json() to support parsing JSON with pyarrow by specifying engine="pyarrow"
Part of the migration of `pandas-stubs` to pandas 2.0.
See #624

loicdiridollou:
- [x] Closes #1033 

### DIFF
--- a/pandas-stubs/io/json/_json.pyi
+++ b/pandas-stubs/io/json/_json.pyi
@@ -42,7 +42,7 @@ def read_json(
         Literal["strict", "ignore", "replace", "backslashreplace", "surrogateescape"]
         | None
     ) = ...,
-    lines: bool,
+    lines: Literal[True],
     chunksize: int,
     compression: CompressionOptions = ...,
     nrows: int | None = ...,
@@ -92,7 +92,7 @@ def read_json(
         Literal["strict", "ignore", "replace", "backslashreplace", "surrogateescape"]
         | None
     ) = ...,
-    lines: Literal[True] = True,
+    lines: Literal[True],
     chunksize: int,
     compression: CompressionOptions = ...,
     nrows: int | None = ...,
@@ -117,7 +117,7 @@ def read_json(
         Literal["strict", "ignore", "replace", "backslashreplace", "surrogateescape"]
         | None
     ) = ...,
-    lines: Literal[True] = True,
+    lines: Literal[True],
     chunksize: int,
     compression: CompressionOptions = ...,
     nrows: int | None = ...,
@@ -167,7 +167,7 @@ def read_json(
         Literal["strict", "ignore", "replace", "backslashreplace", "surrogateescape"]
         | None
     ) = ...,
-    lines: Literal[True] = True,
+    lines: Literal[True],
     chunksize: None = ...,
     compression: CompressionOptions = ...,
     nrows: int | None = ...,
@@ -217,7 +217,7 @@ def read_json(
         Literal["strict", "ignore", "replace", "backslashreplace", "surrogateescape"]
         | None
     ) = ...,
-    lines: Literal[True] = True,
+    lines: Literal[True],
     chunksize: None = ...,
     compression: CompressionOptions = ...,
     nrows: int | None = ...,
--- a/pandas-stubs/io/json/_json.pyi
+++ b/pandas-stubs/io/json/_json.pyi
@@ -98,7 +98,32 @@ def read_json(
     nrows: int | None = ...,
     storage_options: StorageOptions = ...,
     dtype_backend: DtypeBackend | NoDefault = ...,
-    engine: Literal["pyarrow"] = "pyarrow",
+    engine: Literal["ujson"] = ...,
+) -> JsonReader[DataFrame]: ...
+@overload
+def read_json(
+    path_or_buf: FilePath | ReadBuffer[bytes],
+    *,
+    orient: JsonFrameOrient | None = ...,
+    typ: Literal["frame"] = ...,
+    dtype: bool | Mapping[HashableT, DtypeArg] | None = ...,
+    convert_axes: bool | None = ...,
+    convert_dates: bool | list[str] = ...,
+    keep_default_dates: bool = ...,
+    precise_float: bool = ...,
+    date_unit: TimeUnit | None = ...,
+    encoding: str | None = ...,
+    encoding_errors: (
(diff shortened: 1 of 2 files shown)


loicdiridollou:
For `engine="pyarrow"`, you are forced to pass `lines=True` and you can not pass a `StringIO` buffer, it has to be a file path or a `ReadBuffer[bytes]`.

pandas-stubs/io/json/_json.pyi
```diff
@@ -96,6 +123,32 @@ def read_json(
...
+        Literal["strict", "ignore", "replace", "backslashreplace", "surrogateescape"]
+        | None
+    ) = ...,
+    lines: Literal[True] = ...,
+    chunksize: None = ...,
+    compression: CompressionOptions = ...,
+    nrows: int | None = ...,
+    storage_options: StorageOptions = ...,
+    dtype_backend: DtypeBackend | NoDefault = ...,
+    engine: Literal["pyarrow"] = ...,
```
Dr-Irv, line 151: Since the default value of `lines` is `False`, if you have `...` for both `lines` and `engine`, it means that you don't have to specify either.  So I think you don't want the ellipses here on either argument.

pandas-stubs/io/json/_json.pyi
```diff
@@ -72,6 +98,7 @@ def read_json(
     nrows: int | None = ...,
     storage_options: StorageOptions = ...,
     dtype_backend: DtypeBackend | NoDefault = ...,
+    engine: Literal["pyarrow"] = ...,
```
Dr-Irv, line 101: I think the ellipses here should be removed.

pandas-stubs/io/json/_json.pyi
```diff
@@ -42,16 +42,42 @@ def read_json(
...
+        Literal["strict", "ignore", "replace", "backslashreplace", "surrogateescape"]
+        | None
+    ) = ...,
     lines: Literal[True],
     chunksize: int,
     compression: CompressionOptions = ...,
     nrows: int | None = ...,
     storage_options: StorageOptions = ...,
     dtype_backend: DtypeBackend | NoDefault = ...,
+    engine: Literal["pyarrow"] = ...,
```
Dr-Irv, line 76: remove ellipses

pandas-stubs/io/json/_json.pyi
```diff
@@ -120,6 +173,32 @@ def read_json(
...
+        Literal["strict", "ignore", "replace", "backslashreplace", "surrogateescape"]
+        | None
+    ) = ...,
+    lines: Literal[True] = ...,
+    chunksize: None = ...,
+    compression: CompressionOptions = ...,
+    nrows: int | None = ...,
+    storage_options: StorageOptions = ...,
+    dtype_backend: DtypeBackend | NoDefault = ...,
+    engine: Literal["pyarrow"] = ...,
```
Dr-Irv, line 201: same comment about the ellipses

tests/test_io.py
```diff
@@ -1625,3 +1625,28 @@ def test_read_excel_index_col() -> None:
...
+    data_lines = b"""{"col 1":"a","col 2":"b"}
+    {"col 1":"c","col 2":"d"}"""
+    dd = io.BytesIO(data_lines)
+    check(
+        assert_type(
+            pd.read_json(dd, lines=True, engine="pyarrow"),
+            pd.DataFrame,
+        ),
+        pd.DataFrame,
+    )
```
Dr-Irv, line 1652: You should add a test with `TYPE_CHECKING_INVALID_USAGE` that makes sure that we disallow `lines=False` with `engine="pyarrow"` .

Dr-Irv (changes requested):
I'm a little concerned about the misuse of ellipses with default arguments.  Ellipses should only be used in an argument when the argument is optional.  When you want a specific result to happen as a result of the argument being specified, you don't use ellipses.  The overloads that require values to be specified (i.e., the ones without ellipses) should come before the ones that use ellipses.  And the ones with ellipses should have "broad" types.  So writing something like `engine: Literal["pyarrow"] = ...` can't be correct, because the default value of `engine` is `ujson`, so if the stub is to work without specification of that parameter it would be `engine: Literal["ujson", "pyarrow"] = ...` .

pandas-stubs/io/json/_json.pyi
```diff
@@ -42,16 +42,42 @@ def read_json(
         Literal["strict", "ignore", "replace", "backslashreplace", "surrogateescape"]
         | None
     ) = ...,
+    lines: bool,
```
Dr-Irv, line 45: Should be `lines: Literal[True]`, because you can only specify `chunksize` when `lines == True`

pandas-stubs/io/json/_json.pyi
```diff
@@ -66,12 +92,38 @@ def read_json(
         Literal["strict", "ignore", "replace", "backslashreplace", "surrogateescape"]
         | None
     ) = ...,
-    lines: Literal[True],
+    lines: Literal[True] = True,
```
Dr-Irv, line 95: remove `= True`

pandas-stubs/io/json/_json.pyi
```diff
@@ -66,12 +92,38 @@ def read_json(
...
+    convert_dates: bool | list[str] = ...,
+    keep_default_dates: bool = ...,
+    precise_float: bool = ...,
+    date_unit: TimeUnit | None = ...,
+    encoding: str | None = ...,
+    encoding_errors: (
+        Literal["strict", "ignore", "replace", "backslashreplace", "surrogateescape"]
+        | None
+    ) = ...,
+    lines: Literal[True] = True,
```
Dr-Irv, line 120: remove `= True`

pandas-stubs/io/json/_json.pyi
```diff
@@ -96,6 +148,32 @@ def read_json(
...
```
Dr-Irv, line 170: remove `= True`

pandas-stubs/io/json/_json.pyi
```diff
@@ -120,6 +198,32 @@ def read_json(
...
```
Dr-Irv, line 220: remove `= True`

Dr-Irv (changes requested):
Almost there

Dr-Irv (approved):
thanks @loicdiridollou

PR #1036, opened 2024-11-16
GH1034 Migrate old tests in _json.py to new check(assert(...

### LINKED ISSUE #1034: Migrate tests in io/json/_json.py to the new `check(assert_type...` style

loicdiridollou:
- [x] Closes #1034

Simple test migration to the new `check(assert_type` fashion, no logic change.

### DIFF
--- a/tests/test_io.py
+++ b/tests/test_io.py
@@ -709,16 +709,20 @@ def test_types_read_csv() -> None:
             pd.read_csv(path, names="abcd")  # type: ignore[call-overload] # pyright: ignore[reportArgumentType]
             pd.read_csv(path, usecols="abcd")  # type: ignore[call-overload] # pyright: ignore[reportArgumentType]
 
-        tfr1: TextFileReader = pd.read_csv(path, nrows=2, iterator=True, chunksize=3)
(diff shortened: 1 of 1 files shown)


loicdiridollou:
> Thanks for doing this clean up. The tests are ones we migrated from another repo when we started this project.
> 
> In lines 712 to 722, I think you can convert that to use the `check(assert_type(...))` pattern.
> 
> In lines 923-924, there is this:
> 
> ```python
>     s: Series
>     s = DF.iloc[:, 0]
> ```
> 
> I think we can leave out the `s: Series` part because with `s = DF.iloc[:, 0]`, I think the type inference will infer that it is a `Series`.

It is inferred as a Series so I was able to remove it.
For the first one, I addressed those, I missed those since I just addressed when we defined a variable and did not use it anywhere else, if you see any more, feel free to point me there.

Dr-Irv (approved):
Thanks @loicdiridollou

Dr-Irv:
> if you see any more, feel free to point me there.

I did a search for `:` and didn't see anything else, so we're in good shape.  Thanks for doing this.  We have to clean up some of the other tests as well, but that's not as important as getting coverage for 2.x features, etc.

PR #1044, opened 2024-11-21
GH1043 Update _typing.pyi: add "cross" option to JoinHow

### LINKED ISSUE #1043: [mypy] Dataframe.Join with "cross" option for "how" argument throws error even though it's valid
**Describe the bug**
Using a "cross" join makes mypy throw a warning. 

**To Reproduce**
Create a file example.py with following contents:

```
import pandas as pd

df1 = pd.DataFrame({"A":[1,2], "B": ["test", "test"]})
df2 = pd.DataFrame({"C":[2, 3], "D": ["test", "test"]})
df3 = df1.join(df2,how="cross")
print(df3)
```

Run mypy on it: "mypy example.py "

Output:
example.py:5: error: Argument "how" to "join" of "DataFrame" has incompatible type "Literal['cross']"; expected "Literal['left', 'right', 'outer', 'inner']"  [arg-type]

Expected output:
output with 0 errors

 - OS: MacOs
 - OS Version: Sonoma 14.6
 - python version: 3.11.10
 - version of type checker: mypy 1.11.2
 - version of installed `pandas-stubs`: 2.2.2.240603

**Additional context**

brzegorzTesco:
- [ ] Closes #1043 (Replace xxxx with the Github issue number)

I haven't added tests since JoinHow doesn't seem to be tested in any case. Should I add some? Merge tests unimpacted.

### DIFF
--- a/pandas-stubs/_typing.pyi
+++ b/pandas-stubs/_typing.pyi
@@ -729,8 +729,8 @@ InterpolateOptions: TypeAlias = Literal[
 ReplaceMethod: TypeAlias = Literal["pad", "ffill", "bfill"]
 SortKind: TypeAlias = Literal["quicksort", "mergesort", "heapsort", "stable"]
 NaPosition: TypeAlias = Literal["first", "last"]
-JoinHow: TypeAlias = Literal["left", "right", "outer", "inner"]
-MergeHow: TypeAlias = JoinHow | Literal["cross"]
+JoinHow: TypeAlias = Literal["left", "right", "outer", "inner", "cross"]
+MergeHow: TypeAlias = JoinHow
 JsonFrameOrient: TypeAlias = Literal[
     "split", "records", "index", "columns", "values", "table"
 ]
--- a/pandas-stubs/_typing.pyi
+++ b/pandas-stubs/_typing.pyi
@@ -729,8 +729,8 @@ InterpolateOptions: TypeAlias = Literal[
 ReplaceMethod: TypeAlias = Literal["pad", "ffill", "bfill"]
 SortKind: TypeAlias = Literal["quicksort", "mergesort", "heapsort", "stable"]
 NaPosition: TypeAlias = Literal["first", "last"]
-JoinHow: TypeAlias = Literal["left", "right", "outer", "inner", "cross"]
-MergeHow: TypeAlias = JoinHow
+JoinHow: TypeAlias = Literal["left", "right", "outer", "inner"]
+MergeHow: TypeAlias = JoinHow | Literal["cross"]
 JsonFrameOrient: TypeAlias = Literal[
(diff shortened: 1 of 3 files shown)


pandas-stubs/_typing.pyi
```diff
@@ -729,8 +729,8 @@ InterpolateOptions: TypeAlias = Literal[
 ReplaceMethod: TypeAlias = Literal["pad", "ffill", "bfill"]
 SortKind: TypeAlias = Literal["quicksort", "mergesort", "heapsort", "stable"]
 NaPosition: TypeAlias = Literal["first", "last"]
-JoinHow: TypeAlias = Literal["left", "right", "outer", "inner"]
-MergeHow: TypeAlias = JoinHow | Literal["cross"]
+JoinHow: TypeAlias = Literal["left", "right", "outer", "inner", "cross"]
+MergeHow: TypeAlias = JoinHow
```
Dr-Irv, line 733: This fix is not correct, because `JoinHow` is used as the type of the `join` argument in the `align()` method.

The right fix is to change the typing of `DataFrame.join()` to use the existing `MergeHow` type.

You would then add a test like your example that shows that `"cross"` is accepted as a value for the `how` argument of `DataFrame.join()`

Dr-Irv (changes requested):

pandas-stubs/_typing.pyi
brzegorzTesco: Thank you for the quick feedback, I've done as you said.

tests/test_frame.py
```diff
@@ -2345,6 +2345,16 @@ def test_join() -> None:
...
 
 
+def test_types_join() -> None:
+    df1 = pd.DataFrame({"A":[1,2], "B": ["test", "test"]})
+    df2 = pd.DataFrame({"C":[2, 3], "D": ["test", "test"]})
+    df1.join(df2, how="cross")
+    df1.join(df2, how="inner")
+    df1.join(df2, how="outer")
+    df1.join(df2, how="left")
+    df1.join(df2, how="right")
```
Dr-Irv, line 2355: Can you use the `check(assert_type(...` pattern here?

Dr-Irv (changes requested):

Dr-Irv (changes requested):
CI is failing because you have to fix formatting.  See https://github.com/pandas-dev/pandas-stubs/actions/runs/12008445166/job/33489766438?pr=1044

Best to create a dev environment via the instructions here so that pre-commit will pick up things like this:
https://github.com/pandas-dev/pandas-stubs/blob/main/docs/setup.md

brzegorzTesco:
Thank you for the link, I haven't found it on my own previously. I've set up an dev environment, this commit should pass pre-commit checks.

PR #1046, opened 2024-11-21
GH1045 Split overload of groupby on as_index for all cases

### LINKED ISSUE #1045: Changes to `groupby.size()` don't work for all possible groupby
**Describe the bug**
```python
import pandas as pd
ddf = pd.DataFrame({"a": [1, 1, 2], "b": [2, 3, 2]})
reveal_type(ddf.groupby(["a", "b"]).size())

```
pyright reports
```text
 - information: Type of "ddf.groupby(["a", "b"]).size()" is "Unknown"
  c:\Code\pandas-stubs\play\issue1044.py:3:37 - error: Cannot access attribute "size" for class "DataFrameGroupBy[tuple[Unknown, ...], bool]"
    Could not bind method "size" because "DataFrameGroupBy[tuple[Unknown, ...], bool]" is not assignable to parameter "self"
      "DataFrameGroupBy[tuple[Unknown, ...], bool]" is not assignable to "DataFrameGroupBy[tuple[Unknown, ...], Literal[True]]"
        Type parameter "_TT@DataFrameGroupBy" is invariant, but "bool" is not the same as "Literal[True]"
    Could not bind method "size" because "DataFrameGroupBy[tuple[Unknown, ...], bool]" is not assignable to parameter "self"
      "DataFrameGroupBy[tuple[Unknown, ...], bool]" is not assignable to "DataFrameGroupBy[tuple[Unknown, ...], Literal[False]]"
        Type parameter "_TT@DataFrameGroupBy" is invariant, but "bool" is not the same as "Literal[False]"
    Could not bind method "size" because "DataFrameGroupBy[tuple[Unknown, ...], bool]" is not assignable to parameter "self"
      "DataFrameGroupBy[tuple[Unknown, ...], bool]" is not assignable to "DataFrameGroupBy[Timestamp, Literal[True]]" (reportAttributeAccessIssue)
```
mypy reports
```python
note: Revealed type is "Any"
```

The fix here is that any of the methods in `frame.pyi` for `def groupby()` that return `DataFrameGroupBy[SOMETYPE, bool]` need to be split into 2 where there is one overload that returns `DataFrameGroupBy[SOMETYPE, True]` and another that returns `DataFrameGroupBy[SOMETYPE, False]` like is done with `DataFrameGroupBy[Scalar, True]` and `DataFrameGroupBy[Scalar, False]` .

 - OS:  Windows 11
 - python version   3.10
 - version of type checker   pyright 1.1.389, mypy 1.13.0
 - version of installed `pandas-stubs`:   Dev version

**Additional context**

This was missed in PR #1014 by @loicdiridollou .  Hopefully he can fix soon, since the prerelease of VS Code pylance has this bug.

loicdiridollou:
- [x] Closes #1045 

### DIFF
--- a/tests/test_frame.py
+++ b/tests/test_frame.py
@@ -504,8 +504,8 @@ def test_types_mean() -> None:
     s2: pd.Series = df.mean(axis=0)
     df2: pd.DataFrame = df.groupby(level=0).mean()
     if TYPE_CHECKING_INVALID_USAGE:
-        df3: pd.DataFrame = df.groupby(axis=1, level=0).mean()  # type: ignore[call-overload] # pyright: ignore[reportArgumentType]
-        df4: pd.DataFrame = df.groupby(axis=1, level=0, dropna=True).mean()  # type: ignore[call-overload] # pyright: ignore[reportArgumentType]
+        df3: pd.DataFrame = df.groupby(axis=1, level=0).mean()  # type: ignore[call-overload] # pyright: ignore[reportArgumentType, reportCallIssue]
+        df4: pd.DataFrame = df.groupby(axis=1, level=0, dropna=True).mean()  # type: ignore[call-overload] # pyright: ignore[reportArgumentType, reportCallIssue]
     s3: pd.Series = df.mean(axis=1, skipna=True, numeric_only=False)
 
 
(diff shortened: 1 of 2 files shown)


tests/test_frame.py
```diff
@@ -1066,6 +1066,25 @@ def test_types_groupby_as_index() -> None:
...
+        pd.DataFrame,
+    )
+    check(
+        assert_type(
+            df.groupby("a", as_index=True).size(),
+            "pd.Series[int]",
+        ),
+        pd.Series,
+    )
+
```
Dr-Irv, line 1086: can  you add a test that doesn't include the `as_index` argument?

Dr-Irv (changes requested):

tests/test_frame.py
loicdiridollou: Sorry the issue wasn't there, I did a bad copy paste so it was the same test.
Now the tests are different. which includes the test case from the issue.
As well I have added a third case in both tests where we don't specify as_index as you mentioned above.

Dr-Irv (approved):
Thanks @loicdiridollou

PR #1047, opened 2024-11-21
Add @ operator type hints for Series

### LINKED ISSUE #1041: @ multiplication of Series is not inferred to a type (but Unknown)
**Describe the bug**
When multiplying two series with the `@` operator, no type is inferred and the resulting type `Unknown` will be assigned.

**To Reproduce**
```python

import pandas as pd

s1 = pd.Series([0, 1, 2, 3])
s2 = pd.Series([-1, 2, -3, 4])
sc2 = s1 @ s2
reveal_type(sc2) ■ Type of "sc2" is "Unknown"
```

I believe it should return a Scalar at minimum, in some cases Series but it seems like the type hinting of `@` is not in the stubs.

 - OS: MacOS
 - OS Version 15.0.1
 - python version 3.13
 - version of type checker 1.13.0
 - version of installed `pandas-stubs` 2.2.3.241009

loicdiridollou:
- [x] Closes #1041


Dr-Irv (approved):
thanks @loicdiridollou

PR #1054, opened 2024-11-23
GH1053 @ for DataFrame, eval in place for DataFrame, test migrations

### LINKED ISSUE #1053: Implement `__matmul__`, typehint pd.DataFrame.eval `inplace`  and test conversion for `test_frame.py`
**Describe the bug**

- @ operator on dataframe is not typehinted properly.
- `df.eval("...", inplace=True)` returns Unknown
- Convert test for types into check/assert_type framework

**To Reproduce**
```python
import pandas as pd

df1 = pd.DataFrame([[0, 1, -2, -1], [1, 1, 1, 1]])
df2 = pd.DataFrame([[0, 1], [1, 2], [-1, -1], [2, 0]])

check(assert_type(df1 @ df2, pd.DataFrame ),pd.DataFrame )
check(assert_type(df1.eval("c = 0 + 1", inplace=True), None), type(None))
```

 - OS Version [e.g. 22] 15
 - python version 3.12
 - version of type checker 1.13
 - version of installed `pandas-stubs` 2.2.3.20241009

loicdiridollou:
- [x] Closes #1053


pandas-stubs/core/frame.pyi
```diff
@@ -620,7 +625,10 @@ class DataFrame(NDFrame, OpsMixin):
     def query(
         self, expr: _str, *, inplace: Literal[False] = ..., **kwargs
     ) -> DataFrame: ...
-    def eval(self, expr: _str, *, inplace: _bool = ..., **kwargs): ...
+    @overload
+    def eval(self, expr: _str, *, inplace: Literal[False] = ..., **kwargs): ...
+    @overload
+    def eval(self, expr: _str, *, inplace: Literal[True] = ..., **kwargs) -> None: ...
```
Dr-Irv, line 631: swap these 2 overloads, and with `inplace: Literal[True]`, remove the ellipses.

And for `Literal[False] = ...`, we should return `Scalar | np.ndarray | pd.DataFrame | pd.Series`

tests/test_frame.py
```diff
@@ -422,65 +425,90 @@ def test_types_drop_duplicates() -> None:
...
+    check(assert_type(df.query("col1 > col2"), pd.DataFrame), pd.DataFrame)
+    check(assert_type(df.query("col1 % col2 == 0", inplace=True), None), type(None))
 
 
 def test_types_eval() -> None:
     df = pd.DataFrame(data={"col1": [1, 2, 3, 4], "col2": [3, 0, 1, 7]})
-    df.eval("col1 > col2")
-    res: None = df.eval("C = col1 % col2 == 0", inplace=True)
+    check(assert_type(df.eval("E = col1 > col2", inplace=True), None), type(None))
+    check(assert_type(df.eval("C = col1 % col2 == 0", inplace=True), None), type(None))
```
Dr-Irv, line 487: should add tests for `inplace = False`

tests/test_frame.py
```diff
@@ -1174,10 +1229,13 @@ def test_types_groupby() -> None:
...
         ):
             if PD_LTE_22:
-                df7: pd.DataFrame = df.groupby(by="col1").apply(sum)
-    df8: pd.DataFrame = df.groupby("col1").transform("sum")
+                check(
+                    assert_type(df.groupby(by="col1").apply(sum), pd.DataFrame),
+                    pd.DataFrame,
+                )
+    check(assert_type(df.groupby("col1").transform("sum"), pd.DataFrame), pd.DataFrame)
     s1: pd.Series = df.set_index("col1")["col2"]
```
Dr-Irv, line 1237: why not convert this line?

tests/test_frame.py
```diff
@@ -2071,17 +2139,23 @@ def test_types_regressions() -> None:
...
     df = pd.concat([s1, s2], axis=1)
-    ss1: pd.Series = pd.concat([s1, s2], axis=0)
-    ss2: pd.Series = pd.concat([s1, s2])
+    # TODO the inference here returns Any, should return Series
+    ts1: pd.Series = pd.concat([s1, s2], axis=0)
+    ts2: pd.Series = pd.concat([s1, s2])
 
     # https://github.com/microsoft/python-type-stubs/issues/110
+    # TODO the type inference below is broken, a Timestamp is not inferring as datetime.date
     d: datetime.date = pd.Timestamp("2021-01-01")
```
Dr-Irv, line 2148: Use this:
```python
    check(assert_type(pd.Timestamp("2021-01-01"), pd.Timestamp), datetime.date)
```
You can't use `assert_type()` to assert that an object is a subclass of a type.

Dr-Irv (changes requested):

pandas-stubs/core/frame.pyi
loicdiridollou: I was not totally sure about the returns of the case for `False` but thanks for adding the details, it is implemented.
For the first point about the order, do you mind clarifying how it changes the results to have them swapped?

tests/test_frame.py
```diff
@@ -2071,17 +2139,23 @@ def test_types_regressions() -> None:
...
     tslist: list[pd.Timestamp] = list(pd.to_datetime(["2022-01-01", "2022-01-02"]))
-    sseries: pd.Series = pd.Series(tslist)
+    sseries = pd.Series(tslist)
     with pytest_warns_bounded(FutureWarning, "'d' is deprecated", lower="2.2.99"):
-        foo = sseries + pd.Timedelta(1, "d")
+        sseries + pd.Timedelta(1, "d")
 
-    sseries_plus1: pd.Series = sseries + pd.Timedelta(1, "D")
+    check(
+        assert_type(sseries + pd.Timedelta(1, "D"), TimestampSeries),
```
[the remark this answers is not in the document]
loicdiridollou: I think it is because the import is done according to the TYPE_CHECKING flag but might be wrong, indeed in most other cases I saw it between quotes.

tests/test_frame.py
loicdiridollou: That is something I missed, I was going by "every time we define a variable that we don't reuse", in that case `s1` was reused below.

Dr-Irv (approved):
thanks @loicdiridollou

pandas-stubs/core/frame.pyi
Dr-Irv: The matching of overloads is not defined by the typing spec.  There is an effort to document that better.  The general rule is that any call to a function/method will pick the first overload that matches.

So the convention to use is to put the narrower matches before wider matches.  So with `inplace: Literal[True]` being first, it says match `inplace = True`, and if that matches, don't try anything else.  When the second overload is `inplace: Literal[False] = ...`, it is saying match this overload if the value is `False` or unspecified.  But this match is in essence wider than just `inplace: Literal[True]` because a user doesn't have to specify the value.  It would also be valid to write `inplace: bool = ...` as the last overload. 

We may not be consistent with this practice throughout `pandas-stubs`, but with reviews, I'm trying to get there.

tests/test_frame.py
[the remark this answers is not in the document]
Dr-Irv: It's because of the `else` in the `if TYPE_CHECKING` flag at the top where it is defined in case there is no type checking.  So all is OK

PR #1056, opened 2024-11-25
Change signature of StringMethods.rsplit to match pandas

### LINKED ISSUE #1032: Incorrect signature for `pandas.Series.str.rsplit`
**Describe the bug**
The stubs for `pandas_stubs.core.strings.StringMethods.rsplit` includes a `regex: bool` parameter. The function `pandas.Series.str.rsplit` contains no such parameter.

**To Reproduce**
```import pandas as pd

ser = pd.Series(["foo.bar.baz", "qux.foobar"])
ser.str.rsplit(".", regex=False)
```

I'm using `pyright`, which does not raise any error, even though it should since actually running the code results in `TypeError: StringMethods.rsplit() got an unexpected keyword argument 'regex'`.
 - OS: MacOS
 - OS Version: Sonoma 14.6.1
 - Python 3.12.1
 - pyright 1.1.389
 - pandas-stubs 2.2.3.241009

**Additional context**
The confusion seems to stem from the fact that the parallel method `pandas.Series.str.split` does include a `regex: bool` parameter.

kroche98:
This removes the nonexistent `regex: bool` parameter from the `StringMethods.rsplit` function.

- [ ] Closes #1032

I haven't added any tests, as it doesn't seem the parameter types of any string accessor methods are tested.


Dr-Irv:
> I haven't added any tests, as it doesn't seem the parameter types of any string accessor methods are tested.

We do have tests for `str.rsplit()` in `test_series.py`.  Having said that, adding a test for an incorrect parameter isn't really needed.

Dr-Irv (approved):
thanks @kroche98

PR #1058, opened 2024-11-27
GH1055 Add pandas.api.typing to pandas-stubs

### LINKED ISSUE #1055: Add pandas.api.typing module
The `pandas.api.typing` module was added in pandas 2.1, but is not present in pandas-stubs

**To Reproduce**
```python
from pandas.api.typing import NAType
```
mypy gives `error: Library stubs not installed for "pandas.api.typing"  [import-untyped]`

 - OS: Windows
 - OS Version 10
 - Python version 3.12.6
 - mypy version 1.13.0
 - pandas-stubs version 2.2.3.241009

loicdiridollou:
- [x] Closes #1055 

### DIFF
--- a/tests/test_api_typing.py
+++ b/tests/test_api_typing.py
@@ -3,7 +3,6 @@
 from typing import (
     TYPE_CHECKING,
     Literal,
-    assert_type,
 )
 
 import numpy as np
@@ -15,6 +14,7 @@ from pandas.api.typing import (
     NAType,
     SeriesGroupBy,
 )
+from typing_extensions import assert_type
 
 from pandas._typing import Scalar
 
--- a/tests/test_api_typing.py
+++ b/tests/test_api_typing.py
@@ -1,0 +1,63 @@
+"""Test module for classes in pandas.api.typing."""
+
+from typing import (
+    TYPE_CHECKING,
+    Literal,
+    assert_type,
+)
+
+import numpy as np
+import pandas as pd
+from pandas.api.typing import (
+    DataFrameGroupBy,
+    DatetimeIndexResamplerGroupby,
+    FrozenList,
+    NAType,
+    SeriesGroupBy,
+)
+
+from pandas._typing import Scalar
+
+from tests import check
+
+if TYPE_CHECKING:
+    from pandas.core.groupby.groupby import _ResamplerGroupBy  # noqa: F401
+
+
+def test_dataframegroupby():
+    df = pd.DataFrame({"a": [1, 2, 3]})
+    check(
+        assert_type(df.groupby("a"), DataFrameGroupBy[Scalar, Literal[True]]),
+        DataFrameGroupBy,
+    )
+
+
+def test_seriesgroupby():
+    sr: pd.Series[int] = pd.Series([1, 2, 3], index=pd.Index(["a", "b", "a"]))
+    check(
+        assert_type(sr.groupby("a"), SeriesGroupBy[int, Scalar]),
+        SeriesGroupBy,
+    )
+
+
+def test_frozenlist() -> None:
+    flst = FrozenList([1, 2, 3])
+    check(assert_type(flst, FrozenList), FrozenList)
+
+
+def tests_datetimeindexersamplergroupby() -> None:
+    idx = pd.date_range("1999-1-1", periods=365, freq="D")
+    df = pd.DataFrame(
+        np.random.standard_normal((365, 2)), index=idx, columns=["col1", "col2"]
+    )
+    gb_df = df.groupby("col2")
+    check(
+        assert_type(gb_df.resample("ME"), "_ResamplerGroupBy[pd.DataFrame]"),
+        DatetimeIndexResamplerGroupby,
+        pd.DataFrame,
+    )
+
+
+def test_natype() -> None:
+    i64dt = pd.Int64Dtype()
+    check(assert_type(i64dt.na_value, NAType), NAType)
--- a/tests/test_api_typing.py
+++ b/tests/test_api_typing.py
@@ -1,8 +1,5 @@
 """Test module for classes in pandas.api.typing."""
 
-from typing import Literal  # noqa: F401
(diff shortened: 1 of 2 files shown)


loicdiridollou:
Haven't written tests for every case, I first want to get feedback on the structure and if I am going the right direction.

tests/test_api_typing.py
```diff
@@ -0,0 +1,55 @@
...
+from tests import check
+
+if TYPE_CHECKING:
+    from pandas.core.groupby.groupby import _ResamplerGroupBy  # noqa: F401
+
+
+def test_dataframegroupby():
+    df = pd.DataFrame({"a": [1, 2, 3]})
+    check(
+        assert_type(df.groupby("a"), "DataFrameGroupBy[Scalar, Literal[True]]"),
```
Dr-Irv, line 27: We want to make sure that the type the stubs returns is `DataFrameGroupBy`, so use that, not the generic version.

tests/test_api_typing.py
```diff
@@ -0,0 +1,55 @@
...
+        DataFrameGroupBy,
+    )
+
+
+def test_seriesgroupby():
+    sr: pd.Series[int] = pd.Series([1, 2, 3], index=pd.Index(["a", "b", "a"]))
+    check(
+        assert_type(sr.groupby(level=0), "SeriesGroupBy[int, Scalar]"),
```
Dr-Irv, line 35: same comment here - use `assert_type` with `SeriesGroupBy` as the second argument

tests/test_api_typing.py
```diff
@@ -0,0 +1,55 @@
...
+
+
+def tests_datetimeindexersamplergroupby() -> None:
+    idx = pd.date_range("1999-1-1", periods=365, freq="D")
+    df = pd.DataFrame(
+        np.random.standard_normal((365, 2)), index=idx, columns=["col1", "col2"]
+    )
+    gb_df = df.groupby("col2")
+    check(
+        assert_type(gb_df.resample("ME"), "_ResamplerGroupBy[pd.DataFrame]"),
```
Dr-Irv, line 47: same comment - should be able to `assert_type` with `ResamplerGroupBy` as the second argument, and you then avoid `_ResamplerGroupBy`

Dr-Irv (changes requested):
If we can't do `assert_type` on the `pandas.api.typing` objects, then we may have to change the stubs to support that.

Dr-Irv (commented):
So this structure is fine.  Now you need to add tests for all of the other things in `pandas.api.typing`

Dr-Irv:
So I see that your tests failed and I played around with it.  The purpose of the types in `pandas.api.typing` is to allow people to use the types in functions.  Because we are doing things with `Generic`, we have to change the test.  E.g., for `DataFrameGroupBy`:
```python
def test_dataframegroupby():
    group = df.groupby("a")

    def f1(gb: DataFrameGroupBy):
        check(group, DataFrameGroupBy)

    f1(group)
```
We can't use the `check(assert_type` pattern here because of the `Generic`.  But for this particular part of the API, it would be good to add a test like the one above, because that is the anticipated usage.  This will do a typing check via the function call and an `isinstance` check as a result of the use of `check` .

You'd only have to do this for the ones that are `Generic` inside of the stubs.  Otherwise the standard `check(assert_type(...` pattern is fine, like you did with `NAType`

loicdiridollou:
Thanks for the trick here!
One issue is that one of the type can only be obtained through a deprecated resampling, not sure what to do about this one.

tests/test_api_typing.py
```diff
@@ -0,0 +1,220 @@
...
+    TimeGrouper,
+    Window,
+)
+import pytest
+from typing_extensions import assert_type
+
+from tests import check
+
+from pandas.io.json._json import read_json
+from pandas.io.stata import StataReader
```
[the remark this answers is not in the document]
loicdiridollou: Mistake when I left the lsp auto import, will correct.

tests/test_api_typing.py
```diff
@@ -0,0 +1,220 @@
...
+    )
+    gb_df = df.groupby("col2")
+
+    # TODO the groupby here is too wide and returns _ResamplerGroupby alias
+    # def f1(gb: DatetimeIndexResamplerGroupby):
+    #     check(gb, DatetimeIndexResamplerGroupby)
+    #
+    # f1(gb_df.resample("ME"))
+
+    check(gb_df.resample("ME"), DatetimeIndexResamplerGroupby)
```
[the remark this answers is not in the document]
loicdiridollou: I was able to create a Union type for those three and add it to the function definition, though we can't use it as is in the check since it does not accept union types as klass but only individual type.

tests/test_api_typing.py
```diff
@@ -0,0 +1,220 @@
...
+
+
+def test_statereader(tmp_path: Path) -> None:
+    df = pd.DataFrame([[1, 2], [3, 4]], columns=["col_1", "col_2"])
+    time_stamp = pd.Timestamp(2000, 2, 29, 14, 21)
+    variable_labels = {"col_1": "This is an example"}
+    path = tmp_path / "file"
+    df.to_stata(
+        path, time_stamp=time_stamp, variable_labels=variable_labels, version=None
+    )
```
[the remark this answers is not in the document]
loicdiridollou: Indeed I missed that one, will fix.

tests/test_api_typing.py
```diff
@@ -30,7 +29,12 @@
 from tests import check
 
 from pandas.io.json._json import read_json
-from pandas.io.stata import StataReader
+
+ResamplerGroupBy = (
```
Dr-Irv, line 33:
```suggestion
ResamplerGroupBy: TypeAlias = (
```

Dr-Irv (changes requested):
one small change to use `TypeAlias`

Dr-Irv (approved):
thanks @loicdiridollou

PR #1067, opened 2024-12-02
adjust test for slice due to typeshed changes

Dr-Irv:
A change was made to `typeshed` to make `slice` a `Generic`, which is causing the `mypy_nightly` tests to fail.  More discussion here:
https://github.com/python/mypy/issues/18225

This fix should allow the CI to pass as well as the `mypy_nightly` runs.  See comment in code.

### DIFF
--- a/tests/test_frame.py
+++ b/tests/test_frame.py
@@ -2398,7 +2398,13 @@ def test_indexslice_getitem():
         .set_index(["x", "y"])
     )
     ind = pd.Index([2, 3])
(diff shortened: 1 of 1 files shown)


Dr-Irv:
Ooops.  Deleted my branch by mistake, but restored it.  So should be good to go

twoertwein (approved):

twoertwein:
Thank you for navigating the subtle differences between mypy/mpy nightly/pyright!

PR #1073, opened 2024-12-16
Add MaskType to other cases

### LINKED ISSUE #868: Check usage of `Series[bool]` and see if `MaskType` is more appropriate
In #862 a change was made to replace `Series[bool]` with `MaskType` for `.loc`.  There are other places in the stubs where `Series[bool]` is used, and we should investigate which of those usages as arguments (not as return types) should be replaced with `MaskType`

loicdiridollou:
- [x] Closes #868 

### DIFF
--- a/pandas-stubs/core/indexes/base.pyi
+++ b/pandas-stubs/core/indexes/base.pyi
@@ -48,6 +48,7 @@ from pandas._typing import (
     HashableT,
     Label,
     Level,
+    MaskType,
     NaPosition,
     TimedeltaDtypeArg,
     TimestampDtypeArg,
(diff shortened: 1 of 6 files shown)


loicdiridollou:
There was also one other place (https://github.com/pandas-dev/pandas-stubs/blob/af9b633a1234a4ccd2c554777374ff46a744acc6/pandas-stubs/core/indexes/accessors.pyi#L370) but I was a little unsure if `MaskType` was the appropriate behavior there, open to feedback.

Dr-Irv (changes requested):
Take a look at `core/series.pyi`.  The logical operations like `__and__()`, `__or__()`, etc. have the same issue.  Can you make the change there as well?

Dr-Irv (approved):
Thanks @loicdiridollou

PR #1092, opened 2025-01-10
Add "Typing :: Stubs Only" trove classifier

yangdanny97:
That classifier is specifically for standalone type stubs packages, so I think pandas-stubs should have it.

See https://github.com/pypa/trove-classifiers/issues/84


Dr-Irv (approved):
thanks @yangdanny97

PR #1093, opened 2025-01-11
GH1089 Migrate frame/series tests to new framework

### LINKED ISSUE #1089: Fix pandas nightly test build and migrate tests over new framework
pandas3.0.0dev for nightly has been breaking DeprecationWarning tests.

loicdiridollou:
- [x] Closes #1089 

### DIFF
--- a/tests/test_frame.py
+++ b/tests/test_frame.py
@@ -2409,7 +2409,7 @@ def test_indexslice_getitem():
     ind = pd.Index([2, 3])
     check(
         assert_type(
-            pd.IndexSlice[ind, :], tuple["pd.Index[int]", "slice[None, None, None]"]
+            pd.IndexSlice[ind, :], tuple["pd.Index[int]", "slice[None, None, None]"]  # type: ignore[type-arg]
         ),
         tuple,
     )
@@ -2451,10 +2451,10 @@ def test_sum_get_add() -> None:
 
 
 def test_getset_untyped() -> None:
-    result: int = 10
     df = pd.DataFrame({"x": [1, 2, 3, 4, 5], "y": [10, 20, 30, 40, 50]})
     # Tests that Dataframe.__getitem__ needs to return untyped series.
-    result = df["x"].max()
+    # TODO this typecheck is actually bogus as the right part is "Unknown"
+    result: pd.Series = df["x"].max()
 
 
 def test_getmultiindex_columns() -> None:
@@ -2965,7 +2965,9 @@ def test_groupby_apply() -> None:
             pd.Series,
         )
 
-    lfunc: Callable[[pd.DataFrame], float] = lambda x: x.sum().mean()
+    def lfunc(x: pd.DataFrame) -> float:
+        return x.sum().mean()
+
     with pytest_warns_bounded(
         DeprecationWarning,
         "DataFrameGroupBy.apply operated on the grouping columns.",
--- a/tests/test_series.py
+++ b/tests/test_series.py
@@ -145,20 +145,20 @@ def test_types_all() -> None:
 
 def test_types_csv() -> None:
     s = pd.Series(data=[1, 2, 3])
-    csv_df: str = s.to_csv()
+    check(assert_type(s.to_csv(), str), str)
 
     with ensure_clean() as path:
         s.to_csv(path)
-        s2: pd.DataFrame = pd.read_csv(path)
+        check(assert_type(pd.read_csv(path), pd.DataFrame), pd.DataFrame)
 
     with ensure_clean() as path:
         s.to_csv(Path(path))
-        s3: pd.DataFrame = pd.read_csv(Path(path))
+        check(assert_type(pd.read_csv(Path(path)), pd.DataFrame), pd.DataFrame)
 
     # This keyword was added in 1.1.0 https://pandas.pydata.org/docs/whatsnew/v1.1.0.html
     with ensure_clean() as path:
         s.to_csv(path, errors="replace")
-        s4: pd.DataFrame = pd.read_csv(path)
+        check(assert_type(pd.read_csv(path), pd.DataFrame), pd.DataFrame)
 
 
 def test_types_copy() -> None:
@@ -229,11 +229,11 @@ def test_types_boolean_indexing() -> None:
 def test_types_df_to_df_comparison() -> None:
     s = pd.Series(data={"col1": [1, 2]})
     s2 = pd.Series(data={"col1": [3, 2]})
-    res_gt: pd.Series = s > s2
-    res_ge: pd.Series = s >= s2
-    res_lt: pd.Series = s < s2
-    res_le: pd.Series = s <= s2
-    res_e: pd.Series = s == s2
+    check(assert_type(s > s2, "pd.Series[bool]"), pd.Series, np.bool_)
+    check(assert_type(s >= s2, "pd.Series[bool]"), pd.Series, np.bool_)
+    check(assert_type(s < s2, "pd.Series[bool]"), pd.Series, np.bool_)
+    check(assert_type(s <= s2, "pd.Series[bool]"), pd.Series, np.bool_)
+    check(assert_type(s == s2, "pd.Series[bool]"), pd.Series, np.bool_)
 
 
 def test_types_head_tail() -> None:
@@ -309,7 +309,11 @@ def test_types_drop_multilevel() -> None:
         codes=[[0, 0, 0, 1, 1, 1], [0, 1, 2, 0, 1, 2]],
     )
     s = pd.Series(data=[1, 2, 3, 4, 5, 6], index=index)
-    res: pd.Series = s.drop(labels="first", level=1)
+    check(
+        assert_type(s.drop(labels="first", level=1), "pd.Series[int]"),
+        pd.Series,
+        np.int64,
+    )
 
 
 def test_types_drop_duplicates() -> None:
@@ -382,7 +386,11 @@ def test_types_sort_index() -> None:
 # This was added in 1.1.0 https://pandas.pydata.org/docs/whatsnew/v1.1.0.html
 def test_types_sort_index_with_key() -> None:
     s = pd.Series([1, 2, 3], index=["a", "B", "c"])
-    res: pd.Series = s.sort_index(key=lambda k: k.str.lower())
+    check(
+        assert_type(s.sort_index(key=lambda k: k.str.lower()), "pd.Series[int]"),
+        pd.Series,
+        np.int64,
+    )
 
 
 def test_types_sort_values() -> None:
@@ -412,7 +420,11 @@ def test_types_sort_values() -> None:
 # This was added in 1.1.0 https://pandas.pydata.org/docs/whatsnew/v1.1.0.html
 def test_types_sort_values_with_key() -> None:
     s = pd.Series([1, 2, 3], index=[2, 3, 1])
-    res: pd.Series = s.sort_values(key=lambda k: -k)
+    check(
+        assert_type(s.sort_values(key=lambda k: -k), "pd.Series[int]"),
+        pd.Series,
+        np.integer,
+    )
 
 
 def test_types_shift() -> None:
@@ -435,18 +447,32 @@ def test_types_rank() -> None:
 
 def test_types_mean() -> None:
     s = pd.Series([1, 2, 3, np.nan])
-    f1: float = s.mean()
-    s1: pd.Series = s.groupby(level=0).mean()
-    f2: float = s.mean(skipna=False)
-    f3: float = s.mean(numeric_only=False)
+    check(assert_type(s.mean(), float), float)
+    check(
+        assert_type(
+            s.groupby(level=0).mean(),  # pyright: ignore[reportAssertTypeFailure]
+            "pd.Series[float]",
+        ),
+        pd.Series,
+        float,
+    )
+    check(assert_type(s.mean(skipna=False), float), float)
+    check(assert_type(s.mean(numeric_only=False), float), float)
 
 
 def test_types_median() -> None:
     s = pd.Series([1, 2, 3, np.nan])
-    f1: float = s.median()
-    s1: pd.Series = s.groupby(level=0).median()
-    f2: float = s.median(skipna=False)
-    f3: float = s.median(numeric_only=False)
+    check(assert_type(s.median(), float), float)
+    check(
+        assert_type(
+            s.groupby(level=0).median(),  # pyright: ignore[reportAssertTypeFailure]
+            "pd.Series[float]",
+        ),
+        pd.Series,
+        float,
+    )
+    check(assert_type(s.median(skipna=False), float), float)
+    check(assert_type(s.median(numeric_only=False), float), float)
 
 
 def test_types_sum() -> None:
@@ -624,26 +650,36 @@ def test_types_element_wise_arithmetic() -> None:
     s = pd.Series([0, 1, -10])
     s2 = pd.Series([7, -5, 10])
 
-    res_add1: pd.Series = s + s2
-    res_add2: pd.Series = s.add(s2, fill_value=0)
+    check(assert_type(s + s2, "pd.Series[int]"), pd.Series, np.integer)
+    check(assert_type(s.add(s2, fill_value=0), "pd.Series[int]"), pd.Series, np.integer)
 
-    res_sub: pd.Series = s - s2
-    res_sub2: pd.Series = s.sub(s2, fill_value=0)
+    check(assert_type(s - s2, pd.Series), pd.Series, np.integer)
+    check(assert_type(s.sub(s2, fill_value=0), "pd.Series[int]"), pd.Series, np.integer)
 
-    res_mul: pd.Series = s * s2
-    res_mul2: pd.Series = s.mul(s2, fill_value=0)
+    check(assert_type(s * s2, pd.Series), pd.Series, np.integer)
+    check(assert_type(s.mul(s2, fill_value=0), pd.Series), pd.Series, np.integer)
 
-    res_div: pd.Series = s / s2
-    res_div2: pd.Series = s.div(s2, fill_value=0)
+    check(assert_type(s / s2, pd.Series), pd.Series, np.float64)
+    check(
+        assert_type(s.div(s2, fill_value=0), "pd.Series[float]"), pd.Series, np.float64
+    )
 
-    res_floordiv: pd.Series = s // s2
-    res_floordiv2: pd.Series = s.floordiv(s2, fill_value=0)
+    check(assert_type(s // s2, "pd.Series[int]"), pd.Series, np.integer)
+    check(
+        assert_type(s.floordiv(s2, fill_value=0), "pd.Series[int]"),
+        pd.Series,
+        np.integer,
+    )
 
-    res_mod: pd.Series = s % s2
-    res_mod2: pd.Series = s.mod(s2, fill_value=0)
+    check(assert_type(s % s2, "pd.Series[int]"), pd.Series, np.integer)
+    check(assert_type(s.mod(s2, fill_value=0), "pd.Series[int]"), pd.Series, np.integer)
 
-    res_pow: pd.Series = s ** s2.abs()
-    res_pow2: pd.Series = s.pow(s2.abs(), fill_value=0)
+    check(assert_type(s ** s2.abs(), "pd.Series[int]"), pd.Series, np.integer)
+    check(
+        assert_type(s.pow(s2.abs(), fill_value=0), "pd.Series[int]"),
+        pd.Series,
+        np.integer,
+    )
 
     check(assert_type(divmod(s, s2), tuple["pd.Series[int]", "pd.Series[int]"]), tuple)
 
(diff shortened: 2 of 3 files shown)


tests/test_series.py
```diff
@@ -1105,8 +1147,8 @@ def test_types_getitem() -> None:
     s = pd.Series({"key": [0, 1, 2, 3]})
     key: list[int] = s["key"]
     s2 = pd.Series([0, 1, 2, 3])
-    value: int = s2[0]
-    s3: pd.Series = s[:2]
+    check(assert_type(s2[0], int), np.integer)
+    check(assert_type(s[:2], pd.Series), pd.Series)
```
Dr-Irv, line 1151: We should add tests here for `s[0]`  (type check as `Any`, but is a list that is returned) and `s2[:2]`  (type check as `Series[int]`, check the same)

Dr-Irv (changes requested):

tests/test_frame.py
```diff
@@ -2965,7 +2965,9 @@ def sum_mean(x: pd.DataFrame) -> float:
             pd.Series,
         )
 
-    lfunc: Callable[[pd.DataFrame], float] = lambda x: x.sum().mean()
+    def lfunc(x: pd.DataFrame) -> float:
+        return x.sum().mean()
+
```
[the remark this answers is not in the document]
loicdiridollou: My bad, it was giving me a lint warning not to define a lambda, will revert that and ignore the warning (E731).

tests/test_series.py
```diff
@@ -435,18 +447,32 @@ def test_types_rank() -> None:
...
 def test_types_mean() -> None:
     s = pd.Series([1, 2, 3, np.nan])
-    f1: float = s.mean()
-    s1: pd.Series = s.groupby(level=0).mean()
-    f2: float = s.mean(skipna=False)
-    f3: float = s.mean(numeric_only=False)
+    check(assert_type(s.mean(), float), float)
+    check(
+        assert_type(
+            s.groupby(level=0).mean(),  # pyright: ignore[reportAssertTypeFailure]
```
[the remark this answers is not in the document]
loicdiridollou: Great, will hold for your PR to be fixed to merge this one!

tests/test_series.py
```diff
@@ -624,63 +650,79 @@ def test_types_element_wise_arithmetic() -> None:
...
 
 
 # GH 103
 def test_types_complex_arithmetic() -> None:
     c = 1 + 1j
     s = pd.Series([1.0, 2.0, 3.0])
-    x = s + c
-    y = s - c
+    check(assert_type(s + c, pd.Series), pd.Series)
+    check(assert_type(s - c, pd.Series), pd.Series)
```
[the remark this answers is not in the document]
loicdiridollou: image
Indeed, I will create a new issue for those.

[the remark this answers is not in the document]
loicdiridollou: https://github.com/pandas-dev/pandas-stubs/issues/1098

tests/test_series.py
loicdiridollou: Done, good point!

loicdiridollou: Actually accessing s with `0` will cause an warning in python 3.11 and higher:
```
FutureWarning: Series.__getitem__ treating keys as positions is deprecated. In a future version, integer keys will always be treated as labels (consistent with DataFrame behavior). To access a value by position, use `ser.iloc[pos]`
```
I didi add the tests for `s2[:2]`

tests/test_series.py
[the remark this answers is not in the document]
Dr-Irv: I'm leaving this as unresolved, because we need to get the other PR approved and merged

tests/test_series.py
```diff
@@ -624,63 +650,79 @@ def test_types_element_wise_arithmetic() -> None:
...
 
-    res_add1: pd.Series = s + s2
-    res_add2: pd.Series = s.add(s2, fill_value=0)
+    check(assert_type(s + s2, "pd.Series[int]"), pd.Series, np.integer)
+    check(assert_type(s.add(s2, fill_value=0), "pd.Series[int]"), pd.Series, np.integer)
 
-    res_sub: pd.Series = s - s2
-    res_sub2: pd.Series = s.sub(s2, fill_value=0)
+    check(assert_type(s - s2, pd.Series), pd.Series, np.integer)
+    check(assert_type(s.sub(s2, fill_value=0), "pd.Series[int]"), pd.Series, np.integer)
```
Dr-Irv, line 657: I would have thought that both of these should type check as `pd.Series[int]`

tests/test_series.py
```diff
@@ -624,63 +650,79 @@ def test_types_element_wise_arithmetic() -> None:
...
 
-    res_mul: pd.Series = s * s2
-    res_mul2: pd.Series = s.mul(s2, fill_value=0)
+    check(assert_type(s * s2, pd.Series), pd.Series, np.integer)
+    check(assert_type(s.mul(s2, fill_value=0), pd.Series), pd.Series, np.integer)
```
Dr-Irv, line 660: I would hope these would match `Series[int]`.  If not, then we should update the PYI file so that multiplying a `Series[int]` times `Series[int]` creates a `Series[int]`

tests/test_series.py
```diff
@@ -624,63 +650,79 @@ def test_types_element_wise_arithmetic() -> None:
...
 
-    res_div: pd.Series = s / s2
-    res_div2: pd.Series = s.div(s2, fill_value=0)
+    check(assert_type(s / s2, pd.Series), pd.Series, np.float64)
+    check(
+        assert_type(s.div(s2, fill_value=0), "pd.Series[float]"), pd.Series, np.float64
+    )
```
Dr-Irv, line 665: These should match `Series[float]`.  If not, then we need to update the PYI files.

tests/test_series.py
```diff
@@ -624,63 +650,79 @@ def test_types_element_wise_arithmetic() -> None:
...
     s = pd.Series([0, 1, -10])
 
-    res_add1: pd.Series = s + 1
-    res_add2: pd.Series = s.add(1, fill_value=0)
+    check(assert_type(s + 1, "pd.Series[int]"), pd.Series, np.integer)
+    check(assert_type(s.add(1, fill_value=0), "pd.Series[int]"), pd.Series, np.integer)
 
-    res_sub: pd.Series = s - 1
-    res_sub2: pd.Series = s.sub(1, fill_value=0)
+    check(assert_type(s - 1, pd.Series), pd.Series, np.integer)
```
Dr-Irv, line 693: This should be `pd.Series[int]`

tests/test_series.py
```diff
@@ -624,63 +650,79 @@ def test_types_element_wise_arithmetic() -> None:
...
+    check(assert_type(s.sub(1, fill_value=0), "pd.Series[int]"), pd.Series, np.integer)
 
-    res_mul: pd.Series = s * 2
-    res_mul2: pd.Series = s.mul(2, fill_value=0)
+    check(assert_type(s * 2, pd.Series), pd.Series, np.integer)
+    check(assert_type(s.mul(2, fill_value=0), pd.Series), pd.Series, np.integer)
```
Dr-Irv, line 697: These should be `pd.Series[int]`

tests/test_series.py
```diff
@@ -624,63 +650,79 @@ def test_types_element_wise_arithmetic() -> None:
...
 
-    res_div: pd.Series = s / 2
-    res_div2: pd.Series = s.div(2, fill_value=0)
+    check(assert_type(s / 2, pd.Series), pd.Series, np.float64)
```
Dr-Irv, line 699: Should be `pd.Series[float]`

tests/test_series.py
```diff
@@ -624,63 +650,79 @@ def test_types_element_wise_arithmetic() -> None:
...
+    check(assert_type(s.mod(2, fill_value=0), "pd.Series[int]"), pd.Series, np.integer)
 
-    res_pow: pd.Series = s**2
-    res_pow1: pd.Series = s**0
-    res_pow2: pd.Series = s**0.213
-    res_pow3: pd.Series = s.pow(0.5)
+    check(assert_type(s**2, "pd.Series[int]"), pd.Series, np.integer)
+    check(assert_type(s**0, "pd.Series[int]"), pd.Series, np.integer)
+    check(assert_type(s**0.213, "pd.Series[int]"), pd.Series, np.float64)
+    check(assert_type(s.pow(0.5), "pd.Series[int]"), pd.Series, np.float64)
```
Dr-Irv, line 717: These should be `pd.Series[float]` because it is an int series raised to a float.

tests/test_series.py
```diff
@@ -1264,9 +1307,9 @@ def test_types_dot() -> None:
     s2 = pd.Series([-1, 2, -3, 4])
     df1 = pd.DataFrame([[0, 1], [-2, 3], [4, -5], [6, 7]])
     n1 = np.array([[0, 1], [1, 2], [-1, -1], [2, 0]])
-    check(assert_type(s1.dot(s2), Scalar), np.int64)
-    check(assert_type(s1 @ s2, Scalar), np.int64)
-    check(assert_type(s1.dot(df1), "pd.Series[int]"), pd.Series, np.int64)
+    check(assert_type(s1.dot(s2), Scalar), np.integer)
+    check(assert_type(s1 @ s2, Scalar), np.integer)
+    check(assert_type(s1.dot(df1), "pd.Series[int]"), pd.Series, np.integer)
```
Dr-Irv, line 1312: These are inconsistent, and I think the PYI has to be fixed.  Because if we are doing `Series[int].dot(Series[int])`, result should be `int`.  And `s1.dot(df1)` should be `Series[float]` not `Series[int]`, because we don't know the types inside of the dataframe.

Dr-Irv (changes requested):
For the ones where the results aren't right (mostly the arithmetic ones), the PR should include the fixes to the PYI files to make those results come out right.  I don't want to approve the test changes when the results aren't what we expect.

Could include the fix for `complex` here as well (even though you created a separate issue)

tests/test_series.py
loicdiridollou: Apparently pyright is not happy about it, let me add it to the other issue:
```
Diagnostics:                                                                    
"assert_type" mismatch: expected "Series[int]" but received "Series[Unknown]" [reportAssertTypeFailure]
```

tests/test_series.py
loicdiridollou: Yes same issue here as above, will add it to the new issue, it detects `Series[Unknown]`:
```
tests/test_series.py:660: error: Expression is of type "Series[Any]", not "Series[int]"  [assert-type]
```

tests/test_series.py
loicdiridollou: Same issue here, seems like the `__sub__`, `__mul__` and `__div__` are missing hints, adding to the other issue.

tests/test_series.py
loicdiridollou: Same issue here as well

tests/test_series.py
loicdiridollou: Same as well

tests/test_series.py
loicdiridollou: I fixed this one just to make sure that was the issue, will leave the other ones to the other issue.

tests/test_series.py
loicdiridollou: Will add it to the other issue and we can investigate deeper to avoid having too many changes in this one as there are already quite a few).

loicdiridollou:
Will address the changes in this PR (although I created a new issue), seems more correct but agree that it will be slightly larger to review.

Dr-Irv:
> Will address the changes in this PR (although I created a new issue), seems more correct but agree that it will be slightly larger to review.

You have 2 options:

1. Include the changes to the PYI files in this PR so then the "correct" tests pass.
2. Leave the "failing" tests alone (i.e., don't convert them) and fix them in a PR that has the PYI changes.

I'm fine either way, although I think it would be easier to do (1).  No issue for me in reviewing that large of a PR.

tests/test_series.py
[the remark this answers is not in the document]
loicdiridollou: Now fixed

tests/test_series.py
loicdiridollou: This is done

tests/test_series.py
```diff
@@ -1288,7 +1337,8 @@ def test_series_min_max_sub_axis() -> None:
     sd = s1 / s2
     check(assert_type(sa, pd.Series), pd.Series)
     check(assert_type(ss, pd.Series), pd.Series)
-    check(assert_type(sm, pd.Series), pd.Series)
+    # TODO GH1089 This should not match to Series[int]
+    check(assert_type(sm, pd.Series), pd.Series)  # pyright: ignore
```
loicdiridollou, line 1341: @Dr-Irv running into a bit of an issue here, it matches to `Series[int]` so not sure where I got the typehinting wrong, any advice you may have is welcome!

tests/test_series.py
loicdiridollou: It seems like the typehinting is not revealing the Series[float] here, not sure where the mistake is

pandas-stubs/core/series.pyi
```diff
@@ -1716,6 +1739,16 @@ class Series(IndexOpsMixin[S1], NDFrame):
         self, other: Timestamp | datetime | TimestampSeries
     ) -> TimedeltaSeries: ...
```
Dr-Irv, line 1740: If you add this overload, it will fix the problem with the test in line 1339 in `test_series.py`
```python
    @overload
    def __sub__(self, other: S1 | Self) -> Self: ...
```

pandas-stubs/core/series.pyi
```diff
@@ -1716,6 +1739,16 @@ class Series(IndexOpsMixin[S1], NDFrame):
...
     @overload
+    def __sub__(
+        self: Series[int],
+        other: int,
```
Dr-Irv, line 1744:
```suggestion
        other: int | Series[int]
```
This will fix the problem you had with the test in line 656 of `test_series.py`

pandas-stubs/core/series.pyi
```diff
@@ -1623,6 +1628,11 @@ class Series(IndexOpsMixin[S1], NDFrame):
         self, other: int | np_ndarray_anyint | Series[int]
     ) -> Series[int]: ...
     # def __array__(self, dtype: Optional[_bool] = ...) -> _np_ndarray
+    @overload
+    def __div__(self: Series[int], other: Series[int]) -> Series[float]: ...
+    @overload
+    def __div__(self: Series[int], other: int) -> Series[float]: ...
+    @overload
     def __div__(self, other: num | _ListLike | Series[S1]) -> Series[S1]: ...
```
Dr-Irv, line 1636: We need to remove `__div__()`.  It is python 2.  In python 3, we should be using `__truediv__` .  Same goes for removing `__rdiv__` and using `__rtruediv__` instead.  So if you can migrate the overloads accordingly, I think this will get fixed.

I just learned that by looking up the issue!

pandas-stubs/core/series.pyi
```diff
@@ -1643,6 +1653,14 @@ class Series(IndexOpsMixin[S1], NDFrame):
         self, other: timedelta | Timedelta | TimedeltaSeries | np.timedelta64
     ) -> TimedeltaSeries: ...
```
Dr-Irv, line 1654: Add this overload.
```python
    @overload
    def __mul__(self, other: S1 | Self) -> Self: ...
```
This will fix the issue in line 1342 in `test_series.py`

Dr-Irv (changes requested):
I think I figured out your issues by pulling your PR and trying things.  I only tested with pyright in VS Code, so when you run the full tests, something might turn up with `mypy`.

Also - a request - can you NOT resolve conversations.  I will do it once I see that my comment has been resolved.

tests/test_series.py
```diff
@@ -1309,7 +1315,7 @@ def test_types_dot() -> None:
     n1 = np.array([[0, 1], [1, 2], [-1, -1], [2, 0]])
     check(assert_type(s1.dot(s2), Scalar), np.integer)
     check(assert_type(s1 @ s2, Scalar), np.integer)
-    check(assert_type(s1.dot(df1), "pd.Series[int]"), pd.Series, np.integer)
+    check(assert_type(s1.dot(df1), pd.Series), pd.Series, np.integer)
```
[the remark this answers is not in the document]
loicdiridollou: Addressed, that is a good idea!

tests/test_series.py
```diff
@@ -1287,9 +1335,15 @@ def test_series_min_max_sub_axis() -> None:
     sm = s1 * s2
     sd = s1 / s2
     check(assert_type(sa, pd.Series), pd.Series)
-    check(assert_type(ss, pd.Series), pd.Series)
+    check(
+        assert_type(ss, pd.Series),  # pyright: ignore[reportAssertTypeFailure]
```
loicdiridollou, line 1339: @Dr-Irv only thing that I am blocked on right now is these ones, they still get mapped to `Series[int]` although the two elements are `Series[Unknown]`

Dr-Irv (changes requested):
For safety, we should duplicate the tests in `test_types_element_wise_arithmetic()` but where `s` and `s2` have parameters `dtype=float`.  Then the results should all be `Series[float]`

Also, worth doing a set where `s` is `Series[int]` and `s2` is `Series[float]` and vice versa.

And we should do ones with the following.  If you create a `DataFrame` that contains numbers, let's say with a column named `s3`   then `df["s3"]` is `Series[Unknown]` in `pyright`, So now we want to test the arithmetic operations between `Series[int]` and `df["s3"]`, `Series[float]` and `df["s3"]` and `df["s3"]` and `df["s4"]` (where `s4` is another column of numbers).  You'd have to do the tests with all the symmetries.


So create a bunch of new copies of `test_types_element_wise_arithmetic()` (with different names) that test these different combinations as the first and second operands of the arithmetic operation, where `Series[Unknown]` comes from a column in a `DataFrame`
- `Series[int]` and `Series[float]`
- `Series[float]` and `Series[int]`
- `Series[float]` and `Series[float]`
- `Series[int]` and `Series[Unknown]`
- `Series[Unknown]` and `Series[int]`
- `Series[Unknown]` and `Series[Unknown]`
- `Series[float]` and `Series[Unknown]`
- `Series[Unknown]` and `Series[float]`

It might be worth just moving `test_types_element_wise_arithmetic()` into a new `test_series_arithmetic.py` file and just put all the new 8 versions there.

Except for division, the results involving `Series[Unknown]` should just be `Series`

pandas-stubs/core/series.pyi
```diff
@@ -1741,16 +1733,23 @@ class Series(IndexOpsMixin[S1], NDFrame):
...
-        other: int,
+        other: int | Series[int],
     ) -> Series[int]: ...
     @overload
     def __sub__(
         self,
         other: complex,
     ) -> Series[complex]: ...
     @overload
+    def __sub__(self, other: S1 | Self) -> Self: ...
```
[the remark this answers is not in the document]
loicdiridollou: Indeed that was the issue, thanks!

loicdiridollou:
I haven't added all the tests yet, this is quite a bit of work so will probably delay it to another PR and fix in priority based on if it causes more issues for the users.

Dr-Irv:
> I haven't added all the tests yet, this is quite a bit of work so will probably delay it to another PR and fix in priority based on if it causes more issues for the users.

I'm a bit hesitant to approve without the complete set of tests, because people do `s = df["x"]` a lot, and `s` is then `Series[Unknown]` and then they do arithmetic.  So I would like the full set of tests with the `Series[Unknown]` in there to be included.

I know it's a lot of work and really appreciate your hard work on this!

loicdiridollou:
@Dr-Irv almost there (sorry was out on a break), I am just having a hard time wrapping up the int/float on the left and Unknown on the right. Do you have an idea how to take that in consideration? Thanks!

Dr-Irv:
> @Dr-Irv almost there (sorry was out on a break), I am just having a hard time wrapping up the int/float on the left and Unknown on the right. Do you have an idea how to take that in consideration? Thanks!

I found something that worked for `__add__()`, but I just tested it in `pyright` inside VS Code, so don't know what will happen with `mypy`.  You'd also have to make similar changes for the reverse methods (e.g., `__radd__()`) and also for the other arithmetic operators.

The idea is that when you do `df["a"]`, we return a new type called `UnknownSeries` that is just a subclass of `Series[Any]`.  Then when we do the overloads, we can explicitly test for `UnknownSeries` as a result.  It's similar to `TimestampSeries`, etc.  

If we do that, ideally, we should change any method in all PYI files (and the associated tests) that has `-> Series:` to then return `UnknownSeries`.  

It might be worth to do a separate PR first that makes the change to do `UnknownSeries` in all the right places, and adjusts the tests that do `assert_type(somemethod, pd.Series)` to `assert_type(somemethod, UnknownSeries)` so we can see how that looks, and if it looks OK, we then come back to this PR where we can use `UnknownSeries` in all the right places.

I've placed those simple changes to make `__add__()` work when the second argument is unknown and the first one is not unknown here:  https://github.com/Dr-Irv/pandas-stubs/tree/pr1093

Let me know if you're willing to give the `UnknownSeries` idea a try by making the changes throughout, otherwise I can try it myself.

loicdiridollou:
Got it, that makes sense! I did not want to go that route too much originally since it was conflicting with the following issue https://github.com/pandas-dev/pandas-stubs/issues/718 but I will give it a try, I think it should work as you recommended it!
Thanks for the help there!

PR #1097, opened 2025-01-13
handle pyright inferring mixed float and int lists

Dr-Irv:
`pyright` with strict type checking is inferring that a list of mixed int and float is `list[int | float]`, so created overloads for `Series.__new__()` that makes that a `Series[float]`.


twoertwein (approved):
It would be nice if we could re-write all the overloads more generically, but I assume the union issue (since we use `TypeVar with bounds`) and the Series sub-classes prevent us from using a more generic approach

```py
    def __new__(
        cls,
        data: Sequence[S1],
        index: Axes | None = ...,
        *,
        dtype: type[S1] = ...,
        name: Hashable = ...,
        copy: bool = ...,
    ) -> Series[S1]: ...
```

PR #1105, opened 2025-02-04
Simplify and fix overloads for methods with `inplace` parameter

brianhelba:
The return value of these methods varies on the `inplace` argument.

### DIFF
--- a/pandas-stubs/core/series.pyi
+++ b/pandas-stubs/core/series.pyi
@@ -488,11 +488,10 @@ class Series(IndexOpsMixin[S1], NDFrame):
     def index(self) -> Index | MultiIndex: ...
     @index.setter
     def index(self, idx: Index) -> None: ...
-    # TODO: combine Level | Sequence[Level] github.com/python/mypy/issues/14311
     @overload
     def reset_index(
         self,
-        level: Sequence[Level] = ...,
+        level: Sequence[Level] | Level | None = ...,
         *,
         drop: Literal[False] = ...,
         name: Level = ...,
@@ -502,7 +501,7 @@ class Series(IndexOpsMixin[S1], NDFrame):
     @overload
     def reset_index(
         self,
-        level: Sequence[Level] = ...,
+        level: Sequence[Level] | Level | None = ...,
         *,
         drop: Literal[True],
         name: Level = ...,
@@ -512,37 +511,7 @@ class Series(IndexOpsMixin[S1], NDFrame):
     @overload
     def reset_index(
         self,
-        level: Sequence[Level] = ...,
-        *,
-        drop: bool = ...,
-        name: Level = ...,
-        inplace: Literal[True],
-        allow_duplicates: bool = ...,
-    ) -> None: ...
-    @overload
-    def reset_index(
-        self,
-        level: Level | None = ...,
-        *,
-        drop: Literal[False] = ...,
-        name: Level = ...,
-        inplace: Literal[False] = ...,
-        allow_duplicates: bool = ...,
-    ) -> DataFrame: ...
-    @overload
-    def reset_index(
-        self,
-        level: Level | None = ...,
-        *,
-        drop: Literal[True],
-        name: Level = ...,
-        inplace: Literal[False] = ...,
-        allow_duplicates: bool = ...,
-    ) -> Series[S1]: ...
-    @overload
-    def reset_index(
-        self,
-        level: Level | None = ...,
+        level: Sequence[Level] | Level | None = ...,
         *,
         drop: bool = ...,
         name: Level = ...,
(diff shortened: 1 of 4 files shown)


pandas-stubs/core/series.pyi
```diff
@@ -511,37 +510,7 @@ class Series(IndexOpsMixin[S1], NDFrame):
     @overload
     def reset_index(
         self,
-        level: Sequence[Level] = ...,
-        *,
-        drop: bool = ...,
-        name: Level = ...,
-        inplace: Literal[True],
-        allow_duplicates: bool = ...,
```
Dr-Irv, line 519: I don't think you can remove this overload.

pandas-stubs/core/series.pyi
```diff
@@ -1113,38 +1048,16 @@ class Series(IndexOpsMixin[S1], NDFrame):
...
         index: Renamer | Hashable | None = ...,
         *,
         axis: Axis | None = ...,
         copy: bool = ...,
-        inplace: bool = ...,
+        inplace: Literal[False] = ...,
         level: Level | None = ...,
         errors: IgnoreRaise = ...,
-    ) -> Series | None: ...
+    ) -> Series[S1]: ...
```
Dr-Irv, line 1060: This should be `Self`

Dr-Irv (changes requested):
Tests are failing.  Please make sure to run the tests locally before resubmitting.

I understand what you are trying to do here, and it seems correct, but I also would like to know what prompted you to make these changes.  Is there a test we can add that fails with the current stubs but passes with your changes?

pandas-stubs/core/series.pyi
brianhelba: Fixed, sorry for the mistake.

pandas-stubs/core/series.pyi
brianhelba: This seems safe to simplify. [There are only 3 return possibilities](https://pandas.pydata.org/docs/reference/api/pandas.Series.reset_index.html#pandas.Series.reset_index):
> **Series or DataFrame or None**
> When _drop_ is False (the default), a DataFrame is returned. The newly created columns will come first in the DataFrame, followed by the original Series values. When _drop_ is True, a _Series_ is returned. In either case, if `inplace=True`, no value is returned.

The return value can be fully determined by the value of `drop` and `inplace`, otherwise the signatures are identical.

---

Note, the previous 6 overloads with a split for `level: Sequence[Level]` and `level: Level | None` had the same other parameters and possible return values (`DataFrame`, `Series[S1]`, or `None`).

I assumed this duplication was a result of some auto-generated code, but if there's something else I'm failing to understand, please let me know.

brianhelba:
@Dr-Irv Thanks for the quick review and sorry for my delay in replying. I've just updated the PR and tests are now passing.

> I understand what you are trying to do here, and it seems correct, but I also would like to know what prompted you to make these changes. Is there a test we can add that fails with the current stubs but passes with your changes?

My initial reason for wanting this change was a mistake on my part (I thought that the overloads for `Series.rename` were not returning the correct type, but it was actually a separate bug in my project :sweat_smile:). However, after looking at this codebase, I realized that the overloads for `inplace` seemed needlessly complicated, so I wanted to help clean it up.

There are probably a few minor improvements here as well (adding overloads or return type annotations to methods which lacked them before), but it seemed unnecessary to try and make separate PRs (one for new annotations, one for cleanup).

brianhelba:
> Is there a test we can add that fails with the current stubs but passes with your changes?

We could probably add one for `DataFrame.where` or `DataFrame.mask` (places that totally lacked an overload before), but as mentioned, my primary motivation was just cleanup. Removing lines of code is often just as helpful as adding them.

pandas-stubs/core/series.pyi
Dr-Irv: These stubs were "taken over" from a project that Microsoft initiated, and it seems that the duplication was from that project.  So thanks for picking that up.

Dr-Irv:
> My initial reason for wanting this change was a mistake on my part (I thought that the overloads for `Series.rename` were not returning the correct type, but it was actually a separate bug in my project 😅). However, after looking at this codebase, I realized that the overloads for `inplace` seemed needlessly complicated, so I wanted to help clean it up.

No problem.  That makes sense.  We have some legacy code in the stubs that needs more and more cleanup.

> 
> There are probably a few minor improvements here as well (adding overloads or return type annotations to methods which lacked them before), but it seemed unnecessary to try and make separate PRs (one for new annotations, one for cleanup).

More PRs are always welcome!

Dr-Irv (approved):
Thanks @brianhelba

PR #1107, opened 2025-02-07
Type `tolerance` in `reindex_like`, remove unused function

MarcoGorelli:

### DIFF
--- a/pandas-stubs/core/construction.pyi
+++ b/pandas-stubs/core/construction.pyi
@@ -2,13 +2,6 @@ from collections.abc import Sequence
 
 import numpy as np
 from pandas.core.arrays.base import ExtensionArray
-from pandas.core.indexes.api import Index
-from pandas.core.series import Series
-
-from pandas._typing import (
-    ArrayLike,
-    Dtype,
-)
 
 from pandas.core.dtypes.dtypes import ExtensionDtype
 
@@ -22,12 +15,3 @@ def sanitize_array(
     data, index, dtype=..., copy: bool = ..., raise_cast_failure: bool = ...
 ): ...
 def is_empty_data(data) -> bool: ...
-def create_series_with_explicit_dtype(
-    data=...,
-    index: ArrayLike | Index | None = ...,
-    dtype: Dtype | None = ...,
-    name: str | None = ...,
-    copy: bool = ...,
-    fastpath: bool = ...,
-    dtype_if_empty: Dtype = ...,
-) -> Series: ...
--- a/pandas-stubs/core/generic.pyi
+++ b/pandas-stubs/core/generic.pyi
@@ -294,14 +294,6 @@ class NDFrame(indexing.IndexingMixin):
     ) -> _str: ...
     def take(self, indices, axis=..., **kwargs) -> Self: ...
     def __delitem__(self, idx: Hashable) -> None: ...
-    def reindex_like(
-        self,
-        other,
-        method: _str | None = ...,
-        copy: _bool = ...,
-        limit=...,
-        tolerance=...,
-    ) -> Self: ...
     @overload
     def drop(
         self,
--- a/pandas-stubs/core/generic.pyi
+++ b/pandas-stubs/core/generic.pyi
@@ -29,6 +29,7 @@ from typing_extensions import (
 from pandas._libs.lib import NoDefault
 from pandas._typing import (
     S1,
+    AnyArrayLike,
     ArrayLike,
     Axis,
     AxisIndex,
@@ -48,6 +49,7 @@ from pandas._typing import (
     Level,
     P,
     ReplaceMethod,
+    Scalar,
     SortKind,
     StorageOptions,
     T,
@@ -300,7 +302,7 @@ class NDFrame(indexing.IndexingMixin):
         method: _str | None = ...,
         copy: _bool = ...,
         limit=...,
-        tolerance=...,
+        tolerance: Scalar | AnyArrayLike | Sequence[Scalar] = ...,
     ) -> Self: ...
     @overload
     def drop(
--- a/pandas-stubs/core/generic.pyi
+++ b/pandas-stubs/core/generic.pyi
@@ -29,7 +29,6 @@ from typing_extensions import (
 from pandas._libs.lib import NoDefault
 from pandas._typing import (
     S1,
-    AnyArrayLike,
     ArrayLike,
     Axis,
     AxisIndex,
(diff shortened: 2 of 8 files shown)


pandas-stubs/core/construction.pyi
```diff
@@ -22,12 +15,3 @@ def sanitize_array(
     data, index, dtype=..., copy: bool = ..., raise_cast_failure: bool = ...
 ): ...
 def is_empty_data(data) -> bool: ...
-def create_series_with_explicit_dtype(
```
MarcoGorelli, line 25: drive-by: this was removed from pandas in https://github.com/pandas-dev/pandas/pull/49342/files

pandas-stubs/core/indexes/datetimes.pyi
```diff
@@ -39,7 +38,7 @@ from pandas.tseries.offsets import BaseOffset
 class DatetimeIndex(DatetimeTimedeltaMixin[Timestamp], DatetimeIndexProperties):
     def __init__(
         self,
-        data: ArrayLike | AnyArrayLike | list | tuple,
+        data: AnyArrayLike | list | tuple,
```
MarcoGorelli, line 41: also drive-by: `AnyArrayLike` is more generic than `ArrayLike`, so `ArrayLike` can be removed from here, right?

pandas-stubs/core/reshape/encoding.pyi
```diff
@@ -3,20 +3,17 @@ from collections.abc import (
...
-    ArrayLike,
+    AnyArrayLike,
     Dtype,
     HashableT1,
     HashableT2,
 )
 
 def get_dummies(
-    data: ArrayLike | DataFrame | Series,
+    data: AnyArrayLike | DataFrame,
```
MarcoGorelli, line 16: Index is supported here too, so I think `AnyArrayLike` could replace `ArrayLike` and `Series`?

pandas-stubs/core/series.pyi
```diff
@@ -1151,7 +1151,7 @@ class Series(IndexOpsMixin[S1], NDFrame):
         method: _str | FillnaOptions | Literal["nearest"] | None = ...,
         copy: _bool = ...,
         limit: int | None = ...,
-        tolerance: float | None = ...,
+        tolerance: Scalar | AnyArrayLike | Sequence[Scalar] = ...,
```
MarcoGorelli, line 1154: the test I've added would have failed on `main` without this change

pandas-stubs/core/indexes/datetimes.pyi
Dr-Irv: No.   From `_typing.pyi`  :
```python
ArrayLike: TypeAlias = ExtensionArray | np.ndarray[Any, Any]
AnyArrayLike: TypeAlias = Index[Any] | Series[Any] | np.ndarray[Any, Any]
```
So `ArrayLike` has `ExtensionArray` but not `Index` or `Series`

Although maybe we should change that in this PR in `_typing.py` so it corresponds to pandas so that `AnyArrayLike` is a union of `ArrayLike` with `Index[Any]` and `Series[Any]`

pandas-stubs/core/reshape/encoding.pyi
Dr-Irv: docs don't mention `Index`, so maybe open an issue there.  I'm fine with this change.

pandas-stubs/core/generic.pyi
```diff
@@ -300,7 +302,7 @@ class NDFrame(indexing.IndexingMixin):
         method: _str | None = ...,
         copy: _bool = ...,
         limit=...,
-        tolerance=...,
+        tolerance: Scalar | AnyArrayLike | Sequence[Scalar] = ...,
```
Dr-Irv, line 305: Since `reindex_like` is defined in both `series.pyi` and `frame.pyi`, we can remove it here.  Can you do that in this PR?

tests/test_frame.py
```diff
@@ -2663,6 +2663,13 @@ def test_frame_reindex() -> None:
     df.reindex([2, 1, 0])
 
 
+def test_frame_reindex_like() -> None:
+    # GH 84
+    df = pd.DataFrame({"a": [1, 2, 3]}, index=[0, 1, 2])
+    other = pd.DataFrame({"a": [1, 2]}, index=[1, 0])
+    df.reindex_like(other, method="nearest", tolerance=[0.5, 0.2])
```
Dr-Irv, line 2670:
```suggestion
    check(assert_type(df.reindex_like(other, method="nearest", tolerance=[0.5, 0.2]), pd.DataFrame), pd.DataFrame)
```

tests/test_series.py
```diff
@@ -3552,3 +3552,14 @@ def test_series_int_float() -> None:
     check(
         assert_type(pd.Series([1, 2.0, 3]), "pd.Series[float]"), pd.Series, np.float64
     )
+
+
+def test_series_reindex() -> None:
+    s = pd.Series([1, 2, 3], index=[0, 1, 2])
+    s.reindex([2, 1, 0])
```
Dr-Irv, line 3559:
```suggestion
    check(assert_type(s.reindex([2, 1, 0]), "pd.Series[int]"), pd.Series, np.integer)
```

tests/test_series.py
```diff
@@ -3552,3 +3552,14 @@ def test_series_int_float() -> None:
...
+
+
+def test_series_reindex_like() -> None:
+    s = pd.Series([1, 2, 3], index=[0, 1, 2])
+    other = pd.Series([1, 2], index=[1, 0])
+    s.reindex_like(other, method="nearest", tolerance=[0.5, 0.2])
```
Dr-Irv, line 3565:
```suggestion
    check(assert_type(s.reindex_like(other, method="nearest", tolerance=[0.5, 0.2]), "pd.Series[int]"), pd.Series, np.integer)
```

Dr-Irv (changes requested):
Thanks for the PR.  Have a few changes to make.

pandas-stubs/core/indexes/datetimes.pyi
MarcoGorelli: ah thanks i was looking at the pandas `AnyArrayLike` and noticed that that one included `ArrayLike`

I'll leave this out of the current PR then 👍

Dr-Irv (approved):
Thanks @MarcoGorelli

PR #1108, opened 2025-02-07
type freq in shift

### LINKED ISSUE #1111: Series.shift does not always preserve input dtype, but is typed as if it does
**Describe the bug**
``

**To Reproduce**

```python
import pandas as pd


reveal_type(pd.Series([1,2,3]).shift())
#  t.py:4: note: Revealed type is "pandas.core.series.Series[builtins.int]"
```
however, the result is actually `Series['float']`

It's not clear to me if/how pandas-stubs distinguishes ints and nullable ints https://github.com/pandas-dev/pandas-stubs/issues/1110

 - OS: [e.g. Windows, Linux, MacOS] linux
 - OS Version [e.g. 22] 22.04
 - python version 3.12.5
 - version of type checker mypy 1.15
 - version of installed `pandas-stubs` 2.2.3.241126

**Additional context**
noticed looking into https://github.com/pandas-dev/pandas-stubs/pull/1108

MarcoGorelli:
closes #1111 


tests/test_series.py
```diff
@@ -416,10 +416,11 @@ def test_types_sort_values_with_key() -> None:
 
 
 def test_types_shift() -> None:
-    s = pd.Series([1, 2, 3])
+    s = pd.Series([1, 2, 3], index=pd.date_range("2020", periods=3))
```
MarcoGorelli, line 419: 🤔 trying to add a test for this, the type actually changes from Series['int'] to Series['float'], taking a look

tests/test_series.py
```diff
@@ -416,10 +416,20 @@ def test_types_sort_values_with_key() -> None:
...
 
 def test_types_shift() -> None:
-    s = pd.Series([1, 2, 3])
-    s.shift()
-    s.shift(axis=0, periods=1)
-    s.shift(-1, fill_value=0)
+    s = pd.Series([1, 2, 3], index=pd.date_range("2020", periods=3))
+    # Return type "Series[int]"" not quite correct
+    # - https://github.com/pandas-dev/pandas-stubs/issues/1111
+    # - https://github.com/pandas-dev/pandas-stubs/issues/1110
```
MarcoGorelli, line 422: I've kept the return type of `shift` as-is, in order to focus on typing `freq`

Happy to update the return type too based on discussion in the issues

Dr-Irv: Based on the tests you listed below, it would be really hard to make sure that we are getting the generic type correct.  Dependent on the underlying index (which we can't track via static typing), you would get `Series[int]` or `Series[float]`.  

I think the best solution here is to change the return type for `shift()` to be `Series`

Alternatively, we could have an overload, where we have something like:
```python
@overload
    def shift(
        self : Series[int],
        periods: int = ...,
        freq=...,
        axis: Axis = ...,
        fill_value: Hashable | None = ...,
    ) -> Series[int] | Series[float]: ...
@overload
    def shift(
        self : Series[int]
        periods: int = ...,
        freq=...,
        axis: Axis = ...,
        fill_value: Hashable | None = ...,
    ) -> Self: ...
```
but I think the first option of returning `Series` is easier to use.  And you can then put that in this PR

Dr-Irv (changes requested):
Let's change `Series.shift()` to return `Series` rather than `Self`

pandas-stubs/core/series.pyi
```diff
@@ -1223,7 +1223,7 @@ class Series(IndexOpsMixin[S1], NDFrame):
         freq: Frequency | timedelta | None = ...,
         axis: AxisIndex = ...,
         fill_value: object | None = ...,
-    ) -> Series[S1]: ...
+    ) -> Series[Any]: ...
```
Dr-Irv, line 1226: Just leave out the `Any` here, i.e., use `-> Series: ...`

tests/test_series.py
```diff
@@ -417,19 +417,16 @@ def test_types_sort_values_with_key() -> None:
 
 def test_types_shift() -> None:
     s = pd.Series([1, 2, 3], index=pd.date_range("2020", periods=3))
-    # Return type "Series[int]"" not quite correct
-    # - https://github.com/pandas-dev/pandas-stubs/issues/1111
-    # - https://github.com/pandas-dev/pandas-stubs/issues/1110
-    check(assert_type(s.shift(), "pd.Series[int]"), pd.Series, np.floating)
+    check(assert_type(s.shift(), "pd.Series[Any]"), pd.Series, np.floating)
```
Dr-Irv, line 420: if you change the PYI file to not reference `Any`, in the tests, you just do `assert_type(s.shift(), pd.Series)` without the quotes

Dr-Irv (changes requested):

tests/test_series.py
```diff
@@ -416,10 +416,15 @@ def test_types_sort_values_with_key() -> None:
...
 
 def test_types_shift() -> None:
-    s = pd.Series([1, 2, 3])
-    s.shift()
-    s.shift(axis=0, periods=1)
-    s.shift(-1, fill_value=0)
+    s = pd.Series([1, 2, 3], index=pd.date_range("2020", periods=3))
+    check(assert_type(s.shift(), pd.Series), pd.Series, np.floating)
+    check(
+        assert_type(s.shift(axis=0, periods=1), "pd.Series[Any]"),
```
MarcoGorelli, line 422:
```suggestion
        assert_type(s.shift(axis=0, periods=1), pd.Series),
```

Dr-Irv (approved):
thanks @MarcoGorelli

PR #1112, opened 2025-02-08
type is_dtype_equal

MarcoGorelli:

There's already a test for this one using `assert_type`;

https://github.com/pandas-dev/pandas-stubs/blob/feebd4707e594fd1ba7d30fd54e38885872d5ddd/tests/test_api_types.py#L125-L126

the function just wasn't typed


MarcoGorelli:
> I'd prefer to just delete the line from the stubs.

sure thing, always happy to delete code

Dr-Irv (changes requested):
We have references to `is_dtype_equal()` in `pandas-stubs\api\types\__init__.pyi` and `pandas-stubs\core\dtypes\api.pyi` that should be removed as well.

tests/extension/decimal/array.py
```diff
@@ -28,7 +28,7 @@
 )
 
 from pandas.core.dtypes.base import ExtensionDtype
-from pandas.core.dtypes.common import (
+from pandas.core.dtypes.common import (  # type: ignore[attr-defined]
```
Dr-Irv: UGH.  Now that I see that this is used in a test (this one is copied from the pandas source), I think we need to open an issue in `pandas` about documenting this.  I'll do that, and keep this PR on hold.

Dr-Irv: Created https://github.com/pandas-dev/pandas/issues/60905

Dr-Irv:
@MarcoGorelli ok to change this from draft - it will get documented

Dr-Irv:
@MarcoGorelli pinging you again about whether you want to take this out of draft status

MarcoGorelli:
sure, thanks - so, undraft for now and then we type it once it's documented?

Dr-Irv:
> sure, thanks - so, undraft for now and then we type it once it's documented?

Sorry - my mistake.  Since we decided to document it, let's type it for now, which was the original PR here.

MarcoGorelli:
sure, done, thanks!

Dr-Irv (approved):
thanks @MarcoGorelli

PR #1119, opened 2025-02-13
type Series.info

MarcoGorelli:

### DIFF
--- a/pandas-stubs/core/series.pyi
+++ b/pandas-stubs/core/series.pyi
@@ -1224,6 +1224,13 @@ class Series(IndexOpsMixin[S1], NDFrame):
         axis: AxisIndex = ...,
         fill_value: object | None = ...,
     ) -> Series: ...
+    def info(
+        self,
+        verbose: bool | None = ...,
+        buf: WriteBuffer[str] = ...,
+        memory_usage: bool | Literal["deep"] | None = ...,
+        show_counts: bool | None = ...,
+    ) -> None: ...
     def memory_usage(self, index: _bool = ..., deep: _bool = ...) -> int: ...
     def isin(self, values: Iterable | Series[S1] | dict) -> Series[_bool]: ...
     def between(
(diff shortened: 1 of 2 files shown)


pandas-stubs/core/series.pyi
```diff
@@ -1224,6 +1224,13 @@ class Series(IndexOpsMixin[S1], NDFrame):
         axis: AxisIndex = ...,
         fill_value: object | None = ...,
     ) -> Series: ...
+    def info(
+        self,
+        verbose: bool | None = ...,
+        buf: WriteBuffer[str] = ...,
+        memory_usage: bool | Literal["deep"] | None = ...,
+        show_counts: bool | None = ...,
+    ) -> None: ...
```
MarcoGorelli, line 1233: technically, `Series.info` also has a `max_cols` argument, but it raises if it gets passed, as it's only available for `DataFrame.info`

OK to leave it out of the signature in the types, so as to discourage its usage?

Dr-Irv:
Our policy is to not put things in the stubs that are not documented.

Maybe that's an error in `pandas` and `Series.info()` should be documented.  Do you want to open an issue there?

pandas-stubs/core/series.pyi
MarcoGorelli: first - thanks for all your reviews here!

second - i think it's meant to be documented, it was added very intentionally in https://github.com/pandas-dev/pandas/pull/37320#pullrequestreview-819720418 - i'll make a pr in pandas

Dr-Irv:
Once that PR in `pandas` is approved/accepted, let me know and I'll here again

MarcoGorelli:
Sure - it's merged now

pandas-stubs/core/series.pyi
Dr-Irv: OK - leaving out `max_cols` sounds like a good idea

Dr-Irv (approved):
thanks @MarcoGorelli

PR #1120, opened 2025-02-14
type datetimeindex constructor

MarcoGorelli:


pandas-stubs/core/indexes/multi.pyi
```diff
@@ -44,7 +44,7 @@ class MultiIndex(Index[Any]):
         names=...,
         dtype=...,
         copy=...,
-        name=...,
+        name: Hashable = ...,
         verify_integrity: bool = ...,
         _set_identity: bool = ...,
     ) -> None: ...
```
Dr-Irv, line 50: We have an opportunity for cleanup here.  No need to have both `__new__()` and `__init__()`.  We should just have `__init__()`.  We use `__new__()` instead of `__init__()` in the stubs when we want to return different subclasses based on the arguments.  But for `MultiIndex`, that isn't the case.  So we can delete `__new__()`.

Ideally, we'd have types for all of the arguments, which will have to be based on the 3.0 docs, because the docs did get updated to correspond to what is in the code (which doesn't have `_set_identity`.  

For this PR, I'm OK if you don't do them all, BUT `names` and `name` need to be `SequenceNotStr[Hashable]`

Dr-Irv (changes requested):
caught one issue regarding `MultiIndex`

pandas-stubs/core/indexes/multi.pyi
MarcoGorelli: thanks Irv! sure, I corrected `name` for multiindex, sorry for not having spotted that

MarcoGorelli: Looks like we might need `MultiIndex.__new__` too, else the test I added fails pyright with
```
  /home/marcogorelli/pandas-stubs-dev/tests/test_indexes.py:70:62 - error: Argument of type "list[str]" cannot be assigned to parameter "name" of type "Hashable" in function "__new__"
    "list[str]" is incompatible with protocol "Hashable"
```

Dr-Irv (changes requested):
The only thing left to do is to remove `__init__()` from `MultiIndex`

Dr-Irv (approved):
thanks @MarcoGorelli

PR #1088, opened 2025-01-09
GH1089 Fix pandas nightly and migrate tests to new framework

loicdiridollou:
- [x] Closes #1089 
- [x] Tests added: Please use `assert_type()` to assert the type of any return value

PR #1121, opened 2025-02-16
GH1088 Test Migration pt1

### LINKED ISSUE #1089: Fix pandas nightly test build and migrate tests over new framework

loicdiridollou:
- [x] Closes #1088 

This is a subset of the changes for GH1089 since it has just become a much bigger issue than I intended in the first place so splitting the changes to only have migration of the tests and another PR will address changes in the pyi files.


tests/test_series.py
```diff
@@ -681,8 +709,8 @@ def test_types_scalar_arithmetic() -> None:
     res_pow3: pd.Series = s.pow(0.5)
```
Dr-Irv, line 709: Why didn't you convert these as well?

tests/test_series.py
```diff
@@ -681,8 +709,8 @@ def test_types_scalar_arithmetic() -> None:
...
 
 
-# GH 103
 def test_types_complex_arithmetic() -> None:
+    # GH 103
     c = 1 + 1j
     s = pd.Series([1.0, 2.0, 3.0])
     x = s + c
```
Dr-Irv, line 716: could convert these as well (with a TODO)

Dr-Irv (commented):
I'm OK with this, but wondering why you didn't convert all of the test statements.

tests/test_series.py
loicdiridollou: Because if I convert them it would have to hint at `pd.Series[float]` but right now the type is `pd.Series[Unknown]` so I would rather fix the typehints properly and add the proper check/asert_type when fixed than putting just `pd.Series` (issue that we had in the other PR).

tests/test_series.py
loicdiridollou: Same as above, right now the typehinting would lead to `pd.Series[Unknown]` so would rather make a separate PR to fix the typehints altogether.

tests/test_series.py
Dr-Irv: OK, that makes sense

tests/test_series.py
Dr-Irv: OK, that makes sense

Dr-Irv (approved):
thanks @loicdiridollou

PR #1123, opened 2025-02-18
Align AnyArrayLike with pandas._typing definition

MarcoGorelli:

This came up in another PR (https://github.com/pandas-dev/pandas-stubs/pull/1107#discussion_r1946700293)

It seems that the definitions have slightly drifted - OK to align them?


Dr-Irv (approved):
Thanks @MarcoGorelli   There is more alignment to do - just need people to work on it!

PR #1125, opened 2025-02-18
Fix comment commands to source branches 

### LINKED ISSUE #1124: Fix the comment command for nightly on a PR
@Dr-Irv I understand the error, the initial issue with the comment commands was that it would checkout the main branch and not the one we want to merge so I fixed this. Now the issue is that it is looking for the branch in the pandas-stubs repo although in the current setting the branch in on my personal repo (loicdiridollou/pandas-stubs).
I will fix this in a separate PR, just need to test it on a dummy repo on my end so would be preferable to merge this one (I tested the changes locally running `poetry run poe pytest --nightly` and it was good.

_Originally posted by @loicdiridollou in https://github.com/pandas-dev/pandas-stubs/issues/1122#issuecomment-2666851568_

loicdiridollou:
Fix comment commands to source branches to run nightly tests in correct repo

- [x] Closes #1124

### DIFF
--- a/.github/workflows/comment_commands.yml
+++ b/.github/workflows/comment_commands.yml
@@ -36,13 +36,16 @@ jobs:
             })
             core.setOutput('sha', pr.data.head.sha)
             core.setOutput('branch', pr.data.head.ref)
+            core.setOutput('repository', pr.data.head.repo.full_name)
 
       - name: Checkout code on the correct branch
         uses: actions/checkout@v4
         with:
           # context is not aware which branch to checkout so it would otherwise
-          # default to main
+          # default to main (we also need repo name to source from the right user
+          # otherwise it will look for the branch in pandas-stubs repo)
           ref: ${{ steps.get-branch-info.outputs.branch }}
+          repository: ${{ steps.get-branch-info.outputs.repository }}
 
       - name: Install project dependencies
         uses: ./.github/setup


Dr-Irv:
Well, I tried it above, and it skipped the job

Dr-Irv:
> Can you retry? It was working fine when I ran it (maybe check for extra spaces) Also those changes will only take effect once on main since it is not an action related to the PR itself (different classification at the header of the workflow file)

shows it is skipped:  https://github.com/pandas-dev/pandas-stubs/actions/runs/13400980753

loicdiridollou:
Okay will need to dig deeper on this one this is quite weird because I can trigger it and you can't

Dr-Irv:
even on  your last trigger, it skipped it:  https://github.com/pandas-dev/pandas-stubs/actions/runs/13401031918

loicdiridollou:
@Dr-Irv I retried triggering the process a couple of times and haven't seen the issues we had earlier where the action would just not run at all. Please let me know if you are good with the changes.
Thanks

Dr-Irv (approved):
thanks @loicdiridollou

Dr-Irv:
I tried it and it seemed to work

Dr-Irv:
So I think if I as a maintainer try it, it doesn't work, but it worked for you because you have access to your repo.  See https://github.com/pandas-dev/pandas-stubs/actions/runs/13635376466

PR #1126, opened 2025-02-19
type multiindex constructors

MarcoGorelli:

### DIFF
--- a/pandas-stubs/core/indexes/multi.pyi
+++ b/pandas-stubs/core/indexes/multi.pyi
@@ -1,6 +1,7 @@
 from collections.abc import (
     Callable,
     Hashable,
+    Iterable,
     Sequence,
 )
 from typing import (
@@ -16,6 +17,7 @@ from pandas.core.indexes.base import Index
 from typing_extensions import Self
 
 from pandas._typing import (
+    AnyArrayLike,
     Dtype,
     DtypeArg,
     HashableT,
@@ -27,31 +29,42 @@ from pandas._typing import (
 class MultiIndex(Index[Any]):
     def __new__(
         cls,
-        levels=...,
-        codes=...,
-        sortorder=...,
+        levels: SequenceNotStr[SequenceNotStr[Hashable]] = ...,
+        codes: SequenceNotStr[SequenceNotStr[int]] = ...,
+        sortorder: int | None = ...,
         names: SequenceNotStr[Hashable] = ...,
-        dtype=...,
-        copy=...,
+        copy: bool = ...,
         name: SequenceNotStr[Hashable] = ...,
         verify_integrity: bool = ...,
         _set_identity: bool = ...,
     ) -> Self: ...
     @classmethod
     def from_arrays(
-        cls, arrays, sortorder=..., names: SequenceNotStr[Hashable] = ...
+        cls,
+        arrays: SequenceNotStr[SequenceNotStr[Hashable]] | SequenceNotStr[AnyArrayLike],
+        sortorder: int | None = ...,
+        names: SequenceNotStr[Hashable] = ...,
     ) -> Self: ...
     @classmethod
     def from_tuples(
-        cls, tuples, sortorder=..., names: SequenceNotStr[Hashable] = ...
+        cls,
+        tuples: Iterable[tuple[Hashable, ...]],
+        sortorder: int | None = ...,
+        names: SequenceNotStr[Hashable] = ...,
     ) -> Self: ...
     @classmethod
     def from_product(
-        cls, iterables, sortorder=..., names: SequenceNotStr[Hashable] = ...
+        cls,
+        iterables: SequenceNotStr[Iterable[Hashable]],
+        sortorder: int | None = ...,
+        names: SequenceNotStr[Hashable] = ...,
     ) -> Self: ...
     @classmethod
     def from_frame(
-        cls, df, sortorder=..., names: SequenceNotStr[Hashable] = ...
+        cls,
+        df: pd.DataFrame,
+        sortorder: int | None = ...,
+        names: SequenceNotStr[Hashable] = ...,
     ) -> Self: ...
     @property
     def shape(self): ...
--- a/pandas-stubs/core/indexes/multi.pyi
+++ b/pandas-stubs/core/indexes/multi.pyi
@@ -1,7 +1,6 @@
 from collections.abc import (
     Callable,
     Hashable,
-    Iterable,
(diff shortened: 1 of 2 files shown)


pandas-stubs/core/indexes/multi.pyi
```diff
@@ -27,31 +29,42 @@ from pandas._typing import (
...
+        copy: bool = ...,
         name: SequenceNotStr[Hashable] = ...,
         verify_integrity: bool = ...,
         _set_identity: bool = ...,
     ) -> Self: ...
     @classmethod
     def from_arrays(
-        cls, arrays, sortorder=..., names: SequenceNotStr[Hashable] = ...
+        cls,
+        arrays: SequenceNotStr[SequenceNotStr[Hashable]] | SequenceNotStr[AnyArrayLike],
```
MarcoGorelli, line 44: This one's a bit odd

The name is `from_arrays`, and the pandas docstring says that this should be a sequence of array-like

But, the pandas docs (and tests) have several examples of passing lists, which are not `ArrayLike`

`lists` are not included in `AnyArrayLike`, so I've used `SequenceNotStr[Hashable] | AnyArrayLike`

pandas-stubs/core/indexes/multi.pyi
```diff
@@ -27,31 +29,42 @@ from pandas._typing import (
...
         name: SequenceNotStr[Hashable] = ...,
         verify_integrity: bool = ...,
         _set_identity: bool = ...,
     ) -> Self: ...
     @classmethod
     def from_arrays(
-        cls, arrays, sortorder=..., names: SequenceNotStr[Hashable] = ...
+        cls,
+        arrays: SequenceNotStr[SequenceNotStr[Hashable] | AnyArrayLike],
+        sortorder: int | None = ...,
```
MarcoGorelli, line 45: This one's a bit odd

The name is `from_arrays`, and the pandas docstring says that this should be a sequence of array-like

But, the pandas docs (and tests) have several examples of passing lists, which are not `ArrayLike`

`lists` are not included in `AnyArrayLike`, so I've used `SequenceNotStr[Hashable] | AnyArrayLike`

pandas-stubs/core/indexes/multi.pyi
```diff
@@ -27,31 +29,42 @@ from pandas._typing import (
 class MultiIndex(Index[Any]):
     def __new__(
         cls,
-        levels=...,
-        codes=...,
-        sortorder=...,
+        levels: SequenceNotStr[SequenceNotStr[Hashable] | AnyArrayLike] = ...,
```
Dr-Irv:
```suggestion
        levels: Sequence[SequenceNotStr[Hashable]] | AnyArrayLike] = ...,
```
The outer sequence doesn't need to be `NotStr` because the inner one is already not a string.

pandas-stubs/core/indexes/multi.pyi
```diff
@@ -27,31 +29,42 @@ from pandas._typing import (
...
+        codes: SequenceNotStr[SequenceNotStr[int]] = ...,
```
Dr-Irv:
```suggestion
        codes: Sequence[Sequence[int]] = ...,
```
We know it has to be a sequence of sequence of int - don't need `SequenceNotStr` here.

We only use that when we want to allow a sequence that is specifically not a specific string.

pandas-stubs/core/indexes/multi.pyi
```diff
@@ -27,31 +29,42 @@ from pandas._typing import (
...
+        sortorder: int | None = ...,
         names: SequenceNotStr[Hashable] = ...,
-        dtype=...,
-        copy=...,
+        copy: bool = ...,
         name: SequenceNotStr[Hashable] = ...,
         verify_integrity: bool = ...,
         _set_identity: bool = ...,
```
Dr-Irv, line 39: Remove `_set_identity` - it's not in there any more

pandas-stubs/core/indexes/multi.pyi
Dr-Irv: I think it should be `Sequence[Axes]` because this works:
```python
>>> pd.MultiIndex.from_arrays([range(2), [3,4]], names=["a", "b"])
MultiIndex([(0, 3),
            (1, 4)],
           names=['a', 'b'])
>>> pd.MultiIndex.from_arrays([{1:"x", 2:"y"}, [3,4]], names=["a", "b"])
MultiIndex([(1, 3),
            (2, 4)],
           names=['a', 'b'])
```

pandas-stubs/core/indexes/multi.pyi
```diff
@@ -27,31 +29,42 @@ from pandas._typing import (
...
+        cls,
+        arrays: SequenceNotStr[SequenceNotStr[Hashable] | AnyArrayLike],
+        sortorder: int | None = ...,
+        names: SequenceNotStr[Hashable] = ...,
     ) -> Self: ...
     @classmethod
     def from_tuples(
-        cls, tuples, sortorder=..., names: SequenceNotStr[Hashable] = ...
+        cls,
+        tuples: Iterable[tuple[Hashable, ...]],
```
Dr-Irv, line 51:
```suggestion
        tuples: Sequence[tuple[Hashable, ...]],
```
We don't want to allow sets.

pandas-stubs/core/indexes/multi.pyi
```diff
@@ -27,31 +29,42 @@ from pandas._typing import (
...
+        sortorder: int | None = ...,
+        names: SequenceNotStr[Hashable] = ...,
     ) -> Self: ...
     @classmethod
     def from_product(
-        cls, iterables, sortorder=..., names: SequenceNotStr[Hashable] = ...
+        cls,
+        iterables: SequenceNotStr[Iterable[Hashable]],
```
Dr-Irv, line 58:
```suggestion
        iterables: Sequence[SequenceNotStr[Hashable]],
```
We don't want to allow sets.

Dr-Irv (changes requested):

pandas-stubs/core/indexes/multi.pyi
Dr-Irv: > 🤔 maybe `ListLike` would be more suitable

Problem is that `ListLike` allows a single string, so we can't use it.  I think `Axes` is the right choice.

MarcoGorelli:
thanks for your review and explanations!

Dr-Irv (approved):
thanks @MarcoGorelli

PR #1129, opened 2025-02-21
type dataframe.replace

MarcoGorelli:

### DIFF
--- a/pandas-stubs/core/frame.pyi
+++ b/pandas-stubs/core/frame.pyi
@@ -8,7 +8,6 @@ from collections.abc import (
     Sequence,
 )
 import datetime as dt
-from re import Pattern
 import sys
 from typing import (
     Any,
@@ -113,6 +112,7 @@ from pandas._typing import (
     RandomState,
     ReadBuffer,
     Renamer,
+    ReplaceValue,
     Scalar,
     ScalarT,
     SequenceNotStr,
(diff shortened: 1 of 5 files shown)


pandas-stubs/core/frame.pyi
```diff
@@ -799,23 +799,53 @@ class DataFrame(NDFrame, OpsMixin, _GetItemHack):
...
+            Scalar
+            | NAType
+            | Sequence[Scalar | Pattern]
+            | Mapping[Hashable, Scalar | Pattern]
+            | Pattern
+            | None
+        ) = ...,
+        value: (
+            Scalar | NAType | Sequence[Scalar] | Mapping[Hashable, Scalar] | None
+        ) = ...,
```
MarcoGorelli, line 812: `value` was originally annotated to include `Pattern`, but is that right?

It can accept regex strings, sure:
```python
df = pd.DataFrame({'a': ['a', 'ab', 'ba']})
print(df.replace(f'^a(.*)', r'\1', regex=True))
```

But trying to use `df.replace(f'^a(.*)', re.compile(r'\1'), regex=True)` raises, and it doesn't seem to useful to use `re.compile` here if we can't include captured groups in `value`

There's no tests which pass a compiled regex to `value`. I think it's OK to remove the `Pattern` annotation from `value`?

Dr-Irv: Yes.  This was something that was in the Microsoft stubs that we copied over.

pandas-stubs/core/frame.pyi
```diff
@@ -799,23 +799,53 @@ class DataFrame(NDFrame, OpsMixin, _GetItemHack):
...
         *,
         inplace: Literal[True],
         limit: int | None = ...,
```
Dr-Irv, line 815: `limit` is deprecated, so we should remove it.

pandas-stubs/core/frame.pyi
```diff
@@ -799,23 +799,53 @@ class DataFrame(NDFrame, OpsMixin, _GetItemHack):
...
-        regex=...,
+        regex: (
+            bool
+            | str
+            | Pattern
+            | Sequence[str | Pattern]
+            | Mapping[Hashable, str | Pattern]
+        ) = ...,
         method: ReplaceMethod = ...,
```
Dr-Irv, line 823: `method` is deprecated so we should remove it

Dr-Irv (changes requested):
Aside from removing the deprecated `limit` and `method` arguments, can you make what is in `series.pyi` for `Series.replace()` have arguments that correspond?

MarcoGorelli:
Thanks @Dr-Irv !

It was doing my head in trying to complete this so I made an alias `ReplaceValue`, is it alright?

Dr-Irv:
> Thanks @Dr-Irv !
> 
> It was doing my head in trying to cover all the possibilities in all arguments so I made an alias `ReplaceValue`, is it alright?

Yes.  Let me know when it's ready for review

MarcoGorelli:
Ready for review, thanks!

PR #1130, opened 2025-02-25
type bins in seriesgroupby.value_counts

MarcoGorelli:


Dr-Irv (approved):
Thanks @MarcoGorelli .  Took a while for me to figure out where `bins` came from, as it isn't documented in 2.2., but that was a doc error already fixed in `main`

PR #1136, opened 2025-03-02
type dataframe.transpose

MarcoGorelli:


pandas-stubs/core/series.pyi
```diff
@@ -1598,7 +1598,7 @@ class Series(IndexOpsMixin[S1], NDFrame):
         bins: int | None = ...,
         dropna: _bool = ...,
     ) -> Series[float]: ...
-    def transpose(self, *args, **kwargs) -> Series[S1]: ...
+    def transpose(self, axes: None = ...) -> Series[S1]: ...
```
MarcoGorelli, line 1601: The only argument that pandas actually lets us pass at runtime is `axes=None`

https://github.com/pandas-dev/pandas/blob/d831326e11971af61307e4fc940855556ef3cdb7/pandas/compat/numpy/function.py#L317-L320

For Series and Index, it can be passed positionally or by name

For DataFrame, it can only be passed positionally

pandas-stubs/core/arrays/sparse/array.pyi
```diff
@@ -53,7 +56,7 @@ class SparseArray(ExtensionArray, ExtensionOpsMixin):
     def sum(self, axis: int = ..., *args, **kwargs): ...
     def cumsum(self, axis: int = ..., *args, **kwargs): ...
     def mean(self, axis: int = ..., *args, **kwargs): ...
-    def transpose(self, *axes): ...
+    def transpose(self, *axes: Any) -> Self: ...
```
MarcoGorelli, line 59: `axes` is totally ignored in `SparseArray.transpose`

Dr-Irv, line 59: We should delete this.  `SparseArray.transpose()` is not documented.  In fact, only the constructor is documented, so we should delete all the methods for `SparseArray` except `__init__()`.  Can you do that here?

Dr-Irv (changes requested):
I think we need to delete `transpose()` from everywhere except in `pandas/core/frame.pyi` because only `DataFrame.transpose()` is documented.  And even for that method, the `axes` argument isn't documented, so we should leave it out of the pandas stubs.

pandas-stubs/core/frame.pyi
```diff
@@ -656,7 +656,7 @@ class DataFrame(NDFrame, OpsMixin, _GetItemHack):
         show_counts: bool | None = ...,
     ) -> None: ...
     def memory_usage(self, index: _bool = ..., deep: _bool = ...) -> Series: ...
-    def transpose(self, *args, copy: _bool = ...) -> Self: ...
+    def transpose(self, *args: Any, copy: _bool = ...) -> Self: ...
```
MarcoGorelli, line 659: `args` are ignored, ok to use `Any` here?

Dr-Irv: yes

Dr-Irv (approved):
thanks @MarcoGorelli

PR #1144, opened 2025-03-02
GH1052 python 3.13 support

### LINKED ISSUE #1052: Python 3.13 support plan
Is there any roadmap for Python 3.13 support?
I ran some quick experiments on my side and it would require some package updates (python-calamine, tables, scipy).
It seems like building for 3.13 on Windows is causing issue, open to feedback.
https://github.com/loicdiridollou/pandas-stubs/actions/runs/11981847576/job/33408829249

It is failing to install `tables` on Windows due to missing a hdf5 install:
```

  .. ERROR:: Could not find a local HDF5 installation headers (hdrdir=None) and/or libdir (libdir=False). 

     You may need to explicitly state where your local HDF5 headers and

     library can be found by setting the ``HDF5_DIR`` environment

     variable or by using the ``--hdf5`` command-line option.
```

loicdiridollou:
- [x] Closes #1052 

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -33,7 +33,6 @@ packages = [{ "include" = "pandas-stubs" }]
 python = ">=3.10"
 types-pytz = ">= 2022.1.1"
 numpy = ">= 1.23.5"
-h5py = {version = "^3.13.0", platform = "win64"}
 
 [tool.poetry.group.dev.dependencies]
 mypy = "1.15.0"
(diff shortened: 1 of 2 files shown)


loicdiridollou:
Unsure why it fails on windows-latest, will dig into it. Otherwise looks clean on other setups so encouraging!

pyproject.toml
```diff
@@ -48,7 +49,8 @@ black = ">=23.3.0"
 isort = ">=5.12.0"
 openpyxl = ">=3.0.10"
 # for tables, MacOS gives random CI failures on 3.9.2
-tables = { version = "==3.10.1", python = "<4" }   # 3.8.0 depends on blosc2 which caps python to <4
+# for tables, Windows build fails on 3.10.1
+tables = { version = ">=3.10.1", python = "<4" }  # 3.8.0 depends on blosc2 which caps python to <4
```
Dr-Irv, line 53: remove the comment and make it `>= 3.10.2` since that is what worked for Windows with 3.13 anyway and 3.10.2 is needed with python 3.13.

Dr-Irv (changes requested):

Issue #1133, opened 2025-02-28
Should `Series[Any]` be used internally instead of `Series`?


MarcoGorelli:
Currently, `Series` is used in several places where the inner type of the Series isn't known, e.g.:

https://github.com/pandas-dev/pandas-stubs/blob/25fe8aae43dd8b4f018303376456f2a434fea15f/pandas-stubs/core/series.pyi#L864-L871

There's a couple of issues I'm running into with this

First, the pyright-strict job marks this as partially unknown:

```
/home/runner/work/pandas-stubs/pandas-stubs/tests/test_series.py:1039:5 - error: Type of "compare" is partially unknown
Type of "compare" is "Overload[(other: Series[Unknown], align_axis: Literal['index', 0], keep_shape: bool = ..., keep_equal: bool = ...) -> Series[Unknown], (other: Series[Unknown], align_axis: Literal['columns', 1] = ..., keep_shape: bool = ..., keep_equal: bool = ...) -> DataFrame]" (reportUnknownMemberType)
```

Second, when using `pyright` with `--verifytypes` to look for uncovered parts of the public API, this is flagged as "unknown type":

```json
            {
                "category": "function",
                "name": "pandas.testing.assert_series_equal",
                "referenceCount": 1,
                "isExported": true,
                "isTypeKnown": false,
                "isTypeAmbiguous": false,
                "diagnostics": [
                    {
                        "file": "/home/marcogorelli/type_coverage_py/.pyright_env_pandas/lib/python3.12/site-packages/pandas/_testing/__init__.pyi",
                        "severity": "error",
                        "message": "Type of parameter \"left\" is partially unknown\n  Parameter type is \"Series[Unknown]\"\n    Type argument 1 for class \"Series\" has unknown type",
                        "range": {
                            "start": {
                                "line": 4,
                                "character": 27
                            },
                            "end": {
                                "line": 4,
                                "character": 46
                            }
                        }
                    },
```

Would it be OK to use `Series[Any]` instead of just `Series` in such cases? Or, as some libraries do, to introduce a type alias `Incomplete: TypeAlias = Any` to mean "we should be able to narrow down the type but for now we're not doing so" and use that in some cases

The latter use-case (`--verifytypes`) can, I think, really help to prioritise which stubs to add

Dr-Irv:
We've had discussion about this in another PR.  See https://github.com/pandas-dev/pandas-stubs/pull/1093#issuecomment-2640276735

[...]

PR #1146, opened 2025-03-06
Introduce UnknownSeries and UnknownIndex, type `core.strings.pyi` using them

MarcoGorelli:
One step towards https://github.com/pandas-dev/pandas-stubs/issues/1133

I think one way to address this issue could be to do it incrementally - when you type a module strictly, add that to the `pyproject.toml` so that it stays strictly typed. Then gradually the partially unknown types will go away

🤔 this isn't quite working, trying to fix it up


pandas-stubs/_typing.pyi
```diff
@@ -576,6 +576,9 @@ S2 = TypeVar(
     | list[str],
 )
 
+UnknownSeries: TypeAlias = Series[Any]
+UnknownIndex: TypeAlias = Index[Any]
```
Dr-Irv, line 580: I would suggest making `UnknownSeries` and `UnknownIndex` classes, e.g.,
```python
class UnknownSeries(Series[Any]):
    ...
```

pyproject.toml
```diff
@@ -201,6 +201,12 @@ show_error_context = false
 show_column_numbers = false
 show_error_codes = true
 
+[[tool.mypy.overrides]]
```
Dr-Irv, line 204: shouldn't this be for `pyright`, not `mypy` ??

pandas-stubs/_typing.pyi
MarcoGorelli: sure - just for my edification, what's the difference/advantage?

Dr-Irv: I think that the type checker will treat it as a class.  When you create the `TypeAlias`, it is still going to see the type as partially unknown.  But the class is an actual type that is unknown.

pandas-stubs/_typing.pyi
```diff
@@ -576,6 +576,9 @@ S2 = TypeVar(
     | list[str],
 )
 
+class UnknownSeries(Series[Any]): ...
+class UnknownIndex(Index[Any]): ...
```
Dr-Irv, line 580: I think these should move to `core/series.pyi` and `core/indexes/base.pyi`, respectively

pandas-stubs/core/strings.pyi
```diff
@@ -58,7 +63,7 @@ class StringMethods(NoNewAttributesMixin, Generic[T, _TS, _TM, _TS2]):
     @overload
     def cat(
         self,
-        others: Series | pd.Index | pd.DataFrame | np.ndarray | list[Any],
+        others: UnknownIndex | pd.DataFrame | npt.NDArray[Any] | list[Any],
```
Dr-Irv, line 66: Shouldn't `UnknownSeries` also be in here?

tests/test_series.py
```diff
@@ -86,6 +87,7 @@
         UIntDtypeArg,
         VoidDtypeArg,
     )
+    from pandas._typing import UnknownSeries  # noqa: F401
```
Dr-Irv, line 90: I think you can move this up to the previous import in lines 63-73

Dr-Irv (changes requested):

tests/test_series.py
MarcoGorelli: I think that, but then `ruff` removes it as it's unused (as it's only used in `check`, which `ruff` doesn't recognise as usage? 

I think it's similar to

https://github.com/pandas-dev/pandas-stubs/blob/fd6188aec2e823679bbdb9416c88ce9e3b9f988a/tests/test_series.py#L91

pandas-stubs/core/strings.pyi
MarcoGorelli: yup, thanks!

pandas-stubs/core/strings.pyi
```diff
@@ -160,7 +167,7 @@ class StringMethods(NoNewAttributesMixin, Generic[T, _TS, _TM, _TS2]):
     def count(self, pat: str, flags: int = ...) -> Series[int]: ...
     def startswith(self, pat: str | tuple[str, ...], na: Any = ...) -> Series[bool]: ...
     def endswith(self, pat: str | tuple[str, ...], na: Any = ...) -> Series[bool]: ...
-    def findall(self, pat: str, flags: int = ...) -> Series: ...
+    def findall(self, pat: str, flags: int = ...) -> UnknownSeries: ...
```
MarcoGorelli, line 170: this can be narrowed down actually, it should be `Series[list[str]]`, looks like that pattern is used elsewhere (e.g. `rsplit`)

pandas-stubs/_typing.pyi
Dr-Irv: You have two options then.  

1. Allow `UnknownSeries` and `Series[str]` (and whatever other bound versions of `Series[S1]` are needed) in the argument list
2. Allow `UnknownSeries` and `Series[S1]` which would allow all known and unknown series.

If the `TypeAlias` solves the strict problem, that's great, but I'd be surprised if it does.

tests/test_series.py
Dr-Irv: If you do
```python
if TYPE_CHECKING:
    from pandas.core.series import (
        OffsetSeries,
        TimedeltaSeries,
        TimestampSeries,
        UnknownSeries,
    )
else:
    TimedeltaSeries: TypeAlias = pd.Series
    TimestampSeries: TypeAlias = pd.Series
    OffsetSeries: TypeAlias = pd.Series
    UnknownSeries: TypeAlias = pd.Series
```
Then take away the quotes from `UnknownSeries` in the test, ruff won't remove it.

pandas-stubs/core/strings.pyi
Dr-Irv: OK - go for it

pandas-stubs/_typing.pyi
MarcoGorelli: Sure, just tried it and it looks like it works

Simple reproducer:

with `foo/__init__.py` as
```python


def func(a: list) -> None:
    return None

__all__ = ['func']
```
I get (`pyright --verifytypes foo`)
```
Symbols used in public interface:
foo.func
  /home/marcogorelli/tmp-repo/foo/__init__.py:4:5 - error: Type of parameter "a" is partially unknown
    Parameter type is "list[Unknown]"
      Type argument 1 for class "list" has unknown type
```

But if I instead have
```python

from typing import TypeAlias, Any

UnknownList: TypeAlias = list[Any]

def func(a: UnknownList) -> None:
    return None

__all__ = ['func']
```
then I get
```
Symbols exported by "foo": 2
  With known type: 2
  With ambiguous type: 0
  With unknown type: 0
```

Dr-Irv: Great.  That's good to know.

tests/test_series.py
```diff
@@ -1665,8 +1669,17 @@ def test_series_overloads_cat():
...
+        ),
+        UnknownSeries,
+    )
+    check(
+        assert_type(
+            s.str.cat(pd.Series(["A", "B", "C", "D", "E", "F", "G"]), sep=";"),
+            UnknownSeries,
+        ),
+        UnknownSeries,
     )
```
MarcoGorelli, line 1683: the return type of `cat` is the TypeVar `T`, so I think the intention might have been to preserve the input Series type?

anyway, i think that's out of scope for this PR, but I've opened https://github.com/pandas-dev/pandas-stubs/issues/1149to track it

pandas-stubs/core/strings.pyi
```diff
@@ -58,7 +62,9 @@ class StringMethods(NoNewAttributesMixin, Generic[T, _TS, _TM, _TS2]):
     @overload
     def cat(
         self,
-        others: Series | pd.Index | pd.DataFrame | np.ndarray | list[Any],
+        others: (
+            Series[str] | Index[str] | pd.DataFrame | npt.NDArray[np.str_] | list[str]
```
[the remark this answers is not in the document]
MarcoGorelli: It'll still be supported to pass that, because `Any` turns off the type checker for the inner type

minimal reproducer: https://mypy-play.net/?mypy=latest&python=3.12&gist=2186fa9a76052e0309b9f8af68e75172

Dr-Irv (changes requested):
I couldn't comment on the code changes suggested below, but I'd like to suggest the following:

1. Change the references to `StringMethods` in `core/series.pyi` and `core/indexes/base.pyi` to make `Series[str]` and `Index[str]` the first argument.  Then in `core/strings.pyi` the first argument of the `Generic` called `T` will get bound to that type.
2. Update the tests for the string methods in `test_series.py` and `test_indexes.py` to test for the return type of `Series[str]` and `Index[str]` as appropriate.
3. For `test_indexes.py`, we could use a set of tests on the string methods similar to the ones in `test_series.py`

If you think we should do this in a separate PR, I'm OK with that as well.

pandas-stubs/core/strings.pyi
```diff
@@ -172,15 +176,27 @@ class StringMethods(NoNewAttributesMixin, Generic[T, _TS, _TM, _TS2]):
...
-    def extract(self, pat: str, flags: int = ..., *, expand: Literal[False]) -> T: ...
+    def extract(
+        self, pat: str, flags: int = ..., *, expand: Literal[False]
+    ) -> Series[type[object]]: ...
     def extractall(self, pat: str, flags: int = ...) -> pd.DataFrame: ...
-    def find(self, sub: str, start: int = ..., end: int | None = ...) -> T: ...
-    def rfind(self, sub: str, start: int = ..., end: int | None = ...) -> T: ...
+    def find(
+        self, sub: str, start: int = ..., end: int | None = ...
+    ) -> Series[int]: ...
```
Dr-Irv, line 189: This won't work for `Index`:
```python
>>> ind = pd.Index(["abc", "cba", "def"])
>>> ind.str.find("a")
Index([0, 2, -1], dtype='int64')
```
So I think for `StringMethods`, you have to add an extra argument to the `Generic` parameter that would be `Series[int]` or `Index[int]` based on the return types of the methods that return the sequences of `int`

Similar to how it is done for `expand`

tests/test_string_accessors.py
```diff
@@ -0,0 +1,398 @@
...
+    check(
+        assert_type(
+            s.str.cat(pd.Series(["A", "B", "C", "D", "E", "F", "G"]), sep=";"),
+            "pd.Series[str]",
+        ),
+        pd.Series,
+        str,
+    )
+    unknown_s = pd.DataFrame({"a": list("abcdefg")})["a"]
+    check(assert_type(s.str.cat(unknown_s, sep=";"), "pd.Series[str]"), pd.Series, str)
```
Dr-Irv, line 327: I think you should also test things like:
```python
check(assert_type(unknown_s.str.cat(s, sep=";"), "pd.Series[str]"), pd.Series, str)
check(assert_type(unknown_s.str.cat(unknown_s, sep=";"), "pd.Series[str]"), pd.Series, str)
```
Ideally, we should do this for all of the methods to make sure that when an `UnknownSeries` or `UnknownIndex` is passed, then we get an `pd.Series[str]` and `Index[str]` respectively.

pandas-stubs/core/series.pyi
```diff
@@ -1156,7 +1156,16 @@ class Series(IndexOpsMixin[S1], NDFrame):
...
+    ) -> StringMethods[
+        Self,
+        DataFrame,
+        Series[bool],
+        Series[list[str]],
+        Series[int],
+        Series[bytes],
+        Series[str],
+        Series[type[object]],
+    ]: ...
```
Dr-Irv, line 1168: Might be worth doing this (and similar for `Index`):
```python
@overload
@property
 def str(
        self: Series[str]
    ) -> StringMethods[
        Self,
        DataFrame,
        Series[bool],
        Series[list[str]],
        Series[int],
        Series[bytes],
        Series[str],
        Series[type[object]],
    ]: ...
@overload
@property
 def str(
        self: Series[S1]
    ) -> Never: ...
```
I *think* this will cause both `Series[str]` and `UnknownSeries` (and `Series[Any]`) to match the first overload, but the second overload should be flagged as incorrect, if, for example, you had
```python
s = pd.Series([1,2,3])
s.str.startswith("abc")
```
which fails at runtime, but we can detect with typing,  and we can test for this (see tests that use `TYPE_CHECKING_INVALID_USAGE`) .

The overloads will overlap, but we can put in a `# type: ignore` and `#pyright: ignore` to handle that.

Dr-Irv (changes requested):
Thanks for doing all of this work.  It is a nice improvement to the stubs. 

I think that all the methods that have `->T` should be `-> _TSTR`

Because we know these are string methods, so even if the type of the `Series` (or `Index`) is unknown, we know we will be returning `Series[str]` or `Index[str]`

pandas-stubs/core/series.pyi
MarcoGorelli: this doesn't work for me, I get
```
  /home/marcogorelli/pandas-stubs-dev/tests/test_string_accessors.py:308:23 - error: "assert_type" mismatch: expected "str" but received "Never" (reportAssertTypeFailure)
  /home/marcogorelli/pandas-stubs-dev/tests/test_string_accessors.py:309:23 - error: "assert_type" mismatch: expected "str" but received "Never" (reportAssertTypeFailure)
  /home/marcogorelli/pandas-stubs-dev/tests/test_string_accessors.py:312:13 - error: "assert_type" mismatch: expected "Series[str]" but received "Never" (reportAssertTypeFailure)
  /home/marcogorelli/pandas-stubs-dev/tests/test_string_accessors.py:320:13 - error: "assert_type" mismatch: expected "Series[str]" but received "Never" (reportAssertTypeFailure)
  /home/marcogorelli/pandas-stubs-dev/tests/test_string_accessors.py:327:23 - error: "assert_type" mismatch: expected "Series[str]" but received "Never" (reportAssertTypeFailure)
  /home/marcogorelli/pandas-stubs-dev/tests/test_string_accessors.py:358:23 - error: "assert_type" mismatch: expected "DataFrame" but received "Never" (reportAssertTypeFailure)
  /home/marcogorelli/pandas-stubs-dev/tests/test_string_accessors.py:360:21 - error: "assert_type" mismatch: expected "DataFrame" but received "Never" (reportAssertTypeFailure)
  /home/marcogorelli/pandas-stubs-dev/tests/test_string_accessors.py:364:13 - error: "assert_type" mismatch: expected "Series[type[object]]" but received "Never" (reportAssertTypeFailure)
  /home/marcogorelli/pandas-stubs-dev/tests/test_string_accessors.py:371:13 - error: "assert_type" mismatch: expected "Series[type[object]]" but received "Never" (reportAssertTypeFailure)
```

MarcoGorelli:
thanks, have updated

> I think that all the methods that have ->T should be -> _TSTR

I think the only exception is `str.slice`, which preserves the type. but I've gone ahead and done this for others 👍

pandas-stubs/core/series.pyi
Dr-Irv: I got something to work for `Series` .  See https://github.com/Dr-Irv/pandas-stubs/tree/pr1146 and this commit: https://github.com/pandas-dev/pandas-stubs/commit/164bc27185bf213fd7da942414bf2e2935bf8a1d

Had to find something I wrote about in `mypy` a while back:  https://github.com/python/mypy/issues/9937

So you can do something similar for `Index`.  And ideally add some tests for some of the other string methods when the `Series` or `Index` has an unknown subtype.

MarcoGorelli:
Regarding https://github.com/pandas-dev/pandas-stubs/pull/1146#discussion_r1987936530, is it OK if we leave that to a separate PR please?

Partially because I feel like the scope here keeps increasing, and partially because I'm not sure it's correct - for example, if I have
```python
import pandas as pd
from typing import Any

def func(a: pd.Series[Any]) -> None:
    reveal_type(a.str.upper())
```
then with that commit, we get `Revealed type is "Any"`, whereas without it we get `Revealed type is "pandas.core.series.Series[builtins.str]"`

Dr-Irv:
> Regarding [#1146 (comment)](https://github.com/pandas-dev/pandas-stubs/pull/1146#discussion_r1987936530), is it OK if we leave that to a separate PR please?
> 
> Partially because I feel like the scope here keeps increasing, and partially because I'm not sure it's correct - for example, if I have
> 
> ```python
> import pandas as pd
> from typing import Any
> 
> def func(a: pd.Series[Any]) -> None:
>     reveal_type(a.str.upper())
> ```
> 
> then with that commit, we get `Revealed type is "Any"`, whereas without it we get `Revealed type is "pandas.core.series.Series[builtins.str]"`

OK.  It's a `mypy` bug.  Ugh.

tests/test_string_accessors.py
```diff
@@ -0,0 +1,417 @@
...
+    _check(assert_type(s.str.title(), "pd.Series[str]"))
+    _check(
+        assert_type(s.str.translate({241: "n"}), "pd.Series[str]"),
+    )
+    _check(assert_type(s.str.upper(), "pd.Series[str]"))
+    _check(assert_type(s.str.wrap(80), "pd.Series[str]"))
+    _check(assert_type(s.str.zfill(10), "pd.Series[str]"))
+    s_bytes = pd.Series([b"a1", b"b2", b"c3"])
+    _check(assert_type(s_bytes.str.decode("utf-8"), "pd.Series[str]"))
+    s_list = pd.Series([["apple", "banana"], ["cherry", "date"], [1, "eggplant"]])
```
Dr-Irv, line 142:
```suggestion
    s_list = pd.Series([["apple", "banana"], ["cherry", "date"], ["pear", "eggplant"]])
```
I think we want to test that `s_list` is of type `Series[list[str]]`, but with the way it is written, it will get typed as `Series[Any]`

tests/test_string_accessors.py
```diff
@@ -0,0 +1,417 @@
...
+    _check(
+        assert_type(idx.str.translate({241: "n"}), "pd.Index[str]"),
+    )
+    _check(assert_type(idx.str.upper(), "pd.Index[str]"))
+    _check(assert_type(idx.str.wrap(80), "pd.Index[str]"))
+    _check(assert_type(idx.str.zfill(10), "pd.Index[str]"))
+    idx_bytes = pd.Index([b"a1", b"b2", b"c3"])
+    _check(assert_type(idx_bytes.str.decode("utf-8"), "pd.Index[str]"))
+    idx_list: "pd.Index[list]" = pd.Index(
+        [["apple", "banana"], ["cherry", "date"], [1, "eggplant"]]
```
Dr-Irv, line 178: same comment here

Dr-Irv (changes requested):
just a couple of tests to change, otherwise OK

Dr-Irv (approved):
Thanks @MarcoGorelli .  Long journey, but a really nice improvement to the stubs!

MarcoGorelli:
thanks for your careful review, much appreciated!

PR #1151, opened 2025-03-07
make tslibs strptime, timedeltas, and timestamps pass with pyright-strict

MarcoGorelli:
towards #1133


MarcoGorelli:
thanks for your review!

> Then df.index will be unknown - we can't track the type, but you might pass it to some of those methods where we used the Index type

Yup - in that case, the inner type will be ignored. I've added a test for this anyway 👍

MarcoGorelli:
🤔 not sure what to make of the CI failure - we get `  /home/runner/work/pandas-stubs/pandas-stubs/pandas-stubs/_libs/tslibs/timestamps.pyi:19:21 - error: Type of "TimeZones" is unknown (reportUnknownVariableType)` but only on 3.11

tests/test_scalars.py
```diff
@@ -1273,6 +1283,8 @@ def test_timestamp_cmp() -> None:
 
     check(assert_type(ts > c_datetimeindex, np_ndarray_bool), np.ndarray, np.bool_)
     check(assert_type(ts <= c_datetimeindex, np_ndarray_bool), np.ndarray, np.bool_)
+    check(assert_type(ts > c_unknown_index, np_ndarray_bool), np.ndarray, np.bool_)
+    check(assert_type(ts <= c_unknown_index, np_ndarray_bool), np.ndarray, np.bool_)
```
Dr-Irv, line 1287: Interestingly, these are getting handled by the relational operators in `Index`.

Would be more complete to include them all, i.e., add `<`, `>=`, `==` and `!=`

MarcoGorelli: > these are getting handled by the relational operators in Index.

Ah, interesting! I hadn't realised that if an overload couldn't be matched, then the opposite operation from the right-hand-side would be used

updating now, thanks!

Dr-Irv (commented):
So in the other PR for `StringMethods`, I changed `DataFrame.__getitem__()` to return `UnknownSeries`, which allows matching it in an overload.

So here, I think if you change `DataFrame.index` to be `UnknownIndex` and use `UnknownIndex` instead of `Index[Any]` things would work.  Just a hunch.

MarcoGorelli:
ok thanks, i'll resume this when `UnknownIndex` and `UnknownSeries` are available

Dr-Irv:
> which I find bizarre because locally it passes for me even on 3.11 🤔

And it worked fine on Windows in CI.

I got it to work by taking the import of `TimeZones` out of `timestamps.pyi` and copying it from `_typing.pyi` into `timestamps.pyi` and then it got past the CI.  Not ideal.

So I am guessing that because `timestamps.pyi` is set up to be strict, but `_typing.pyi` i not, that is why it got confused.

But why that only happens on 3.11 on Mac and Ubuntu is really odd.

pandas-stubs/_libs/tslibs/strptime.pyi
```diff
@@ -1,3 +1,4 @@
+# pyright
```
Dr-Irv, line 1: don't think we need this

pandas-stubs/_libs/tslibs/timestamps.pyi
```diff
@@ -49,6 +49,9 @@ _Ambiguous: TypeAlias = bool | Literal["raise", "NaT"]
 _Nonexistent: TypeAlias = (
     Literal["raise", "NaT", "shift_backward", "shift_forward"] | Timedelta | timedelta
 )
+# Repeated from `_typing.pyi` so as to satisfy mixed strict / non-strict paths.
+# https://github.com/pandas-dev/pandas-stubs/pull/1151#issuecomment-2715130190
+TimeZones: TypeAlias = str | _tzinfo | None | int
```
Dr-Irv, line 54: Can you add a comment in `_typing.pyi` that says that `TimeZones` is duplicated here, in case we ever change it there, we know to change it here as well.

tests/test_scalars.py
```diff
@@ -1253,6 +1261,8 @@ def test_timestamp_cmp() -> None:
     c_np_dt64 = np.datetime64(1, "ns")
     c_dt_datetime = dt.datetime(year=2000, month=1, day=1)
     c_datetimeindex = pd.DatetimeIndex(["2000-1-1"])
+    # DatetimeIndex, but the type checker detects it to be Index[Unknown].
```
Dr-Irv, line 1264:
```suggestion
    # DatetimeIndex, but the type checker detects it to be UnknownIndex
```

tests/test_scalars.py
```diff
@@ -1358,6 +1372,13 @@ def test_timestamp_cmp() -> None:
         assert_type(ts != c_datetimeindex, np_ndarray_bool), np.ndarray, np.bool_
     )
     assert (eq_arr != ne_arr).all()
+    eq_arr = check(
+        assert_type(ts == c_unknown_index, np_ndarray_bool), np.ndarray, np.bool_
+    )
+    ne_arr = check(
+        assert_type(ts != c_unknown_index, np_ndarray_bool), np.ndarray, np.bool_
+    )
```
Dr-Irv, line 1380: can you add checks for the reverse operations, e.g., `c_unknown_index == ts` ?

tests/test_scalars.py
```diff
@@ -1313,6 +1325,8 @@ def test_timestamp_cmp() -> None:
 
     check(assert_type(ts >= c_datetimeindex, np_ndarray_bool), np.ndarray, np.bool_)
     check(assert_type(ts < c_datetimeindex, np_ndarray_bool), np.ndarray, np.bool_)
+    check(assert_type(ts >= c_unknown_index, np_ndarray_bool), np.ndarray, np.bool_)
+    check(assert_type(ts < c_unknown_index, np_ndarray_bool), np.ndarray, np.bool_)
```
Dr-Irv, line 1329: can you add checks for the reverse operations, i.e., with `c_unknown_index` on the LHS?

Dr-Irv (changes requested):
looks like we are getting closer.  I suggested a few extra tests and some other cleanup.

Issue #1166, opened 2025-03-13
SqlAlchemy update to 2.0.39 breaks read_sql


loicdiridollou:
**Describe the bug**

```
Poe => mypy

===========================================
Beginning: 'Run mypy on 'tests' (using the local stubs) and on the local stubs'
===========================================

tests/test_io.py:1419: error: Argument 1 to "read_sql" has incompatible type "Select[Row[tuple[int]]] | FromStatement[Row[tuple[int]]] | UpdateBase"; expected "str | Selectable | TextClause | Select[Any] | FromStatement[Any]"  [arg-type]
Found 1 error in 1 file (checked 226 source files)
```

Running the test suite will raise an error with the latest update of SqlAlchemy to 2.0.39.
For now I have restricted the version of the package to stop at 2.0.38, the typehint needs to be fixed.

**Please complete the following information:**
 - OS: all
 - OS Version [e.g. 22]
 - python version 3.10-3.12
 - version of type checker latest
 - version of installed `pandas-stubs` latest

**Additional context**

PR #1167, opened 2025-03-13
issue 1166 Update test to use SQLAlchemy 2.0 compatible select statement

shirzady1934:
This PR resolves https://github.com/pandas-dev/pandas-stubs/issues/1166 by update test to use SQLAlchemy 2.0 compatible select statement


tests/test_io.py
```diff
@@ -1415,9 +1415,8 @@ class Temp(Base):
 
             Session = sqlalchemy.orm.sessionmaker(engine)
             with Session() as session:
-                pd.read_sql(
-                    session.query(Temp.quantity).statement, session.connection()
-                )
+                stmt = sqlalchemy.select(Temp.quantity)
```
loicdiridollou, line 1418: Here you are modifying the type of `stmt` so you are not really testing what used to be tested. What is needed is to update the stubs so that this test passes.
Without changing this test, what needs to be done is to update a pyi file.
For that I would recommend you look into what is the new type of stmt with the update of sqlalchemy, you will see that stmt has now an extra possible type `UpdateBase` and what how that impacts the stubs in `pandas-stubs/io/sql.pyi`. There you see that the Type Alias `_SQLStatement` is too restrictive and you need to allow UpdateBase in that TypeAlias too.
Feel free to let me know if something is unclear.

loicdiridollou (changes requested):
One small change to make sure we are testing the right thing.

shirzady1934:
@loicdiridollou  I reverted my previous commits and added extra possible type UpdateBase to _SQLStatement, is that ok?

pandas-stubs/io/sql.pyi
```diff
@@ -33,6 +34,7 @@ _SQLStatement: TypeAlias = (
     | sqlalchemy.sql.expression.TextClause
     | sqlalchemy.sql.Select
     | FromStatement
+    | UpdateBase
```
Dr-Irv, line 37: Instead of importing `UpdateBase`, use the pattern as in lines 33 and 34, and place `sqlalchemy.sql.expression.UpdateBase` here

Dr-Irv (changes requested):

Dr-Irv (changes requested):
while you were working on this, we put a constraint in `pyproject.toml` on `sqlalchemy` so that our CI would pass. 

Can you change this line in that file:
```text
SQLAlchemy = ">=2.0.12,<2.0.39"
```
to
```text
SQLAlchemy = ">=2.0.39"
```

pyproject.toml
```diff
@@ -59,7 +59,7 @@ xarray = ">=22.6.0"
 tabulate = ">=0.8.10"
 jinja2 = ">=3.1"
 scipy = { version = ">=1.9.1", python = "<3.14" }
-SQLAlchemy = ">=2.0.12,<2.0.39"
+SQLAlchemy = ">2.0.39"
```
Dr-Irv, line 62:
```suggestion
SQLAlchemy = ">=2.0.39"
```

you missed the `=` sign.

Dr-Irv (changes requested):

Dr-Irv (approved):
thanks @shirzady1934

Issue #1169, opened 2025-03-13
type `kwargs` in `pct_change` according to params in `shift`


MarcoGorelli:
Opening this so I can point to it as a todo. See https://github.com/pandas-dev/pandas-stubs/issues/1168 for context

PR #1170, opened 2025-03-13
type args and kwargs as Any in series.pyi

### LINKED ISSUE #1168: question: how to type compatibility-only `args` / `kwargs`?
I'm trying to use `pyright --verifytypes` to check which public methods are untyped, and a lot of the results come from `kwargs` and `args`.

There seem to be 3 cases for them:
1. they're undocumented, and just silently exist to allow for numpy-compatible code. For example: `Series.std`
2. they get passed down to a third-party function. For example: most plotting funtions
3. when they get passed to other pandas functions which have known args/kwargs. For example, the `kwargs` in `Series.pct_change`.

How should `kwargs` and `args` get typed in these cases? I'm thinking:
1., 2.: use `Any`
3.: make a `Protocol` with only the args/kwargs which can be accepted

I think if we can get these typed (even if just with `Any`) then pyright's verifytypes should become a lot more useful to identify missing parts of the public API surface

repo where I'm developing this type-coverage-measurement functionality: https://github.com/MarcoGorelli/measuring-pandas-type-completeness (note: it's currently quite patchy, but I've already found it useful to decide what to focus on)

MarcoGorelli:
ref #1168

These are all cases where `kwargs` / `args` just exist for compatibility with numpy

Exceptions are:
- `to_numpy`, where the kwargs get passed to the extensionarray. Given that third parties can develop their own extensionarrays, I think this one really should be `Any`, we can't know what arguments third party EA authors will accept?
- `pct_change`: this could be more precise, have opened https://github.com/pandas-dev/pandas-stubs/issues/1169 about that

### DIFF
--- a/pandas-stubs/core/series.pyi
+++ b/pandas-stubs/core/series.pyi
@@ -456,12 +456,14 @@ class Series(IndexOpsMixin[S1], NDFrame):
     def ravel(self, order: _str = ...) -> np.ndarray: ...
     def __len__(self) -> int: ...
     def view(self, dtype=...) -> Series[S1]: ...
-    def __array_ufunc__(self, ufunc: Callable, method: _str, *inputs, **kwargs): ...
+    def __array_ufunc__(
+        self, ufunc: Callable, method: _str, *inputs: Any, **kwargs: Any
+    ): ...
     def __array__(self, dtype=...) -> np.ndarray: ...
     @property
     def axes(self) -> list: ...
     def take(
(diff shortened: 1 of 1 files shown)


Dr-Irv (approved):
Thanks @MarcoGorelli

Dr-Irv:
@MarcoGorelli before I merge, this will close #1168, but isn't that issue covering more than just `core/series.pyi` ?  If so, I think it should remain open.

MarcoGorelli:
yup thanks!

PR #1177, opened 2025-03-18
use UnknownSeries in series.pyi

MarcoGorelli:


Dr-Irv (changes requested):
I think we may want to add a test like this (maybe more tests - not sure):
```python
def test_unknown() -> None:
    s = pd.Series(pd.NA)

    def foo(sf: pd.Series):
        pass

    foo(s)

    check(assert_type(s + pd.Series([1]), pd.Series), pd.Series)

```
Idea is to make sure that an `UnknownSeries` can be passed and that arithmetic works.  I'm fine for now if you just add the above.

Eventually, we need to deal with this comment in a PR:  https://github.com/pandas-dev/pandas-stubs/pull/1093#issuecomment-2640276735  (and other comments in that PR).

But the incremental approach is fine.  Would be interesting to convert all references in the stubs from plain `Series` to `UnknownSeries`.

Dr-Irv (approved):
thanks @MarcoGorelli

PR #1188, opened 2025-04-12
GH1180 Clean up all/any methods for Series and DataFrame

### LINKED ISSUE #1180: Wrong type hint for Series.all() and Series.any()
The Series.all() and Series.any() have the type _bool for the output. However, the real output type is np.bool_. This causes issues when trying to compile a function using mypyc.

**To Reproduce**
Compile the following code
```
def test_function(df: pd.Dataframe):
# some code
is_all_nan = df["column_name"].isna().all()
# some more code
```
Typechecker mypy/mypyc
TypeError: bool object expected; got numpy.bool_

 - OS: Windows
 - OS Version: 11
 - python version 3.11.6=7
 - mypy version 1.8.0
 - pandas-stubs version 2.2.3.250308

loicdiridollou:
- [x] Closes #1180 

Technically we should fix the `pd.DataFrame.all` but that would mean having a `pd.Series[np.bool]`, yet `np.bool` is not a type of `S1` so will leave it there for a bit, it has some FIXME statements there for us to know.


pandas-stubs/core/frame.pyi
```diff
@@ -1660,7 +1660,8 @@ class DataFrame(NDFrame, OpsMixin, _GetItemHack):
         bool_only: _bool | None = ...,
         skipna: _bool = ...,
         **kwargs: Any,
-    ) -> _bool: ...
+    ) -> np.bool: ...
+    # FIXME the type below is not correct, should be pd.Series[np.bool]
```
Dr-Irv, line 1664: The way I look at it is that we are using `Series[bool]` to correspond to whatever `bool` is stored inside - being a `python` one, a `numpy` one, or even if we use `BooleanDtype`, so I don't think this comment is necessary.

Dr-Irv (changes requested):
I think you should remove the `FIXME` comments.  Otherwise OK

pandas-stubs/core/frame.pyi
loicdiridollou: Actually the type checker will not accept that, because pd.Series can only be subscribed with S1 which contains bool (the generic boolean from python) but not np.bool.
Yet pd.DataFrame.any will return at runtime pd.Series[np.bool] (but this is not accepted since np.bool is not a subtype of S1).
Let me know if this is not clear,
image

Dr-Irv: The type returned from `DataFrame.any()` and `DataFrame.all()` should be `Series[_bool]` .  Then the tests should use that.

loicdiridollou: My issue is that _bool does not contain `np.bool` which is the type we get at runtime. Happy to keep the stubs as is but that would mean that runtime type does not align with static type.
Or we can add np.bool to _bool, open to suggestion.

Dr-Irv: I don't think it matters.  We have checks like this:
```python
check(assert_type(df.any(), "pd.Series[bool]"), pd.Series, np.bool_)
```
So even though `np.bool_` is in the `Series`, we call the type of the `Series[bool]`

It's similar to this:
```python
    check(assert_type(df.value_counts(), "pd.Series[int]"), pd.Series, np.integer)
```
What's inside the series are numpy integers, but we call that `Series[int]`

loicdiridollou: Okay I see your point, let me fix the comments, thanks for the more detailed vision on this issue!

Dr-Irv (approved):
Thanks @loicdiridollou

Issue #1189, opened 2025-04-16
subtracing a TimeStampSeries from a Timestamp not supported


Dr-Irv:
**Describe the bug**
Type checking missing for Timestamp - TimestampSeries

**To Reproduce**

```python
import pandas as pd
ts1 = pd.to_datetime(pd.Series(["2022-03-05", "2022-03-06"]))
assert isinstance(ts1.iloc[0], pd.Timestamp)
one_ts = ts1.iloc[0]
reveal_type(one_ts - ts1)
```
revealed type should be `TimedeltaSeries`.

Note that `ts1 - one_ts` works fine.

Need to add an overload in `pandas-stubs/_libs/tslibs/timestamps.pyi`

**Please complete the following information:**
 - OS: Windows
 - OS Version 11
 - python version  3.10
 - version of type checker  pyright 1.386
 - version of installed `pandas-stubs`  - development copy

loicdiridollou:
#1191

PR #1191, opened 2025-04-19
GH1189/GH1181 Timestamp subtraction and index shift

### LINKED ISSUE #1181: pd.Index().shift() claims to return None
**Describe the bug**
The stubs claim that shift returns None, when it returns a pd.Index of the same type as the original_index

**To Reproduce**
```python
from datetime import datetime

import pandas as pd

original_index = pd.date_range(
    start=datetime(2024, 2, 2),
    end=datetime(2024, 4, 2),
    freq="15min",
    inclusive="left",
    tz="Europe/Berlin",
)

new_index = original_index.shift(1)
print(original_index, new_index)
print(f"{type(original_index)}, {type(new_index)}")
```

I get this **mypy** error:
> example.py:13: error: "shift" of "Index" does not return a value (it only ever returns None)  [func-returns-value]

- macOS 15.4 (24E248)
- Python 3.11.11
- I am using mypy 1.11.1 and pandas-stubs 2.2.3.250308.

loicdiridollou:
- [x] Closes #1181 and #1189


pandas-stubs/core/indexes/base.pyi
```diff
@@ -393,7 +393,7 @@ class Index(IndexOpsMixin[S1]):
     def asof_locs(self, where, mask): ...
     def sort_values(self, return_indexer: bool = ..., ascending: bool = ...): ...
     def sort(self, *args, **kwargs) -> None: ...
-    def shift(self, periods: int = ..., freq=...) -> None: ...
+    def shift(self, periods: int = ..., freq=...) -> Self: ...
```
loicdiridollou, line 396: One small question here, `pd.Index.shift` is technically not implemented so wondering if it should be `NoReturn` here and `Self` for `pd.DatetimeIndex`, `pd.TimedeltaIndex` and `pd.PeriodIndex` or if this is fine as is (technically it would fail at runtime anyway so the typing would not help much to spot the bug).

Dr-Irv: I think we should take `shift()` out of here and only include it in the stubs for those 3 types.

Dr-Irv (changes requested):

pandas-stubs/core/indexes/base.pyi
loicdiridollou: Agreed! Thanks for the direction

Dr-Irv (approved):
thanks @loicdiridollou

PR #1192, opened 2025-04-20
GH1182 Add numpy integer, floating, complex to Scalar

### LINKED ISSUE #1182: `na_value` in `to_numpy` does not accept NumPy number types
**Describe the bug**
Calling `to_numpy` with `na_value` set to a NumPy type gives a type error. This works at runtime.

**To Reproduce**

    ```python
    import numpy as np
    import pandas as pd
    
    data_array = pd.arrays.IntegerArray(
        np.array([1, 2, -1], dtype=np.int32),
        mask=np.array([0, 0, 1], dtype=np.bool),
    )
    data = data_array.to_numpy(
        dtype=np.int32,
        na_value=np.int32(42),
    )
    ```

    I am using MyPy.

    ```
    Argument "na_value" to "to_numpy" of "BaseMaskedArray" has incompatible type "signedinteger[_32Bit]"; expected "str | bytes | date | datetime | timedelta | <7 more items> | complex"
    ```

 - OS: Windows
 - OS Version: 11
 - python version: 3.11.11
 - version of type checker: 1.15.0
 - version of installed `pandas-stubs`: 2.2.3.250308

loicdiridollou:
- [x] Closes #1182 

### DIFF
--- a/pandas-stubs/_typing.pyi
+++ b/pandas-stubs/_typing.pyi
@@ -518,7 +518,9 @@ IndexIterScalar: TypeAlias = (
     | Timestamp
     | Timedelta
 )
-Scalar: TypeAlias = IndexIterScalar | complex
+Scalar: TypeAlias = (
+    IndexIterScalar | complex | np.integer | np.floating | np.complexfloating
+)
 ScalarT = TypeVar("ScalarT", bound=Scalar)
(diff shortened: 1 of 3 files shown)


Dr-Irv (approved):
Thanks @loicdiridollou

PR #1193, opened 2025-04-20
GH1173 Experiment with fuller typing

### LINKED ISSUE #1173: type `kwargs` in `DataFrame.query` according to `DataFrame.eval`

loicdiridollou:
- [x] Closes #1173
- [] Tests added: Please use `assert_type()` to assert the type of any return value

@MarcoGorelli I took a try at your idea of typehinting the whole set of arguments in `pd.DataFrame.query` but stumbling upon the issue when the user passes a dictionary (which is still an allowed behavior).
Wondering if there is something I am missing here, please let me know!


Dr-Irv:
> @MarcoGorelli I took a try at your idea of typehinting the whole set of arguments in `pd.DataFrame.query` but stumbling upon the issue when the user passes a dictionary (which is still an allowed behavior).
> Wondering if there is something I am missing here, please let me know!

You'd still need to have the `**kwargs` as an overload.  If you want to allow the other arguments, you can have that (although I don't think we should do that because it's not documented), but you'd still need the `**kwargs` overload

pandas-stubs/core/frame.pyi
```diff
@@ -676,6 +676,13 @@ class DataFrame(NDFrame, OpsMixin, _GetItemHack):
         self,
         expr: _str,
         *,
+        parser: Literal["pandas", "python"] = ...,
+        engine: Literal["python", "numexpr"] | None = ...,
+        local_dict: dict[_str, Any] | None = ...,
+        global_dict: dict[_str, Any] | None = ...,
+        resolvers: list[Mapping] | None = ...,
+        level: int = ...,
+        target: object | None = ...,
```
Dr-Irv, line 685: Can you indicate where you found this list of possible keyword arguments?

loicdiridollou: I sourced them by tracing the execution to:
https://github.com/pandas-dev/pandas/blob/337d40e5d55f7787e48f029486f47fd5a053bc80/pandas/core/computation/eval.py#L177-L187

loicdiridollou:
I was able to make some progress here, let me know if this is what you envisioned and I will clean it up.

MarcoGorelli:
Hey - yeah, this is what I was thinking - I don't really understand why `**kwargs` would be needed at all, given that the accepted arguments by `eval` are limited https://pandas.pydata.org/docs/reference/api/pandas.eval.html#pandas.eval , although based on

>  If you want to allow the other arguments, you can have that (although I don't think we should do that because it's not documented)

it looks like Irv disagrees and that what I've suggested may be against the pandas-stubs philosophy - which is fair enough

Dr-Irv:
> Hey - yeah, this is what I was thinking - I don't really understand why `**kwargs` would be needed at all, given that the accepted arguments by `eval` are limited https://pandas.pydata.org/docs/reference/api/pandas.eval.html#pandas.eval , although based on
> 
> > If you want to allow the other arguments, you can have that (although I don't think we should do that because it's not documented)
> 
> it looks like Irv disagrees and that what I've suggested may be against the pandas-stubs philosophy - which is fair enough

So I looked more carefully at the pandas docs, which do say that the accepted `**kwargs` are from `eval()`.   Not the greatest on the docs front, and that could be improved.  Maybe one of you could create an issue there??

So I'll look more carefully at this PR now with that in mind.

pandas-stubs/core/frame.pyi
```diff
@@ -698,16 +698,44 @@ class DataFrame(NDFrame, OpsMixin, _GetItemHack):
...
-        **kwargs: Any,  # TODO: make more precise https://github.com/pandas-dev/pandas-stubs/issues/1173
+    ) -> None: ...
+    @overload
+    def query(
+        self,
+        expr: _str,
+        *,
+        inplace: Literal[True],
+        **kwargs: Any,
     ) -> None: ...
```
Dr-Irv, line 717: Not clear why you need this overload when the previous overload covers it?

pandas-stubs/core/frame.pyi
```diff
@@ -698,16 +698,44 @@ class DataFrame(NDFrame, OpsMixin, _GetItemHack):
...
+        target: object | None = ...,
+    ) -> Self: ...
+    @overload
+    def query(
+        self,
+        expr: _str,
+        *,
+        inplace: Literal[False] = ...,
+        **kwargs: Any,
     ) -> Self: ...
```
Dr-Irv, line 739: same comment - not sure why we need this overload when the previous one covers it.

tests/test_frame.py
```diff
@@ -514,6 +514,29 @@ def test_types_query() -> None:
...
+            df.query("col1 > col2", parser="pandas", engine="numexpr"), pd.DataFrame
+        ),
+        pd.DataFrame,
+    )
+    check(
+        assert_type(
+            df.query("col1 > col2", parser="pandas", engine="numexpr"), pd.DataFrame
+        ),
+        pd.DataFrame,
+    )
```
Dr-Irv, line 530: duplicate tests

Dr-Irv (changes requested):

tests/test_frame.py
loicdiridollou: Fixed

pandas-stubs/core/frame.pyi
loicdiridollou: The goal was to still allow for someone passing kwargs as a dictionary instead of the individual arguments since it is the way documented in the docs like `df.query("col1 > col2", **kwargs)` that you would pass from another function.

loicdiridollou: The idea is that someone may put a wrapper like: 
```python
def my_own_query(df, expr, **kwargs):
    return df.query(expr, **kwargs)
```
If you drop the second overload this will raise an error.

loicdiridollou:
I will raise an issue on the pandas side if it is not simpler to have all the arguments directly in the function. It is a good point since it should not be too hard to maintain.

pandas-stubs/core/frame.pyi
loicdiridollou: This test would raise a mypy/pyright error without the other overload.

https://github.com/pandas-dev/pandas-stubs/blob/d19ac8994dd69d4a1663310b45ce6cd288bd2cac/tests/test_frame.py#L532-L536

Dr-Irv: My take on this is that we shouldn't support the `kwargs` being passed in a `dict`, because they can't be type-checked. The docs say what `kwargs` are acceptable, via the `eval()` link, so let's not include this test (and the overloads supporting `kwargs`

pandas-stubs/core/frame.pyi
Dr-Irv: I think the goal here is to restrict the possible keyword arguments, so we should remove the overload.

Dr-Irv:
> I will raise an issue on the pandas side if it is not simpler to have all the arguments directly in the function. It is a good point since it should not be too hard to maintain.

It's a docs issue, so raise an issue there that it would be more useful to list the arguments under `query()` rather than have people have to click on `eval()`

loicdiridollou:
Opened the issue on pandas repo, will mark this PR as blocked for now until the pandas team can clarify orientation on this.

Dr-Irv:
> Opened the issue on pandas repo, will mark this PR as blocked for now until the pandas team can clarify orientation on this.

I think we should remove the `kwargs` from the stubs.  The docs do say that the only valid `kwargs` are the ones from `eval()`, so let's just include them.  If we get complaints, we point them to that part of the docs.

Dr-Irv (approved):
thanks @loicdiridollou

PR #1194, opened 2025-04-22
Improve parameter types for `DataFrame.pct_change` and `Series.pct_change`

### LINKED ISSUE #1169: type `kwargs` in `pct_change` according to params in `shift`

yangdanny97:
This PR does the following for the two functions in the title:
- replace kwargs with keyword-only parameters for `pct_change` based on parameters from `shift`
- replace `...` w/ simple defaults when applicable based on docs: https://pandas.pydata.org/docs/reference/api/pandas.Series.pct_change.html
- make `fill_method` for `Series.pct_change` be `None`, just like `DataFrame.pct_change` since all other values are deprecated

- [ ] Closes #1169

cc @MarcoGorelli should these have tests? they're not overloaded so I'm not sure if it's necessary

### DIFF
--- a/pandas-stubs/core/frame.pyi
+++ b/pandas-stubs/core/frame.pyi
@@ -15,8 +15,8 @@ from typing import (
     Generic,
     Literal,
     NoReturn,
-    overload,
     Unpack,
+    overload,
 )
 
 from _typing import TimeZones
@@ -1989,10 +1989,10 @@ class DataFrame(NDFrame, OpsMixin, _GetItemHack):
     def ne(self, other, axis: Axis = ..., level: Level | None = ...) -> Self: ...
     def pct_change(
         self,
-        periods: int = ...,
+        periods: int = 1,
         fill_method: None = ...,
         limit: int | None = ...,
-        freq=...,
+        freq: Frequency | dt.timedelta | None = None,
         **kwargs: Unpack[PctChangeKwargs],
     ) -> Self: ...
     def pop(self, item: _str) -> Series: ...
--- a/pandas-stubs/core/frame.pyi
+++ b/pandas-stubs/core/frame.pyi
@@ -16,6 +16,7 @@ from typing import (
     Literal,
     NoReturn,
     overload,
+    Unpack,
 )
 
 from _typing import TimeZones
@@ -113,6 +114,7 @@ from pandas._typing import (
     NaPosition,
     NDFrameT,
     ParquetEngine,
+    PctChangeKwargs,
     QuantileInterpolation,
     RandomState,
     ReadBuffer,
@@ -1991,7 +1993,7 @@ class DataFrame(NDFrame, OpsMixin, _GetItemHack):
         fill_method: None = ...,
         limit: int | None = ...,
         freq=...,
-        **kwargs: Any,  # TODO: make more precise https://github.com/pandas-dev/pandas-stubs/issues/1169
+        **kwargs: Unpack[PctChangeKwargs],
     ) -> Self: ...
     def pop(self, item: _str) -> Series: ...
     def pow(
--- a/pandas-stubs/core/series.pyi
+++ b/pandas-stubs/core/series.pyi
@@ -22,6 +22,7 @@ from typing import (
     Literal,
     NoReturn,
     overload,
+    Unpack,
 )
 
 from _typing import (
@@ -143,6 +144,7 @@ from pandas._typing import (
     MaskType,
     NaPosition,
     ObjectDtypeArg,
+    PctChangeKwargs,
     QuantileInterpolation,
     RandomState,
     Renamer,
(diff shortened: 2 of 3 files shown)


pandas-stubs/core/frame.pyi
```diff
@@ -1987,11 +1987,13 @@ class DataFrame(NDFrame, OpsMixin, _GetItemHack):
     def ne(self, other, axis: Axis = ..., level: Level | None = ...) -> Self: ...
     def pct_change(
         self,
-        periods: int = ...,
+        periods: int = 1,
```
Dr-Irv, line 1990: You can't do this in a stubs.  Please leave it as `periods: int = ...` because the `periods` argument is optional.

pandas-stubs/core/frame.pyi
```diff
@@ -1987,11 +1987,13 @@ class DataFrame(NDFrame, OpsMixin, _GetItemHack):
...
         fill_method: None = ...,
         limit: int | None = ...,
```
Dr-Irv, line 1992: `limit` has been deprecated, so remove this

pandas-stubs/core/frame.pyi
```diff
@@ -1987,11 +1987,13 @@ class DataFrame(NDFrame, OpsMixin, _GetItemHack):
...
-        freq=...,
-        **kwargs: Any,  # TODO: make more precise https://github.com/pandas-dev/pandas-stubs/issues/1169
+        freq: Frequency | dt.timedelta | None = None,
```
Dr-Irv, line 1993:
```suggestion
        freq: DateOffset| dt.timedelta | None = None,
```
`Frequency` is too wide.  Should be changed in `shift` too.  That's old code.

pandas-stubs/core/frame.pyi
```diff
@@ -1987,11 +1987,13 @@ class DataFrame(NDFrame, OpsMixin, _GetItemHack):
...
+        *,
+        axis: AxisIndex = ...,
+        fill_value: object | None = ...,
```
Dr-Irv, line 1996:
```suggestion
        fill_value: Scalar | NAType | None = ...,
```
`object` is too wide.

pandas-stubs/core/series.pyi
```diff
@@ -1548,11 +1548,13 @@ class Series(IndexOpsMixin[S1], NDFrame):
...
-        fill_method: _str = ...,
+        periods: int = 1,
+        fill_method: None = ...,
         limit: int | None = ...,
-        freq=...,
-        **kwargs: Any,  # TODO: make more precise https://github.com/pandas-dev/pandas-stubs/issues/1169
+        freq: Frequency | timedelta | None = None,
+        *,
+        axis: AxisIndex = ...,
+        fill_value: object | None = ...,
```
Dr-Irv, line 1557: Same comment as above on `DataFrame.pct_change()` .  Also, remove `axis` as it is not used in `Series.pct_change()`.

Dr-Irv (changes requested):
Can you also add tests for the various arguments in `test_frame.py:test_dataframe_pct_change()` ?

pandas-stubs/core/frame.pyi
yangdanny97: Sorry, could you please clarify? The argument is still optional when written like this.

I see a few other places where the default is written when it's a simple literal, like https://github.com/pandas-dev/pandas-stubs/blob/main/pandas-stubs/core/frame.pyi#L317

yangdanny97: Understood, thank you

tests/test_frame.py
```diff
@@ -2428,7 +2428,10 @@ def test_groupby_series_methods() -> None:
 
 def test_dataframe_pct_change() -> None:
     df = pd.DataFrame({"x": [1, 2, 2, 3, 3], "y": [10, 20, 30, 40, 50]})
+    df.pct_change()
     df.pct_change(fill_method=None)
+    df.pct_change(axis="columns", periods=-1)
+    df.pct_change(fill_value=0)
```
Dr-Irv, line 2434: I know these tests were "old", but this is a chance to use our current convention, e.g.
```python
check(assert_type(df.pct_change(), pd.DataFrame), pd.DataFrame)
```

Dr-Irv (changes requested):
can you also add tests for `Series.pct_change()` in `test_series.py` ?

Use the `check(assert_type(...` pattern

yangdanny97:
I had to change the return type of `Series.pct_change` to get the tests to pass, I don't think the result always has the same dtype as the input.

Based on the [docs](https://pandas.pydata.org/docs/reference/api/pandas.Series.pct_change.html), the output seems to be a series of floats, but `shift` is typed with `UnknownSeries` as the return so maybe this one should also be `UnknownSeries`...

@Dr-Irv  thoughts?

Dr-Irv:
> I had to change the return type of `Series.pct_change` to get the tests to pass, I don't think the result always has the same dtype as the input.
> 
> Based on the [docs](https://pandas.pydata.org/docs/reference/api/pandas.Series.pct_change.html), the output seems to be a series of floats, but `shift` is typed with `UnknownSeries` as the return so maybe this one should also be `UnknownSeries`...
> 
> @Dr-Irv thoughts?

You did the right thing.  `Series.pct_change()` will return a `Series[float]`, so I'm good with that.  The previous declaration was just wrong, and existed from when we first created the repo.

PR #1195, opened 2025-04-22
GH1174 more precise kwargs in `DataFrame.to_clipboard`

### LINKED ISSUE #1174: type `kwargs` in `DataFrame.to_clipboard` according to `read_csv`

yangdanny97:
This PR types the keyword parameters in DataFrame.to_clipboard according to to_csv

- [ ] Closes #1174


pandas-stubs/core/frame.pyi
```diff
@@ -2228,9 +2230,29 @@ class DataFrame(NDFrame, OpsMixin, _GetItemHack):
     def take(self, indices: list, axis: Axis = ..., **kwargs: Any) -> Self: ...
     def to_clipboard(
         self,
-        excel: _bool = ...,
-        sep: _str | None = ...,
-        **kwargs: Any,  # TODO: make more precise https://github.com/pandas-dev/pandas-stubs/issues/1174
+        excel: _bool = True,
```
Dr-Irv, line 2233:
```suggestion
        excel: _bool = ...,
```
default args don't go in stubs

pandas-stubs/core/frame.pyi
```diff
@@ -2228,9 +2230,29 @@ class DataFrame(NDFrame, OpsMixin, _GetItemHack):
...
+        sep: _str | None = None,
+        *,
+        path_or_buf: FilePath | WriteBuffer[bytes] | WriteBuffer[str] | None = None,
```
Dr-Irv, line 2236: This argument is incorrect with `to_clipboard`

Dr-Irv (changes requested):
If you want to copy the arguments from `to_csv()`, use what is in `frame.pyi`, not what is in the source code.

Also, add tests for some of the `to_csv()` arguments with `to_clipboard()`, in `test_io.py:test_clipboard()`

Dr-Irv (changes requested):
If you want to copy the arguments from `to_csv()`, use what is in `generic.pyi`, not what is in the source code.

Also, add tests for some of the `to_csv()` arguments with `to_clipboard()`, in `test_io.py:test_clipboard()`

I think it would be better to delete `to_clipboard()` from `frame.pyi` and `series.pyi` and just create the single definition in `generic.pyi`.

pandas-stubs/core/frame.pyi
yangdanny97: Ok, I wasn't sure since I saw a few functions with parameter default literals, but most were `...`.

The [official guide for writing stubs](https://typing.python.org/en/latest/guides/writing_stubs.html#functions-and-methods) says

> For default values, use the literal values of “simple” default values (None, bools, ints, bytes, strings, and floats). Use the ellipsis literal ... in place of more complex default values. 

IMO it helps with developer experience in the IDE a bit since otherwise they'd be forced to open up the docs to see the default, but I'm happy to split the defaults off into a separate change or add a note to never write defaults into the contributing guide.

Dr-Irv (approved):
Thanks @yangdanny97

PR #1197, opened 2025-04-22
GH1196 Add typing to pd.Index.inser/delete and derived classes

### LINKED ISSUE #1196: Type pd.Index.insert
Currenty `pd.Index.insert` returns unknowns while it should return `pd.Index`.

loicdiridollou:
- [x] Closes #1196


Dr-Irv (approved):
thanks @loicdiridollou

PR #1198, opened 2025-04-25
GH941 add `UnknownSeries` overload for `Series.map`

### LINKED ISSUE #941: `Series.map()` needs type for `arg` and result
Specifically i am trying to typehint this function:

```
def map_hitgroup(series: pd.Series[int]) -> pd.Series[str]:
    hitgroup_mapping: dict[int, str] = {
        0: "generic",
        1: "head",
        2: "chest",
        3: "stomach",
        4: "left arm",
        5: "right arm",
        6: "left leg",
        7: "right leg",
        8: "neck",
        10: "gear",
    }
    return series.map(
        hitgroup_mapping
    )
```

There are 2 issues here. The type hint for series.map says that the return type (parameter) is the same as the initial one, which is not the case here (but it still works when running it).

The other is that the type hint does not specify the [argument type](https://github.com/pandas-dev/pandas-stubs/blob/main/pandas-stubs/core/series.pyi#L906).

Which results in the classic unknown warning.

Is there anything that i can currently do to adress this or is there just no way with the stubs and pyright strict mode.

_Originally posted by @JanEricNitschke in https://github.com/pandas-dev/pandas-stubs/discussions/940_

yangdanny97:
Implements the suggested solution + tests in https://github.com/pandas-dev/pandas-stubs/issues/941#issuecomment-2289142525

- [ ] Closes #941

### DIFF
--- a/pandas-stubs/core/series.pyi
+++ b/pandas-stubs/core/series.pyi
@@ -965,6 +965,12 @@ class Series(IndexOpsMixin[S1], NDFrame):
         na_action: None = ...,
     ) -> Series[S2]: ...
     @overload
+    def map(
+        self,
+        arg: Callable[[Any], Any] | Mapping[Any, Any] | UnknownSeries,
+        na_action: Literal["ignore"] | None = ...,
+    ) -> UnknownSeries: ...
+    @overload
     def aggregate(
         self: Series[int],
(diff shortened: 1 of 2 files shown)


Dr-Irv (approved):
thanks @yangdanny97

PR #1199, opened 2025-04-25
GH1103 Fixes for `Dataframe.squeeze` and `Series.squeeze`

### LINKED ISSUE #1103: Series.squeeze return type should be a Series or Scalar
**Describe the bug**

The return type for DataFrame/Series.squeeze is currently specified as Scalar, which is incomplete. According to the official documentation (pandas.DataFrame.squeeze and pandas.Series.squeeze), the correct return type should be DataFrame, Series, or scalar.

Pandas documentation:
 - https://pandas.pydata.org/docs/reference/api/pandas.DataFrame.squeeze.html 
 - https://pandas.pydata.org/docs/reference/api/pandas.Series.squeeze.html

Admittedly how a series squeeze returns a DataFrame needs further investigation, but Series is definitely possible ;)

mypy error:

```bash
error: Incompatible return value type (got "str | bytes | date | timedelta | datetime64[date | int | None] | timedelta64[timedelta | int | None] | int | float | complex", expected "Series[Any]")  [return-value]
```

**To Reproduce**

Example test.py:

```python
import pandas as pd

def foo() -> pd.Series:
    df = pd.DataFrame({
        "A": [1, 2, 3]
    })

    return df["A"].squeeze()


print(
    foo()
)
```

Run it, just to see the expected output (a series):

```bash
(.venv) ~/test-case $ python test.py
0    1
1    2
2    3
Name: A, dtype: int64
```

mypy expects a scalar:

```bash
(.venv) ~/test-case $ mypy test.py
test.py:8: error: Incompatible return value type (got "str | bytes | date | timedelta | datetime64[date | int | None] | timedelta64[timedelta | int | None] | int | float | complex", expected "Series[Any]")  [return-value]
```

versions:

```bash
(.venv) ~/test-case $ mypy --version
mypy 1.14.1 (compiled: yes)
(.venv) ~/test-case $ pip freeze
mypy==1.14.1
mypy-extensions==1.0.0
numpy==2.2.2
pandas==2.2.3
pandas-stubs==2.2.3.241126
python-dateutil==2.9.0.post0
pytz==2025.1
six==1.17.0
types-pytz==2024.2.0.20241221
typing_extensions==4.12.2
tzdata==2025.1
(.venv) ~/test-case $ python --version
Python 3.11.10
```

 - OS: MacOS
 - OS Version 15.2
 - python version 3.11
 - version of type 1.14.1
 - version of installed `pandas-stubs` 2.2.3.241126

yangdanny97:
This PR does the following based on #1103 and https://pandas.pydata.org/docs/reference/api/pandas.DataFrame.squeeze.html
- add a return type annotation to Dataframe.squeeze
- remove the axis parameter for Series.squeeze, as the docs say "For Series this parameter is unused and defaults to None."
- add Series as a possible return type for Series.squeeze. I think since it only happens when the object is returned unchanged, the dtype is the same as the original.
- add unit tests for the different possible returns (I'm not sure whether `assert_type` provides any value here, since the return is just a union)

- [ ] Closes #1103


tests/test_frame.py
```diff
@@ -3231,6 +3231,17 @@ def test_resample() -> None:
...
 
+def test_squeeze() -> None:
+    df1 = pd.DataFrame({"a": [1, 2], "b": [3, 4]})
+    check(df1.squeeze(), pd.DataFrame)
+    df2 = pd.DataFrame({"a": [1, 2]})
+    check(df2.squeeze(), pd.Series)
+    df3 = pd.DataFrame({"a": [1], "b": [2]})
+    check(df3.squeeze(), pd.Series, np.integer)
+    df4 = pd.DataFrame({"a": [1]})
+    check(df4.squeeze(), np.integer)
```
Dr-Irv, line 3242: You need to do `check(assert_type(...` here and in the tests in `test_series.py`.  Since the types are unions, you will have to import `Union` so that it works at runtime.

Dr-Irv (approved):
thanks @yangdanny97

PR #1200, opened 2025-04-25
GH730 allow str for float format to string

### LINKED ISSUE #730: `float_format` in `.to_string` can be `str`
**Describe the bug**
The change in #721 broke this for mypy. Although the current docs say "one-parameter function, optional, default None", you are allowed to pass a string that specifies the format like in `.to_csv`.

**To Reproduce**
These usages all work with both frame and series.
```python
# t.py
import pandas as pd

df = pd.DataFrame({"a": [0.1, 1.0, 3.14]})

print(df.to_string(float_format="%.3g"))
print(df.to_string(float_format="{:.3g}".format))
print(df.to_string(float_format=lambda x: f"{x:.3e}"))
print(df.to_string(float_format=pd.io.formats.format.EngFormatter(3, True)))

print(df.a.to_string(float_format="%.3g"))
print(df.a.to_string(float_format="{:.3g}".format))
print(df.a.to_string(float_format=lambda x: f"{x:.3e}"))
print(df.a.to_string(float_format=pd.io.formats.format.EngFormatter(3, True)))
```
```
> mypy t.py
t.py:5: error: No overload variant of "to_string" of "DataFrame" matches argument type "str"  [call-overload]
t.py:5: note: Possible overload variants:
t.py:5: note:     def [HashableT1 <: Hashable, HashableT2 <: Hashable] to_string(self, buf: Union[str, PathLike[str], WriteBuffer[str]], columns: Union[List[HashableT1], Index, Series[Any], None] = ..., col_space: Union[int, List[int], Dict[HashableT2, int], None] = ..., header: Union[bool, List[str], Tuple[str, ...]] = ..., index: bool = ..., na_rep: str = ..., formatters: Union[List[Callable[..., Any]], Tuple[Callable[..., Any], ...], Mapping[Union[str, int], Callable[..., Any]], None] = ..., float_format: Optional[Callable[[float], str]] = ..., sparsify: Optional[bool] = ..., index_names: bool = ..., justify: Optional[str] = ..., max_rows: Optional[int] = ..., max_cols: Optional[int] = ..., show_dimensions: bool = ..., decimal: str = ..., line_width: Optional[int] = ..., min_rows: Optional[int] = ..., max_colwidth: Optional[int] = ..., encoding: Optional[str] = ...) -> None
t.py:5: note:     def [HashableT <: Hashable] to_string(self, buf: None = ..., columns: Union[List[HashableT], Index, Series[Any], None] = ..., col_space: Union[int, List[int], Dict[Hashable, int], None] = ..., header: Union[bool, Sequence[str]] = ..., index: bool = ..., na_rep: str = ..., formatters: Union[List[Callable[..., Any]], Tuple[Callable[..., Any], ...], Mapping[Union[str, int], Callable[..., Any]], None] = ..., float_format: Optional[Callable[[float], str]] = ..., sparsify: Optional[bool] = ..., index_names: bool = ..., justify: Optional[str] = ..., max_rows: Optional[int] = ..., max_cols: Optional[int] = ..., show_dimensions: bool = ..., decimal: str = ..., line_width: Optional[int] = ..., min_rows: Optional[int] = ..., max_colwidth: Optional[int] = ..., encoding: Optional[str] = ...) -> str
t.py:10: error: No overload variant of "to_string" of "Series" matches argument type "str"  [call-overload]
t.py:10: note: Possible overload variants:
t.py:10: note:     def to_string(self, buf: Union[str, PathLike[str], WriteBuffer[str]], na_rep: str = ..., float_format: Callable[[float], str] = ..., header: bool = ..., index: bool = ..., length: bool = ..., dtype: bool = ..., name: bool = ..., max_rows: Optional[int] = ..., min_rows: Optional[int] = ...) -> None
t.py:10: note:     def to_string(self, buf: None = ..., na_rep: str = ..., float_format: Callable[[float], str] = ..., header: bool = ..., index: bool = ..., length: bool = ..., dtype: bool = ..., name: bool = ..., max_rows: Optional[int] = ..., min_rows: Optional[int] = ...) -> str
```

 - OS: [e.g. Windows, Linux, MacOS] Windows
 - OS Version [e.g. 22] 10
 - python version 3.10.11
 - version of type checker mypy 1.3.0
 - version of installed `pandas-stubs` 2.0.2.230605
 -  pandas 2.0.2

**Additional context**
Maybe [`FloatFormatType`](https://github.com/pandas-dev/pandas/blob/abcd440e11d061171b0a4eeb058748629a288568/pandas/_typing.py#L303) can be used (again) but could be updated to be more specific about the callable (`Callable[[float], str]`).

And perhaps the pandas docs should be updated? Unless support for passing format string was supposed to have been removed?

loicdiridollou:
- [x] Closes #730 

Should fix the issue, yet passing an `str` to `float_format` will depend on the underlying type, for example for a float it will work totally fine but for numpy Float64 it won't work.
So the typehints are correct to the extent that it also works at runtime.

Docs are not fixed yet, seems like the issue on the pandas repo is stale https://github.com/pandas-dev/pandas/issues/53675


tests/test_frame.py
```diff
@@ -1698,6 +1699,28 @@ def test_types_to_string() -> None:
...
+    """Test the different argument types for float_format."""
+    df = pd.DataFrame(
+        {"values": [2.304, 1.1, 3487392, 13.4732894237, 14.3, 18.0, 17.434, 19.3]}
+    )
+    check(assert_type(df.to_string(), str), str)
+
+    def _formatter(x) -> str:
+        return f"{x:.2f}"
+
+    check(assert_type(df.to_string(), str), str)
```
Dr-Irv, line 1712: you can delete this.  It's the same as line 1707

tests/test_series.py
```diff
@@ -2932,6 +2934,29 @@ def test_to_string() -> None:
...
+    """Test the different argument types for float_format."""
+    sr = pd.Series(
+        [2.304, 1.1, 3487392, 13.4732894237, 14.3, 18.0, 17.434, 19.3], name="values"
+    )
+    check(assert_type(sr.to_string(), str), str)
+
+    def _formatter(x) -> str:
+        return f"{x:.2f}"
+
+    check(assert_type(sr.to_string(), str), str)
```
Dr-Irv, line 2947: delete - same as line 2942

tests/test_series.py
```diff
@@ -2932,6 +2934,29 @@ def test_to_string() -> None:
...
+    check(assert_type(sr.to_string(float_format=_formatter), str), str)
+    check(
+        assert_type(
+            sr.to_string(float_format=EngFormatter(accuracy=2, use_eng_prefix=False)),
+            str,
+        ),
+        str,
+    )
+    check(assert_type(sr.to_string(float_format="%.2f"), str), str)
+    check(assert_type(sr.to_string(float_format="%.2f"), str), str)
```
Dr-Irv, line 2957: delete - repeated line

pandas-stubs/core/frame.pyi
```diff
@@ -2311,7 +2314,7 @@ class DataFrame(NDFrame, OpsMixin, _GetItemHack):
         index: _bool = ...,
         na_rep: _str = ...,
         formatters: FormattersType | None = ...,
-        float_format: Callable[[float], str] | None = ...,
+        float_format: FloatFormatType | Callable[[float], str] | None = ...,
```
Dr-Irv, line 2317: `FloatFormatType` is too wide, as it includes any `Callable`.  

Can you change `FloatFormatType` to use `Callable[[float], str]` instead of `Callable` in `_typing.pyi`, and then you don't need the `Callable` declaration here or elsewhere?

Dr-Irv (changes requested):

pandas-stubs/core/frame.pyi
loicdiridollou: I was going to suggest that indeed since we should collapse those two parts.
This is addressed.

Dr-Irv (approved):
thanks @loicdiridollou

PR #1205, opened 2025-04-27
move private imports from testing into tests/__init__.py

Dr-Irv:
In preparation for an eventual change in `pyright` that would ban private imports, I have moved the private imports for the tests into `tests/__init__.py`.  

Also will be working to make some of the current private imports public as reflected in the discussion here:  
https://github.com/pandas-dev/pandas/issues/55231#issuecomment-2833691635

### DIFF
--- a/pandas-stubs/_typing.pyi
+++ b/pandas-stubs/_typing.pyi
@@ -31,6 +31,7 @@ from pandas.core.generic import NDFrame
 from pandas.core.groupby.grouper import Grouper
 from pandas.core.indexes.base import Index
 from pandas.core.series import Series
+from pandas.core.tools.datetimes import FulldatetimeDict
 from typing_extensions import (
     ParamSpec,
     TypeAlias,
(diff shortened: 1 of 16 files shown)


twoertwein:
Thank you @Dr-Irv ! Just had a quick look: the changes look great! I will review/approve it later today.

I assume we are still allowed to import from _typing inside the stubs?

Dr-Irv:
> I assume we are still allowed to import from _typing inside the stubs?

Yes.  The issue that came up with pyright (and they removed the feature in 1.1.400 at my request - but want to put it back), is that if you installed `pandas-stubs` and had code that did an import from `pandas._typing`, it would report that as an import from a private module.

So this PR is "centralizing" all of the imports from the private module, except for ones that I am proposing to put somewhere else.

twoertwein (approved):
Looks great and the new pyright feature seems also useful to avoid using private APIs! I wouldn't be surprised if some people directly import from `_libs`.

PR #1206, opened 2025-04-29
GH984 Add overload for DataFrame.clip and update those for Series.clip

### LINKED ISSUE #984: Create overloads for `DataFrame.clip()`
From discussion in #982 .  See all comments there.

              It's possible (but not required) that you could also handle this via typing checks by creating overloads that are the acceptable combinations of the arguments `lower`, `upper` and `axis`

_Originally posted by @Dr-Irv in https://github.com/pandas-dev/pandas-stubs/pull/982#discussion_r1725010182_

loicdiridollou:
- [x] Closes #984 

Turns out there is a nasty edge case when you pass everything as None for upper and lower and ask for `inplace=True`, it will return `Self` and not `None`.

### DIFF
--- a/tests/test_frame.py
+++ b/tests/test_frame.py
@@ -659,23 +659,124 @@ def test_types_quantile() -> None:
     df.quantile(np.array([0.25, 0.75]))
 
 
-@pytest.mark.parametrize("lower", [None, 5, pd.Series([3, 4])])
-@pytest.mark.parametrize("upper", [None, 15, pd.Series([12, 13])])
-@pytest.mark.parametrize("axis", [None, 0, "index"])
-def test_types_clip(lower, upper, axis) -> None:
-    def is_none_or_numeric(val: Any) -> bool:
-        return val is None or isinstance(val, int | float)
-
+def test_dataframe_clip() -> None:
+    """Test different clipping combinations for dataframe."""
     df = pd.DataFrame(data={"col1": [20, 12], "col2": [3, 14]})
-    uses_array = not (is_none_or_numeric(lower) and is_none_or_numeric(upper))
-    if uses_array and axis is None:
-        with pytest.raises(ValueError):
-            df.clip(lower=lower, upper=upper, axis=axis)
-    else:
-        check(
-            assert_type(df.clip(lower=lower, upper=upper, axis=axis), pd.DataFrame),
-            pd.DataFrame,
-        )
+    if TYPE_CHECKING_INVALID_USAGE:
+        df.clip(lower=pd.Series([4, 5]), upper=None, axis=None)  # type: ignore[call-overload] # pyright: ignore[reportCallIssue, reportArgumentType]
+        df.clip(lower=None, upper=pd.Series([4, 5]), axis=None)  # type: ignore[call-overload] # pyright: ignore[reportCallIssue, reportArgumentType]
+        df.clip(lower=pd.Series([1, 2]), upper=pd.Series([4, 5]), axis=None)  # type: ignore[call-overload] # pyright: ignore[reportCallIssue, reportArgumentType]
+        df.copy().clip(lower=pd.Series([1, 2]), upper=None, axis=None, inplace=True)  # type: ignore[call-overload] # pyright: ignore[reportCallIssue, reportArgumentType]
+        df.copy().clip(lower=None, upper=pd.Series([1, 2]), axis=None, inplace=True)  # type: ignore[call-overload] # pyright: ignore[reportCallIssue, reportArgumentType]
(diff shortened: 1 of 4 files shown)


Dr-Irv (changes requested):
can you add tests where either `lower` or `upper` is not specified?

I think that might lead to some changes in how you wrote the overloads.

loicdiridollou:
Turns out no changes were needed but agreed that this time we check all the cases.

Dr-Irv:
I'm a little confused here, as you closed the PR and then reopened it.  Should I review?

loicdiridollou:
Yes it is open for review, I have accidentally pressed the close button, my
mistake.

tests/test_frame.py
```diff
@@ -778,6 +778,117 @@ def test_dataframe_clip() -> None:
...
+        assert_type(df.clip(upper=15, axis="index", inplace=True), None),
+        type(None),
+    )
+    check(
+        assert_type(df.clip(upper=pd.Series([1, 2]), axis="index", inplace=True), None),
+        type(None),
+    )
+    check(assert_type(df.clip(upper=None, axis="index"), pd.DataFrame), pd.DataFrame)
+    check(assert_type(df.clip(upper=15, axis="index"), pd.DataFrame), pd.DataFrame)
+    check(assert_type(df.clip(upper=None, axis="index"), pd.DataFrame), pd.DataFrame)
```
Dr-Irv, line 810: lines 808 and 810 are the same

tests/test_frame.py
```diff
@@ -778,6 +778,117 @@ def test_dataframe_clip() -> None:
...
+        pd.DataFrame,
+    )
+    check(
+        assert_type(df.clip(lower=pd.Series([1, 2]), axis="index"), pd.DataFrame),
+        pd.DataFrame,
+    )
+    check(
+        assert_type(df.clip(lower=None, axis="index"), pd.DataFrame),
+        pd.DataFrame,
+    )
```
Dr-Irv, line 881: lines 863 and 871 and 879  are the same test

tests/test_frame.py
```diff
@@ -659,23 +659,235 @@ def test_types_quantile() -> None:
...
+    check(
+        assert_type(df.clip(lower=5, axis="index", inplace=True), None),
+        type(None),
+    )
+    check(
+        assert_type(df.clip(lower=pd.Series([1, 2]), axis="index", inplace=True), None),
+        type(None),
+    )
+    check(
+        assert_type(df.clip(lower=None, axis="index", inplace=True), pd.DataFrame),
```
Dr-Irv, line 859: lines 847 and 859 are the same test

tests/test_frame.py
```diff
@@ -659,23 +659,235 @@ def test_types_quantile() -> None:
...
+        assert_type(df.clip(lower=None, axis="index"), pd.DataFrame),
+        pd.DataFrame,
+    )
+    check(
+        assert_type(df.clip(lower=None, axis=0, inplace=True), pd.DataFrame),
+        pd.DataFrame,
+    )
+    check(assert_type(df.clip(lower=5, axis=0, inplace=True), None), type(None))
+    check(
+        assert_type(df.clip(lower=None, axis=0, inplace=True), pd.DataFrame),
```
Dr-Irv, line 888: lines 883 and 888 are the same test

tests/test_frame.py
```diff
@@ -659,23 +659,235 @@ def test_types_quantile() -> None:
...
+        assert_type(df.clip(lower=None, axis=None, inplace=True), pd.DataFrame),
+        pd.DataFrame,
+    )
+    check(
+        assert_type(df.clip(lower=5, axis=None, inplace=True), None),
+        type(None),
+    )
+    check(
+        assert_type(df.clip(lower=None, axis=None, inplace=True), pd.DataFrame),
+        pd.DataFrame,
```
Dr-Irv, line 837: lines 828 and 836 are the same test

Dr-Irv (changes requested):
I think you may also need similar overloads for `clip` in `series.pyi` and then similar tests that omit the `lower` or `upper` arguments.

loicdiridollou:
Since clip for series does not use the axis when passing a list like
argument to lower or upper the overloads are already there. Let me clean up
the tests and add for those without upper or lower passed and we should be
in a good place, the big work was on the data frame side.

loicdiridollou:
Turns out the axis parameter does not need anything specific (to split the overloads) since it is not used (unless you pass something not allowed).

Dr-Irv (approved):
thanks @loicdiridollou

PR #1207, opened 2025-05-02
fix: Series initialiser also takes positional arguments

cmp0xff:
# Describe the bug
All arguments in `pandas.Series.__init__` [are positional](https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.Series.html), but in the stubs some arguments are key-word only.

# To Reproduce
## Provide a minimal runnable `pandas` example that is not properly checked by the stubs.
```py
import pandas as pd

pd.Series([], pd.DatetimeIndex([]), "float", "col")
```
## Indicate which type checker you are using (`mypy` or  `pyright`).
`mypy`
## Show the error message received from that type checker while checking your example.
```
error: No overload variant of "Series" matches argument types "list[Never]", "DatetimeIndex", "str", "str"  [call-overload]
note:     def [S1: str | bytes | date | time | bool | <11 more items>] Series(data: ndarray[Any, dtype[floating[_64Bit]]], index: ExtensionArray | ndarray[Any, Any] | Index[Any] | Series[Any] | list[Any] | dict[Any, Any] | range | tuple[Any, ...] | None = ..., *, dtype: ExtensionDtype | str | dtype[generic] | type[str] | type[complex] | type[bool] | type[object] = ..., name: Hashable = ..., copy: 
bool = ...) -> Series[float]
note:     def [S1: str | bytes | date | time | bool | <11 more items>] Series(data: Sequence[Never], index: ExtensionArray | ndarray[Any, Any] | Index[Any] | Series[Any] | list[Any] | dict[Any, Any] | range | tuple[Any, ...] | None = ..., *, dtype: ExtensionDtype | str | dtype[generic] | type[str] | type[complex] | type[bool] | type[object] = ..., name: Hashable = ..., copy: bool = ...) -> Series[Any]
note:     def [S1: str | bytes | date | time | bool | <11 more items>] Series(data: Sequence[list[str]], index: ExtensionArray | ndarray[Any, Any] | Index[Any] | Series[Any] | list[Any] | dict[Any, Any] | range | tuple[Any, ...] | None = ..., *, dtype: ExtensionDtype | str | dtype[generic] | type[str] | type[complex] | type[bool] | type[object] = ..., name: Hashable = ..., copy: bool = ...) -> Series[list[str]]
note:     def [S1: str | bytes | date | time | bool | <11 more items>] Series(data: Sequence[str], index: ExtensionArray | ndarray[Any, Any] | Index[Any] | Series[Any] | list[Any] | dict[Any, Any] | range | tuple[Any, ...] | None = ..., *, dtype: ExtensionDtype | str | dtype[generic] | type[str] | type[complex] | type[bool] | type[object] = ..., name: Hashable = ..., copy: bool = ...) -> Series[str]
```

# Please complete the following information:
## OS
`Windows`
## OS Version
`10`
## python version
`13.2`
## version of type checker
`1.15.0`
## version of installed `pandas-stubs`
`2.2.3.250308`


cmp0xff:
Whoa, I'll need to think about this PR. Didn't expect so many errors.

pandas-stubs/core/series.pyi
```diff
@@ -303,8 +298,7 @@ class Series(IndexOpsMixin[S1], NDFrame):
         cls,
         data: _ListLike,
         index: Axes | None = ...,
-        *,
-        dtype: TimestampDtypeArg,
+        dtype: TimestampDtypeArg = ...,
```
Dr-Irv, line 301: this is the change that probably caused the first issue I saw.  You changed `*, dtype: TimestampDtypeArg` to `dtype: TimestampDtypeArg =...` which is wrong.  The way it was before, the stubs are written to say that if a user specifies a `dtype` argument with type `TimestampDtypeArg`, it will match this overload.  By putting the `...` in there, it becomes optional so other things will match this overload and then return a `TimestampSeries`

pandas-stubs/core/series.pyi
```diff
@@ -354,8 +345,7 @@ class Series(IndexOpsMixin[S1], NDFrame):
         cls,
         data: Scalar | _ListLike | dict[HashableT1, Any] | None,
         index: Axes | None = ...,
-        *,
-        dtype: type[S1],
+        dtype: type[S1] = ...,
```
Dr-Irv, line 348: same comment

pandas-stubs/core/series.pyi
cmp0xff: 41288292e01d16a13221b373f8ba0eee07fb1be3 490c1699ebe87ba84e5b46387698ffc051f25347

pandas-stubs/core/series.pyi
cmp0xff: 41288292e01d16a13221b373f8ba0eee07fb1be3 490c1699ebe87ba84e5b46387698ffc051f25347

Dr-Irv (approved):
thanks @cmp0xff

PR #1211, opened 2025-05-10
remove Any from np.ndarray declarations in _typing.pyi

### LINKED ISSUE #1210: future numpy update needed
numpy is changing typing and revealed an issue in pandas-stubs.  See https://github.com/numpy/numpy/pull/28940#issuecomment-2868897126

I think the solution is to not reference `np.ndarray[Any, Any]` in `_typing.pyi` and just use `np.ndarray` .

We can try that change now, and then whenever `numpy` does a new release, see if it is still broken.

Dr-Irv:
- [X] Closes #1210 
- [ ] Tests added: 
   -  None needed


twoertwein:
Thank you @Dr-Irv !

PR #1219, opened 2025-05-18
align typing with pandas source

### LINKED ISSUE #128: CLEAN:  Reconcile types in pandas._typing.py and pandas-stubs._typing.pyi

Dr-Irv:
Possibly closes #128 

I went through the current pandas source in `pandas/_typing.py` and separated the types declared there from the "private" types only used in the stubs, which now appear at the bottom of the file, and added documentation where appropriate to indicate where things are different.

Unclear whether the "private" types in the stubs should all be preceded by underscores.  They are imported from `pandas/_typing`, so that makes them private anyway.  When I wrote #128, I suggested to make them all private.  Now that I have more experience with typing, I don't think it's necessary.

By doing this, I have a list of types that should be "public" which will eventually go in `pandas.api.typing.aliases` .  That will then allow me to work on https://github.com/pandas-dev/pandas/issues/55231 

Asking for opinions of @loicdiridollou and a review/approval/merge from @twoertwein


pandas-stubs/core/arrays/interval.pyi
```diff
@@ -15,10 +18,14 @@ from pandas._libs.interval import (
 from pandas._typing import (
     Axis,
     Scalar,
+    ScalarIndexer,
+    SequenceIndexer,
     TakeIndexer,
     np_ndarray_bool,
 )
 
+IntervalOrNA: TypeAlias = Interval | float
```
loicdiridollou, line 27: Do we need to keep this in this file or should we move it to _typing?

loicdiridollou:
Great job in here, I think this is a major improvement. Hopefully that can unblock the other issue in pandas repo.
Added a few comments about some types being defined in files instead of _typing.pyi.
Not totally sure what is the exact direction that would fit the team at pandas but this is a ton of work already here and we can continue to use the `stubtest` to clean up some of the function signatures that have drifted from the pandas implementaion.

Dr-Irv:
> Added a few comments about some types being defined in files instead of _typing.pyi.

For now, I'd rather mirror what is in the `pandas` source, where both of those types are defined in the respective files.  That's where I picked them up from.  If someone really wants access to them, they can import them from those files, which are not in private modules.

pandas-stubs/_typing.pyi
```diff
@@ -53,36 +53,126 @@ from pandas.core.dtypes.dtypes import (
...
+    @overload
+    def __getitem__(self, index: slice, /) -> Sequence[_T_co]: ...
+    def __contains__(self, value: object, /) -> bool: ...
+    def __len__(self) -> int: ...
+    def __iter__(self) -> Iterator[_T_co]: ...
+    def index(self, value: Any, start: int = ..., stop: int = ..., /) -> int: ...
+    def count(self, value: Any, /) -> int: ...
+    def __reversed__(self) -> Iterator[_T_co]: ...
+
+ListLike: TypeAlias = AnyArrayLike | SequenceNotStr | range
```
twoertwein, line 86: Not sure, does SequenceNotStr need a paramter here? Something like `ListLike: TypeAlias = AnyArrayLike | SequenceNotStr[Any] | range`

pandas-stubs/_typing.pyi
```diff
@@ -53,36 +53,126 @@ from pandas.core.dtypes.dtypes import (
...
+    | float
+    | str
+)
+TimestampNonexistent: TypeAlias = (
+    Literal["shift_forward", "shift_backward", "NaT", "raise"]
+    | Timedelta
+    | datetime.timedelta
+)
+TimedeltaConvertibleTypes: TypeAlias = (
+    Timedelta | datetime.timedelta | np.timedelta64 | np.int64 | float | str
```
twoertwein, line 131: Could probably replace `np.int64` with `np.integer`?

pandas-stubs/_typing.pyi
```diff
@@ -53,36 +53,126 @@ from pandas.core.dtypes.dtypes import (
...
 )
-DictConvertible: TypeAlias = FulldatetimeDict | DataFrame
 
-CorrelationMethod: TypeAlias = (
-    Literal["pearson", "kendall", "spearman"]
-    | Callable[[np.ndarray, np.ndarray], float]
+Timezone: TypeAlias = str | tzinfo  # Not used
```
twoertwein, line 134: Does "Not used" mean, it is declared in pandas but we don't use it in pandas-stubs?

pandas-stubs/_typing.pyi
```diff
@@ -392,74 +479,40 @@ AstypeArg: TypeAlias = (
...
-    def truncate(self, size: int | None = ..., /) -> int: ...
-
-FilePath: TypeAlias = str | PathLike[str]
+# DtypeArg specifies all allowable dtypes in a functions its dtype argument
+DtypeArg: TypeAlias = Dtype | Mapping[Hashable, Dtype]
+DtypeObj: TypeAlias = np.dtype[np.generic] | ExtensionDtype
 
-_T_co = TypeVar("_T_co", covariant=True)
+# converters
+ConvertersArg: TypeAlias = dict[Hashable, Callable[[Dtype], Dtype]]
```
twoertwein, line 487: I think this might need to be `Mapping[Hashable, Callable[[Dtype], Dtype]]` or `dict[HashableT, Callable[[Dtype], Dtype]]`  to allow sub-classes when this is used outside of bidirectional type inference (assigned to a variable and then used as an argument).

I believe we discussed Mapping+Hashable vs dict+HasableT at somepoint. I forgot which one we ended up with (feel like dict as pandas did not always accept sub-classes).

pandas-stubs/_typing.pyi
```diff
@@ -392,74 +479,40 @@ AstypeArg: TypeAlias = (
...
-    def __contains__(self, value: object, /) -> bool: ...
-    def __len__(self) -> int: ...
-    def __iter__(self) -> Iterator[_T_co]: ...
-    def index(self, value: Any, start: int = ..., stop: int = ..., /) -> int: ...
-    def count(self, value: Any, /) -> int: ...
-    def __reversed__(self) -> Iterator[_T_co]: ...
+# parse_dates
+ParseDatesArg: TypeAlias = (
+    bool | list[Hashable] | list[list[Hashable]] | dict[Hashable, list[Hashable]]
+)
```
twoertwein, line 492: Might need HashableT (and Mapping/dict + HashableT2 for the dict values)

pandas-stubs/core/arrays/datetimelike.pyi
```diff
@@ -1,17 +1,31 @@
...
+    DatetimeLikeScalar,
+    PositionalIndexerTuple,
+    ScalarIndexer,
+    SequenceIndexer,
+    TimeAmbiguous,
+    TimeNonexistent,
+    TimeUnit,
+)
+
+DTScalarOrNaT: TypeAlias = DatetimeLikeScalar | NaTType
```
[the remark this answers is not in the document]
twoertwein: Seems to be used only here so it might be nice to keep it here. I think this needs to be `_DTScalarOrNaT`.

pandas-stubs/core/arrays/interval.pyi
twoertwein: probably `_IntervalOrNA`

pandas-stubs/core/arrays/sparse/array.pyi
```diff
@@ -1,8 +1,23 @@
...
     ExtensionOpsMixin,
 )
+from typing_extensions import Self
+
+from pandas._typing import (
+    ScalarIndexer,
+    SequenceIndexer,
+)
+
+class ellipsis(Enum):
```
twoertwein, line 19: is this class public in pandas?

Dr-Irv: This works in `pandas`:
```
from pandas.core.arrays.sparse.array import ellipsis
```

pandas-stubs/core/arrays/datetimelike.pyi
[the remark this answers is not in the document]
Dr-Irv: I want to be consistent with what is in `pandas` for right now.  This statement currently works:  `from pandas.core.arrays.datetimelike import DTScalarOrNaT`

pandas-stubs/core/arrays/interval.pyi
Dr-Irv: I want to be consistent with pandas for now.  This statement works:  `from pandas.core.arrays.interval import IntervalOrNA`

pandas-stubs/_typing.pyi
Dr-Irv: I think you're right.  Will fix that.

pandas-stubs/_typing.pyi
Dr-Irv: done

pandas-stubs/_typing.pyi
Dr-Irv: I clarified the comment.  It is declared in pandas, but not used.

pandas-stubs/_typing.pyi
Dr-Irv: I copied it from pandas.  Will change to `Mapping` here.

pandas-stubs/_typing.pyi
Dr-Irv: ok

Dr-Irv:
@twoertwein  thanks for the review.  I've pushed a new version that addresses your comments (and commented on a few other comments where I didn't make a change)

pandas-stubs/core/arrays/interval.pyi
twoertwein: Is this a shift in policy (from public == what is documented on the web) to more how the python community/type checkers defines public/private? That is fine with me, just want to make sure if that is the case.

Dr-Irv: > Is this a shift in policy (from public == what is documented on the web) to more how the python community/type checkers defines public/private? That is fine with me, just want to make sure if that is the case.

Not really.  I just want to be consistent right now with the types in `pandas`.  I think we need to change some of types likes these that are in `pandas` to be private (precede with underscores), which is a separate effort.

twoertwein (approved):

twoertwein:
Thanks @Dr-Irv !

PR #1223, opened 2025-05-19
GH1221 Fix typing module exposure in pandas.api

### LINKED ISSUE #1221: Is there some obvious reason why `api.typing` is not present in the `api` interface stub?
Is there some obvious reason why `api.typing` is not present in the `api` interface stub?
https://github.com/pandas-dev/pandas-stubs/blob/86073150f5e1b607c37a536ec99204ed21bad0cb/pandas-stubs/api/__init__.pyi
Was it missed in https://github.com/pandas-dev/pandas-stubs/pull/1058?
I see `types` there but not `typing`, both are public interfaces according to docs.

I just wanted to use `pandas.api.typing.NaTType` in my code but my static type checker complained that `typing` is not part of `pandas.api`.

_Originally posted by @Molkree in https://github.com/pandas-dev/pandas-stubs/discussions/1220_

loicdiridollou:
- [x] Closes #1221

### DIFF
--- a/pandas-stubs/api/__init__.pyi
+++ b/pandas-stubs/api/__init__.pyi
@@ -3,4 +3,5 @@ from pandas.api import (
     indexers as indexers,
     interchange as interchange,
     types as types,
+    typing as typing,
 )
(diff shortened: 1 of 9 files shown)


Dr-Irv (approved):
thanks @loicdiridollou 

This will close the original issue.  I will create a new issue based on the discussion there of other modules.

PR #1225, opened 2025-05-19
GH1217 Modify items/iterros/itertuples methods to return Iterator

### LINKED ISSUE #1217: Return type for DataFrame/Series iteration methods should be Iterator, not Iterable
**Describe the bug**

Several iteration methods in pandas(-stubs) specify their return type as `Iterable` while they are implemented as generators or return zip objects which qualify them as `Iterator`. This flags straight-forward use with `next`, which should be perfectly valid. Most of them are _documented_ to return an Iterator as per the docstring and it seems reasonable to assume that they actually return iterators. 

1. `DataFrame.itertuples` - Documented to "Iterate over" and return an "iterator", returns a `zip` object
2. `DataFrame.iterrows` - Documented to "Iterate over", Generator function
3. `DataFrame.items` - Documented to "Iterate over", Generator function
4. `Series.items` - Documented to "Lazily iterate over", Generator function

**To Reproduce**
```
import pandas as pd

def foo(df: pd.DataFrame):
    next(df.items())
```
(mypy)
```
test.py:4: error: No overload variant of "next" matches argument type "Iterable[tuple[Hashable, Series[Any]]]"  [call-overload]
test.py:4: note: Possible overload variants:
test.py:4: note:     def [_T] next(SupportsNext[_T], /) -> _T
test.py:4: note:     def [_T, _VT] next(SupportsNext[_T], _VT, /) -> _T | _VT
```

 - OS: Linux
 - python 3.12.3
 - mypy 1.15.0
 - pandas-stubs 2.2.3.250308

**Additional context**

- This also affects the type within the main pandas code, should I make another issue there?
- Related documentation issue: https://github.com/pandas-dev/pandas/issues/27861

loicdiridollou:
- [x] Closes #1217 


tests/test_frame.py
```diff
@@ -599,7 +599,7 @@ def test_types_median() -> None:
 def test_types_iterrows() -> None:
     df = pd.DataFrame(data={"col1": [2, 1], "col2": [3, 4]})
     check(
-        assert_type(df.iterrows(), "Iterable[tuple[Hashable, pd.Series]]"),
+        assert_type(df.iterrows(), "Iterator[tuple[Hashable, pd.Series]]"),
         Iterable,
         tuple,
     )
```
Dr-Irv, line 605: Can we add a test of this form for `iterrows()` and `itertuples()`:

```python
for t1, t2 in df.iterrows():
    check(assert_type(t1, Hashable), Hashable)
    check(assert_type(t2, pd.Series), pd.Series)
```

Dr-Irv (changes requested):

tests/test_frame.py
```diff
@@ -625,6 +628,9 @@ def test_types_itertuples() -> None:
         object,
     )
 
+    for t1 in df.itertuples():
+        check(assert_type(t1, _PandasNamedTuple), _PandasNamedTuple)
+
```
Dr-Irv, line 633: We shouldn't be using `_PandasNamedTuple` here in the call to `check` because what is happening internally is that pandas is creating a new class of type `pandas.core.foo.Pandas`, and if you use the `name="Foobar"` argument, the class is of the type `pandas.core.foo.Foobar`

But it's a dynamic class, so I think something like this would be appropriate:
```python
for t1 in df.itertuples():
    assert_type(t1, _PandasNamedTuple)
    assert(t1.__class__.__name__ == "Pandas")
    assert(isinstance(t1.Index, int))
    assert(isinstance(t1.col1, int))  
    assert(isinstance(t1.col2, int))
    for k in [0, 1, 2]:
        assert(isinstance(t1[k], int))
```

And you could add a similar test for having `name="Foobar"` as the argument, and only the line that checks the name of the class would need to be changed.

Dr-Irv (changes requested):

Dr-Irv (changes requested):

tests/test_frame.py
```diff
@@ -629,7 +629,22 @@ def test_types_itertuples() -> None:
     )
 
     for t1 in df.itertuples():
-        check(assert_type(t1, _PandasNamedTuple), _PandasNamedTuple)
+        assert_type(t1, _PandasNamedTuple)
+        assert t1.__class__.__name__ == "Pandas"
+        assert isinstance(t1.Index, int)
+        assert isinstance(t1.col1, int)
+        assert isinstance(t1.col2, int)
+        for k in [0, 1]:
```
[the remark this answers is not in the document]
loicdiridollou: Ahh yes correct I thought it was iterating over the number of elements.

Dr-Irv (approved):
thanks @loicdiridollou

PR #1228, opened 2025-05-21
GH1226 Add Styler.map to the stubs

### LINKED ISSUE #1226: `pandas.io.formats.style.Styler.map` is missing
**Describe the bug**
[`pandas.io.formats.style.Styler.map`](https://pandas.pydata.org/docs/reference/api/pandas.io.formats.style.Styler.map.html) method is missing from [the stub](https://github.com/pandas-dev/pandas-stubs/blob/93a9e11f4e4656a935853d3814ac9073adf4c9cc/pandas-stubs/io/formats/style.pyi#L56)

**To Reproduce**
```py
import pandas as pd

df = pd.DataFrame(data={"col1": [1, -2], "col2": [-3, 4]})
df.style.map(lambda v: "color: red;" if v < 0 else None)
```
Last line raises type checker errors. mypy: ` error: "Styler" has no attribute "map"  [attr-defined]`, pylance: `Cannot access attribute "map" for class "Styler"`

 - OS: WSL
 - OS Version [e.g. 22]: Ubuntu 2024
 - python version: 3.13
 - version of type checker: mypy 1.15.0/pylance 2025.5.1 with pyright 1.1.400
 - version of installed `pandas-stubs`: 2.2.3.250308 but the issue is present on the latest https://github.com/pandas-dev/pandas-stubs/commit/93a9e11f4e4656a935853d3814ac9073adf4c9cc

loicdiridollou:
- [x] Closes #1226


Dr-Irv (changes requested):

pandas-stubs/io/formats/style.pyi
```diff
@@ -71,6 +71,9 @@ class Styler(StylerRenderer):
         formatter: ExtFormatter | None = ...,
     ) -> None: ...
     def concat(self, other: Styler) -> Styler: ...
+    def map(
+        self, func: Callable, subset: Subset | None = ..., **kwargs: Any
```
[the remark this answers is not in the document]
loicdiridollou: I think this is a little tricky since the example on the docs mention that you can pass kwargs to the function so the signature would be something like, Callable[[Scalar, kwargs], str | None].
I will type it in as Callable[[Scalar], str | None] since the docs itself mention the return type None and then we can see if users complain. What do you think?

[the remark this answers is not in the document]
Dr-Irv: > I think this is a little tricky since the example on the docs mention that you can pass kwargs to the function so the signature would be something like, Callable[[Scalar, kwargs], str | None]. I will type it in as Callable[[Scalar], str | None] since the docs itself mention the return type None and then we can see if users complain. What do you think?

Can we do this with an overload?  If `kwargs` is not specified, then we use a single argument in the `Callable`.  If `kwargs` is specified as a `dict`, then a different `Callable` signature is needed.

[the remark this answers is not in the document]
Molkree: @Dr-Irv 
> kwargs has to be a dict

As far as I know kwargs are annotated with the type of the values so if it can be only ints, then it's `kwargs: int`, if it can be anything then it's `kwargs: Any`

I am pretty sure we don't expect only dictionaries to be passed in kwargs here so `Any` seems correct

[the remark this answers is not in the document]
loicdiridollou: The example in the docs passes a str and not a dict so indeed we may be a bit lax.

pandas-stubs/io/formats/style.pyi
```diff
@@ -71,6 +71,19 @@ class Styler(StylerRenderer):
...
+    @overload
+    def map(
+        self,
+        func: Callable[[Scalar], str | None],
+        subset: Subset | None = ...,
+    ) -> Styler: ...
+    @overload
+    def map(
+        self,
+        func: Callable[..., str | None],
```
loicdiridollou, line 83: @Dr-Irv I could not find how to type a function that takes Scalar and some other stuff, I tried `[Scalar, ...]` but it was not happy, feel free to recommend something better.

Molkree: > [Callable cannot express complex signatures](https://docs.python.org/3/library/typing.html#annotating-callable-objects:~:text=Callable%20cannot%20express%20complex%20signatures)

I think this is usually done using `Protocol`, I tried writing this:
```py
    class _MapCallable(Protocol):
        def __call__(self, first_arg: Scalar, /, *args: Any, **kwargs: Any) -> str | None: ...

    def map(
        self,
        func: _MapCallable,
        subset: Subset | None = ...,
        **kwargs: Any,
    ) -> Styler: ...
```
but it fails the test
```
tests/test_styler.py:256: error: Expression is of type "Any", not "Styler"  [assert-type]
tests/test_styler.py:257: error: No overload variant of "map" of "Styler" matches argument types "Callable[[str | bytes | date | datetime | timedelta | <7 more items> | complex | integer[Any] | floating[Any] | complexfloating[Any, Any], str], str | None]", "str"  [call-overload]
tests/test_styler.py:257: note: Possible overload variants:
tests/test_styler.py:257: note:     def map(self, func: Callable[[str | bytes | date | datetime | timedelta | <7 more items> | complex | integer[Any] | floating[Any] | complexfloating[Any, Any]], str | None], subset: _IndexSlice | slice[Any, Any, Any] | tuple[slice[Any, Any, Any], ...] | list[Any] | Index[Any] | None = ...) -> Styler
tests/test_styler.py:257: note:     def map(self, func: _MapCallable, subset: _IndexSlice | slice[Any, Any, Any] | tuple[slice[Any, Any, Any], ...] | list[Any] | Index[Any] | None = ..., **kwargs: Any) -> Styler
```

Maybe this will be helpful to figure something out.

Dr-Irv (changes requested):

pandas-stubs/io/formats/style.pyi
Molkree: > Note that `pyright` accepts the func without the `/` defined, but `mypy` doesn't, so this is a workaround to make `mypy` happy.

I had to double check that and I agree that this is mypy bug. According to [these typing rules](https://typing.python.org/en/latest/spec/callables.html#parameter-kinds) callable with standard argument of matching type should be assignable to the positional-only argument of matching type, same for kwargs but because we typed them as `Any` everything should be assignable to everything.

I think it's this bug specifically: https://github.com/python/mypy/issues/12525 (or a variation of it, error messages are not the same)

You learn something new everyday!

loicdiridollou: Thanks for all the feedback here, indeed I learn about those intricacies and mypy issues!
My only fear is that it would force people to have the argument-based function with the `/`, but I think it is worth trying out and seeing if we get complaints from the user base.

Dr-Irv: > Thanks for all the feedback here, indeed I learn about those intricacies and mypy issues! My only fear is that it would force people to have the argument-based function with the `/`, but I think it is worth trying out and seeing if we get complaints from the user base.

It only forces people who use `mypy` to include `/`.  If you use `pyright`, it isn't required.

Dr-Irv (approved):
thanks @loicdiridollou

PR #1231, opened 2025-05-29
GH1230 Improve types for pd.options.display.* properties

### LINKED ISSUE #1230: Type checks for pd.options.display.* are not accurate
**Describe the bug**
Various display options underneath `pd.options.display.` do not have correct annotations per pandas documentation.

The following columns show mismatches between [documented behavior](https://pandas.pydata.org/docs/reference/api/pandas.describe_option.html#pandas.describe_option) compared to type check:
```
pd.describe_option('display.(?:colheader_justify|large_repr|max_columns|max_colwidth|max_dir_items|max_rows|max_seq_items|memory_usage|min_rows|show_dimensions)')
```

**To Reproduce**
```python
import pandas as pd

# Overly restrictive types set, expect errors to not be raised
pd.options.display.max_columns = None
pd.options.display.max_colwidth = None
pd.options.display.max_dir_items = None
pd.options.display.max_rows = None
pd.options.display.max_seq_items = None
pd.options.display.min_rows = None

# Overly permissive values permitted, expect errors to be raised
pd.options.display.colheader_justify = "invalid_value"
pd.options.display.large_repr = "invalid_value"
pd.options.display.memory_usage = "invalid_value"
pd.options.display.show_dimensions = "invalid_value"
```

```
test.py:4: error: Incompatible types in assignment (expression has type "None", variable has type "int")  [assignment]
    pd.options.display.max_columns = None
                                     ^~~~
test.py:5: error: Incompatible types in assignment (expression has type "None", variable has type "int")  [assignment]
    pd.options.display.max_colwidth = None
                                      ^~~~
test.py:6: error: Incompatible types in assignment (expression has type "None", variable has type "int")  [assignment]
    pd.options.display.max_dir_items = None
                                       ^~~~
test.py:7: error: Incompatible types in assignment (expression has type "None", variable has type "int")  [assignment]
    pd.options.display.max_rows = None
                                  ^~~~
test.py:8: error: Incompatible types in assignment (expression has type "None", variable has type "int")  [assignment]
    pd.options.display.max_seq_items = None
                                       ^~~~
test.py:9: error: Incompatible types in assignment (expression has type "None", variable has type "int")  [assignment]
    pd.options.display.min_rows = None
                                  ^~~~
test.py:14: error: Incompatible types in assignment (expression has type "str", variable has type "bool")  [assignment]
    pd.options.display.memory_usage = "invalid_value"
                                      ^~~~~~~~~~~~~~~
Found 7 errors in 1 file (checked 1 source file)
```

 - OS: MacOS
 - OS Version: 15.5
 - python version: 3.12
 - pandas version: 2.2.3
 - version of type checker: mypy 1.9.0
 - version of installed `pandas-stubs`: 2.2.3.250308

**Additional context**
I'm happy to do PR, just let me know if that's preferred

mike-flowers-airbnb:
- [X] Closes #1230

Updated based off documentation in [config_init.py](https://github.com/pandas-dev/pandas/blob/main/pandas/core/config_init.py)

Let me know if you'd like me to update the second test to also use `assert_type` on each value.

### DIFF
--- a/pandas-stubs/_config/config.pyi
+++ b/pandas-stubs/_config/config.pyi
@@ -48,7 +48,7 @@ class DisplayUnicode(DictWrapper):
     east_asian_width: bool
 
 class Display(DictWrapper):
-    chop_threshold: int | None
+    chop_threshold: float | None
     colheader_justify: Literal["left", "right"]
     date_dayfirst: bool
     date_yearfirst: bool
--- a/pandas-stubs/_config/config.pyi
+++ b/pandas-stubs/_config/config.pyi
@@ -49,30 +49,30 @@ class DisplayUnicode(DictWrapper):
 
 class Display(DictWrapper):
     chop_threshold: int | None
-    colheader_justify: str
+    colheader_justify: Literal["left", "right"]
     date_dayfirst: bool
     date_yearfirst: bool
     encoding: str
     expand_frame_repr: bool
     float_format: Callable[[float], str] | None
     html: DisplayHTML
-    large_repr: str
+    large_repr: Literal["truncate", "info"]
     latex: DisplayLaTeX
     max_categories: int
-    max_columns: int
-    max_colwidth: int
-    max_dir_items: int
+    max_columns: int | None
+    max_colwidth: int | None
+    max_dir_items: int | None
     max_info_columns: int
     max_info_rows: int
-    max_rows: int
-    max_seq_items: int
-    memory_usage: bool
-    min_rows: int
+    max_rows: int | None
+    max_seq_items: int | None
+    memory_usage: bool | Literal["deep"] | None
+    min_rows: int | None
     multi_sparse: bool
     notebook_repr_html: bool
     pprint_nest_depth: int
     precision: int
-    show_dimensions: str
+    show_dimensions: bool | Literal["truncate"]
     unicode: DisplayUnicode
     width: int
 
--- a/tests/test_config.py
+++ b/tests/test_config.py
@@ -54,13 +54,13 @@ def test_display_float_format():
 def test_display_types_none_allowed_get_options():
     # GH 1230
     # Initial values
-    assert_type(pd.options.display.chop_threshold, Optional[float])
-    assert_type(pd.options.display.max_columns, Optional[int])
-    assert_type(pd.options.display.max_colwidth, Optional[int])
-    assert_type(pd.options.display.max_dir_items, Optional[int])
-    assert_type(pd.options.display.max_rows, Optional[int])
-    assert_type(pd.options.display.max_seq_items, Optional[int])
-    assert_type(pd.options.display.min_rows, Optional[int])
+    check(assert_type(pd.options.display.chop_threshold, Optional[float]), type(None))
+    check(assert_type(pd.options.display.max_columns, Optional[int]), int)
+    check(assert_type(pd.options.display.max_colwidth, Optional[int]), int)
+    check(assert_type(pd.options.display.max_dir_items, Optional[int]), int)
+    check(assert_type(pd.options.display.max_rows, Optional[int]), int)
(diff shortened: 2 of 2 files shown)


tests/test_config.py
```diff
@@ -48,3 +48,46 @@ def test_display_float_format():
...
+        "min_rows",
+    ]
+    for prop in properties_to_test:
+        # Test with None
+        setattr(pd.options.display, prop, None)
+        check(assert_type(pd.get_option(f"display.{prop}"), Any), type(None))
+
+        # Test with integer values
+        setattr(pd.options.display, prop, 100)
+        check(assert_type(pd.get_option(f"display.{prop}"), Any), int)
```
Dr-Irv, line 70: For testing the types,  you can't use a loop.  You should have an explicit line with each of the items.  e.g.  something like (for the first option):
```python
    pd.options.display.max_columns = None
    check(assert_type(pd.options.display.max_columns, Union[int, None]), type(None))
    pd.options.display.max_columns = 100
    check(assert_type(pd.options.display.max_columns, Union[int, None]), int)

```

tests/test_config.py
```diff
@@ -48,3 +48,46 @@ def test_display_float_format():
...
+    # Test memory_usage with allowed values
+    pd.options.display.memory_usage = True
+    pd.options.display.memory_usage = False
+    pd.options.display.memory_usage = "deep"
+    pd.options.display.memory_usage = None
+
+    # Test show_dimensions with allowed values
+    pd.options.display.show_dimensions = True
+    pd.options.display.show_dimensions = False
+    pd.options.display.show_dimensions = "truncate"
```
Dr-Irv, line 93: For each of these, you should do a `check(assert_type(...))` after setting the option to make sure you get the same value back and that the type matches.

Dr-Irv (changes requested):
Changes in tests.  I don't have time right now to verify the changes to the types in `Display` - will do that in next review.

tests/test_config.py
mike-flowers-airbnb: Thanks @Dr-Irv , I flattened the loop so that's resolved. Unfortunately, I've run into two issues with the current proposed testing.
1) The type identified in `assert_type` is strictly the type that is being set rather than all possible values (which makes sense as it knows the exact value at that point in time
2) For the integer checks there's differences here between `pyright` and `mypy`. `mypy` expects the type to be `int` but `pyright` expect it to be `Literal[100]`. I'm honestly not sure how to handle this situation. Any ideas?

tests/test_config.py
mike-flowers-airbnb: Added these, also added a `check(assert_type(...))` before each to ensure all potential types are visible.

tests/test_config.py
Dr-Irv: `pyright` is doing bi-directional type inference.  I think if you separate the statements into 2 separate test functions, it will work:
```python
def test_get_options():
    check(assert_type(pd.options.display.max_columns, Union[int, None]), int)
    # Add other options here

def test_set_options():
    pd.options.display.max_columns = None
    pd.options.display.max_columns = 100
    # Add other options here
```

pandas-stubs/_config/config.pyi
```diff
@@ -49,30 +49,30 @@ class DisplayUnicode(DictWrapper):
 
 class Display(DictWrapper):
     chop_threshold: int | None
```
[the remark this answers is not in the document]
mike-flowers-airbnb: Absolutely right, fixed, thanks!

tests/test_config.py
mike-flowers-airbnb: Thanks @Dr-Irv , split the test into separate ones and both are passing now. Let me know if this is in line with what you were thinking.

Dr-Irv (changes requested):

Dr-Irv (approved):
thanks @mike-flowers-airbnb

PR #1232, opened 2025-05-29
Use `default` in `TypeVar` so `Series` defaults to `Series[Any]`, and `Index` to `Index[Any]`

### LINKED ISSUE #1133: Should `Series[Any]` be used internally instead of `Series`?

MarcoGorelli:
bit of an experiment

closes #1133

I saw that NumPy follow this pattern, for example here:

https://github.com/numpy/numpy/blob/3363b38d99f87601869c2c9d68c2b16e54509675/numpy/__init__.pyi#L788-L792

and they import `TypeVar` from `typing_extensions` (which is a dependency of mypy anyway)

This would close 1133 for free, and I think improve ergonomics


Dr-Irv:
I'm fine with this.  Need to run `black`.  The new `default` for `TypeVar` is in python 3.13, so that's why you have to import `TypeVar` from `typing_extensions`

MarcoGorelli:
cool, thanks! there's a couple of test failures I haven't figured out, will investigate
```
tests/test_scalars.py:1382: error: Expression is of type "Any", not "ndarray[tuple[int, ...], dtype[numpy.bool[builtins.bool]]]"  [assert-type]
tests/test_scalars.py:1385: error: Expression is of type "Any", not "ndarray[tuple[int, ...], dtype[numpy.bool[builtins.bool]]]"  [assert-type]
```

pandas-stubs/core/groupby/generic.pyi
```diff
@@ -168,7 +168,7 @@ class SeriesGroupBy(GroupBy[Series[S1]], Generic[S1, ByT]):
         self,
     ) -> Iterator[tuple[ByT, Series[S1]]]: ...
 
-_TT = TypeVar("_TT", bound=Literal[True, False])
+_TT = TypeVar("_TT", bound=Literal[True, False], default=Literal[True])
```
Dr-Irv, line 171: I'm a bit surprised that you didn't have to change the `import` of `TypeVar` to come from `typing_extensions` here since you are using the `default` feature.

Dr-Irv (commented):
Looks OK so far.

tests/test_frame.py
```diff
@@ -3102,29 +3102,29 @@ def test_frame_stack() -> None:
...
-                df_multi_level_cols2.stack(0), Union[pd.DataFrame, "pd.Series[Any]"]
+                df_multi_level_cols2.stack(0), Union[pd.DataFrame, "pd.Series"]
             ),
             pd.DataFrame,
         )
         check(
             assert_type(
                 df_multi_level_cols2.stack([0, 1]),
-                Union[pd.DataFrame, "pd.Series[Any]"],
+                Union[pd.DataFrame, "pd.Series"],
```
Dr-Irv: Any of the places where we used to have `"pd.Series[Any]"` that you changed to `"pd.Series"` can have the quotes removed.  The quotes were there because of the `[Any]` part.

pandas-stubs/core/frame.pyi
```diff
@@ -1684,6 +1686,8 @@ class DataFrame(NDFrame, OpsMixin, _GetItemHack):
     @property
     def iloc(self) -> _iLocIndexerFrame[Self]: ...
     @property
+    # mypy complains if we use Index[Any] instead of UnknownIndex here, even though
+    # the latter is aliased to the former ¯\_(ツ)_/¯.
     def index(self) -> UnknownIndex: ...
```
MarcoGorelli, line 1691: this is a complete mystery to me - any ideas?

Dr-Irv: What if you just used `Index` ?

MarcoGorelli: same
```
tests/test_scalars.py:1380: error: Expression is of type "Any", not "ndarray[tuple[int, ...], dtype[numpy.bool[builtins.bool]]]"  [assert-type]
tests/test_scalars.py:1383: error: Expression is of type "Any", not "ndarray[tuple[int, ...], dtype[numpy.bool[builtins.bool]]]"  [assert-type]
```

the only difference i applied was
```diff
--- a/pandas-stubs/core/frame.pyi
+++ b/pandas-stubs/core/frame.pyi
@@ -1688,7 +1688,7 @@ class DataFrame(NDFrame, OpsMixin, _GetItemHack):
     @property
     # mypy complains if we use Index[Any] instead of UnknownIndex here, even though
     # the latter is aliased to the former ¯\_(ツ)_/¯.
-    def index(self) -> UnknownIndex: ...
+    def index(self) -> Index: ...
```

Dr-Irv: Maybe leave it as `Index` in `frame.pyi` there, but change the `__eq__()` and `__ne__()` methods in `timestamps.pyi` to use `Index[Any]` instead of `UnknownIndex`

MarcoGorelli: Not sure I follow, sorry
- in `timestamps.pyi` it's already using `Index` in this PR (not `UnknownIndex`)
- if I leave it as `Index` in `frame.pyi`, then mypy complains (though pyright is fine with it)

Dr-Irv: I was suggesting using `Index` in `frame.pyi` and `Index[Any]` in the `__eq__()` and `__ne__()` methods in `timestamps.pyi`

The 2 lines that are failing are related to those 2 methods.

MarcoGorelli: Thanks - I've tried applying the following diff:
```diff
diff --git a/pandas-stubs/_libs/tslibs/timestamps.pyi b/pandas-stubs/_libs/tslibs/timestamps.pyi
index 795f98a..5a026c2 100644
--- a/pandas-stubs/_libs/tslibs/timestamps.pyi
+++ b/pandas-stubs/_libs/tslibs/timestamps.pyi
@@ -14,6 +14,7 @@ from typing import (
     Literal,
     SupportsIndex,
     overload,
+    Any,
 )
 
 import numpy as np
@@ -236,7 +237,7 @@ class Timestamp(datetime, SupportsIndex):
     @overload
     def __eq__(self, other: TimestampSeries) -> Series[bool]: ...  # type: ignore[overload-overlap]
     @overload
-    def __eq__(self, other: npt.NDArray[np.datetime64] | Index) -> np_ndarray_bool: ...  # type: ignore[overload-overlap]
+    def __eq__(self, other: npt.NDArray[np.datetime64] | Index[Any]) -> np_ndarray_bool: ...  # type: ignore[overload-overlap]
     @overload
     def __eq__(self, other: object) -> Literal[False]: ...
     @overload
@@ -244,7 +245,7 @@ class Timestamp(datetime, SupportsIndex):
     @overload
     def __ne__(self, other: TimestampSeries) -> Series[bool]: ...  # type: ignore[overload-overlap]
     @overload
-    def __ne__(self, other: npt.NDArray[np.datetime64] | Index) -> np_ndarray_bool: ...  # type: ignore[overload-overlap]
+    def __ne__(self, other: npt.NDArray[np.datetime64] | Index[Any]) -> np_ndarray_bool: ...  # type: ignore[overload-overlap]
     @overload
     def __ne__(self, other: object) -> Literal[True]: ...
     def __hash__(self) -> int: ...
diff --git a/pandas-stubs/core/frame.pyi b/pandas-stubs/core/frame.pyi
index 2a9875d..b7ca108 100644
--- a/pandas-stubs/core/frame.pyi
+++ b/pandas-stubs/core/frame.pyi
@@ -1688,7 +1688,7 @@ class DataFrame(NDFrame, OpsMixin, _GetItemHack):
     @property
     # mypy complains if we use Index[Any] instead of UnknownIndex here, even though
     # the latter is aliased to the former ¯\_(ツ)_/¯.
-    def index(self) -> UnknownIndex: ...
+    def index(self) -> Index: ...
     @index.setter
     def index(self, idx: Index) -> None: ...
     @property
```
is this what you meant? Because then, the problem persists

Dr-Irv: > is this what you meant? Because then, the problem persists

yes, that's what I meant.  I guess we have to keep `UnknownIndex` around.

So is the current version ready for final review?

MarcoGorelli: yup, thanks!

Dr-Irv (approved):
thanks @MarcoGorelli .  Nice new feature of typing to include.

MarcoGorelli:
yay, thanks!

PR #1233, opened 2025-05-29
Remove some unnecessary S1 TypeVars

MarcoGorelli:
While exploring #1232 I noticed a few places where `S1` is used unnecessarily, because it only appears in one argument and in a method of a class that's not generic in `S1`


Dr-Irv (approved):
Thanks @MarcoGorelli

PR #1237, opened 2025-05-30
GH1159 Disallow StringMethods for Series and Index when subtype is not str

### LINKED ISSUE #1159: Disallow StringMethods for Series and Index when subtype is not `str`, but allow for `Any`
In #1146 tried to get `StringMethods` to know that passing `Series[float]` (for example) is wrong.  

I got something to work for `Series` with `pyright` .  See https://github.com/Dr-Irv/pandas-stubs/tree/pr1146 and this commit: https://github.com/pandas-dev/pandas-stubs/commit/164bc27185bf213fd7da942414bf2e2935bf8a1d

Would like to add this to the stubs.

But as noted here:  https://github.com/pandas-dev/pandas-stubs/pull/1146#issuecomment-2713697759 it fails in `mypy` which seems to be due to https://github.com/python/mypy/issues/15921

loicdiridollou:
- [x] Closes #1159

### DIFF
--- a/pandas-stubs/core/frame.pyi
+++ b/pandas-stubs/core/frame.pyi
@@ -256,9 +256,9 @@ class _LocIndexerFrame(_LocIndexer, Generic[_T]):
             ]
             | None
         ),
-    ) -> Series: ...
+    ) -> UnknownSeries: ...
     @overload
-    def __getitem__(self, idx: tuple[Scalar, slice]) -> Series | _T: ...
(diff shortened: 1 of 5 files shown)


loicdiridollou:
Work is mostly inspired by @Dr-Irv contributions, I am going to investigate why it is failing on 3.10 and 3.11 but not on 3.12 and 3.13.

loicdiridollou:
Will close this one for now, haven't been able to carve out time for it.

PR #1240, opened 2025-06-05
GH1222 Candidate for ty on pandas-stubs

### LINKED ISSUE #1222: Test ty on the CI
If [ty](https://github.com/astral-sh/ty) becomes as popular as ruff, it would be good to test against ty on the CI.

Since ty is still in an alpha stage, this should at most be an optional check for now.

loicdiridollou:
- [x] Closes #1222 partially

Proposal to have ty on pandas-stubs, on tests there are still a lot of errors that come back as TODO from ty and the inference is a bit unclear for many tests (or it comes back as Unknown so would need more investigation and learning on my end).


pandas-stubs/_typing.pyi
```diff
@@ -513,7 +514,7 @@ IndexKeyFunc: TypeAlias = Callable[[Index], Index | AnyArrayLike] | None
 
 # types of `func` kwarg for DataFrame.aggregate and Series.aggregate
 # More specific than what is in pandas
-AggFuncTypeBase: TypeAlias = Callable | str | np.ufunc
+AggFuncTypeBase: TypeAlias = Union[Callable, str, np.ufunc]  # noqa: UP007
```
loicdiridollou, line 518: Here it complains at a later line that it does not accept the `X | Y` for in `list`, will open a bug in `ty`

pandas-stubs/core/arrays/datetimes.pyi
```diff
@@ -28,7 +28,7 @@ class DatetimeArray(DatetimeLikeArrayMixin, TimelikeOps, DatelikeOps):
     @tz.setter
     def tz(self, value) -> None: ...
     @property
-    def tzinfo(self) -> tzinfo | None: ...
+    def tzinfo(self) -> _tzinfo | None: ...
```
loicdiridollou, line 31: The type `tzinfo` conflicts with the property.

pandas-stubs/core/groupby/groupby.pyi
```diff
@@ -393,3 +330,66 @@ class GroupBy(BaseGroupBy[NDFrameT]):
         weights: Sequence | Series | None = ...,
         random_state: RandomState | None = ...,
     ) -> NDFrameT: ...
+
+_GroupByT = TypeVar("_GroupByT", bound=GroupBy)
```
loicdiridollou, line 334: Just an ordering issue here

pandas-stubs/core/indexes/base.pyi
```diff
@@ -272,10 +272,10 @@ class Index(IndexOpsMixin[S1]):
         Self,
         MultiIndex,
         np_ndarray_bool,
-        Index[list[str]],
+        Index[list[_str]],
```
loicdiridollou, line 275: Same here, type `str` conflicts with the property.

pandas-stubs/core/series.pyi
```diff
@@ -158,7 +158,7 @@ from pandas._typing import (
     StrDtypeArg,
     StrLike,
     Suffixes,
-    T,
+    T as _T,
```
loicdiridollou, line 161: Same here property `T` for transpose conflicts with type `T`.

pandas-stubs/io/formats/style_render.pyi
```diff
@@ -46,12 +50,12 @@ CSSStyles: TypeAlias = list[CSSDict]
 Subset: TypeAlias = _IndexSlice | slice | tuple[slice, ...] | list[HashableT] | Index
 
 class StylerRenderer:
-    loader: jinja2.loaders.PackageLoader
```
loicdiridollou, line 49: It was raising an `unresolved-attribute` error here, importing directly seems to have fixed the issue.

.github/workflows/test.yml
```diff
@@ -35,6 +35,9 @@ jobs:
       - name: Run mypy on 'tests' (using the local stubs) and on the local stubs
         run: poetry run poe mypy
 
+      - name: Run ty on 'pandas-stubs' (using the local stubs) and on the local stubs
```
loicdiridollou, line 38: Currently running it as part of the CI but could be a blocker if the project makes a lot of changes or if we bump into something they don't handle yet as it is still an alpha.
We could make it part of the optional tests so we see it at merge and we keep track of the evolution like we do for nightly.

pandas-stubs/_typing.pyi
```diff
@@ -463,6 +464,10 @@ VoidDtypeArg: TypeAlias = (
     | Literal["V", "void", "void0"]
 )
 
+# DtypeArg specifies all allowable dtypes in a functions its dtype argument
+DtypeArg: TypeAlias = Dtype | Mapping[Hashable, Dtype]
+DtypeObj: TypeAlias = np.dtype[np.generic] | ExtensionDtype
+
```
Dr-Irv, line 470: Why move this?  I want to keep the order of things close to what is in the pandas source.

pandas-stubs/core/arrays/datetimes.pyi
Dr-Irv: OK - that's a reasonable change

pandas-stubs/core/generic.pyi
```diff
@@ -69,8 +69,8 @@ class NDFrame(indexing.IndexingMixin):
     def set_flags(
         self,
         *,
-        copy: bool = ...,
-        allows_duplicate_labels: bool | None = ...,
+        copy: _bool = ...,
+        allows_duplicate_labels: _bool | None = ...,
```
Dr-Irv, line 73: I don't see why this change is necessary.

pandas-stubs/core/indexes/base.pyi
Dr-Irv: ok

pandas-stubs/core/series.pyi
Dr-Irv: OK

pandas-stubs/core/series.pyi
```diff
@@ -213,7 +213,7 @@ class _LocIndexerSeries(_LocIndexer, Generic[S1]):
         idx: (
             MaskType
             | Index
-            | SequenceNotStr[float | str | Timestamp]
+            | SequenceNotStr[float | _str | Timestamp]
```
Dr-Irv, line 216: in lines 182-183, we should declare `_bool` and `_str` as `TypeAlias`

pandas-stubs/core/strings.pyi
```diff
@@ -46,12 +46,14 @@ _T_STR = TypeVar("_T_STR", bound=Series[str] | Index[str])
 # Used for the result of str.partition
 _T_OBJECT = TypeVar("_T_OBJECT", bound=Series[type[object]] | Index[type[object]])
 
+_slice = slice
```
Dr-Irv:
```suggestion
_slice: TypeAlias = slice
```

pandas-stubs/io/formats/style_render.pyi
Dr-Irv: OK

pandas-stubs/_typing.pyi
loicdiridollou: AsTypeArg uses the DtypeObj which is defined after it so ty was complained that it was not defined so I just reordered the definition of DtypeObj above AsTypeArg.

pandas-stubs/core/generic.pyi
loicdiridollou: Book is a method for NDFrame so it was confusing ty to have the method and the type named the same (see line 90)

pandas-stubs/core/series.pyi
loicdiridollou: Will fix those

loicdiridollou:
Agreed with adding it to the CI, for how it would only be for pandas-stubs and not the tests as the project is still experimental there is a lot to do and understand on the tests side. Worst case if we see it as a blocker on velocity we can bring it to merge only. Let me do the changes

pandas-stubs/_typing.pyi
Dr-Irv: can you add a comment about that, because I'd prefer we don't use `Union` in `_typing.pyi`

pandas-stubs/core/groupby/groupby.pyi
Dr-Irv: ok

pandas-stubs/_typing.pyi
Dr-Irv: ok

pandas-stubs/core/generic.pyi
Dr-Irv: ok - I got confused - I was looking at the source for the wrong `generic.pyi`

pandas-stubs/_typing.pyi
loicdiridollou: Yes I will add the issue comment when I open the issue on the ty repo so we can track it (totally agree it is not pretty).

pandas-stubs/core/series.pyi
loicdiridollou: This is addressed.

Dr-Irv (approved):
Thanks @loicdiridollou  you'll have to keep an eye on version improvements on `ty` so that `pyproject.toml` gets properly updated.

PR #1242, opened 2025-06-06
GH456 First attempt GroupBy.transform improved typing

### LINKED ISSUE #456: Create more precise annotations for DataFrame/Series GroupBy operations (agg, apply, transform)

loicdiridollou:
- [x] Closes #456

### DIFF
--- a/pandas-stubs/_typing.pyi
+++ b/pandas-stubs/_typing.pyi
@@ -925,44 +925,6 @@ GroupByObjectNonScalar: TypeAlias = (
     | list[Grouper]
 )
 GroupByObject: TypeAlias = Scalar | Index | GroupByObjectNonScalar | Series
-GroupByFuncStrs: TypeAlias = Literal[
-    # Reduction/aggregation functions
-    "all",
-    "any",
-    "corrwith",
-    "count",
-    "first",
-    "idxmax",
-    "idxmin",
-    "last",
-    "max",
-    "mean",
-    "median",
-    "min",
-    "nunique",
-    "prod",
-    "quantile",
-    "sem",
-    "size",
-    "skew",
-    "std",
-    "sum",
-    "var",
-    # Transformation functions
-    "bfill",
-    "cumcount",
-    "cummax",
-    "cummin",
-    "cumprod",
-    "cumsum",
-    "diff",
-    "ffill",
-    "fillna",
-    "ngroup",
-    "pct_change",
-    "rank",
-    "shift",
-]
 
 StataDateFormat: TypeAlias = Literal[
     "tc",
--- a/pandas-stubs/_typing.pyi
+++ b/pandas-stubs/_typing.pyi
@@ -925,6 +925,44 @@ GroupByObjectNonScalar: TypeAlias = (
     | list[Grouper]
 )
 GroupByObject: TypeAlias = Scalar | Index | GroupByObjectNonScalar | Series
+GroupByFuncStrs: TypeAlias = Literal[
+    # Reduction/aggregation functions
+    "all",
+    "any",
+    "corrwith",
+    "count",
+    "first",
+    "idxmax",
+    "idxmin",
+    "last",
+    "max",
+    "mean",
+    "median",
+    "min",
+    "nunique",
+    "prod",
+    "quantile",
+    "sem",
+    "size",
+    "skew",
+    "std",
+    "sum",
+    "var",
+    # Transformation functions
+    "bfill",
+    "cumcount",
+    "cummax",
+    "cummin",
+    "cumprod",
+    "cumsum",
+    "diff",
+    "ffill",
+    "fillna",
+    "ngroup",
+    "pct_change",
+    "rank",
+    "shift",
+]
 
 StataDateFormat: TypeAlias = Literal[
     "tc",
--- a/pandas-stubs/core/groupby/base.pyi
+++ b/pandas-stubs/core/groupby/base.pyi
@@ -1,7 +1,56 @@
 from collections.abc import Hashable
 import dataclasses
+from typing import (
+    Literal,
+    TypeAlias,
+)
 
 @dataclasses.dataclass(order=True, frozen=True)
 class OutputKey:
     label: Hashable
     position: int
+
+reduction_kernels: TypeAlias = Literal[
+    "all",
+    "any",
+    "corrwith",
+    "count",
+    "first",
+    "idxmax",
+    "idxmin",
+    "last",
+    "max",
+    "mean",
+    "median",
+    "min",
+    "nunique",
+    "prod",
+    # as long as `quantile`'s signature accepts only
+    # a single quantile value, it's a reduction.
+    # GH#27526 might change that.
+    "quantile",
+    "sem",
+    "size",
+    "skew",
+    "std",
+    "sum",
+    "var",
+]
+
+transformation_kernels: TypeAlias = Literal[
+    "bfill",
+    "cumcount",
+    "cummax",
+    "cummin",
+    "cumprod",
+    "cumsum",
+    "diff",
+    "ffill",
+    "fillna",
+    "ngroup",
+    "pct_change",
+    "rank",
+    "shift",
+]
+
+transform_kernel_allowlist: TypeAlias = reduction_kernels | transformation_kernels
--- a/pandas-stubs/core/groupby/base.pyi
+++ b/pandas-stubs/core/groupby/base.pyi
@@ -10,7 +10,7 @@ class OutputKey:
     label: Hashable
     position: int
 
-reduction_kernels: TypeAlias = Literal[
+ReductionKernelType: TypeAlias = Literal[
     "all",
     "any",
     "corrwith",
@@ -37,7 +37,7 @@ reduction_kernels: TypeAlias = Literal[
     "var",
 ]
 
-transformation_kernels: TypeAlias = Literal[
+TransformationKernelType: TypeAlias = Literal[
     "bfill",
     "cumcount",
     "cummax",
@@ -53,4 +53,4 @@ transformation_kernels: TypeAlias = Literal[
     "shift",
 ]
 
-transform_kernel_allowlist: TypeAlias = reduction_kernels | transformation_kernels
+TransformReductionListType: TypeAlias = ReductionKernelType | TransformationKernelType
--- a/pandas-stubs/core/groupby/generic.pyi
+++ b/pandas-stubs/core/groupby/generic.pyi
@@ -73,6 +73,15 @@ class SeriesGroupBy(GroupBy[Series[S1]], Generic[S1, ByT]):
     @overload
     def aggregate(
         self,
+        func: Callable[[Series], S2],
+        *args,
+        engine: WindowingEngine = ...,
+        engine_kwargs: WindowingEngineKwargs = ...,
+        **kwargs,
+    ) -> Series[S2]: ...
+    @overload
+    def aggregate(
+        self,
         func: list[AggFuncTypeBase],
         /,
         *args,
--- a/pandas-stubs/core/groupby/generic.pyi
+++ b/pandas-stubs/core/groupby/generic.pyi
@@ -7,6 +7,7 @@ from collections.abc import (
 )
 from typing import (
     Any,
+    Concatenate,
     Generic,
     Literal,
     NamedTuple,
@@ -31,15 +32,18 @@ from typing_extensions import (
 from pandas._libs.tslibs.timestamps import Timestamp
 from pandas._typing import (
     S1,
+    S2,
     AggFuncTypeBase,
     AggFuncTypeFrame,
     ByT,
     CorrelationMethod,
     Dtype,
+    GroupByFuncStrs,
     IndexLabel,
     Level,
     ListLike,
     NsmallestNlargestKeep,
+    P,
     Scalar,
     TakeIndexer,
     WindowingEngine,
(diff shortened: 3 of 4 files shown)


pandas-stubs/_typing.pyi
```diff
@@ -925,6 +925,44 @@ GroupByObjectNonScalar: TypeAlias = (
     | list[Grouper]
 )
 GroupByObject: TypeAlias = Scalar | Index | GroupByObjectNonScalar | Series
+GroupByFuncStrs: TypeAlias = Literal[
```
loicdiridollou: We can't import `transform_kernel_allowlist` from `pandas.core.groupby.base` so redefining it here.

pandas-stubs/core/groupby/generic.pyi
```diff
@@ -72,14 +76,24 @@ class SeriesGroupBy(GroupBy[Series[S1]], Generic[S1, ByT]):
         **kwargs,
     ) -> Series: ...
     agg = aggregate
+    @overload
     def transform(
         self,
-        func: Callable | str,
-        *args,
+        func: Callable[Concatenate[Series[S1], P], Series[S2]],
+        *args: Any,
```
loicdiridollou, line 108: Can only define it as `Any` here, can't do `P.args` since engine argument gets in the middle.

loicdiridollou: Could work if we rearranged the order, not sure if it is doable, open to suggestion.

Dr-Irv: I think if you put a `/` before engine, then it knows that `engine` has to be a keyword argument (which is true)

pandas-stubs/core/groupby/generic.pyi
```diff
@@ -72,14 +76,24 @@ class SeriesGroupBy(GroupBy[Series[S1]], Generic[S1, ByT]):
...
-        **kwargs,
+        **kwargs: Any,
+    ) -> Series[S2]: ...
+    @overload
+    def transform(
+        self,
+        func: Callable,
+        *args: Any,
+        **kwargs: Any,
     ) -> Series: ...
```
Dr-Irv: should this be `UnknownSeries` ?

pandas-stubs/core/groupby/generic.pyi
```diff
@@ -72,14 +76,24 @@ class SeriesGroupBy(GroupBy[Series[S1]], Generic[S1, ByT]):
...
+    @overload
+    def transform(self, func: GroupByFuncStrs, *args, **kwargs) -> Series: ...
```
Dr-Irv: Should this be `UnknownSeries` ?

pandas-stubs/core/groupby/generic.pyi
```diff
@@ -206,14 +220,24 @@ class DataFrameGroupBy(GroupBy[DataFrame], Generic[ByT, _TT]):
...
     ) -> DataFrame: ...
     agg = aggregate
+    @overload
     def transform(
         self,
-        func: Callable | str,
-        *args,
+        func: Callable[Concatenate[DataFrame, P], DataFrame],
+        *args: Any,
         engine: WindowingEngine = ...,
```
Dr-Irv, line 255: same comment - add the `/` before `engine`

tests/test_series.py
```diff
@@ -1084,6 +1084,24 @@ def test_types_groupby_agg() -> None:
...
+    def transform_func(
+        x: pd.Series[int], pos_arg: bool, kw_arg: str
+    ) -> pd.Series[float]:
+        return x / (2.0 if pos_arg else 1.0)
+
+    check(
+        assert_type(
+            s.groupby(lambda x: x).transform(transform_func, True, kw_arg="foo"),
+            "pd.Series[float]",
+        ),
```
Dr-Irv, line 1108: add a test where `engine` is specified.

pandas-stubs/_typing.pyi
Dr-Irv: I think you should just copy what is in the source `pandas.core.groupby.base.py` into `pandas.core.groupby.base.pyi` so they are equivalent, but change the `frozenset` to be `Literal` 

Then don't add anything to `_typing.pyi`

pandas-stubs/core/groupby/generic.pyi
loicdiridollou: Yes indeed, about the `/` shouldn't it be before `*args`? I get lint error if I put it after it.

pandas-stubs/core/groupby/generic.pyi
loicdiridollou: Yes, fixed

loicdiridollou:
For 
```
I think you should just copy what is in the source pandas.core.groupby.base.py into pandas.core.groupby.base.pyi so they are equivalent, but change the frozenset to be Literal
```
is there a convention to name the TypeAlias or should we keep the same name as in the pandas code?

Dr-Irv:
> For
> 
> ```
> I think you should just copy what is in the source pandas.core.groupby.base.py into pandas.core.groupby.base.pyi so they are equivalent, but change the frozenset to be Literal
> ```
> 
> is there a convention to name the TypeAlias or should we keep the same name as in the pandas code?

Probably shouldn't use the same name, because then the types won't match.  Not that anyone should be importing that anyway.  So keep the same name and append `_type` after the name.

pandas-stubs/core/groupby/base.pyi
```diff
@@ -1,7 +1,56 @@
...
+    Literal,
+    TypeAlias,
+)
 
 @dataclasses.dataclass(order=True, frozen=True)
 class OutputKey:
     label: Hashable
     position: int
+
+reduction_kernels: TypeAlias = Literal[
```
Dr-Irv: as commented elsewhere, lets call this `reduction_kernels_type` so that the types of the objects line up.

tests/test_series.py
```diff
@@ -1078,25 +1078,90 @@ def test_types_groupby_agg() -> None:
         r"The provided callable <built-in function (min|sum)> is currently using",
         upper="2.2.99",
     ):
-        check(assert_type(s.groupby(level=0).agg(sum), pd.Series), pd.Series)
+
+        def sum_sr(s: pd.Series[int]) -> int:
+            # type of `sum` not well inferred by mypy
+            return sum(s)
```
Dr-Irv: why not use `s.sum()` ?

tests/test_series.py
```diff
@@ -1078,25 +1078,90 @@ def test_types_groupby_agg() -> None:
...
     check(
         assert_type(s.groupby(level=0).aggregate(["min", "sum"]), pd.DataFrame),
         pd.DataFrame,
     )
+
+    def func(s: pd.Series[int]) -> float:
+        return s.astype(float).min()
+
+    s = pd.Series([1, 2, 3, 4])
+    s.groupby([1, 1, 2, 2]).agg(lambda x: x.astype(float).min())
```
Dr-Irv: don't you want a `check(assert_type(...` here?

tests/test_series.py
```diff
@@ -1078,25 +1078,90 @@ def test_types_groupby_agg() -> None:
...
     with pytest_warns_bounded(
         FutureWarning,
         r"The provided callable <built-in function (min|sum)> is currently using",
         upper="2.2.99",
     ):
-        check(assert_type(s.groupby(level=0).aggregate(sum), pd.Series), pd.Series)
+
+        def sum_sr(s: pd.Series[int]) -> int:
+            # type of `sum` not well inferred by mypy
+            return sum(s)
```
Dr-Irv: use `s.sum()`

Dr-Irv (changes requested):

pandas-stubs/core/groupby/generic.pyi
Dr-Irv: OK - I wasn't sure

tests/test_series.py
loicdiridollou: The issue was if I passed `sum` the builtin one in the aggregate method directly, here is does not change anything.

tests/test_series.py
loicdiridollou: Correct my mistake

loicdiridollou: Turns out the inference on the fly of lambdas is not super clear so you need to define the function on the side to have the right types.

tests/test_series.py
```diff
@@ -1078,25 +1078,94 @@ def test_types_groupby_agg() -> None:
...
+            assert_type(s.groupby(level=0).agg(sum_sr), "pd.Series[int]"),
+            pd.Series,
+            np.integer,
+        )
         check(
             assert_type(s.groupby(level=0).agg([min, sum]), pd.DataFrame), pd.DataFrame
         )
 
 
+def test_types_groupby_transform() -> None:
```
Dr-Irv, line 1096: I think you should add tests for  two of the string transform arguments (e.g., "mean", "first")

Dr-Irv (changes requested):

tests/test_series.py
Dr-Irv: Yes, that is an issue with lambda functions.

Dr-Irv: Actually, I think you can have a test of 
```python
check(assert_type(    s.groupby([1, 1, 2, 2]).agg(lambda x: x.astype(float).min()), pd.Series), pd.Series)
``` 
which would be worthwhile

pandas-stubs/core/groupby/generic.pyi
```diff
@@ -70,16 +84,18 @@ class SeriesGroupBy(GroupBy[Series[S1]], Generic[S1, ByT]):
     def aggregate(
         self,
         func: AggFuncTypeBase | None = ...,
+        /,
         *args,
```
Dr-Irv, line 97: Before this overload, you could add this overload:
```python
    @overload
    def aggregate(
        self,
        func: Callable[[Series], S2],
        *args,
        engine: WindowingEngine = ...,
        engine_kwargs: WindowingEngineKwargs = ...,
        **kwargs,
    ) -> Series[S2]: ...
```
Then you know that if you start with a `Series` with a known type, then the return type would be inferred from the callable.  And it works with a lambda function, e.g.:
```python
    s = pd.Series([1, 2, 3, 4])
    q = s.groupby([1, 1, 2, 2]).agg(lambda x: x.astype(float).min())
```
In this case, `q` would have type `Series[float]`, which is what you want.

loicdiridollou: image
Even with the new overload mypy still complains (pyright does not). I think it does not recognize the lamda as returning S2 (float).

Dr-Irv: I think that's because the type of `new_func` isn't clear.

But I think it would work if you did `check(assert_type(s.groupby([1,1,2,2]).agg(lambda x: x.astype(float).min()), "pd.Series[int]"), pd.Series, int)`

Because then it can know that `x` is a `Series[int]` and that the `lambda` becomes `Series[int]`

Can you try that?

loicdiridollou: I tried that for the last push, see https://github.com/pandas-dev/pandas-stubs/blob/f9863d03c387dcca86da8f86675a105a5d837786/tests/test_series.py#L1167
It fails in all CI:
```

tests/test_series.py:1167: error: Expression is of type "Series[Any]", not "Series[float]"  [assert-type]
Found 1 error in 1 file (checked 224 source files)
```

loicdiridollou: When I look with how `mypy` reads the type of the lambda, it has no idea about the type of `x`:
```
tests/test_series.py:1168: note: Revealed type is "def (x: Any) -> Any"
```
so that may explain why it fails on lambda expressions whatsoever.

Dr-Irv: OK - so we can leave the `lambda` test in, but just have it `assert_type()` against `Series` instead of `Series[float]`

tests/test_series.py
```diff
@@ -1147,6 +1165,9 @@ def func(s: pd.Series[int]) -> float:
         np.floating,
     )
 
+    # test below passes with mypy but pyright correctly sees it as pd.Series[float]
+    # check(assert_type(s.groupby([1,1,2,2]).agg(lambda x: x.astype(float).min()), pd.Series), pd.Series, float)
```
Dr-Irv: Keep the commented test in there so it is still there and executes, since it works for both type checkers, but comment out the one that is "better" that has `pyright` infer it as `Series[float]`.

Dr-Irv (changes requested):

tests/test_series.py
loicdiridollou: I am forced to comment it out because `pyright` sees it as `pd.Series[float]` but `mypy` sees it as `pd.Series` so in both versions the CI will fail either for mypy or pyright step. What would you recommend doing?

Dr-Irv: I guess you have to keep it commented out.  Do you have a test like this that passes both checkers:
```python
func: Callable[[pd.Series], float] = lambda x: x.astype(float).min()
check(assert_type(s.groupby([1,1,2,2]).agg(func), "pd.Series[float]"), pd.Series, float)
```
So you can have the "preferred" version in there commented out, but I think the above test would pass both type checkers.

loicdiridollou: Actually that also fails to pass with mypy (pyright is fine with it).

Dr-Irv: I tried a bunch of ideas and couldn't get it to work.  It's probably a mypy bug, but I couldn't come up with a simple example that illustrates the problem.

tests/test_series.py
```diff
@@ -1078,25 +1078,111 @@ def test_types_groupby_agg() -> None:
...
+    )
+    check(
+        assert_type(
+            s.groupby(level=0).aggregate(func, engine="cython"), "pd.Series[float]"
+        ),
+        pd.Series,
+        np.floating,
+    )
+
+    # test below passes with mypy but pyright correctly sees it as pd.Series[float]
```
Dr-Irv: Just have to change the comment to say "fails with mypy"

Dr-Irv (changes requested):
Just have to adjust that comment, then we are good to go.

Dr-Irv (approved):
thanks @loicdiridollou

PR #1244, opened 2025-06-07
Replace aliases to builtins with imports

hamdanal:

These aliases are not needed as they are there to avoid name collisions with methods defined in the class. Name collisions can be avoided by importing these symbols from `builtins` under a different name. This has no functional change to type checkers but has two tiny advantages:
1. IDEs now display the type as builtin type instead of an alias that doesn't exist at runtime
2. Linters like ruff can now understand that the symbol used is exactly the builtin one and can detect more bad usages (see the comment below)

### DIFF
--- a/pandas-stubs/core/frame.pyi
+++ b/pandas-stubs/core/frame.pyi
@@ -1,3 +1,7 @@
+from builtins import (
+    bool as _bool,
+    str as _str,
+)
 from collections.abc import (
     Callable,
     Hashable,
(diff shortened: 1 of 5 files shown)


pandas-stubs/core/frame.pyi
```diff
@@ -2148,7 +2149,7 @@ class DataFrame(NDFrame, OpsMixin, _GetItemHack):
     def reindex_like(
         self,
         other: DataFrame,
-        method: _str | FillnaOptions | Literal["nearest"] | None = ...,
+        method: FillnaOptions | Literal["nearest"] | None = ...,
```
hamdanal, line 2152: ruff can now detect that `_str` shadows the literals we accept so it started complaining here

Dr-Irv: and this is a better change because the set of strings that are accepted is limited

PR #1257, opened 2025-06-19
fix for DataFrameGroupBy.value_counts

Dr-Irv:
Discovered that `DataFrameGroupBy.value_counts()` was not matching when you ask for a subset of the columns.

Due to `__getitem__()` having `bool` instead of the passed in value whether an index is included.

@loicdiridollou since you worked on this, can you review and approve, and then I will merge


loicdiridollou (approved):
Looks good to me, thanks @Dr-Irv

PR #1259, opened 2025-06-20
Update types for `Series.unstack`, `DataFrame.stack`, and `DataFrame.unstack`

### LINKED ISSUE #1256: Incorrect type for `Series.unstack`
**Describe the bug**

As noted in https://github.com/pandas-dev/pandas-stubs/issues/703, providing a list of ints or strings to `unstack` leads to a type error:

> I've identified a few places (see attachment). However, I am not sure the patch is correct/complete. It is interesting to see that `stack` is annotated with `level: Level | list[Level] = ...` while `unstack` has `level: Level = ...` while IMHO both should be `IndexLabel`.

It seems as though the fix in https://github.com/pandas-dev/pandas-stubs/pull/837 only touched `groupby`.

stephenlrandall:
- [x] Closes #1256
- [x] Tests added: One for `Series.unstack`, and modified the test for `DataFrame.unstack`


tests/test_series.py
```diff
@@ -3902,11 +3902,22 @@ def test_series_items() -> None:
...
     check(
         assert_type(pd.Timestamp(0) + s.cumsum(), "TimestampSeries"),
         pd.Series,
         pd.Timestamp,
     )
+
+
+def test_series_unstack() -> None:
+    df = pd.DataFrame([[1, 3, 5], [2, 4, 6]])
+    s = df.transpose().stack([*range(df.index.nlevels)])
```
Dr-Irv, line 3916: Can you do `check(assert_type(s, pd.Series), pd.Series, np.int64)` here?  (I think that's the right type)

Dr-Irv (changes requested):

tests/test_series.py
stephenlrandall: That fails with:

```
tests/test_series.py:3917: error: Expression is of type "DataFrame | Series[Any]", not "Series[Any]"  [assert-type]
Found 1 error in 1 file (checked 224 source files)
```

What I can do instead is:

```python
def test_series_unstack() -> None:
    df = pd.DataFrame([[1, 3, 5], [2, 4, 6]])
    s = cast(pd.Series, df.transpose().stack([*range(df.index.nlevels)]))
    check(assert_type(s, pd.Series), pd.Series, np.int64)
    check(
        assert_type(
            s.unstack([*range(s.index.nlevels // 2)]),
            pd.DataFrame,
        ),
        pd.DataFrame,
    )
```

where I explicitly cast the stacked object to a `Series`. Or would you rather have something else there?

Dr-Irv: Replace `s = cast(pd.Series, df.transpose().stack([*range(df.index.nlevels)]))` with 

```python
s = df.transpose().stack([*range(df.index.nlevels)]))
check(assert_type(s, Union[pd.Series, pd.DataFrame), pd.Series)
```

This then says from a typing perspective, you could get the `Union`, but at runtime, you will get a `Series`

stephenlrandall: Makes sense -- updated the test.

Dr-Irv (approved):
thanks @stephenlrandall

PR #1250, opened 2025-06-13
GH1248/GH1249 DataFrame.assign and MultiIndex.from_product Series/Index

loicdiridollou:
- [x] Closes #1248 and #1249
- [x] Tests added: Please use `assert_type()` to assert the type of any return value

PR #1260, opened 2025-06-23
fix: assign Sequences

cmp0xff:
# Describe the bug

Following #1250, `DataFrame.assign()` should allow `Sequence`s, e.g. `tuple`s

# To Reproduce

```py
import pandas as pd
df = pd.DataFrame({"a": (1,2,3)})
df = df.assign(d=lambda df: (10,20,30))
```
 
Need to update `IntoColumn` in `_typing.pyi` to include `Sequence[Scalar]`

# Please complete the following information
- OS: Windows 11
- python version: 3.10
- `mypy` 1.16.0
- version of installed `pandas-stubs` :  development version


Dr-Irv (approved):
thanks @cmp0xff

PR #1261, opened 2025-06-24
GH1255 Improve index typing for Series

### LINKED ISSUE #1255: `Series` `index` argument should accept iterables of certain types
**Describe the bug**
For the `index` argument, we should support iterables, but need to be careful that we don't allow plain strings.
```python

index={"a":3, "c": 4}

s1 = pd.Series([1, 2], index=index)
s2 = pd.Series([1, 2], index=index.keys())

t = pd.Series([1,2], index="ab")  # This should fail
```

Currently, `s1` and `s2` are rejected by pyright, but should be accepted.  `t` is rejected as well, but I'm including it because any change we make has to make sure this still fails.

To do this, we could modify the `AxesData` type in `_typing.pyi` to include `Iterable[SeriesDType]`, but if we do that, then a plain `str` is accepted.  To avoid that problem, I think adding a `Series.__new__()` overload with `index: str` returning `Never` right at the top (and similar for `Index.__new__()` with the `data` argument) would force that to be rejected.

The above tests should be included in a PR, where `s1` and `s2` are accepted, but `t` is not.

loicdiridollou:
- [x] Closes #1255

### DIFF
--- a/pandas-stubs/_typing.pyi
+++ b/pandas-stubs/_typing.pyi
@@ -1,4 +1,7 @@
-from builtins import type as type_t
+from builtins import (
+    str,
+    type as type_t,
+)
 from collections.abc import (
     Callable,
     Hashable,
--- a/pandas-stubs/_typing.pyi
+++ b/pandas-stubs/_typing.pyi
@@ -1,11 +1,9 @@
-from builtins import (
-    str,
-    type as type_t,
-)
+from builtins import type as type_t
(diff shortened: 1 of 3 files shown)


Dr-Irv:
pre-commit failing - just run `black`

loicdiridollou:
Thanks, accidentally deleted a line and I fixed it wrong. Should be good this time!

PR #1262, opened 2025-06-24
GH851 Change use of NoDefault to make it more private

### LINKED ISSUE #851: Change use of `NoDefault` to be private and maybe indicate it shouldn't be passed
`NoDefault` should never be used by users, but is useful when using `stubtest` to detect inconsistencies between the stubs and the source.  See https://github.com/pandas-dev/pandas-stubs/pull/848#issuecomment-1890472766

Should consider renaming `NoDefault` in the stubs to be private (e.g., `_NoDefault), and maybe even something like `_NoDefaultDoNotUse` .

loicdiridollou:
- [x] Closes #851

It turns out you can't actually make it fully private with `_` at the start of the name as it will raise a lint with PYI (it does not see it used in the file itself although it is used outside the file).
```
pandas-stubs/_libs/lib.pyi:17:1: PYI047 Private TypeAlias `_NoDefaultDoNotUse` is never used
   |
16 | no_default: Final = _NoDefault.no_default
17 | _NoDefaultDoNotUse: TypeAlias = Literal[_NoDefault.no_default]
   | ^^^^^^^^^^^^^^^^^^ PYI047
18 |
19 | def infer_dtype(value: object, skipna: bool = ...) -> str: ...
   |

Found 1 error. 
```


Dr-Irv (approved):
thanks @loicdiridollou

PR #1263, opened 2025-06-25
Run Pyrefly on pandas-stubs in CI

yangdanny97:
Similar to https://github.com/pandas-dev/pandas-stubs/pull/1240, this PR adds Pyrefly to run on just the stubs directory, since there are very few false-positives now (I've filed https://github.com/facebook/pyrefly/issues/569 to address some of these).

I tested the CI and added the suppressions on my fork here: https://github.com/yangdanny97/pandas-stubs/pull/1

There are some known gaps that prevent us from checking the tests directory without a lot of false-positives, so that can be saved for later.


Dr-Irv:
One thing to consider.  In CI, we have tests that separately test the stubs locally, but also test them when installed (for `mypy` and `pyright`). There have been occasions when the type checker passes the stubs locally, but fails when installed as a package.  So you may want to add a test for `pyrefly` in that case, although maybe that doesn't make sense because you can't have `pyrefly` pass tests on user code (`pandas-stubs/tests`)

Dr-Irv:
> I was thinking that we test the installed stubs once we're able to check the `tests` directory. Without user code, I don't think we would emit errors for an installed dependency.
> 
> We're still some ways off from being able to do that (there are a few hundred false positives right now), so I figured we could run it on just the stubs for now and still get some value out of it.
> 
> Wdyt?

OK - that's fine by me.

yangdanny97:
Yep, I'll be keeping an eye on new versions.

Pyrefly has a subcommand that automatically adds/removes suppression comments to help with upgrades.

PR #1265, opened 2025-06-28
GH1264 Version 2.0 cleanup

loicdiridollou:


pandas-stubs/core/indexes/base.pyi
```diff
@@ -347,11 +347,16 @@ class Index(IndexOpsMixin[S1]):
     def __neg__(self) -> Self: ...
     def __nonzero__(self) -> None: ...
     __bool__ = ...
-    def union(self, other: list[HashableT] | Index, sort=...) -> Index: ...
-    def intersection(self, other: list[S1] | Self, sort: bool = ...) -> Self: ...
+    def union(
+        self, other: list[HashableT] | Index, sort: bool | None = ...
```
Dr-Irv:
```suggestion
        self, other: list[HashableT] | Self, sort: bool | None = ...
```
Might as well be consistent with these different operations

pandas-stubs/core/series.pyi
```diff
@@ -2324,6 +2324,13 @@ class TimedeltaSeries(Series[Timedelta]):
         *args: Any,
         **kwargs: Any,
     ) -> TimedeltaSeries: ...
+    def cumprod(
+        self,
+        axis: AxisIndex | None = ...,
```
Dr-Irv: Remove `None`

tests/test_indexes.py
```diff
@@ -312,6 +316,20 @@ def test_range_index_union():
...
+            pd.RangeIndex(0, 10).union(["a", "b", "c"], sort=True),
+            Union[pd.Index, "pd.Index[int]", pd.RangeIndex],
+        ),
+        pd.Index,
+    )
+    check(
+        assert_type(
+            pd.RangeIndex(0, 10).union(["a", "b", "c"], sort=False),
+            Union[pd.Index, "pd.Index[int]", pd.RangeIndex],
+        ),
```
Dr-Irv: This doesn't seem right, and I think you'll need some specific overloads here for `RangeIndex` (and `Sequence[int]` for that matter).   If `RangeIndex` has a set operation with either a `Sequence[int]`, `Series[int]`, `Index[int]` or `RangeIndex`, the result is `Union[RangeIndex, Index[int]]`.  With any other type, it is just `Index`

tests/test_series.py
```diff
@@ -3941,3 +3941,8 @@ def test_series_index_type() -> None:
 
     if TYPE_CHECKING_INVALID_USAGE:
         t = pd.Series([1, 2], index="ab")  # type: ignore[call-overload] # pyright: ignore[reportCallIssue, reportArgumentType]
+
+
+def test_timedelta_index_cumsum() -> None:
```
Dr-Irv:
```suggestion
def test_timedelta_invalid_cumprod() -> None:
```

tests/test_series.py
```diff
@@ -3941,3 +3941,8 @@ def test_series_index_type() -> None:
...
+    if TYPE_CHECKING_INVALID_USAGE:
+        assert_type(pd.Series([pd.Timedelta(0), pd.Timedelta(1)]).cumprod(), Never)
```
Dr-Irv, line 3966: Since this is invalid usage,  you should add similar tests for  `TimestampSeries`, `PeriodSeries`, `OffsetSeries`, and maybe `Series[str]` (not sure if that latter one is possible)

Dr-Irv (changes requested):

tests/test_indexes.py
loicdiridollou: Indeed that was the wrong copy paste, this is fixed.

pandas-stubs/core/series.pyi
loicdiridollou: Since the default argument is `None`, wouldn't that cause issues?

tests/test_series.py
loicdiridollou: Actually wherever `*` is not defined on the type it will not work so Timestamp, Period, Offset series won't work, fixing it.

pandas-stubs/core/series.pyi
Dr-Irv: According to https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.Series.cumprod.html#pandas.Series.cumprod the default argument is 0

loicdiridollou: I see what you mean, I was looking at the 2.3.0 code which still has None as default but on main it is now 0 (I would not see anyone use None by default and otherwise it will make it a good time to fix it).

PR #1268, opened 2025-06-28
allow args and kwargs in groupby.apply

### LINKED ISSUE #1266: GroupBy.apply: `error: No overloads for "apply" match the provided arguments (reportCallIssue)`
**To Reproduce**
```python

import pandas as pd

df = pd.DataFrame({
    "group": ["A", "A", "B", "B", "C"],
    "value": [10, 15, 10, 25, 30]
})


def add_constant_to_mean(group: pd.DataFrame, constant: int) -> pd.DataFrame:
    mean_val = group["value"].mean()
    group["adjusted"] = mean_val + constant
    return group

result = df.groupby("group", group_keys=False).apply(add_constant_to_mean, constant=5)
```

Linting using pyright (version 1.1.393) and `pandas-stubs` 2.2.3.250527:
```
error: No overloads for "apply" match the provided arguments (reportCallIssue)
```

Thanks!!

Dr-Irv:
- [x] Closes #1266(
- [x] Tests added: `test_groupby.py:test_dataframe_apply_kwargs()`

### DIFF
--- a/pandas-stubs/core/groupby/generic.pyi
+++ b/pandas-stubs/core/groupby/generic.pyi
@@ -11,6 +11,7 @@ from typing import (
     Generic,
     Literal,
     NamedTuple,
+    Protocol,
     TypeVar,
     final,
     overload,
@@ -208,26 +209,35 @@ class SeriesGroupBy(GroupBy[Series[S2]], Generic[S2, ByT]):
 
 _TT = TypeVar("_TT", bound=Literal[True, False])
 
+class DFCallable1(Protocol):
+    def __call__(self, df: DataFrame, /, *args, **kwargs) -> Scalar | list | dict: ...
+
+class DFCallable2(Protocol):
+    def __call__(self, df: DataFrame, /, *args, **kwargs) -> DataFrame | Series: ...
+
+class DFCallable3(Protocol):
+    def __call__(self, df: Iterable, /, *args, **kwargs) -> float: ...
+
 class DataFrameGroupBy(GroupBy[DataFrame], Generic[ByT, _TT]):
     # error: Overload 3 for "apply" will never be used because its parameters overlap overload 1
     @overload  # type: ignore[override]
     def apply(
         self,
-        func: Callable[[DataFrame], Scalar | list | dict],
+        func: DFCallable1,
         *args,
         **kwargs,
     ) -> Series: ...
     @overload
     def apply(
         self,
-        func: Callable[[DataFrame], Series | DataFrame],
+        func: DFCallable2,
         *args,
         **kwargs,
     ) -> DataFrame: ...
     @overload
     def apply(  # pyright: ignore[reportOverlappingOverload]
         self,
-        func: Callable[[Iterable], float],
+        func: DFCallable3,
         *args,
         **kwargs,
     ) -> DataFrame: ...
--- a/tests/test_groupby.py
+++ b/tests/test_groupby.py
@@ -1102,3 +1102,23 @@ def test_dataframe_value_counts() -> None:
         Series,
         np.int64,
     )
+
+
+def test_dataframe_apply_kwargs() -> None:
+    # GH 1266
+    df = DataFrame({"group": ["A", "A", "B", "B", "C"], "value": [10, 15, 10, 25, 30]})
+
+    def add_constant_to_mean(group: DataFrame, constant: int) -> DataFrame:
+        mean_val = group["value"].mean()
+        group["adjusted"] = mean_val + constant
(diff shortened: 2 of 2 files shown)


schorlton-bugseq:
Thank you!

pandas-stubs/core/groupby/generic.pyi
```diff
@@ -208,26 +209,35 @@ class SeriesGroupBy(GroupBy[Series[S2]], Generic[S2, ByT]):
 
 _TT = TypeVar("_TT", bound=Literal[True, False])
 
+class DFCallable1(Protocol):
+    def __call__(self, df: DataFrame, /, *args, **kwargs) -> Scalar | list | dict: ...
```
twoertwein: Could also add the `include_groups` keyword  https://pandas.pydata.org/docs/reference/api/pandas.core.groupby.DataFrameGroupBy.apply.html

twoertwein: (Weird deprecation warning: " Setting include_groups to True is deprecated. Only the value False will be allowed in a future version of pandas.")

pandas-stubs/core/groupby/generic.pyi
```diff
@@ -208,26 +209,35 @@ class SeriesGroupBy(GroupBy[Series[S2]], Generic[S2, ByT]):
...
+
 class DataFrameGroupBy(GroupBy[DataFrame], Generic[ByT, _TT]):
     # error: Overload 3 for "apply" will never be used because its parameters overlap overload 1
     @overload  # type: ignore[override]
     def apply(
         self,
-        func: Callable[[DataFrame], Scalar | list | dict],
+        func: DFCallable1,
         *args,
         **kwargs,
```
twoertwein: Could probably type args + kwargs here using `typing.ParamSpec` as they have to be compatible with `func`

pandas-stubs/core/groupby/generic.pyi
Dr-Irv: Yes, I think this is a temporary parameter to handle the changeover in behavior, so I don't think we should put it in the stubs.

pandas-stubs/core/groupby/generic.pyi
Dr-Irv: OK, I've done that.  Also removed `include_groups` in the tests.

twoertwein (approved):
Two small optional changes (happy to merge without them):

- `include_groups` is not deprecated, only `include_groups=True`, I would be fine keeping the test and adding `include_groups: Literal[False]` in the signature
- it might be nice to have a test where a kwargs/args is incompatible with the `func`

Dr-Irv:
> Two small optional changes (happy to merge without them):
> 
> * `include_groups` is not deprecated, only `include_groups=True`, I would be fine keeping the test and adding `include_groups: Literal[False]` in the signature

Yes, but there are 2 issues with that.
1.  The argument will become irrelevant since only a value of `False` is accepted.  So we shouldn't encourage its use.
2. If we use `ParamSpec`, we can't have a keyword argument between `*args` and `**kwargs` in the typing stubs.  See https://github.com/python/typing/discussions/1905


> * it might be nice to have a test where a kwargs/args is incompatible with the `func`

Yes, although I'm not sure how to write one!  I just wanted to handle the use case reported.

twoertwein:
> 1. The argument will become irrelevant since only a value of `False` is accepted.  So we shouldn't encourage its use.
> 2. If we use `ParamSpec`, we can't have a keyword argument between `*args` and `**kwargs` in the typing stubs.  See [Is `typing.Protocol` the only way to concatenate the keyword-only arguments in `ParamSpec`? python/typing#1905](https://github.com/python/typing/discussions/1905)

Thank you for the explanations!

> Yes, although I'm not sure how to write one! I just wanted to handle the use case reported.

Something like 
```py
df.groupby("group", group_keys=False)[["group", "value"]].apply(
    add_constant_to_mean, constant="5". # typing: ignore
)
```
should hopefully be flagged as wrong.

Dr-Irv:
> Something like
> 
> ```python
> df.groupby("group", group_keys=False)[["group", "value"]].apply(
>     add_constant_to_mean, constant="5"  # typing: ignore
> )
> ```
> 
> should hopefully be flagged as wrong.

Thanks for the suggestion.  I've incorporated into the last commit

twoertwein:
Thanks @Dr-Irv !

DeflateAwning:
This PR is the bane of my existence. Can you please add back the overload for `include_groups: Literal[False]`

Issue #1098, opened 2025-01-14
Addition of a Series[int] with a complex returns Series[unknown]


loicdiridollou:
```python
import pandas as pd

c = 1 + 1j
s = pd.Series([1.0, 2.0, 3.0])

check(assert_type(s + c, "pd.Series[complex]"), pd.Series) ■ "assert_type" mismatch: expected "Series[complex]" but received "Series[unknown]"
```

Please complete the following information:

OS: Darwin
OS Version [e.g. 22]: 15.2
python version: 3.12.7
version of type checker: mypy latest
version of installed pandas-stubs: latest
Additional context

version of pandas: 2.2.3
mypy option: strict=False

Dr-Irv:
Need to update the various operators to handle arguments that are `complex`

loicdiridollou:
Also another issue about the __sub__ operator between two `Series[int]`:
```python
s = pd.Series([0, 1, -10])
s2 = pd.Series([7, -5, 10])

check(assert_type(s - s2, "pd.Series[int]"), pd.Series, np.integer)
```

loicdiridollou:
```python
s1 = pd.Series([0, 1, 2, 3])
s2 = pd.Series([-1, 2, -3, 4])
df1 = pd.DataFrame([[0, 1], [-2, 3], [4, -5], [6, 7]])
n1 = np.array([[0, 1], [1, 2], [-1, -1], [2, 0]])
check(assert_type(s1.dot(s2), Scalar), np.int64)
check(assert_type(s1 @ s2, Scalar), np.int64)
check(assert_type(s1.dot(df1), "pd.Series[int]"), pd.Series, np.int64)
check(assert_type(s1.dot(s2), Scalar), np.integer)
check(assert_type(s1 @ s2, Scalar), np.integer)
check(assert_type(s1.dot(df1), "pd.Series[int]"), pd.Series, np.integer)  # should be float here as we don't know the type of the df
```

shirzady1934:
Hey @loicdiridollou  I fixed this for __add__ but, for __mul__  we cannot expect check(assert_type(s * s2, "pd.Series[int]"), pd.Series, np.integer) to pass without a type error because, as shown in 
https://github.com/pandas-dev/pandas-stubs/blob/7328e89be08d75bdac41fd1da7ab28199211d72c/tests/test_series.py#L620
 the test verifies that the square of a Series returns a float. If both operands are treated as integers, this type mismatch will trigger an assertion failure.

[...]

PR #1273, opened 2025-07-12
fix: #718 drop TimestampSeries and TimedeltaSeries

cmp0xff:


tests/test_timefuncs.py
```diff
@@ -1371,7 +1508,7 @@ def test_mean_median_std() -> None:
             pd.Timestamp("2021-01-03"),
         ]
     )
-    check(assert_type(s2.mean(), pd.Timestamp), pd.Timestamp)
+    # check(assert_type(s2.mean(), pd.Timestamp), pd.Timestamp)
```
cmp0xff: Problem: `Series[Timestamp].mean()` now returns `Any`

tests/test_timefuncs.py
```diff
@@ -1332,35 +1469,35 @@ def test_timedelta64_and_arithmatic_operator() -> None:
...
     tss = pd.Series(pd.date_range(start="2012-01-01", periods=10, freq="W-MON"))
     plus = tds + tss
-    check(assert_type(plus, "TimestampSeries"), pd.Series, pd.Timestamp)
+    check(assert_type(plus, "pd.Series[pd.Timestamp]"), pd.Series, pd.Timestamp)
 
 
 def test_mean_median_std() -> None:
     s = pd.Series([pd.Timedelta("1 ns"), pd.Timedelta("2 ns"), pd.Timedelta("3 ns")])
-    check(assert_type(s.mean(), pd.Timedelta), pd.Timedelta)
+    # check(assert_type(s.mean(), pd.Timedelta), pd.Timedelta)
```
cmp0xff: Problem: `Series[Timedelta].mean()` now returns `Any`

tests/test_timefuncs.py
```diff
@@ -543,37 +598,121 @@ def test_series_dt_accessors() -> None:
...
+        assert_type(
+            s5.dt.as_unit("us"),  # pyright: ignore[reportAssertTypeFailure]
+            "pd.Series[pd.Timedelta]",
+        ),
+        pd.Series,
+        pd.Timedelta,
+    )
+    check(
+        assert_type(
+            s5.dt.as_unit("ns"),  # pyright: ignore[reportAssertTypeFailure]
```
cmp0xff: Problem: pyright does not recognise the [descriptor trick](https://github.com/microsoft/pyright/issues/3071#issuecomment-1043978070)

tests/test_timefuncs.py
```diff
@@ -543,37 +598,121 @@ def test_series_dt_accessors() -> None:
...
+        assert_type(
+            s5.dt.as_unit("ms"),  # pyright: ignore[reportAssertTypeFailure]
+            "pd.Series[pd.Timedelta]",
+        ),
+        pd.Series,
+        pd.Timedelta,
+    )
+    check(
+        assert_type(
+            s5.dt.as_unit("us"),  # pyright: ignore[reportAssertTypeFailure]
```
cmp0xff: Problem: pyright does not recognise the [descriptor trick](https://github.com/microsoft/pyright/issues/3071#issuecomment-1043978070)

tests/test_series.py
```diff
@@ -3892,7 +3897,7 @@ def foo(sf: pd.Series) -> None:
         pass
 
     foo(s)
-    check(assert_type(s + pd.Series([1]), pd.Series), pd.Series)
+    check(assert_type(s + pd.Series([1]), pd.Series), pd.Series)  # type: ignore [assert-type]   # pyright: ignore[reportAssertTypeFailure]
```
cmp0xff: Problem in `__add__` or something else

tests/test_series.py
```diff
@@ -1599,10 +1589,22 @@ def test_series_min_max_sub_axis() -> None:
     ss = s1 - s2
     sm = s1 * s2
     sd = s1 / s2
-    check(assert_type(sa, pd.Series), pd.Series)
-    check(assert_type(ss, pd.Series), pd.Series)
-    check(assert_type(sm, pd.Series), pd.Series)
-    check(assert_type(sd, pd.Series), pd.Series)
+    check(
+        assert_type(sa, pd.Series),  # pyright: ignore[reportAssertTypeFailure]
```
cmp0xff: Problem in `__add__` or something else

tests/test_frame.py
```diff
@@ -2819,9 +2822,20 @@ def test_sum_get_add() -> None:
     summer = df.sum(axis=1)
     check(assert_type(summer, pd.Series), pd.Series)
 
-    check(assert_type(s + summer, pd.Series), pd.Series)
-    check(assert_type(s + df["y"], pd.Series), pd.Series)
-    check(assert_type(summer + summer, pd.Series), pd.Series)
+    check(
+        assert_type(s + summer, pd.Series),  # pyright: ignore[reportAssertTypeFailure]
```
cmp0xff: Problem in `__add__` or something else

tests/test_frame.py
```diff
@@ -2808,7 +2805,13 @@ def test_indexslice_getitem():
 def test_compute_values():
     df = pd.DataFrame({"x": [1, 2, 3, 4]})
     s: pd.Series = pd.Series([10, 20, 30, 40])
-    check(assert_type(df["x"] + s.values, pd.Series), pd.Series, np.int64)
+    check(
+        assert_type(
+            df["x"] + s.values, pd.Series  # pyright: ignore[reportAssertTypeFailure]
```
cmp0xff: Problem in `__add__` or something else

pandas-stubs/core/series.pyi
```diff
@@ -1858,6 +1996,25 @@ class Series(IndexOpsMixin[S1], NDFrame):
...
+    # @overload
+    # def mean(
+    #     self: Series[Timedelta],
+    #     axis: AxisIndex | None = ...,
+    #     skipna: _bool = ...,
+    #     level: None = ...,
+    #     numeric_only: _bool = ...,
+    #     **kwargs: Any,
+    # ) -> Timedelta: ...
+    # @overload
```
cmp0xff: I cannot make `Series[Any].mean()` return float, while `Series[Timestamp].mean()` returns `Timestamp`.

pandas-stubs/core/indexes/accessors.pyi
```diff
@@ -427,3 +428,17 @@ class TimedeltaIndexProperties(
...
+        self, instance: Series[Timestamp], owner: Any
+    ) -> TimestampProperties: ...
+    @overload
+    def __get__(
+        self, instance: Series[Timedelta], owner: Any
+    ) -> TimedeltaProperties: ...
+    @overload
+    def __get__(
+        self, instance: Series, owner: Any
+    ) -> CombinedDatetimelikeProperties: ...
```
cmp0xff: https://github.com/microsoft/pyright/issues/3071#issuecomment-1043978070

tests/test_timefuncs.py
cmp0xff: c99cedcee0c485b3e52cc2944ac87ab85f33c911

pandas-stubs/core/series.pyi
cmp0xff: c99cedcee0c485b3e52cc2944ac87ab85f33c911

tests/test_timefuncs.py
cmp0xff: c99cedcee0c485b3e52cc2944ac87ab85f33c911

tests/test_frame.py
loicdiridollou: Those sound suspicious, I would not let those fall through because that would mean the user will encounter those too.

tests/test_series.py
loicdiridollou: If you need to add the assert type ignore in all the cases that means that there is something to be fixed.
This test replicates a pretty common usage of pandas so I would not let this go unfixed.
(fwiw I have the same issue with #1098 where pyright does not see the right types but mypy sees the correct inference).

tests/test_series.py
loicdiridollou: Seems like this issue is appearing in `__add__` but I believe you will have the same issue with `__sub__`.
Could you duplicate the test and have it with `__sub__` and see if you get the same pyright and mypy error please?

pandas-stubs/core/tools/datetimes.pyi
```diff
@@ -94,7 +91,7 @@ def to_datetime(
     unit: str | None = ...,
     origin: Literal["julian", "unix"] | TimestampConvertibleTypes = ...,
     cache: bool = ...,
-) -> TimestampSeries: ...
+) -> Series[Timestamp]: ...
```
loicdiridollou, line 96: I think you may need to flip the order of this class because the declaration of `Series` otherwise `ty` does not find it.

loicdiridollou (commented):
Overall this is a lot of great work, I have taken a crack at it and was quickly under water with the amount of changes needed.
I believe the `__add__` issue that you have seems similar to what I am experiencing with #1098 and you may have the same problem with `__sub__` and `__truediv__` but you will be able to confirm.
I think we can't let those slip in the current version. Maybe the option is to keep the current TimestampSeries definition as a middle layer and start supporting the `Series[Timestamp]` along the way (not sure if this is possible).

tests/test_timefuncs.py
cmp0xff: 3466be5d18f1ff34cc4e78026729ff1d4f6b5366

tests/test_timefuncs.py
cmp0xff: 3466be5d18f1ff34cc4e78026729ff1d4f6b5366

tests/test_series.py
cmp0xff: e5b7b1ea3414d338b1ef33bcc096b18aa848b0bc

tests/test_frame.py
cmp0xff: e5b7b1ea3414d338b1ef33bcc096b18aa848b0bc

tests/test_frame.py
cmp0xff: e5b7b1ea3414d338b1ef33bcc096b18aa848b0bc

pandas-stubs/core/indexes/accessors.pyi
cmp0xff: 3466be5d18f1ff34cc4e78026729ff1d4f6b5366

tests/test_series.py
cmp0xff: e5b7b1ea3414d338b1ef33bcc096b18aa848b0bc

pandas-stubs/core/tools/datetimes.pyi
cmp0xff: 79ae3a91e97872a528e73c17daf770980e44b4bd and 3466be5d18f1ff34cc4e78026729ff1d4f6b5366

pandas-stubs/core/series.pyi
```diff
@@ -2379,4 +2339,4 @@ class OffsetSeries(Series[BaseOffset]):
 class IntervalSeries(Series[Interval[_OrderableT]], Generic[_OrderableT]):
     @property
     def array(self) -> IntervalArray: ...
-    def diff(self, periods: int = ...) -> Never: ...
+    def diff(self, periods: int = ...) -> Never: ...  # pyrefly: ignore
```
cmp0xff, line 4821: Additional ignore, potential issue

pandas-stubs/core/series.pyi
```diff
@@ -2368,7 +2328,7 @@ class OffsetSeries(Series[BaseOffset]):
     def __radd__(  # pyright: ignore[reportIncompatibleMethodOverride]
         self, other: BaseOffset
     ) -> OffsetSeries: ...
-    def cumprod(
+    def cumprod(  # pyrefly: ignore
```
cmp0xff: Additional ignore, potential issue

tests/test_series.py
```diff
@@ -1601,7 +1591,7 @@ def test_series_min_max_sub_axis() -> None:
     sd = s1 / s2
     check(assert_type(sa, pd.Series), pd.Series)
     check(assert_type(ss, pd.Series), pd.Series)
-    check(assert_type(sm, pd.Series), pd.Series)
+    check(assert_type(sm, pd.Series), pd.Series)  # type: ignore[assert-type]
```
cmp0xff: Additional ignore, maybe mypy is not very clever here

tests/test_series.py
```diff
@@ -3226,23 +3216,25 @@ def test_timedelta_div() -> None:
...
 
     if TYPE_CHECKING_INVALID_USAGE:
-        1 / series  # type: ignore[operator] # pyright: ignore[reportOperatorIssue]
-        [1] / series  # type: ignore[operator] # pyright: ignore[reportOperatorIssue]
-        1 // series  # type: ignore[operator] # pyright: ignore[reportOperatorIssue]
-        [1] // series  # type: ignore[operator] # pyright: ignore[reportOperatorIssue]
+        1 / series
+        [1] / series
+        1 // series
+        [1] // series
```
cmp0xff: I had to remove the ignores, not sure if acceptable

tests/test_timefuncs.py
```diff
@@ -321,9 +302,9 @@ def test_fail_on_adding_two_timestamps() -> None:
     s1 = pd.Series(pd.to_datetime(["2022-05-01", "2022-06-01"]))
     s2 = pd.Series(pd.to_datetime(["2022-05-15", "2022-06-15"]))
     if TYPE_CHECKING_INVALID_USAGE:
-        ssum: pd.Series = s1 + s2  # type: ignore[operator] # pyright: ignore[reportOperatorIssue]
+        ssum: pd.Series = s1 + s2
         ts = pd.Timestamp("2022-06-30")
-        tsum: pd.Series = s1 + ts  # type: ignore[operator] # pyright: ignore[reportOperatorIssue]
+        tsum: pd.Series = s1 + ts
```
cmp0xff: I had to remove the ignores, not sure if acceptable

tests/test_timefuncs.py
```diff
@@ -1332,30 +1388,31 @@ def test_timedelta64_and_arithmatic_operator() -> None:
...
+    check(assert_type((s3 + td), "pd.Series[pd.Timedelta]"), pd.Series, pd.Timedelta)
     check(assert_type((s3 / td), "pd.Series[float]"), pd.Series, float)
     if TYPE_CHECKING_INVALID_USAGE:
-        r1 = s1 * td  # type: ignore[operator] # pyright: ignore[reportOperatorIssue]
-        r2 = s1 / td  # type: ignore[operator] # pyright: ignore[reportOperatorIssue]
-        r3 = s3 * td  # type: ignore[operator] # pyright: ignore[reportOperatorIssue]
+        s1.__mul__(td)
+        r1 = s1 * td  # pyright: ignore[reportOperatorIssue]
+        r2 = s1 / td  # pyright: ignore[reportOperatorIssue]
+        r3 = s3 * td  # pyright: ignore[reportOperatorIssue]
```
cmp0xff: I had to remove the ignores, not sure if acceptable

tests/test_timefuncs.py
```diff
@@ -1332,30 +1388,31 @@ def test_timedelta64_and_arithmatic_operator() -> None:
...
+    check(assert_type((s1 - td), "pd.Series[pd.Timestamp]"), pd.Series, pd.Timestamp)
+    check(assert_type((s1 + td), "pd.Series[pd.Timestamp]"), pd.Series, pd.Timestamp)
+    check(assert_type((s3 - td), "pd.Series[pd.Timedelta]"), pd.Series, pd.Timedelta)
+    check(assert_type((s3 + td), "pd.Series[pd.Timedelta]"), pd.Series, pd.Timedelta)
     check(assert_type((s3 / td), "pd.Series[float]"), pd.Series, float)
     if TYPE_CHECKING_INVALID_USAGE:
-        r1 = s1 * td  # type: ignore[operator] # pyright: ignore[reportOperatorIssue]
-        r2 = s1 / td  # type: ignore[operator] # pyright: ignore[reportOperatorIssue]
-        r3 = s3 * td  # type: ignore[operator] # pyright: ignore[reportOperatorIssue]
+        s1.__mul__(td)
```
cmp0xff:
```suggestion
```

pandas-stubs/core/indexes/accessors.pyi
```diff
@@ -162,7 +164,7 @@ class _DatetimeLikeOps(
 
 _DTTimestampTimedeltaReturnType = TypeVar(
     "_DTTimestampTimedeltaReturnType",
-    bound=Series | TimestampSeries | TimedeltaSeries | DatetimeIndex | TimedeltaIndex,
+    bound=Series | Series[Timestamp] | DatetimeIndex | TimedeltaIndex,
```
twoertwein: I think you can probably remove `Series[Timestamp]` here as there is already `Series`

pandas-stubs/core/series.pyi
```diff
@@ -183,6 +182,16 @@ from pandas.core.dtypes.dtypes import CategoricalDtype
...
 
+_scalar_timedelta: TypeAlias = timedelta | np.timedelta64 | BaseOffset | Timedelta
+_vector_timedelta: TypeAlias = (
+    Sequence[timedelta] | Sequence[Timedelta] | TimedeltaIndex
+)
+_nonseries_timedelta: TypeAlias = _scalar_timedelta | _vector_timedelta
+_nonseries_int: TypeAlias = int | np_ndarray_anyint | Sequence[int]
+
+_T_INT = TypeVar("_T_INT", bound=int)
+_T_STAMP_AND_DELTA = TypeVar("_T_STAMP_AND_DELTA", bound=Timestamp | Timedelta)
```
twoertwein: might be nice to use consistent upper/lower-case across type-aliases/var: `_int_t` or `_intT` (similar to `NDFrameT)`

pandas-stubs/core/series.pyi
twoertwein: (I wouldn't worry about errors specific to pyrefly/ty at this moment)

tests/test_series.py
twoertwein: what does mypy infer?

tests/test_series.py
twoertwein: could potentially add `-> Never` overloads?

twoertwein (commented):
Thank you for working on this!

(It might be too late for that, but it could be nice to split timedelta and timestamp into separate MRs.)

Overload order can be important when we have  `Series[Any]` or `Series[unknown]`. After you get feedback from @Dr-Irv, it might be nice to add new tests for overloads that might have changed their any/unknown behavior.

pandas-stubs/core/series.pyi
cmp0xff: It's not very consistent everywhere, e.g. https://github.com/pandas-dev/pandas-stubs/blob/19767bbf269f754e1e01dbdd47060ef20729624f/pandas-stubs/core/strings.pyi#L42

cmp0xff:
> (It might be too late for that, but it could be nice to split timedelta and timestamp into separate MRs.)

I managed to isolate #1274 which just removes TimestampSeries. This PR can be reviewed after #1274.

tests/test_series.py
cmp0xff: `mypy` says the return type is `Any`

cmp0xff: Related: https://github.com/pandas-dev/pandas-stubs/pull/1274#discussion_r2203306834

cmp0xff: `ss` used to be a problem in #1274, but got fixed here

Dr-Irv:
So with this being dependent on #1274, I won't review this until that one is accepted and merged in.

tests/test_series.py
cmp0xff: Not there anymore

tests/test_series.py
cmp0xff: Now covered by more systematic tests and with ignores

pandas-stubs/core/series.pyi
cmp0xff: Does not exist anymore

cmp0xff:
/pandas_nightly

@Dr-Irv I've finished it from my point of view. There are a lot of changes. How do we proceed? I also added `Index`es to `__mul__`, __truediv__` and `__floordiv__` because they already exist in the original `TimedeltaSeries`.

Dr-Irv:
> @Dr-Irv I've finished it from my point of view. There are a lot of changes. How do we proceed? I also added `Index`es to `__mul__`, **truediv**`and`**floordiv**`because they already exist in the original`TimedeltaSeries`.

I'll review it and we'll take it from there.

pandas-stubs/_libs/tslibs/timestamps.pyi
```diff
@@ -225,8 +222,6 @@ class Timestamp(datetime, SupportsIndex):
     @overload
     def __add__(self, other: timedelta | np.timedelta64 | Tick) -> Self: ...
     @overload
-    def __add__(self, other: TimedeltaSeries) -> Series[Timestamp]: ...
```
Dr-Irv, line 228: Does it matter if we keep this overload in here?  I realize that if you now do `Timestamp + Series[Timedelta]`, that `Series[Timedelta].__radd__(Timestamp)` will get used, but I don't see an issue in keeping the redundancy.

Same goes for `__sub__()` below

pandas-stubs/core/indexes/datetimes.pyi
```diff
@@ -69,15 +66,10 @@ class DatetimeIndex(
 
     # various ignores needed for mypy, as we do want to restrict what can be used in
     # arithmetic for these types
-    @overload  # type: ignore[override]
-    def __add__(self, other: TimedeltaSeries) -> Series[Timestamp]: ...
```
Dr-Irv, line 73: Why remove this?  IMHO, it's better to depend on the forward operator `__add__()` than the reverse operator `__radd__()`, but maybe something else is going on??

Same comment with respect to `__sub__()`

tests/series/arithmetic/timedelta/test_floordiv.py
```diff
@@ -0,0 +1,294 @@
...
+    check(assert_type(left // i, "pd.Series[pd.Timedelta]"), pd.Series, pd.Timedelta)
+    check(assert_type(left // f, "pd.Series[pd.Timedelta]"), pd.Series, pd.Timedelta)
+    if TYPE_CHECKING_INVALID_USAGE:
+        assert_type(left // c, Never)
+        assert_type(left // s, Never)
+    check(assert_type(left // d, "pd.Series[int]"), pd.Series, np.integer)
+
+    # `numpy` typing gives the corresponding `ndarray`s in the static type
+    # checking, where our `__rsub__` cannot override. At runtime, they lead to
+    # errors or pd.Series.
```
Dr-Irv, line 139: I think you need to update this comment.  The comment should be about `__floordiv__` and say that `numpy` typing is resolving this due to `__rfloordiv__()` being defined in `numpy`, even though they don't work at runtime.

tests/series/arithmetic/timedelta/test_truediv.py
```diff
@@ -0,0 +1,367 @@
...
+    check(assert_type(left / i, "pd.Series[pd.Timedelta]"), pd.Series, pd.Timedelta)
+    check(assert_type(left / f, "pd.Series[pd.Timedelta]"), pd.Series, pd.Timedelta)
+    if TYPE_CHECKING_INVALID_USAGE:
+        assert_type(left / c, Never)
+        assert_type(left / s, Never)
+    check(assert_type(left / d, "pd.Series[float]"), pd.Series, np.floating)
+
+    # `numpy` typing gives the corresponding `ndarray`s in the static type
+    # checking, where our `__rsub__` cannot override. At runtime, they lead to
+    # errors or pd.Series.
```
Dr-Irv, line 165: update comment to be about `__truediv__`

tests/series/test_series.py
```diff
@@ -2817,58 +2813,58 @@ def test_astype_timestamp(cast_arg: TimestampDtypeArg, target_type: type) -> Non
 @pytest.mark.parametrize("cast_arg, target_type", ASTYPE_TIMEDELTA_ARGS, ids=repr)
 def test_astype_timedelta(cast_arg: TimedeltaDtypeArg, target_type: type) -> None:
     s = pd.Series([1, 2, 3])
-    check(s.astype(cast_arg), TimedeltaSeries, target_type)
+    check(s.astype(cast_arg), pd.Series, target_type)
```
Dr-Irv, line 2813:
```suggestion
    check(s.astype(cast_arg), pd.Series[pd.Timedelta], target_type)
```
If this isn't working with this change, that's a bug to fix.

tests/test_natype.py
```diff
@@ -43,7 +43,7 @@ def test_arithmetic() -> None:
 
     # __rmul__
     check(assert_type(s_int * na, pd.Series), pd.Series)
-    check(assert_type(idx_int * na, "pd.Index[int]"), pd.Index)
+    check(assert_type(idx_int * na, pd.Index), pd.Index)
```
Dr-Irv, line 46: This shouldn't be changed.

pandas-stubs/core/series.pyi
```diff
@@ -214,6 +214,11 @@ class _SupportsAdd(Protocol[_T_co]):
 class _SupportsMul(Protocol[_T_co]):
     def __mul__(self, value: Self, /) -> _T_co: ...
 
+@type_check_only
+class _SupportsAddRMul(Protocol[_T_co]):
```
```suggestion
class _SupportsAdd_MulFloat(Protocol[_T_co]):
```
Makes the name more descriptive

pandas-stubs/core/series.pyi
```diff
@@ -4001,7 +4507,7 @@ class Series(IndexOpsMixin[S1], NDFrame):
         level: None = ...,
         numeric_only: _bool = False,
         **kwargs: Any,
-    ) -> float: ...
+    ) -> S1: ...
```
Dr-Irv: This isn't entirely correct, because `pd.Series([1,2,3,4]).median()` returns a `float`

pandas-stubs/core/series.pyi
```diff
@@ -4163,7 +4662,23 @@ class Series(IndexOpsMixin[S1], NDFrame):
...
+        self: Series[Timedelta],
+        dtype: None = None,
+        copy: bool = False,
+        na_value: Scalar = ...,
+        **kwargs,
+    ) -> np_1darray[np.timedelta64]: ...
+    @overload
     def to_numpy(
+        self: Series[Timedelta],
+        dtype: np.dtype[GenericT] | SupportsDType[GenericT] | type[GenericT],
```
Dr-Irv, line 4702: Just checking here, but if someone did `pd.Series([pd.Timedelta("1 min")]).to_numpy(dtype=np.timedelta64)`, would that then give a type of `np_1darray[np.timedelta64]` ?

pandas-stubs/_libs/tslibs/timestamps.pyi
cmp0xff: - I would like to reduce redundancy, particularly across different files. Redundancy can cause conflicts and result in type checkers giving `Any`.
- From a hierarchical point of view, `Timedelta`s can be elements of `Series`, and not the other way around. It somehow makes sense to me that `Timedelta` does not know too much about `Series`; `Series` knowing about `Timedelta` is sufficient.
- This can also reduce "cyclic import", which is of course not a big deal in stub files.

pandas-stubs/core/indexes/datetimes.pyi
cmp0xff: - This is the result from a hierarchical point of view again. `Series` contain `Index`es, not the other way around.
- In `Series` we already have many overloads for `__radd__`. Maybe we can move them to `__add__` in other classes in another effort.

At the moment I think it is a good practise to
- have forward and reverse operators in a class and its __components__, so they are together
- define no arithmetic with respect to types that can __contain__ the current class, so there is less redundancy

Alternatively, what you have suggested is
- have forward operators
- eliminate or reduce reverse operators

What would you say?

tests/test_natype.py
cmp0xff: Could you elaborate? I don't get it. All other cases give the generic `pd.Index`, except for this one.

tests/series/arithmetic/timedelta/test_floordiv.py
cmp0xff: 05fad40cce41e740eddfc7467ac4bcabae6f0bd0

pandas-stubs/core/series.pyi
cmp0xff: 05fad40cce41e740eddfc7467ac4bcabae6f0bd0

pandas-stubs/core/series.pyi
cmp0xff: 05fad40cce41e740eddfc7467ac4bcabae6f0bd0

pandas-stubs/core/series.pyi
cmp0xff: Yes it does give `np_1darray[np.timedelta64]` 05fad40cce41e740eddfc7467ac4bcabae6f0bd0

tests/series/arithmetic/timedelta/test_truediv.py
cmp0xff: Should have been fixed in 05fad40cce41e740eddfc7467ac4bcabae6f0bd0 🤔

tests/series/test_series.py
cmp0xff: This is runtime type checking, and does not work with `pd.Series[pd.Timedelta]`. Static type checking is carried out below.

pandas-stubs/_libs/tslibs/timestamps.pyi
Dr-Irv: OK, I'm fine with your reasoning.  As we've seen with `numpy`, making sure that the reverse operator is properly used isn't always straightforward, but if it is passing our tests, then we should be good.

Dr-Irv: Having said that, then `TimedeltaIndex` should eventually be removed as well.

pandas-stubs/core/indexes/datetimes.pyi
Dr-Irv: I'm fine with your suggested practice.

tests/test_natype.py
Dr-Irv: Fair point.  I just try to keep tests unchanged, but as you point out, this is the only one without the specification of the subtype, so your change is fine.

pandas-stubs/core/series.pyi
cmp0xff: Hmm, `mypy` and `pyright` have different opinions. `pyright` agrees with the new test, whereas mypy complains
> Expression is of type "ndarray[tuple[int], dtype[timedelta64[Any]]]", not "ndarray[tuple[int], dtype[timedelta64[timedelta | int | None]]]"  [assert-type]

I have tried to play with `to_numpy` but did not manage to make it work. bad9882b3918c05d7ca1c6b0a2b408b62ff4985f

pandas-stubs/_libs/tslibs/timestamps.pyi
cmp0xff: Unlike `TimedeltaSeries` and `TimestampSeries`, [TimedeltaIndex](https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.TimedeltaIndex.html) and [DatetimeIndex](https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.DatetimeIndex.html) are real classes in Pandas.

Dr-Irv: A bit of splitting hairs here, because when you say "From a hierarchical point of view, `Timedelta`s can be elements of `Series`, and not the other way around. It somehow makes sense to me that `Timedelta` does not know too much about `Series`; `Series` knowing about Timedelta is sufficient."  In the same way, `Timedelta`s can be elements of `Index`, but we happen to call that `TimedeltaIndex` in pandas.  So I could write "`Timedelta` does not know too much about `Index`, and really doesn't know much about `TimedeltaIndex` either.

I'm not going to require this change.  What matters is that the tests work.

tests/series/test_series.py
Dr-Irv: OK, that makes sense.

tests/series/test_agg.py
Dr-Irv, line 1: All of the tests in this file are testing `median`, so rename to `test_median.py` and rename the test functions to be `test_median_XXX` instead of `test_agg_XXX`

ttest.py
Dr-Irv, line 1: remove this file

pandas-stubs/core/series.pyi
Dr-Irv: Maybe the `GenericT` is too wide.  We could borrow from numpy:
```python
_TD64ItemT_co = TypeVar("_TD64ItemT_co", bound=dt.timedelta | int | None, default=dt.timedelta | int | None, covariant=True)
_DT64ItemT_co = TypeVar("_DT64ItemT_co", bound=dt.date | int | None, default=dt.date | int | None, covariant=True)
```
Then use those types in `to_numpy(Series[Timedelta])` and `to_numpy(Series[Timestamp])` respectively instead of `GenericT`

Dr-Irv (commented):
Only a few things

pandas-stubs/core/series.pyi
cmp0xff: It does not work, I added

```py
TD64ItemT = TypeVar("TD64ItemT", bound=datetime.timedelta | int | None, default=datetime.timedelta | int | None)
DT64ItemT = TypeVar("DT64ItemT", bound=datetime.date | int | None, default=datetime.date | int | None)
```

in the stub file, `pyright` complained

```text
Type "TD64ItemT@to_numpy" is not assignable to type "generic[Any]"
    Type "timedelta* | int* | None" is not assignable to type "generic[Any]"
      "int*" is not assignable to "generic[Any]
```

which can be ignore. However in the test, `mypy` still gives

```text
error: Expression is of type "ndarray[tuple[int], dtype[timedelta64[Any]]]", not "ndarray[tuple[int], dtype[timedelta64[timedelta | int | None]]]"  [assert-type]
```

tests/series/test_agg.py
cmp0xff: I think it would be good to have all aggregation tests in one place. ddc41d16d9838b939e3ff31d7e943afe73f8e50a

ttest.py
cmp0xff: Oops. ddc41d16d9838b939e3ff31d7e943afe73f8e50a

pandas-stubs/_libs/tslibs/timestamps.pyi
cmp0xff: Thank you for writing the idea down. I have it in my mind but did now spell it out. I have added it to https://github.com/pandas-dev/pandas-stubs/issues/1378#issuecomment-3337485352.

pandas-stubs/core/series.pyi
```diff
@@ -4163,7 +4689,23 @@ class Series(IndexOpsMixin[S1], NDFrame):
         **kwargs,
     ) -> np_1darray[GenericT]: ...
     @overload
+    def to_numpy(  # pyrefly: ignore[bad-override]
+        self: Series[Timedelta],
+        dtype: None = None,
```
```suggestion
        dtype: None | type[np.timedelta64] = None,
```
If you do this change, it fixes the issue you were having with `mypy` in `test_timefuncs.py`

pandas-stubs/_typing.pyi
```diff
@@ -825,6 +825,14 @@ np_ndarray_td: TypeAlias = npt.NDArray[np.timedelta64]
...
 # Define shape and generic type variables with defaults similar to numpy
 GenericT = TypeVar("GenericT", bound=np.generic, default=Any)
+TD64ItemT = TypeVar(
+    "TD64ItemT",
+    bound=datetime.timedelta | int | None,
+    default=datetime.timedelta | int | None,
+)
+DT64ItemT = TypeVar(
+    "DT64ItemT", bound=datetime.date | int | None, default=datetime.date | int | None
+)
```
Dr-Irv: since we're not using these, you can remove them

pandas-stubs/core/series.pyi
```diff
@@ -4466,7 +4461,7 @@ class Series(IndexOpsMixin[S1], NDFrame):
         level: None = ...,
         numeric_only: _bool = False,
         **kwargs: Any,
-    ) -> float: ...
+    ) -> Any: ...
```
Dr-Irv: I'd rather leave this as `float`.  While it's true that the untyped `Series` coming from a column in a `DataFrame` could be a series of `Timestamp`, most people will want the type of the result of this to be `float`.

pandas-stubs/core/series.pyi
```diff
@@ -4632,7 +4627,26 @@ class Series(IndexOpsMixin[S1], NDFrame):
...
     @overload
     def std(
-        self: Series[Timestamp] | Series[Timedelta],
+        self: Series[Never],
+        axis: AxisIndex | None = 0,
+        skipna: _bool | None = True,
+        ddof: int = 1,
+        numeric_only: _bool = False,
+        **kwargs: Any,
+    ) -> Any: ...
```
Dr-Irv: similar comment - most people will use `std` on floating/int columns, so we can be safe here and make the result `float`

pandas-stubs/core/series.pyi
```diff
@@ -3983,25 +4488,34 @@ class Series(IndexOpsMixin[S1], NDFrame):
         level: None = ...,
         numeric_only: _bool = False,
         **kwargs: Any,
+    ) -> Any: ...
```
Dr-Irv: same comment as for `mean` and `std`.  Most people will want the result to be `float`.  Otherwise they have to `cast` to `Series[float]` to get the right type of the result (or cast the result)

Dr-Irv (changes requested):
figured out the `to_numpy` issue with `mypy`.  

Would rather have `median`, `mean` and `std` return `float` on untyped `Series` as it is more user friendly that way.

pandas-stubs/core/series.pyi
cmp0xff: f094cd75575d9f7e7b59438d0820c7ed91811443

pandas-stubs/core/series.pyi
cmp0xff: f094cd75575d9f7e7b59438d0820c7ed91811443

pandas-stubs/core/series.pyi
cmp0xff: f094cd75575d9f7e7b59438d0820c7ed91811443

pandas-stubs/core/series.pyi
cmp0xff: f094cd75575d9f7e7b59438d0820c7ed91811443

pandas-stubs/core/series.pyi
cmp0xff: f094cd75575d9f7e7b59438d0820c7ed91811443

pandas-stubs/_typing.pyi
cmp0xff: f094cd75575d9f7e7b59438d0820c7ed91811443

Dr-Irv (approved):
thanks @cmp0xff   major accomplishment!

Dr-Irv:
@cmp0xff I have just released v2.3.2.250926 so we'll see what feedback we will get

PR #1277, opened 2025-07-14
GH1269 Add @Final decorator to match pandas code where possible

### LINKED ISSUE #1269: How we handle @final decorator in the stubs
Opening this issue more as a discussion, I was doing some cleanup with the `stubtest` task and I have noticed that some of the methods in pandas annotated with `@final` are not always annotated as such in the stubs.
This generates around 500 warnings with the `stubtest` tasks. I know we have discussed it some times in PRs but wanted to open this issue to clarify how we should stand on it.
The main points are:

- Should we also annotate the stubs with `@final` (it is partially done right now)?
- What is the impact of having it? (or not having it)
- Does it make a difference to have it or not?

loicdiridollou:
- [x] Closes #1269 
- [x] Tests added: Please use `assert_type()` to assert the type of any return value NOTHING TO DO

Still some places where `stubtest` would complain due to inheritance of the current class (like Index and DatetimeIndex).


pandas-stubs/io/common.pyi
```diff
@@ -9,9 +10,14 @@ from pandas._typing import CompressionDict
 class IOHandles(Generic[AnyStr]):
```
twoertwein, line 10: I'm fairly certain this class is considered private, could remove it from the stubs.

twoertwein (approved):
It would be nice to have tests for the finals, but that does not seem worth it for now. Hopefully we can run (parts of) subtests in the future on the CI.

twoertwein:
Thank you @loicdiridollou!

pandas-stubs/io/common.pyi
loicdiridollou: Let me check on that and we could eventually drop from the stubs.

PR #1275, opened 2025-07-13
feat(series): #1098 __add__ and __mul__

cmp0xff:
This PR partially addresses #1098 and updates `__add__` and `__mul__`.

The task proved to be tricky, and I have the feeling that pyright and mypy do not support `self` typing very well. I did not make `__sub__` and `__div__` as well as `__truediv__` work.

# To do
- `add`, `__radd__` and `radd`
- `__sub__`, `sub`, `__rsub__` and `rsub`
- `mul`, `__rmul__` and `rmul`
- `__div__`, `div`, `__rdiv__` and `rdiv`
- Other arithmetic operations

# Checklist
- [x] Tests added: Please use `assert_type()` to assert the type of any return value

PR #1280, opened 2025-07-17
feat(series): #1098 arithmetic truediv

cmp0xff:
- [x] Addresses #1098 and continues the programme proposed in #1275


Dr-Irv (changes requested):
This all looks good, but can you reorganize the operations to keep things together, in this order:

1. `__truediv__()`
2. `truediv()`
3. `__rtruediv__()`
4. `rtruediv()`

Also, you don't need to do the overloads for `div` and `rdiv`.  You can just do
```python
div = truediv
rdiv = rtruediv
```

PR #1282, opened 2025-07-18
feat(index): append

cmp0xff:

`Index.append` used not to be typed. In this PR, typings for `Index.append` are added and tested.


pandas-stubs/core/indexes/base.pyi
```diff
@@ -401,7 +401,12 @@ class Index(IndexOpsMixin[S1]):
     ) -> Self: ...
     @overload
     def __getitem__(self, idx: int | tuple[np_ndarray_anyint, ...]) -> S1: ...
-    def append(self, other): ...
+    @overload
+    def append(self, other: Index[Never]) -> Index: ...
```
twoertwein: Do you need this first overload here?

For any/unknown the second overload might already behaves as expected?

pandas-stubs/core/indexes/base.pyi
```diff
@@ -401,7 +401,12 @@ class Index(IndexOpsMixin[S1]):
...
+    @overload
+    def append(self, other: Index[S1] | Sequence[Index[S1]]) -> Index[S1]: ...
```
twoertwein: it might be nicer here to have a second S1 here as the resulting Index can contain a mix of different types.

`def append(self, other: Index[S2] | Sequence[Index[S2]]) -> Index[S1 | S2]: ...`

pandas-stubs/core/indexes/base.pyi
cmp0xff: 3a3405797429eb262949b6b9d3c7e6d6a0d301bd.

pandas-stubs/core/indexes/base.pyi
cmp0xff: 45e0ad30fd18379d6b97f74d8e2a68ca37c3fa3a, but this one works less well.
- `mypy` is not happy with `Index[int].append(Index[int | str])` and gives `Index[Any]`
- `pyright` is not happy with `Index[int | str].append([Index[int], Index[str]])` and gives `Index[int | Any]`. In particular, the typing for `[Index[int], Index[str]]` seems to be `list[Index[int] | Index[str]]`, instead of `list[Index[int | str]]`.

twoertwein: > * `mypy` is not happy with `Index[int].append(Index[int | str])` and gives `Index[Any]`
> * `pyright` is not happy with `Index[int | str].append([Index[int], Index[str]])` and gives `Index[int | Any]`. In particular, the typing for `[Index[int], Index[str]]` seems to be `list[Index[int] | Index[str]]`, instead of `list[Index[int | str]]`.

While that is annoying for testing on the CI, I think that is the safer choice for user: rather expect a wider type that includes `Any` than suggesting it is a narrower type. This needs input from @Dr-Irv.

If `S1` and `S2` were covariant, it seems to work for at least pyright in a simple toy example (but they are invariant)

```py
from typing import TypeVar, reveal_type, Generic, Sequence

S1 = TypeVar("S1", bound=int | str, covariant=True)
S2 = TypeVar("S2", bound=int | str, covariant=True)

class Index(Generic[S1]):
    def __init__(self, data: list[S1]) -> None: ...

    def append(self: Index[S1], other: Sequence[Index[S2]]) -> Index[S1 | S2]: ...

strings = Index(["a"])
reveal_type(strings)
ints = Index([1])
reveal_type(ints)

reveal_type(strings.append([ints]))
reveal_type(ints.append([strings]))

string_ints = Index(["a", 1])
reveal_type(string_ints)
reveal_type(string_ints.append([ints]))
reveal_type(strings.append([string_ints]))

reveal_type(strings.append([ints]))
reveal_type(strings.append([strings, ints]))

reveal_type(strings.append([ints, strings, string_ints]))
```

cmp0xff: > If `S1` and `S2` were covariant, it seems to work for at least pyright in a simple toy example (but they are invariant)
> 
> ```python
> from __future__ import annotations
> from typing import TypeVar, reveal_type, Generic, Sequence
> 
> S1 = TypeVar("S1", bound=int | str, covariant=True)
> S2 = TypeVar("S2", bound=int | str, covariant=True)
> 
> class Index(Generic[S1]):
>     def __init__(self, data: list[S1]) -> None: ...
> 
>     def append(self: Index[S1], other: Sequence[Index[S2]]) -> Index[S1 | S2]: ...
> ```
Hi, I am new to covariance / contravariance, but I read [PEP484 (covariance-and-contravariance)](https://peps.python.org/pep-0484/#covariance-and-contravariance) and it says `covariant` is for classes, not for functions, where the latter case is prohibited. In your example, `S1` is find, but not `S2`. Could you help me and explain a bit? Thanks.
> ```py
> B_co = TypeVar('B_co', covariant=True)
>
> def bad_func(x: B_co) -> B_co:  # Flagged as error by a type checker
>     ...
> ```

twoertwein: We can't change `S1` to be covariant.  While the following is not exactly what we like to have, it is probably the closest we can get (but it doesn't work with mypy, unless the caller casts).


```py
from typing import TypeVar, reveal_type, Generic, Sequence

S1 = TypeVar("S1", bound=int | str)
IndexT = TypeVar("IndexT", bound="Index")


class Index(Generic[S1]):
    def __init__(self, data: list[S1]) -> None: ...

    def append(self: Index[S1], other: Sequence[IndexT]) -> Index[S1] | IndexT: ...


strings = Index(["a"])
reveal_type(strings)
ints = Index([1])
reveal_type(ints)

reveal_type(strings.append([ints]))
reveal_type(ints.append([strings]))

string_ints = Index(["a", 1])
reveal_type(string_ints)
reveal_type(string_ints.append([ints]))
reveal_type(strings.append([string_ints]))

reveal_type(strings.append([ints]))
reveal_type(strings.append([strings, ints]))

reveal_type(strings.append([ints, strings, string_ints]))
```

cmp0xff: Hi, I ran the script myself. In the most complicated case, I see `Index[int | str] | Index[int] | Index[str]`. To be honest, as a user I would rather see `Index[Unknown]`, because it's simpler, and in both cases I would probably still need a manual `cast`. Nevertheless, 3844062b0df37d810c10744b226d2aa9641c4286

pandas-stubs/core/indexes/base.pyi
```diff
@@ -401,7 +405,14 @@ class Index(IndexOpsMixin[S1]):
     ) -> Self: ...
     @overload
     def __getitem__(self, idx: int | tuple[np_ndarray_anyint, ...]) -> S1: ...
-    def append(self, other): ...
+    @overload
+    def append(self, other: Index[S1] | Sequence[Index[S1]]) -> Self: ...
+    @overload
+    def append(self, other: Index[S2] | Sequence[Index[S2]]) -> Index[S1 | S2]: ...
```
Dr-Irv: I don't think that's right.   I think the result would be `Index[S1] | Index[S2]` .  

I'm not sure how the downstream `Index` stuff would work with the union type inside the generic.

pandas-stubs/core/indexes/base.pyi
```diff
@@ -401,7 +405,14 @@ class Index(IndexOpsMixin[S1]):
...
+    @overload
+    def append(self, other: Sequence[_T_INDEX]) -> Self | _T_INDEX: ...
```
Dr-Irv: Why can't we use `Sequence[Index]` here without the `TypeVar` ?

tests/test_indexes.py
```diff
@@ -1028,6 +1030,66 @@ def test_getitem() -> None:
...
+        assert_type(third.append(cast("list[Index[Any]]", [])), "pd.Index[int | str]"),  # type: ignore[assert-type]
+        pd.Index,
+    )
+    check(assert_type(third.append([first]), "pd.Index[int | str]"), pd.Index)  # type: ignore[assert-type]
+    check(
+        assert_type(  # type: ignore[assert-type]
+            third.append([first, second]),  # pyright: ignore[reportAssertTypeFailure]
+            "pd.Index[int | str]",
+        ),
+        pd.Index,
```
Dr-Irv: these tests need to work without having `ignore` in them.  So you'll need to fix the types in the `append()` declarations to make that happen.

tests/test_indexes.py
```diff
@@ -1028,6 +1030,66 @@ def test_getitem() -> None:
...
+    check(assert_type(first.append([second]), "pd.Index[str]"), pd.Index, str)
+
+
+def test_append_list_str() -> None:
+    """Test pd.Index[list[str]].append"""
+    first = pd.Index([["str", "rts"]])
+    second = pd.Index([["srt", "trs"]])
+    check(assert_type(first.append([]), "pd.Index[list[str]]"), pd.Index, list)
+    check(assert_type(first.append(second), "pd.Index[list[str]]"), pd.Index, list)
+    check(assert_type(first.append([second]), "pd.Index[list[str]]"), pd.Index, list)
```
Dr-Irv: I don't think we should be supporting `Index[list[str]]` because `list[str]` is not hashable and labels in an Index should be hashable.

But this is a bug in `pandas`, I think.  See https://github.com/pandas-dev/pandas/issues/61937

So can you remove this test.

Then we will have to separate out `list[str]` from `S1` and have an `I1` that includes everything in `S1` except `list[str]`, while `S1` includes everything.

So can you make that change as well?

cmp0xff: Removing `Index[list[str]]` is hard. The following code runs:
```py
import pandas as pd

pd.Index(["a_b"]).str.split("_")
```
It produces `Index([['a', 'b']], dtype='object')`. We have `Index[S1]` all over `pandas-stubs`.

pandas-stubs/core/indexes/base.pyi
cmp0xff: When we append `Index([1])` to `Index(["a"])`, we actually get `Index(["a", 1])`, which has the type `Index[str | int]`.

Dr-Irv: If we can't narrow it to a specific type that is one of the types in `S1`, then I think it will need to be just `Index` without specifying the subtype, or `Index[Any]`, or even `UnknownIndex` .

Right now, if you did `Index(["a", 1])`, the revealed type by pyright is `Index[str | int]`, but with mypy it is `Index[Any]`, so I think I want to avoid placing unions inside the generic type because I don't know what else will break and there could be inconsistencies in the type checkers in handling this.

tests/test_indexes.py
Dr-Irv: Yes, but it really isn't supported to have `list[str]` in an Index:
```python
ind = pd.Index(["a_b", "c_d"])
spl = ind.str.split("_")
spl.duplicated()
```
This will create an exception because the list is unhashable.

Might be best in this PR to just leave out the test with `append` and `Index[list[str]]` and not worry for now about splitting it out until the issue in pandas is straightened out.

cmp0xff: ac0857ba032422c8342c616b6479ae2054416d3a

pandas-stubs/core/indexes/base.pyi
cmp0xff: 67e6bde96ea6894aa32b827610347f76c2728989, except for that `mypy` is unable to handle `pd.Index([1, "a"]).append(pd.Index([1, "a"]))` and gives `Index[Any]`.

pandas-stubs/core/indexes/base.pyi
cmp0xff: The idea came from https://github.com/pandas-dev/pandas-stubs/pull/1282#discussion_r2217319354. Removed in 67e6bde96ea6894aa32b827610347f76c2728989.

tests/test_indexes.py
cmp0xff: Three `type: ignore` remains, which I believe are `mypy` bugs. I tried to add
```py
    @overload
    def append(self, other: Sequence[Never]) -> Self: ...
```
as the first overload, which did not help.

pandas-stubs/core/indexes/base.pyi
```diff
@@ -408,11 +405,9 @@ class Index(IndexOpsMixin[S1]):
     @overload
     def append(self, other: Index[S1] | Sequence[Index[S1]]) -> Self: ...
     @overload
-    def append(self, other: Index[S2] | Sequence[Index[S2]]) -> Index[S1 | S2]: ...
-    @overload
-    def append(self, other: Sequence[_T_INDEX]) -> Self | _T_INDEX: ...
+    def append(self, other: Index[S2]) -> Index[S1 | S2]: ...
```
Dr-Irv:
```suggestion
    def append(self, other: Index[S2]) -> Index: ...
```
I really want to avoid having the union types here.

tests/test_indexes.py
```diff
@@ -1036,31 +1036,31 @@ def test_append_mix() -> None:
...
-    check(assert_type(first.append([second]), "pd.Index[int | str]"), pd.Index)
+    check(assert_type(first.append([second]), pd.Index), pd.Index)
 
-    check(assert_type(first.append(third), "pd.Index[int | str]"), pd.Index)  # type: ignore[assert-type]
-    check(assert_type(first.append([third]), "pd.Index[int | str]"), pd.Index)  # type: ignore[assert-type]
     check(
         assert_type(  # type: ignore[assert-type]
-            first.append([second, third]),  # pyright: ignore[reportAssertTypeFailure]
-            "pd.Index[int | str]",
+            first.append(third), "pd.Index[int | str]"
```
Dr-Irv: Should be `Index[Any]` (or just `pd.Index`)

Dr-Irv (changes requested):
Let's not return things like `Index[str | int]` and just make that `Index`

pandas-stubs/core/indexes/base.pyi
cmp0xff: d7a538d85b2a6cb93fc6e18b8afb4b3b289fe03f

tests/test_indexes.py
cmp0xff: d7a538d85b2a6cb93fc6e18b8afb4b3b289fe03f. Now just two `# type: ignore[assert-type]`s remain.

tests/test_indexes.py
```diff
@@ -1028,6 +1030,52 @@ def test_getitem() -> None:
...
+
+    check(
+        assert_type(  # type: ignore[assert-type]
+            third.append([]), "pd.Index[int | str]"
+        ),
+        pd.Index,
+    )
+    check(
+        assert_type(  # type: ignore[assert-type]
+            third.append(cast("list[Index[Any]]", [])), "pd.Index[int | str]"
```
Dr-Irv: Change these to `pd.Index` in the `assert_type` statements, and then I don't think you need the `# type: ignore` statements., i.e., remove the union part of the generic paramter.

cmp0xff: I added comments in fe5f1a9036102a136f82557c7b3e3ace937803a7.
- We have the following overloads
    ```py
    @overload
    def append(self, other: Index[S1] | Sequence[Index[S1]]) -> Self: ...
    @overload
    ```
- `pyright` uses the first overload and gives `Index[str | int]`
- `mypy` uses the second overload and gives `Index[Any]`
- Even if I add the following as the _first_ overload
    ```py
    @overload
    def append(self, other: Sequence[Never]) -> Self: ...
    ```
    `mypy` still uses the _last_ overload and gives `Index[Any]`

I tend to believe this is a `mypy` bug.

cmp0xff: After all, `Index(["a", 1])` was created as the type `Index[str | int]` from the beginning.

cmp0xff: How about 1c75df60c1a4c901204dbef1edc389032b73fd8c? I am trying to get rid of `Any` by using `constraint` instead of `bound` in the `TypeVar`.

Dr-Irv: That is fine.  Have to wonder if we should change `Index` to be based on `C2` instead of `S1`, but that can be another PR.

Dr-Irv (approved):
thanks @cmp0xff

PR #1283, opened 2025-07-19
fix(DataFrame): #799 to_dict

### LINKED ISSUE #799: DataFrame.to_dict(orient="index") should return a nested dictionary
**Describe the bug**
When you do `DataFrame.to_dict(orient="index")`, and you are using default integer index and string columns, you will get a `Dict[int, Dict[str, Any]]` at runtime. The current annotations do not reflect this nested dictionary, they instead return a `Dict[Hashable, Any]`. I understand that the exact type this returns has changed slightly after recent commits by @twoertwein, but the issue still stands.

Expectation: returns `MutableMapping[Hashable, MutableMapping[Hashable, Any]]`

This issue could be broadened to more precisely annotate all possible values of `orient` if desired. For instance, `orient="split"` probably should return a `TypedDict` that has `index` and `columns` as keys.

**To Reproduce**
Any usage of `to_dict(orient="index")` should reproduce this. Below is a screenshot of an example:

image

cmp0xff:
- [x] Closes #799 (Replace xxxx with the Github issue number)


pandas-stubs/core/frame.pyi
```diff
@@ -410,39 +422,55 @@ class DataFrame(NDFrame, OpsMixin, _GetItemHack):
     @overload
     def to_dict(
         self,
-        orient: Literal["dict", "list", "series", "index"],
+        orient: Literal["index"],
+        *,
+        into: _T_MUTABLE_MAPPING | type[_T_MUTABLE_MAPPING],
+        index: Literal[True] = ...,
+    ) -> MutableMapping[Hashable, _T_MUTABLE_MAPPING]: ...
```
cmp0xff: I do now know how to get rid of the `MutableMapping` here...

twoertwein: Is `into` used for the outer mapping, the inner mapping, or both at runtime?

I assume we would need somthing like `_T_MUTABLE_MAPPING[Hashable, dict[Hashable, Any]]`, if TypeVar would support [] (it doesn't) and if the inner mapping is always a dict (I don't know).

I'm not sure if we can annotate both the correct sub-class and also the keys/values.

cmp0xff: > Is `into` used for the outer mapping, the inner mapping, or both at runtime?

Good point! 
```py
pd.DataFrame({("a", "b"): [[1], [2, 3]]}).to_dict("index", into=defaultdict(list))
```
the result is actually
```py
defaultdict(<class 'list'>, {0: {('a', 'b'): [1]}, 1: {('a', 'b'): [2, 3]}})
```
Only the outer mapping is the type specified by `into`. The inner mapping are `dict`s.

pandas-stubs/core/frame.pyi
```diff
@@ -410,51 +425,67 @@ class DataFrame(NDFrame, OpsMixin, _GetItemHack):
...
-    ) -> MutableMapping[Hashable, Any]: ...
+    ) -> MutableMapping[Hashable, dict[Hashable, Any]]: ...
+    @overload
+    def to_dict(
+        self,
+        orient: Literal["index"],
+        *,
+        into: type[dict] = ...,
+        index: Literal[True] = ...,
+    ) -> dict[Hashable, dict[Hashable, Any]]: ...
```
cmp0xff: - Hard-coded `defaultdict`, `OrderedDict` and `dict`
- Otherwise, `to_dict` now gives `MutableMapping`

pandas-stubs/core/frame.pyi
```diff
@@ -392,13 +399,21 @@ class DataFrame(NDFrame, OpsMixin, _GetItemHack):
         na_value: Scalar = ...,
     ) -> np.ndarray: ...
     @overload
+    def to_dict(
+        self,
+        orient: str = ...,
+        *,
+        into: type[defaultdict],
+        index: Literal[True] = ...,
+    ) -> Never: ...
```
cmp0xff, line 463: `into` receiving a type `defaultdict` is forbidden, see https://github.com/pandas-dev/pandas/blob/v2.3.1/pandas/core/common.py#L416-L417. Hence returning `Never`.

pandas-stubs/core/frame.pyi
```diff
@@ -392,13 +399,21 @@ class DataFrame(NDFrame, OpsMixin, _GetItemHack):
...
+    @overload
     def to_dict(
         self,
         orient: Literal["records"],
         *,
-        into: MutableMapping | type[MutableMapping],
+        into: _T_MUTABLE_MAPPING | type[_T_MUTABLE_MAPPING],
         index: Literal[True] = ...,
-    ) -> list[MutableMapping[Hashable, Any]]: ...
+    ) -> list[_T_MUTABLE_MAPPING]: ...
```
cmp0xff, line 471: Made covariant

pandas-stubs/core/frame.pyi
```diff
@@ -410,51 +425,67 @@ class DataFrame(NDFrame, OpsMixin, _GetItemHack):
...
         self,
-        orient: Literal["split", "tight"] = ...,
+        orient: Literal["dict", "list", "series"] = ...,
         *,
-        into: MutableMapping | type[MutableMapping],
-        index: bool = ...,
-    ) -> MutableMapping[Hashable, Any]: ...
+        into: _T_MUTABLE_MAPPING | type[_T_MUTABLE_MAPPING],
+        index: Literal[True] = ...,
+    ) -> _T_MUTABLE_MAPPING: ...
```
cmp0xff: Covariant

pandas-stubs/core/frame.pyi
```diff
@@ -410,51 +425,67 @@ class DataFrame(NDFrame, OpsMixin, _GetItemHack):
...
+    ) -> MutableMapping[str, list]: ...
+    @overload
+    def to_dict(
+        self,
+        orient: Literal["split", "tight"],
         *,
         into: type[dict] = ...,
         index: bool = ...,
-    ) -> dict[Hashable, Any]: ...
+    ) -> dict[str, list]: ...
```
cmp0xff, line 543: The old typing was wrong, now fixed.

pandas-stubs/core/frame.pyi
cmp0xff: With afb5729c46d497a5615323ce3b47bd5c1548d9b1, now if `into` is `defaultdict` or `OrderedDict`, the outer mapping is of the corresponding type, and the inner mapping is correctly `dict`.

If `into` is a more generic `MutableMapping`, however, I do not see a way to make the outer mapping covariant and fix the inner mapping to `dict` at the same time. I choose to make the outer mapping just the generic `MutableMapping`, and the inner mapping `dict`.

tests/test_frame.py
```diff
@@ -3638,33 +3642,86 @@ def test_to_records() -> None:
...
     check(
-        assert_type(DF.to_dict(into=target), MutableMapping[Hashable, Any]), defaultdict
+        assert_type(data.to_dict(into=target), defaultdict[Any, list]),
+        defaultdict,
+        tuple,
     )
-    target = defaultdict(list)
     check(
-        assert_type(DF.to_dict("tight", into=target), MutableMapping[Hashable, Any]),
+        assert_type(  # type: ignore[assert-type]
```
Dr-Irv: I don't want to see `ignore` in test files (unless we are testing invalid usage).  So can you fix this?

Dr-Irv (changes requested):

tests/test_frame.py
cmp0xff: The overload
```py
    def to_dict(
        self,
        orient: Literal["index"],
        *,
        into: defaultdict,
        index: Literal[True] = ...,
    ) -> defaultdict[Hashable, dict[Hashable, Any]]: ...
```
should give the correct typing. However, with the following testing script
```py
import pandas as pd
from collections import defaultdict
from typing import Any, reveal_type

data = pd.DataFrame({("str", "rts"): [[1, 2, 4], [2, 3], [3]]})
target: defaultdict[Any, list] = defaultdict(list)

reveal_type(data.to_dict("index", into=target))
```
`mypy` 1.17.0 gives `attempt.py:8: note: Revealed type is "Any"`. In contract, `pyright` gives `attempt.py:8:13 - information: Type of "data.to_dict("index", into=target)" is "defaultdict[Hashable, dict[Hashable, Any]]"`.

I think it is a `mypy` bug.

Dr-Irv: Yes, was able to recreate at   https://github.com/python/mypy/issues/19525

So in this case, can you put a comment in the test with a link to that issue to indicate why the `# type: ignore` is there?

pandas-stubs/core/frame.pyi
brianschubert: Hi! Just a heads up that this is actually a common misuse of Never that can have some unintended consequences. You can find more information about what problems this can cause in this DPO thread and the links therein: https://discuss.python.org/t/specifying-invalid-signatures-in-overloads/75405. Unfortunately the ability to mark certain overloads as invalid is an open issue in Python's type system. One workaround I've seen other projects use is marking the invalid overload as `@deprecated`, which will make type checkers emit a diagnostic when the overload is matched without otherwise affecting type checking

tests/test_frame.py
cmp0xff: How about 81c57e3aa6c41096b53947d25611b2384656dfd6?

Dr-Irv: that works.  Seems like what you were trying before wasn't correct with `pyright` based on the discussion in the issue I created for `mypy`.

Dr-Irv (approved):
thanks @cmp0xff

cmp0xff:
Hi @Dr-Irv, thank you for the approval. However, regarding https://github.com/pandas-dev/pandas-stubs/pull/1283#discussion_r2238130149, do we have some guideline here?

cmp0xff:
Hi @Dr-Irv , should we resolve the conflict and merge it?

Dr-Irv (approved):
thanks @cmp0xff

PR #1286, opened 2025-07-23
Improve results of stubtest

loicdiridollou:
Clean up some old pandas 2.2 code that was removed, improved the results of the stubtest.


pandas-stubs/core/series.pyi
```diff
@@ -1579,6 +1584,7 @@ class Series(IndexOpsMixin[S1], NDFrame):
         periods: int = ...,
         fill_method: None = ...,
         freq: DateOffset | timedelta | _str | None = ...,
+        limit: int | None = ...,  # deprecated since 2.1
```
twoertwein: not sure if we want to add a deprecated argument. If we add it, we could add it with just its default value `limit: None`

pandas-stubs/io/parquet.pyi
```diff
@@ -14,5 +15,9 @@ def read_parquet(
     engine: ParquetEngine = ...,
     columns: list[str] | None = ...,
     storage_options: StorageOptions = ...,
+    use_nullable_dtypes: bool = ...,  # deprecated since 2.0
```
twoertwein: can probably be removed

pandas-stubs/core/series.pyi
```diff
@@ -1132,6 +1135,7 @@ class Series(IndexOpsMixin[S1], NDFrame):
         freq: DateOffset | timedelta | _str | None = ...,
         axis: Axis = ...,
         fill_value: Scalar | NAType | None = ...,
+        suffix: _str | None = ...,  # has no effect at the moment
```
Dr-Irv: Probably should split into 2 overloads, based on the docs.  One is where `periods` is `int` and there is no `suffix` argument, and the other where `periods` is `Sequence[int]` and then `suffix` is allowed.

I don't understand what you mean by "has no effect at the moment"

pandas-stubs/core/series.pyi
```diff
@@ -1217,7 +1222,7 @@ class Series(IndexOpsMixin[S1], NDFrame):
     def droplevel(self, level: Level | list[Level], axis: AxisIndex = ...) -> Self: ...
     def pop(self, item: Hashable) -> S1: ...
     @final
-    def squeeze(self) -> Series[S1] | Scalar: ...
+    def squeeze(self, axis: Axis = ...) -> Series[S1] | Scalar: ...
```
```suggestion
    def squeeze(self, axis: None = ...) -> Series[S1] | Scalar: ...
```
The only valid value here is `None` for `axis`

pandas-stubs/core/series.pyi
Dr-Irv: we shouldn't add deprecated arguments.  so remove this

pandas-stubs/io/parquet.pyi
Dr-Irv: YEs, I agree.  Should remove.

Dr-Irv (changes requested):

Dr-Irv:
> Thanks @twoertwein for the review, I was debating myself about the deprecated args, they appear in the stubtest since the code is still there. Prob worth opening a discussion with @Dr-Irv how we want to handle it, a part of me want to drop them since we should not use them but the stubtest complains. Not sure there is a perfect answer here, the stubtest is super useful as it uncovers many bits of code that were deleted from the pandas code but it of course can't handle deprecation like we would like.

I prefer to not have deprecated arguments in the stubs, because we want people to stop using them.

Maybe there is a way to tell `stubtest` to ignore certain differences?

pandas-stubs/core/series.pyi
loicdiridollou: If you read the source code it is not used at all.
https://github.com/pandas-dev/pandas/blob/4f952b79f91f4f9d4d86b46d34e78dc657bfd4a2/pandas/core/generic.py#L10122-L10130

Maybe I missed something in the code but not sure.

loicdiridollou:
When it comes to stubtest I don't think there is a way for it to filter since the errors that we get is that the arguments in the stubs are misaligned with the runtime version (it is impossible for stubtest to know what is deprecated in pandas).
That is why I felt like reintroducing the arguments would help clear out the errors but as you said it is not the better way to go.

pandas-stubs/core/series.pyi
Dr-Irv: I created https://github.com/pandas-dev/pandas/issues/61955

so let's delete it from the stubs

Dr-Irv (changes requested):

Dr-Irv (approved):
thanks @loicdiridollou

PR #1289, opened 2025-07-24
fix nightly runs

Dr-Irv:
NIghtly runs failed due to tests related to passing in compiled regex when using pyarrow.  This also identified a couple of bugs in pandas with the new string Dtype

@twoertwein also thanks for reviews while I was on vacation!


twoertwein (approved):

twoertwein:
Thanks @Dr-Irv !

PR #1295, opened 2025-07-27
Rename internal strings accessor module to match runtime

hamdanal:
This is a simple rename of the internal strings accessor module to match the runtime module name for the following reasons:

For users:
- This fixes editors like VS Code not displaying the documentation of the methods like `series.str.replace` on hover because they couldn't match the runtime docstring to the stub which was living in a different module
- The additional nesting (`pandas.core.strings.accessor` instead of `pandas.core.strings`) makes it less desirable for users to reach out to the class that is not considered public api

For the stubs developers:
- When working on the stubs, having the name matching runtime makes it easier to navigate between the source and the stubs in an IDE as the "Go to definition" and "Go to type definition" actions now work correctly
- If and when you start using stubtest as part of your infrastructure, this is one less module to worry about as stubtest will not complain about a missing stub and a non-existent runtime


hamdanal:
This is how VS Code with Pylance is able to display documentation on hover now
image

Dr-Irv (approved):
Thanks @hamdanal

PR #1297, opened 2025-07-28
Fix #849: Update converters type in read_excel for better Pyright compatibility

### LINKED ISSUE #849: Type checking with `converters=` on `read_excel`
**Describe the bug**
Pyright doesn't seem to like something about using `converters=...` with `read_excel`

**To Reproduce**
```py
from functools import partial

import pandas as pd

df_1 = pd.read_excel(
    "foo.csv",
    converters={"field_1": partial(pd.to_datetime, errors="coerce")},
)
df_2 = pd.read_excel(
    "foo.csv",
    sheet_name="sheet",
    converters={"field_1": partial(pd.to_datetime, errors="coerce")},
)

reveal_type(df_1)
reveal_type(df_2)
```
2. Indicate which type checker you are using (`mypy` or  `pyright`).: Both

mypy:
```
$ mypy t.py
t.py:15: note: Revealed type is "pandas.core.frame.DataFrame"
t.py:16: note: Revealed type is "pandas.core.frame.DataFrame"
Success: no issues found in 1 source file
```

pyright:
```
/home/user/t.py
  /home/user/t.py:7:16 - error: Argument of type "dict[str, partial[Timestamp]]" cannot be assigned to parameter "converters" of type "dict[int | str, (object) -> object] | None" in function "read_excel"
    Type "dict[str, partial[Timestamp]]" cannot be assigned to type "dict[int | str, (object) -> object] | None"
      "dict[str, partial[Timestamp]]" is incompatible with "dict[int | str, (object) -> object]"
        Type parameter "_KT@dict" is invariant, but "str" is not the same as "int | str"
        Type parameter "_VT@dict" is invariant, but "partial[Timestamp]" is not the same as "(object) -> object"
      "dict[str, partial[Timestamp]]" is incompatible with "None" (reportGeneralTypeIssues)
  /home/user/t.py:9:8 - error: No overloads for "read_excel" match the provided arguments (reportGeneralTypeIssues)
  /home/user/t.py:12:16 - error: Argument of type "dict[str, partial[Timestamp]]" cannot be assigned to parameter "converters" of type "dict[int | str, (object) -> object] | None" in function "read_excel"
    Type "dict[str, partial[Timestamp]]" cannot be assigned to type "dict[int | str, (object) -> object] | None"
      "dict[str, partial[Timestamp]]" is incompatible with "dict[int | str, (object) -> object]"
        Type parameter "_KT@dict" is invariant, but "str" is not the same as "int | str"
        Type parameter "_VT@dict" is invariant, but "partial[Timestamp]" is not the same as "(object) -> object"
      "dict[str, partial[Timestamp]]" is incompatible with "None" (reportGeneralTypeIssues)
  /home/user/t.py:15:13 - information: Type of "df_1" is "DataFrame"
  /home/user/t.py:16:13 - information: Type of "df_2" is "Unknown"
```

 - WSL - ubuntu 22.04
 - python version `3.10.12`
 - version of type checker: `pyright==1.1.345`, `mypy==1.8.0`
 - version of installed `pandas-stubs`: `pandas_stubs==1.4.4.220919` (also present on latest pandas_stubs version for most recent pandas version)

**Additional context**
Not sure if its something with the converters type or read_excel, but using a lambda instead of partial results in both type checkers not liking it even though `read_csv` has no issue with either ie

```py
converters={"field": lambda s: pd.to_datetime(s, errors="coerce")}
#vs
converters={"field": partial(pd.to_datetime, errors="coerce")}

```

Sachi-Parashar:
### Summary

This PR addresses issue [#849](https://github.com/pandas-dev/pandas-stubs/issues/849), where `pyright` fails to type-check the `converters` argument in `read_excel` when using `functools.partial` or lambda functions.

### Fix

Updated all `read_excel` and `parse` overloads in `pandas/io/excel/_base.pyi`:

```diff
- converters: Mapping[int | str, Callable[[object], object]] | None
+ converters: Mapping[int | str, Callable[[Any], Any]] | None
```
This relaxes the type signature to accept functools.partial and similar callables.

### Test
Added a test tests/test_excel_converters.py with assert_type to confirm the fix works with pyright and mypy.
```python
from functools import partial
import pandas as pd

conv = partial(pd.to_datetime, errors="coerce")
df = pd.read_excel("foo.xlsx", converters={"col1": conv})
assert_type(df, pd.DataFrame)
```
###  Notes
-  Verified with both mypy and pyright. - Worked✅ 
- If needed, I'm open to updating ConvertersArg in _typing.pyi for broader consistency in a follow-up PR.

Closes #849


tests/test_excel_converters.py
```diff
@@ -0,0 +1,9 @@
+from functools import partial
+import pandas as pd
+from typing_extensions import assert_type
+
+partial_func = partial(pd.to_datetime, errors="coerce")
+
+df = pd.read_excel("foo.xlsx", converters={"field_1": partial_func})
+
+assert_type(df, pd.DataFrame)
```
Dr-Irv: This should go in `test_io.py` and be in a `test_converters_partial()` function

pandas-stubs/io/excel/_base.pyi
```diff
@@ -54,7 +54,7 @@ def read_excel(
     usecols: str | UsecolsArgType = ...,
     dtype: str | Dtype | Mapping[str, str | Dtype] | None = ...,
     engine: ExcelReadEngine | None = ...,
-    converters: Mapping[int | str, Callable[[object], object]] | None = ...,
+    converters: Mapping[int | str, Callable[[Any], Any]] | None = ...,
```
Dr-Irv, line 57: can you see if `Callable[[Any], Scalar]` will work here (and throughout)?

Dr-Irv (changes requested):

tests/test_excel_converters.py
Sachi-Parashar: Thanks for the suggestion @Dr-Irv — I’ve moved the test to `test_io.py` as `test_converters_partial()` and deleted the separate test file.

pandas-stubs/io/excel/_base.pyi
Sachi-Parashar: I tried using Callable[[Any], Scalar], but pyright raises an error because it can’t verify the return type of partial(pd.to_datetime) as Scalar.
`Argument of type "dict[str, partial[Timestamp]]" cannot be assigned to parameter "converters" of type "Mapping[int | str, (Any) -> Scalar]"`
So I kept Callable[[Any], Any] to avoid false positives, as it works with both pyright and mypy. Let me know if you have a preferred workaround.

tests/test_io.py
```diff
@@ -1792,3 +1792,13 @@ def test_read_json_engine() -> None:
         pd.read_json(dd, lines=False, engine="pyarrow")  # type: ignore[call-overload] # pyright: ignore[reportArgumentType, reportCallIssue]
         pd.read_json(io.StringIO(data), engine="pyarrow")  # type: ignore[call-overload] # pyright: ignore[reportArgumentType]
         pd.read_json(io.StringIO(data), lines=True, engine="pyarrow")  # type: ignore[call-overload] # pyright: ignore[reportArgumentType, reportCallIssue]
+
+
+def test_converters_partial() -> None:
+    from functools import partial
+    import pandas as pd
+    from typing_extensions import assert_type
```
Dr-Irv: move the first import to top of file.
the second 2 imports are already there at the top, so remove from here.

tests/test_io.py
```diff
@@ -1792,3 +1792,13 @@ def test_read_json_engine() -> None:
...
+
+    partial_func = partial(pd.to_datetime, errors="coerce")
+    df = pd.read_excel("foo.xlsx", converters={"field_1": partial_func})
+    assert_type(df, pd.DataFrame)
```
Dr-Irv: Use `check(assert_type(df, pd.DataFrame), pd.DataFrame)`

tests/test_io.py
```diff
@@ -1792,3 +1792,13 @@ def test_read_json_engine() -> None:
...
         pd.read_json(io.StringIO(data), lines=True, engine="pyarrow")  # type: ignore[call-overload] # pyright: ignore[reportArgumentType, reportCallIssue]
+
+
+def test_converters_partial() -> None:
+    from functools import partial
+    import pandas as pd
+    from typing_extensions import assert_type
+
+    partial_func = partial(pd.to_datetime, errors="coerce")
+    df = pd.read_excel("foo.xlsx", converters={"field_1": partial_func})
```
Dr-Irv: This will need to work at runtime.  If you look at `test_read_excel()`, you will see the use of `ensure_clean()` and that the test needs to first write out a file, and then read back the file.  So you need to change to use that pattern.

pandas-stubs/io/excel/_base.pyi
Dr-Irv: I researched it a bit, and I think your solution is fine.

tests/test_io.py
Sachi-Parashar: Moved import statement to the top of the file.

tests/test_io.py
Sachi-Parashar: Used check(assert_type(df, pd.DataFrame), pd.DataFrame) as suggested.

tests/test_io.py
Sachi-Parashar: Updated the test to write and read using ensure_clean() pattern.

tests/test_io.py
```diff
@@ -304,23 +305,23 @@ def test_sas_bdat() -> None:
     path = pathlib.Path(CWD, "data", "airline.sas7bdat")
     check(assert_type(read_sas(path), DataFrame), DataFrame)
     with check(
-        assert_type(read_sas(path, iterator=True), Union[SAS7BDATReader, XportReader]),
-        SAS7BDATReader,
+            assert_type(read_sas(path, iterator=True), Union[SAS7BDATReader, XportReader]),
+            SAS7BDATReader,
```
Dr-Irv: I don't know why you indented here and elsewhere.  Make sure to install the `pre-commit` and run `black`

Dr-Irv: You can see from the CI that the pre-commit failed because of this

PR #1301, opened 2025-08-01
GH1291 Add BaseOffset for shift operators on Series and DataFrame

### LINKED ISSUE #1291: Series.shift with BDay is not allowed
**Describe the bug**

**To Reproduce**

```python
sr = pd.Series(pd.bdate_range("2023-01-03", "2023-02-06"))
check(assert_type(sr.shift(BDay(1)), "TimestampSeries"), pd.Series, pd.Timestamp)
```

This will raise an issue as only `DateOffset` is allowed in the `period` parameter. The issue is that `BDay` does not inherit from DateOffset but directly from `BaseOffset`.

Is there a reason why we only allow DateOffset and not BaseOffset?

Happy to take care of the PR.

 - OS: MacOS
 - OS Version 15.5
 - python version 3.13
 - version of type checker 1.17.0

loicdiridollou:
- [x] Closes #1291 

### DIFF
--- a/pandas-stubs/core/frame.pyi
+++ b/pandas-stubs/core/frame.pyi
@@ -892,7 +892,7 @@ class DataFrame(NDFrame, OpsMixin, _GetItemHack):
     def shift(
         self,
         periods: int | Sequence[int] = ...,
-        freq: DateOffset | dt.timedelta | _str | None = ...,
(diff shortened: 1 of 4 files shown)


Dr-Irv (approved):
thanks @loicdiridollou

PR #1306, opened 2025-08-03
fix: #1304 add PD_LTE_23

### LINKED ISSUE #1304: Reinstate test for Path.truediv that fails with 3.0
Prior to #1280 , we had this test:
```python
def test_path_div() -> None:
    # GH 682
    folder = Path.cwd()
    files = pd.Series(["a.png", "b.png"])
    if PD_LTE_23:
        # Bug in 3.0 https://github.com/pandas-dev/pandas/issues/61940
        check(assert_type(folder / files, pd.Series), pd.Series, Path)

    folders = pd.Series([folder, folder])
    check(assert_type(folders / Path("a.png"), pd.Series), pd.Series, Path)
```

In #1280, @cmp0xff created this:
https://github.com/pandas-dev/pandas-stubs/blob/9a389ec144ed863d39aabf98d37c598a73f764f7/tests/series/arithmetic/test_truediv.py#L143

The new tests in that function need to be surrounded with the `if PD_LTE_23:` and the comment above

cmp0xff:
- [x] Closes #1304 (Replace xxxx with the Github issue number)

### DIFF
--- a/tests/series/arithmetic/test_truediv.py
+++ b/tests/series/arithmetic/test_truediv.py
@@ -7,7 +7,6 @@ from typing_extensions import assert_type
 
 from tests import (
     PD_LTE_23,
-    WINDOWS,
     check,
 )
 
@@ -168,12 +167,10 @@ def test_truediv_path(tmp_path: Path) -> None:
     fnames = pd.Series(["a.png", "b.gz", "c.txt"])
 
     if PD_LTE_23:
-        # Bug in 3.0 https://github.com/pandas-dev/pandas/issues/61940
+        # Bug in 3.0 https://github.com/pandas-dev/pandas/issues/61940 (pyarrow.lib.ArrowInvalid)
         check(assert_type(fnames / tmp_path, pd.Series), pd.Series, Path)
         check(assert_type(tmp_path / fnames, pd.Series), pd.Series, Path)
 
-    if PD_LTE_23 or not WINDOWS:
-        # pyarrow.lib.ArrowInvalid: Could not convert WindowsPath('...') with type WindowsPath: did not recognize Python value type when inferring an Arrow data type
         check(assert_type(fnames.truediv(tmp_path), pd.Series), pd.Series, Path)
         check(assert_type(fnames.div(tmp_path), pd.Series), pd.Series, Path)
 


tests/series/arithmetic/test_truediv.py
```diff
@@ -140,12 +143,24 @@ def test_truediv_pd_series() -> None:
...
+    check(assert_type(folders / Path("a.png"), pd.Series), pd.Series, Path)
+
+
 def test_truediv_path() -> None:
-    """Test pd.Series / path object"""
-    left, p = pd.Series(["pat", "ath", "path"]), Path()
+    """Test pd.Series / path object.
+
+    Also GH 682."""
+    left, p = pd.Series(["a.png", "b.gz", "c.txt"]), Path.cwd()
```
loicdiridollou: I think it will be a bit cleaner and more consistent with the rest of the tests if you use the `ensure_clean` context manager (so it does not create a folder locally): 
```python
with ensure_clean() as path:
    ...
```
See the following example:
https://github.com/pandas-dev/pandas-stubs/blob/9a389ec144ed863d39aabf98d37c598a73f764f7/tests/test_io.py#L109-L115

tests/series/arithmetic/test_truediv.py
```diff
@@ -140,12 +143,24 @@ def test_truediv_pd_series() -> None:
     check(assert_type(left.rdiv(c), pd.Series), pd.Series)
 
 
+def test_path_div() -> None:
+    # GH 682
+    folder = Path.cwd()
```
loicdiridollou: Same about the use of `ensure_clean`, the rest is good.

tests/series/arithmetic/test_truediv.py
cmp0xff: It turns out `ensure_clean()` makes a string, not a Path. I'll need another machine to work it out more efficiently. I guess I'll need to combine the temp path factory fixture, and the Python temp file factory

Meanwhile, may I ask why do we do this? The tests do not create any new file / path, after all.

loicdiridollou: Indeed considering that you don't need to check if the path exists my comment below with ensure path is actually not super useful, that is my mistake.
Just create a path like `Path('/folder')` or keep `Path.cwd()`, I originally thought your tests created the paths and used them to write things. Maybe it is fine to keep things as you wrote them with cwd as it won't be used so no need to worry about permissions.
Also since you mention the temp path factory fixture, in a test where you may need a tmp directory, pytest has the option to pass `tmp_path` in the function definition to create a temporary folder to dump data: 
```python
def test_dump_data(tmp_path: Path) -> None:
    pd.DataFrame().to_csv(tmp_path / "filename.csv")
```
if ever you need, we used to use a package for it but pytest has it integrated.

loicdiridollou:
Looks like it is failing for nightly run with pandas 3.0 for the test_truediv_path, see https://github.com/pandas-dev/pandas-stubs/actions/runs/16727359383
You can run those tests locally with `poetry run poe pytest —nightly`

Dr-Irv:
> Looks like it is failing for nightly run with pandas 3.0 for the test_truediv_path, see https://github.com/pandas-dev/pandas-stubs/actions/runs/16727359383 You can run those tests locally with `poetry run poe pytest —nightly`

I tested locally like that, and ALL of those tests are failing with the nightly.

Thanks @loicdiridollou for remembering the right "/" command!

Dr-Irv:
> I see the issue, in pandas 3.0 the default type is StringDtype and not object which does not support the div operator. If you do `left.astype(object) / p` you won’t have the issue. So I would wrap the test with PD_LTE_23 since it is not supported with pandas 3.0 as the dtype of the series is changing, but @Dr-Irv can confirm.

Yes, all of those tests need to be wrapped this way until the bug is fixed in pandas.  That's why the comment as a ref to the pandas Github issue.

tests/series/arithmetic/test_truediv.py
```diff
@@ -143,27 +144,37 @@ def test_truediv_pd_series() -> None:
...
-    check(assert_type(left / p, pd.Series), pd.Series, Path)
     if PD_LTE_23:
         # Bug in 3.0 https://github.com/pandas-dev/pandas/issues/61940
-        check(assert_type(p / left, pd.Series), pd.Series, Path)
+        check(assert_type(fnames / tmp_path, pd.Series), pd.Series, Path)
+        check(assert_type(tmp_path / fnames, pd.Series), pd.Series, Path)
 
-    check(assert_type(left.truediv(p), pd.Series), pd.Series, Path)
-    check(assert_type(left.div(p), pd.Series), pd.Series, Path)
+    if PD_LTE_23 or not WINDOWS:
```
Dr-Irv: The `or not WINDOWS` should not be necessary.

cmp0xff: ce7498a

cmp0xff:
Hi @Dr-Irv , it seeems that I cannot trigger the nightly test.

tests/series/arithmetic/test_truediv.py
```diff
@@ -140,15 +143,36 @@ def test_truediv_pd_series() -> None:
...
-    check(assert_type(left.truediv(p), pd.Series), pd.Series, Path)
-    check(assert_type(left.div(p), pd.Series), pd.Series, Path)
+    check(assert_type(folders / fpath, pd.Series), pd.Series, Path)
+    check(assert_type(folders.truediv(fpath), pd.Series), pd.Series, Path)
+    check(assert_type(folders.div(fpath), pd.Series), pd.Series, Path)
 
-    check(assert_type(left.rtruediv(p), pd.Series), pd.Series, Path)
-    check(assert_type(left.rdiv(p), pd.Series), pd.Series, Path)
+    # mypy thinks it's `Path`, in contrast to Series.__rtruediv__(self, other: Path) -> Series: ...
+    check(assert_type(tmp_path / fpaths, pd.Series), pd.Series, Path)  # type: ignore[assert-type]
```
loicdiridollou, line 158: @Dr-Irv will confirm but I think this is an issue we have had in the past with mypy getting confused

tests/series/arithmetic/test_truediv.py
```diff
@@ -140,15 +143,36 @@ def test_truediv_pd_series() -> None:
...
 
-    check(assert_type(left.truediv(p), pd.Series), pd.Series, Path)
-    check(assert_type(left.div(p), pd.Series), pd.Series, Path)
+    check(assert_type(folders / fpath, pd.Series), pd.Series, Path)
+    check(assert_type(folders.truediv(fpath), pd.Series), pd.Series, Path)
+    check(assert_type(folders.div(fpath), pd.Series), pd.Series, Path)
 
-    check(assert_type(left.rtruediv(p), pd.Series), pd.Series, Path)
-    check(assert_type(left.rdiv(p), pd.Series), pd.Series, Path)
+    # mypy thinks it's `Path`, in contrast to Series.__rtruediv__(self, other: Path) -> Series: ...
```
Dr-Irv, line 157: I tried to come up with a simple example to illustrate the issue and couldn't.

loicdiridollou: Issue comes from how mypy reads the `Path.__truediv__` typing and it prioritizes it over Series.__rtruediv__, not expert enough to know why

Dr-Irv: Yes, I don't know why it did that.  Tried some simple experiments and couldn't replicate it.

PR #1308, opened 2025-08-06
GH804 Fix property setter/getter for Index.names

### LINKED ISSUE #804: type annotation for Index/MultiIndex.names is incorrect
**Describe the bug**

type annotation for Index/MultiIndex.names is incorrect

**To Reproduce**

```
In [1]: import pandas as pd

In [2]: df = pd.DataFrame({"a": ["a", "b", "c"], "i": [10, 11, 12]}, index=pd.Index([5, 10, 15], name="idx"))

In [4]: df.index.names
Out[4]: FrozenList(['idx'])

In [5]: df.index.names = (None,)

In [6]: df.index.names
Out[6]: FrozenList([None])
```

`mypy`

bigframes/core/blocks.py:436: error: Incompatible types in assignment (expression has type "tuple[None]", variable has type "list[str]")  [assignment]

 - OS: [e.g. Windows, Linux, MacOS] Linux
 - OS Version [e.g. 22]

Distributor ID: Debian
Description:    Debian GNU/Linux rodete
Release:        n/a
Codename:       rodete

 - python version Python 3.10.9
 - version of type checker mypy==1.5.1
 - version of installed `pandas-stubs` pandas-stubs==2.0.3.230814

**Additional context**

loicdiridollou:
- [x] Closes #804


pandas-stubs/core/indexes/base.pyi
```diff
@@ -306,13 +307,13 @@ class Index(IndexOpsMixin[S1]):
...
-    def name(self) -> Hashable | None: ...
+    def name(self) -> _str | None: ...
     @name.setter
     def name(self, value) -> None: ...
     @property
-    def names(self) -> list[Hashable]: ...
+    def names(self) -> list[_str | None]: ...
     @names.setter
-    def names(self, names: Sequence[Hashable]) -> None: ...
+    def names(self, names: SequenceNotStr[_str | None]) -> None: ...
```
Dr-Irv: All of these should use `Hashable` instead of `_str`

e.g., the following is valid:
```python
df.index.names=[3]
```

loicdiridollou: Yes I realized this morning, all fixed with new test case.

PR #1274, opened 2025-07-13
refactor(series)!: 🕰️ drop TimestampSeries

cmp0xff:
- [x] Closes #1355
- [x] Closes #1359
- [x] Addresses #718
- [x] Tests added: Please use `assert_type()` to assert the type of any return value

PR #1311, opened 2025-08-08
feat(series): addition for bools

cmp0xff:
This is inspired by #1274 and a follow-up for #1275.


Dr-Irv (approved):
thanks @cmp0xff

PR #1312, opened 2025-08-08
feat(series): arithmetic truediv and sub

cmp0xff:
This is inspired by #1274 and a follow-up for #1275.

(I thought cmp0xff/pandas-stubs#1 would automatically redirect to pandas-dev/pandas-stubs, which did not happen)


tests/series/test_series.py
```diff
@@ -819,7 +819,7 @@ def test_types_element_wise_arithmetic() -> None:
     check(assert_type(s.add(s2, fill_value=0), "pd.Series[int]"), pd.Series, np.integer)
 
     # TODO this one below should type pd.Series[int]
```
Dr-Irv: you can remove this comment now

Dr-Irv (changes requested):
In #1311 you added support for `bool` in addition, so don't you want to include the `bool` tests for subtraction as well?

It's really nice that the pattern on these operators is similar so it seems pretty straightforward to get them working.

tests/series/test_series.py
cmp0xff: e3447cf91a71444b2dd1f82e72b8f650c18a0161

Dr-Irv:
> > In #1311 you added support for `bool` in addition, so don't you want to include the `bool` tests for subtraction as well?
> 
> `bool`s can be added but not subtracted.
> 
> ```python
> pd.Series([True]) - pd.Series([True])  # TypeError: numpy boolean subtract, the `-` operator, is not supported, use the bitwise_xor, the `^` operator, or the logical_xor function instead.
> ```

But they can be subtracted from non-bools:
```python
pd.Series([3,4,5])-pd.Series([True, False, True])
0    2
1    4
2    4
dtype: int64
```

So I would think you would want to add those to the tests.

tests/series/arithmetic/complex/test_sub.py
```diff
@@ -0,0 +1,129 @@
...
+from typing_extensions import assert_type
+
+from tests import check
+
+left = pd.Series([1j, 2j, 3j])  # left operand
+
+
+def test_sub_py_scalar() -> None:
+    """Test pd.Series[complex] - Python native scalars"""
+    i, f, c = 1, 1.0, 1j
```
Dr-Irv: I think you could do:
```suggestion
    b, i, f, c = True, 1, 1.0, 1j
```
And then add the tests on `b` here, and similar in the other funcs.

Dr-Irv (changes requested):

tests/series/arithmetic/complex/test_sub.py
cmp0xff: 672009edb82b98d52e45b86330b3c939897f72a8

tests/series/arithmetic/bool/test_sub.py
```diff
@@ -0,0 +1,113 @@
...
+left = pd.Series([True, True, False])  # left operand
+
+
+def test_sub_py_scalar() -> None:
+    """Test pd.Series[bool] - Python native scalars"""
+    b, i, f, c = True, 1, 1.0, 1j
+
+    check(assert_type(left - i, "pd.Series[int]"), pd.Series, np.integer)
+    check(assert_type(left - f, "pd.Series[float]"), pd.Series, np.floating)
+    check(assert_type(left - c, "pd.Series[complex]"), pd.Series, np.complexfloating)
```
Dr-Irv, line 25: You have to add `left-b` to the tests (and similar below) and throughout this file

tests/series/arithmetic/int/test_sub.py
```diff
@@ -10,7 +10,7 @@
 
 def test_sub_py_scalar() -> None:
     """Test pd.Series[int] - Python native scalars"""
-    i, f, c = 1, 1.0, 1j
+    b, i, f, c = True, 1, 1.0, 1j
 
     check(assert_type(left - i, "pd.Series[int]"), pd.Series, np.integer)
     check(assert_type(left - f, "pd.Series[float]"), pd.Series, np.floating)
```
Dr-Irv: You need to add the tests for `b` in here and throughout this file in each of the test functions.

Dr-Irv (changes requested):
Still missing some tests for completeness.

tests/series/arithmetic/bool/test_sub.py
cmp0xff: Should have been fixed by 1e9b11d406ca20accc96c898473e7e38b3acd75a

tests/series/arithmetic/int/test_sub.py
cmp0xff: c8e3ab25b940bef8dcaa51f672a2373842dd994c addresses an issue that appears to originate from `mypy`. All unsatisfactory aspects from my side for now are resolved.

tests/series/arithmetic/bool/test_sub.py
```diff
@@ -0,0 +1,154 @@
...
+
+def test_sub_numpy_array() -> None:
+    """Test pd.Series[bool] - numpy array"""
+    b = np.array([True, False, True], np.bool_)
+    i = np.array([2, 3, 5], np.int64)
+    f = np.array([1.0, 2.0, 3.0], np.float64)
+    c = np.array([1.1j, 2.2j, 4.1j], np.complex128)
+
+    if TYPE_CHECKING_INVALID_USAGE:
+        assert_type(left - b, Never)
```
Dr-Irv, line 87: Below you are able to detect that `left.sub(b)` is invalid and will generate a type error.  How come this isn't true for `left - b` ?

Same comment about `b-left` and `left.rsub(b)`

tests/series/arithmetic/bool/test_sub.py
```diff
@@ -0,0 +1,154 @@
...
+def test_sub_pd_series() -> None:
+    """Test pd.Series[bool] - pandas series"""
+    b = pd.Series([True, False, True])
+    i = pd.Series([2, 3, 5])
+    f = pd.Series([1.0, 2.0, 3.0])
+    c = pd.Series([1.1j, 2.2j, 4.1j])
+
+    # In the following two cases, mypy fails to recognise the second operand as pd.Series[bool]
+    if TYPE_CHECKING_INVALID_USAGE:
+        _ = left - cast("pd.Series[bool]", b)  # type: ignore[redundant-cast,operator] # pyright: ignore[reportUnnecessaryCast,reportOperatorIssue]
```
Dr-Irv: So the reason this is happening is because of the underscore on the LHS.  

If you make the LHS a different variable (e.g. `result = left - b`, then `mypy` will detect it fine.  But each time you do a test like that within the function, you have to use a different variable name on the LHS.

This came up in another PR that I just reviewed.

This will allow you to get rid of the `cast`

Dr-Irv (changes requested):
I found a way to handle differentiating between `bool` and `int`/`float`/`complex` in the stubs so that we can better catch things like `Series[bool].sub(bool)` instead of using `Never`.  See this commit I did on my own branch of your PR:
https://github.com/pandas-dev/pandas-stubs/pull/1329/commits/a8e34a43b4dbabd216d2ec715c8c84cd870ca46c

It only fixes `Series.sub()`.  I'll let you work on `Series.__sub__()` 

It is based on this cool idea:
https://discuss.python.org/t/type-hints-for-bool-vs-int-overload/72239/21

tests/series/arithmetic/bool/test_sub.py
cmp0xff: 0a8689794b47fb7523ce9c2870551779063d6f12

tests/series/arithmetic/bool/test_sub.py
cmp0xff: Now there is no `-> Never` in the subtraction section of `pandas-stubs`. These `Never`s must have been ordered by `numpy`.

pandas-stubs/core/series.pyi
```diff
@@ -2071,6 +2071,77 @@ class Series(IndexOpsMixin[S1], NDFrame):
     @overload
     def __rxor__(self, other: int | np_ndarray_anyint | Series[int]) -> Series[int]: ...
     @overload
+    def __sub__(
+        self, other: Timestamp | datetime | TimestampSeries
+    ) -> TimedeltaSeries: ...
```
Dr-Irv: This is missing from `def sub()`

pandas-stubs/core/series.pyi
```diff
@@ -2081,24 +2152,280 @@ class Series(IndexOpsMixin[S1], NDFrame):
         other: Timedelta | TimedeltaSeries | TimedeltaIndex | np.timedelta64,
     ) -> TimedeltaSeries: ...
```
Dr-Irv, line 2176: The above 2 overloads are missing from `def sub()`

pandas-stubs/core/series.pyi
```diff
@@ -2081,24 +2152,280 @@ class Series(IndexOpsMixin[S1], NDFrame):
...
+        level: Level | None = None,
+        fill_value: float | None = None,
+        axis: int = 0,
+    ) -> Series[complex]: ...
+    @overload
+    def __rsub__(
+        self: Series[Never], other: complex | _ListLike | Series
+    ) -> Series: ...
+    @overload
+    def __rsub__(self, other: Series[Never]) -> Series: ...
```
Dr-Irv, line 2315: This is missing from `def rsub()`.  But I wonder if this `def __rsub__()` is needed, because it would only apply if both args are `Series` and that's handled by `def __sub__()`

pandas-stubs/core/series.pyi
```diff
@@ -2081,24 +2152,280 @@ class Series(IndexOpsMixin[S1], NDFrame):
...
+    ) -> Series[complex]: ...
     @overload
     def __truediv__(
         self: Series[Never], other: complex | _ListLike | Series
     ) -> Series: ...
     @overload
     def __truediv__(self, other: Series[Never]) -> Series: ...
     @overload
-    def __truediv__(self: Series[bool], other: bool | np_ndarray_bool) -> Never: ...
+    def __truediv__(self: Series[bool], other: np_ndarray_bool) -> Never: ...
```
Dr-Irv, line 2494: You still have some `Never` tests for `Series[bool]` / `np_ndarray_bool` in `tests/series/bool/test_truediv.py:test_truediv_numpy_array()`.  Can this overload be removed and those tests be changed to detect it as a typing issue?

pandas-stubs/core/series.pyi
cmp0xff: Added in 44d0be02795c6329320eb83c04083ee76264ab64

I agree that this is handled by `def __sub__()`, but from https://github.com/pandas-dev/pandas-stubs/pull/1314#discussion_r2270673683 I think we would like to be less concise and more consistent, maybe for easier maintenance.

pandas-stubs/core/series.pyi
cmp0xff: Added in 44d0be02795c6329320eb83c04083ee76264ab64

pandas-stubs/core/series.pyi
cmp0xff: Removed in 44d0be02795c6329320eb83c04083ee76264ab64

pandas-stubs/core/series.pyi
cmp0xff: If we remove the operator version, `Series[bool] - np_ndarray_bool` will give `np_ndarray_bool` instead of `Never`. I would prefer `Never`.

The method versions have been removed in 44d0be02795c6329320eb83c04083ee76264ab64, at the price of being less consistent.

pandas-stubs/core/series.pyi
```diff
@@ -2239,6 +2235,22 @@ class Series(IndexOpsMixin[S1], NDFrame):
...
+        axis: int = 0,
+    ) -> Series[complex]: ...
+    @overload
+    def sub(
+        self: Series[Timedelta],
+        other: timedelta | np.timedelta64 | TimedeltaSeries | TimedeltaIndex,
+        level: Level | None = None,
+        fill_value: float | None = None,
+        axis: int = 0,
+    ) -> Series[complex]: ...
```
Dr-Irv: to maintain order, shouldn't these 2 overloads be last for `sub`, like they are for `__sub__()` ?

pandas-stubs/core/series.pyi
```diff
@@ -2071,10 +2071,6 @@ class Series(IndexOpsMixin[S1], NDFrame):
     @overload
     def __rxor__(self, other: int | np_ndarray_anyint | Series[int]) -> Series[int]: ...
     @overload
-    def __sub__(
-        self, other: Timestamp | datetime | TimestampSeries
-    ) -> TimedeltaSeries: ...
```
Dr-Irv: I think this needs to be put back in.  I'm guessing we don't have a test, but here is one that works without this last commit, but gets a failure without:
```python
import pandas as pd
from typing import reveal_type                                                                                                                                  df = pd.DataFrame({"x": pd.to_datetime(["2025-01-02", "2025-03-04"])})                                                                                          y = df["x"] - pd.Timestamp("2024-06-04")

reveal_type(y)         
```
We get `TimedeltaSeries` with main, but not with the above change.

So the fix here is to add the above test, and also include this in `__sub__()` and `sub()`

pandas-stubs/core/series.pyi
Dr-Irv: needs to be put back.  see comment elsewhere

pandas-stubs/core/series.pyi
Dr-Irv: OK - makes sense.

pandas-stubs/core/series.pyi
Dr-Irv: OK, that's fine.  What's happening here is that if we have no declaration for `__sub__(self: Series[bool], other: np_ndarray_bool)`, then it goes to `ndarray[bool].__rsub__(other: Any)` for `numpy` which is why we get that result.

pandas-stubs/core/series.pyi
cmp0xff: Fixed in 61de1a13d80f9ae9af7fc0d5cb530ab8cc76a96f

pandas-stubs/core/series.pyi
cmp0xff: The example is incomplete, but I attempted a fix in 61de1a13d80f9ae9af7fc0d5cb530ab8cc76a96f

Dr-Irv: Ooops sorry about not copying the example correctly!  But you figured it out

Dr-Irv (approved):
Thanks @cmp0xff .  Very nice contribution!

Now only `__mul__()` is left to do.  :-)

PR #1313, opened 2025-08-10
Add check assert_type to tests for frame.py and add __new__ method for DatetimeIndex

loicdiridollou:
Cleaned up tests in tests/test_frame.py as there was no check assert_type in many places.
Added a __new__ method for DatetimeIndex since it would not instantiate to DatetimeIndex but to Series[int].

### DIFF
--- a/pandas-stubs/core/indexes/datetimes.pyi
+++ b/pandas-stubs/core/indexes/datetimes.pyi
@@ -23,6 +23,7 @@ from pandas import (
 from pandas.core.indexes.accessors import DatetimeIndexProperties
 from pandas.core.indexes.datetimelike import DatetimeTimedeltaMixin
 from pandas.core.series import (
+    Series,
     TimedeltaSeries,
     TimestampSeries,
 )
@@ -56,6 +57,19 @@ class DatetimeIndex(DatetimeTimedeltaMixin[Timestamp], DatetimeIndexProperties):
         copy: bool = ...,
         name: Hashable = ...,
     ) -> None: ...
+    @classmethod
+    def __new__(
+        cls,
+        data: AxesData | Series,
+        freq: Frequency = ...,
+        tz: TimeZones = ...,
+        ambiguous: str = ...,
+        dayfirst: bool = ...,
+        yearfirst: bool = ...,
+        dtype: Dtype = ...,
+        copy: bool = ...,
+        name: Hashable = ...,
+    ) -> Self: ...
     def __reduce__(self): ...
     # various ignores needed for mypy, as we do want to restrict what can be used in
     # arithmetic for these types
--- a/tests/test_frame.py
+++ b/tests/test_frame.py
@@ -109,17 +109,44 @@ def getSeriesData() -> dict[str, pd.Series]:
 
 
 def test_types_init() -> None:
-    pd.DataFrame(data={"col1": [1, 2], "col2": [3, 4]})
-    pd.DataFrame(data={"col1": [1, 2], "col2": [3, 4]}, index=[2, 1])
-    pd.DataFrame(data=[[1, 2, 3], [4, 5, 6]])
-    pd.DataFrame(data=itertools.repeat([1, 2, 3], 3))
-    pd.DataFrame(data=(range(i) for i in range(5)))
-    pd.DataFrame(data=[1, 2, 3, 4], dtype=np.int8)
-    pd.DataFrame(
-        np.array([[1, 2, 3], [4, 5, 6], [7, 8, 9]]),
-        columns=["a", "b", "c"],
-        dtype=np.int8,
-        copy=True,
+    check(
+        assert_type(pd.DataFrame(data={"col1": [1, 2], "col2": [3, 4]}), pd.DataFrame),
+        pd.DataFrame,
+    )
+    check(
+        assert_type(
+            pd.DataFrame(data={"col1": [1, 2], "col2": [3, 4]}, index=[2, 1]),
+            pd.DataFrame,
+        ),
+        pd.DataFrame,
+    )
+    check(
+        assert_type(pd.DataFrame(data=[[1, 2, 3], [4, 5, 6]]), pd.DataFrame),
+        pd.DataFrame,
+    )
+    check(
+        assert_type(pd.DataFrame(data=itertools.repeat([1, 2, 3], 3)), pd.DataFrame),
+        pd.DataFrame,
+    )
+    check(
+        assert_type(pd.DataFrame(data=(range(i) for i in range(5))), pd.DataFrame),
+        pd.DataFrame,
+    )
+    check(
+        assert_type(pd.DataFrame(data=[1, 2, 3, 4], dtype=np.int8), pd.DataFrame),
+        pd.DataFrame,
+    )
+    check(
+        assert_type(
+            pd.DataFrame(
+                np.array([[1, 2, 3], [4, 5, 6], [7, 8, 9]]),
+                columns=["a", "b", "c"],
+                dtype=np.int8,
+                copy=True,
+            ),
+            pd.DataFrame,
+        ),
+        pd.DataFrame,
     )
     check(
         assert_type(pd.DataFrame(0, index=[0, 1], columns=[0, 1]), pd.DataFrame),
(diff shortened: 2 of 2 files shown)


tests/test_frame.py
```diff
@@ -2567,11 +2702,14 @@ def test_types_regressions() -> None:
...
     df = pd.DataFrame({"A": [1, 2, 3], "B": [5, 6, 7]})
-    pd.DatetimeIndex(
-        data=df["A"],
-        tz=None,
-        ambiguous="NaT",
-        copy=True,
+    pd.DatetimeIndex(data=df["A"], tz=None, ambiguous="NaT", copy=True),
+
+    check(
+        assert_type(
```
loicdiridollou: This is the reason I added the `__new__` to datetimes.pyi, prob worth putting a test in the relevant test file, please let me know if so.

pandas-stubs/core/indexes/datetimes.pyi
```diff
@@ -56,6 +57,19 @@ class DatetimeIndex(DatetimeTimedeltaMixin[Timestamp], DatetimeIndexProperties):
...
+        data: AxesData | Series,
+        freq: Frequency = ...,
+        tz: TimeZones = ...,
+        ambiguous: str = ...,
+        dayfirst: bool = ...,
+        yearfirst: bool = ...,
+        dtype: Dtype = ...,
+        copy: bool = ...,
+        name: Hashable = ...,
+    ) -> Self: ...
```
Dr-Irv: You can just do this by making this the `__init__()` method, (returning `None`).

So in line 50, just modify `data` to be `data: AxesData | Series`

tests/test_frame.py
```diff
@@ -269,16 +296,22 @@ def test_types_setitem_mask() -> None:
...
 def test_types_iloc_iat() -> None:
     df = pd.DataFrame(data={"col1": [1, 2], "col2": [3, 4]})
-    df.iloc[1, 1]
-    df.iloc[[1], [1]]
+    check(assert_type(df.iloc[1, 1], Scalar), np.integer)
+    check(assert_type(df.iloc[[1], [1]], pd.DataFrame), pd.DataFrame)
+
+    # TODO the below should pass
+    # check(assert_type(df.iat[0, 0], Scalar), np.integer)
     df.iat[0, 0]
```
Dr-Irv: can you fix the declaration of `iat` in `frame.pyi` ?

tests/test_frame.py
```diff
@@ -269,16 +296,22 @@ def test_types_setitem_mask() -> None:
...
 
 def test_types_loc_at() -> None:
     df = pd.DataFrame(data={"col1": [1, 2], "col2": [3, 4]})
-    df.loc[[0], "col1"]
+    check(assert_type(df.loc[[0], "col1"], pd.Series), pd.Series)
+    check(assert_type(df.loc[0, "col1"], Scalar), np.integer)
+
+    # TODO the below should pass
+    # check(assert_type(df.at[0, "col1"], Scalar), np.integer)
     df.at[0, "col1"]
```
Dr-Irv: same comment - can you fix the declaration of `at()` in `frame.pyi` ?

tests/test_frame.py
```diff
@@ -728,35 +769,35 @@ def test_frame_iterator() -> None:
...
+    check(assert_type(df.sum(), pd.Series), pd.Series)
+    check(assert_type(df.sum(axis=1), pd.Series), pd.Series)
 
 
 def test_types_cumsum() -> None:
     df = pd.DataFrame(data={"col1": [2, 1], "col2": [3, 4]})
-    df.cumsum()
-    df.sum(axis=0)
+    check(assert_type(df.cumsum(), pd.DataFrame), pd.DataFrame)
+    check(assert_type(df.sum(axis=0), pd.Series), pd.Series)
```
Dr-Irv: this is an odd test.  I wonder if it should be `df.cumsum(axis=0)`, since the name of the function is `test_types_cumsum()`

tests/test_frame.py
```diff
@@ -975,33 +1016,33 @@ def test_dataframe_clip() -> None:
...
+    check(assert_type(df.std(axis=1, ddof=1), pd.Series), pd.Series)
+    check(assert_type(df.std(skipna=True, numeric_only=False), pd.Series), pd.Series)
 
 
 def test_types_idxmin() -> None:
     df = pd.DataFrame(data={"col1": [2, 1], "col2": [3, 4]})
-    df.idxmin()
-    df.idxmin(axis=0)
+    check(assert_type(df.idxmin(), pd.Series), pd.Series)
+    check(assert_type(df.idxmin(axis=0), pd.Series), pd.Series)
```
Dr-Irv: These should be `Series[int]`.  Can you fix the declaration in `frame.pyi` ?

tests/test_frame.py
```diff
@@ -975,33 +1016,33 @@ def test_dataframe_clip() -> None:
...
 
 
 def test_types_idxmax() -> None:
     df = pd.DataFrame(data={"col1": [2, 1], "col2": [3, 4]})
-    df.idxmax()
-    df.idxmax(axis=0)
+    check(assert_type(df.idxmax(), pd.Series), pd.Series)
+    check(assert_type(df.idxmax(axis=0), pd.Series), pd.Series)
```
Dr-Irv: Should be `Series[int]`

tests/test_frame.py
Dr-Irv: Yes, actually we should probably move that test there.  

Actually, looking at this function `test_types_regressions()`, I think the various tests in there should be moved to the appropriate files.  That function was created before we took over the repo, so splitting the tests here to the right files would be the right thing to do.

tests/test_frame.py
```diff
@@ -2837,10 +2975,11 @@ def test_sum_get_add() -> None:
 
 
 def test_getset_untyped() -> None:
-    result: int = 10
+    """Test that Dataframe.__getitem__ needs to return untyped series."""
     df = pd.DataFrame({"x": [1, 2, 3, 4, 5], "y": [10, 20, 30, 40, 50]})
-    # Tests that Dataframe.__getitem__ needs to return untyped series.
+    result: int = 10
     result = df["x"].max()
```
Dr-Irv: you can delete these 2 lines.

pandas-stubs/core/indexes/datetimes.pyi
loicdiridollou: That was actually my first try and weridly enough it still picked up another definition and matched to `Index[int]` and not `DatetimeIndex`.

tests/test_frame.py
loicdiridollou: I will add similar iAtIndexerFrame like we have for loc (just can't map it to the iLocIndexerFrame due to overriding some methods).

tests/test_frame.py
loicdiridollou: Yes indeed I was confused too but wanted to limit the changes in the PR to keep it small. I will fix it here directly instead of making a separate PR.

tests/test_frame.py
loicdiridollou: Changed for both idxmin and idxmax.

tests/test_frame.py
loicdiridollou: I will break the test function and split the tests where they should go.

pandas-stubs/core/indexes/datetimes.pyi
Dr-Irv: OK, I see what is happening here.  The subclasses eventually go to `Index` and that has `__new__()`.  

Just change the current definition of `__init__()` to be `__new__()` and have it return `Self` and you should be OK, since `AxesData` includes `Series`

pandas-stubs/core/frame.pyi
```diff
@@ -285,6 +287,44 @@ class _LocIndexerFrame(_LocIndexer, Generic[_T]):
...
 
+class _iAtIndexerFrame(_iAtIndexer, Generic[_T]):
+    def __getitem__(self, idx: tuple[int, int]) -> Scalar: ...
+    def __setitem__(
+        self,
+        idx: tuple[int, int],
+        value: Scalar | NAType | NaTType | None,
+    ) -> None: ...
+
+class _AtIndexerFrame(_AtIndexer, Generic[_T]):
```
Dr-Irv: since none of the methods in either of these refer to `_T`,  you can remove the `Generic` part.

pandas-stubs/core/frame.pyi
```diff
@@ -1683,7 +1723,9 @@ class DataFrame(NDFrame, OpsMixin, _GetItemHack):
     def __iter__(self) -> Iterator[Hashable]: ...
     # properties
     @property
-    def at(self): ...  # Not sure what to do with this yet; look at source
+    def at(
+        self,
+    ) -> _AtIndexerFrame: ...  # Not sure what to do with this yet; look at source
```
Dr-Irv: remove the comment

pandas-stubs/core/frame.pyi
```diff
@@ -1695,7 +1737,9 @@ class DataFrame(NDFrame, OpsMixin, _GetItemHack):
     @property
     def empty(self) -> _bool: ...
     @property
-    def iat(self): ...  # Not sure what to do with this yet; look at source
+    def iat(
+        self,
+    ) -> _iAtIndexerFrame: ...  # Not sure what to do with this yet; look at source
```
Dr-Irv: remove the comment

tests/test_frame.py
```diff
@@ -3465,6 +3379,11 @@ def test_frame_scalars_slice() -> None:
     df3 = pd.DataFrame({"x": range(2)}, index=pd.Index(["a", "b"]))
     check(assert_type(df3.loc[str_], Union[pd.Series, pd.DataFrame]), pd.Series)
 
+    # https://github.com/microsoft/python-type-stubs/issues/62
+    df7: pd.DataFrame = pd.DataFrame({"x": [1, 2, 3]}, index=pd.Index(["a", "b", "c"]))
+    index: pd.Index = pd.Index(["b"])
```
Dr-Irv: should remove the `: pd.DataFrame` and `: pd.Index` from this test

pandas-stubs/core/frame.pyi
loicdiridollou: Correct, I had duplicated the _LocIndexerFrame and was removing unused section but forgot about it.

Dr-Irv (approved):
thanks @loicdiridollou

PR #1314, opened 2025-08-11
feat(series): add and truediv for bools

cmp0xff:
This is inspired by https://github.com/pandas-dev/pandas-stubs/pull/1312#issuecomment-3168718063 and a follow-up for #1311.

### DIFF
--- a/pandas-stubs/core/series.pyi
+++ b/pandas-stubs/core/series.pyi
@@ -2238,7 +2238,14 @@ class Series(IndexOpsMixin[S1], NDFrame):
     @overload
     def truediv(
         self: Series[float],
-        other: int | Sequence[int] | np_ndarray_anyint | np_ndarray_float | Series[int],
+        other: (
+            int
+            | Sequence[int]
+            | np_ndarray_bool
+            | np_ndarray_anyint
+            | np_ndarray_float
+            | Series[_T_INT]
+        ),
         level: Level | None = None,
         fill_value: float | None = None,
         axis: AxisIndex = 0,
@@ -2395,7 +2402,14 @@ class Series(IndexOpsMixin[S1], NDFrame):
     @overload
     def rtruediv(
         self: Series[float],
-        other: int | Sequence[int] | np_ndarray_anyint | np_ndarray_float | Series[int],
+        other: (
+            int
+            | Sequence[int]
+            | np_ndarray_bool
+            | np_ndarray_anyint
+            | np_ndarray_float
+            | Series[_T_INT]
+        ),
         level: Level | None = None,
         fill_value: float | None = None,
         axis: AxisIndex = 0,
--- a/pandas-stubs/core/series.pyi
+++ b/pandas-stubs/core/series.pyi
@@ -2101,22 +2101,25 @@ class Series(IndexOpsMixin[S1], NDFrame):
         self: Series[Never], other: complex | _ListLike | Series
     ) -> Series: ...
     @overload
-    def __truediv__(self, other: Series[Never]) -> Series: ...  # type: ignore[overload-overlap]
+    def __truediv__(self, other: Series[Never]) -> Series: ...
     @overload
-    def __truediv__(
-        self: Series[bool],
-        other: bool | np_ndarray_bool | Series[bool],
-    ) -> Never: ...
+    def __truediv__(self: Series[bool], other: bool | np_ndarray_bool) -> Never: ...
     @overload
-    def __truediv__(
+    def __truediv__(  # pyright: ignore[reportOverlappingOverload]
         self: Series[bool],
-        other: int | Sequence[int] | np_ndarray_anyint | np_ndarray_float | Series[int],
+        other: (
+            float
+            | Sequence[float]
+            | np_ndarray_anyint
+            | np_ndarray_float
+            | Series[int]
+            | Series[float]
+        ),
     ) -> Series[float]: ...
     @overload
-    def __truediv__(  # type: ignore[overload-overlap]
-        self: Series[bool],
-        other: _T_NUM_NON_BOOL | Sequence[_T_NUM_NON_BOOL] | Series[_T_NUM_NON_BOOL],
-    ) -> Series[_T_NUM_NON_BOOL]: ...
+    def __truediv__(
+        self: Series[bool], other: complex | Sequence[complex] | Series[complex]
+    ) -> Series[complex]: ...
     @overload
     def __truediv__(
         self: Series[int],
@@ -2283,10 +2286,7 @@ class Series(IndexOpsMixin[S1], NDFrame):
     @overload
     def __rtruediv__(self: Series[Never], other: complex | _ListLike) -> Series: ...
     @overload
-    def __rtruediv__(
-        self: Series[bool],
-        other: bool | np_ndarray_bool,
-    ) -> Never: ...
+    def __rtruediv__(self: Series[bool], other: bool | np_ndarray_bool) -> Never: ...
     @overload
     def __rtruediv__(
         self: Series[bool],
--- a/pandas-stubs/core/series.pyi
+++ b/pandas-stubs/core/series.pyi
@@ -188,7 +188,6 @@ from pandas.core.dtypes.dtypes import CategoricalDtype
 
 from pandas.plotting import PlotAccessor
 
-_T_NUM_NON_BOOL = TypeVar("_T_NUM_NON_BOOL", int, float, complex)
 _T_INT = TypeVar("_T_INT", bound=int)
 _T_COMPLEX = TypeVar("_T_COMPLEX", bound=complex)
 
@@ -2124,8 +2123,8 @@ class Series(IndexOpsMixin[S1], NDFrame):
     def __truediv__(
         self: Series[int],
         other: (
-            _T_INT
-            | Sequence[_T_INT]
+            int
+            | Sequence[int]
             | np_ndarray_bool
             | np_ndarray_anyint
             | np_ndarray_float
@@ -2141,8 +2140,8 @@ class Series(IndexOpsMixin[S1], NDFrame):
     def __truediv__(
         self: Series[float],
         other: (
-            _T_INT
-            | Sequence[_T_INT]
+            int
+            | Sequence[int]
             | np_ndarray_bool
             | np_ndarray_anyint
             | np_ndarray_float
@@ -2154,10 +2153,6 @@ class Series(IndexOpsMixin[S1], NDFrame):
         self: Series[float],
         other: _T_COMPLEX | Sequence[_T_COMPLEX] | Series[_T_COMPLEX],
     ) -> Series[_T_COMPLEX]: ...
-    @overload
-    def __truediv__(
-        self: Series[_T_COMPLEX], other: np_ndarray_complex
-    ) -> Series[complex]: ...
     @overload
     def __truediv__(
         self: Series[complex],
@@ -2171,6 +2166,10 @@ class Series(IndexOpsMixin[S1], NDFrame):
         ),
     ) -> Series[complex]: ...
     @overload
+    def __truediv__(
+        self: Series[_T_COMPLEX], other: np_ndarray_complex
+    ) -> Series[complex]: ...
+    @overload
     def __truediv__(self, other: Path) -> Series: ...
     @overload
     def truediv(
@@ -2181,7 +2180,7 @@ class Series(IndexOpsMixin[S1], NDFrame):
         axis: AxisIndex = 0,
     ) -> Series: ...
     @overload
-    def truediv(  # type: ignore[overload-overlap]
+    def truediv(
         self,
         other: Series[Never],
         level: Level | None = None,
@@ -2191,33 +2190,40 @@ class Series(IndexOpsMixin[S1], NDFrame):
     @overload
     def truediv(
         self: Series[bool],
-        other: bool | np_ndarray_bool | Series[bool],
+        other: bool | np_ndarray_bool,
         level: Level | None = None,
         fill_value: float | None = None,
         axis: AxisIndex = 0,
     ) -> Never: ...
     @overload
-    def truediv(
+    def truediv(  # pyright: ignore[reportOverlappingOverload]
         self: Series[bool],
-        other: int | Sequence[int] | np_ndarray_anyint | np_ndarray_float | Series[int],
+        other: (
+            float
+            | Sequence[float]
+            | np_ndarray_anyint
+            | np_ndarray_float
+            | Series[int]
+            | Series[float]
+        ),
         level: Level | None = None,
         fill_value: float | None = None,
         axis: AxisIndex = 0,
     ) -> Series[float]: ...
     @overload
-    def truediv(  # type: ignore[overload-overlap]
+    def truediv(
         self: Series[bool],
-        other: _T_NUM_NON_BOOL | Sequence[_T_NUM_NON_BOOL] | Series[_T_NUM_NON_BOOL],
+        other: complex | Sequence[complex] | Series[complex],
         level: Level | None = None,
         fill_value: float | None = None,
         axis: AxisIndex = 0,
-    ) -> Series[_T_NUM_NON_BOOL]: ...
+    ) -> Series[complex]: ...
     @overload
     def truediv(
         self: Series[int],
         other: (
-            _T_INT
-            | Sequence[_T_INT]
+            int
+            | Sequence[int]
             | np_ndarray_bool
             | np_ndarray_anyint
             | np_ndarray_float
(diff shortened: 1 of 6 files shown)


tests/series/arithmetic/bool/test_truediv.py
```diff
@@ -0,0 +1,175 @@
...
+from typing_extensions import assert_type
+
+from tests import check
+
+left = pd.Series([True, False, True])  # left operand
+
+
+def test_truediv_py_scalar() -> None:
+    """Test pd.Series[bool] / Python native scalars"""
+    i, f, c = 1, 1.0, 1j
```
Dr-Irv: valid to add a test with `b=True` in the divisor and the dividend

tests/series/arithmetic/bool/test_truediv.py
```diff
@@ -0,0 +1,175 @@
...
+    check(assert_type(left.rdiv(i), "pd.Series[float]"), pd.Series, np.floating)
+    check(assert_type(left.rdiv(f), "pd.Series[float]"), pd.Series, np.floating)
+    check(
+        assert_type(left.rdiv(c), "pd.Series[complex]"), pd.Series, np.complexfloating
+    )
+
+
+def test_truediv_py_sequence() -> None:
+    """Test pd.Series[bool] / Python native sequence"""
+    i, f, c = [2, 3, 5], [1.0, 2.0, 3.0], [1j, 1j, 4j]
```
Dr-Irv: valid to have `b = [True, False, True]` here

tests/series/arithmetic/bool/test_truediv.py
```diff
@@ -0,0 +1,175 @@
...
+    check(assert_type(left.rdiv(i), "pd.Series[float]"), pd.Series, np.floating)
+    check(assert_type(left.rdiv(f), "pd.Series[float]"), pd.Series, np.floating)
+    check(
+        assert_type(left.rdiv(c), "pd.Series[complex]"), pd.Series, np.complexfloating
+    )
+
+
+def test_truediv_numpy_array() -> None:
+    """Test pd.Series[bool] / numpy array"""
+    i = np.array([2, 3, 5], np.int64)
```
Dr-Irv, line 116: can add a `b=np.array([True, False, True], np.bool_)` here

tests/series/arithmetic/bool/test_truediv.py
```diff
@@ -0,0 +1,175 @@
...
+    check(assert_type(left.rdiv(i), "pd.Series[float]"), pd.Series, np.floating)
+    check(assert_type(left.rdiv(f), "pd.Series[float]"), pd.Series, np.floating)
+    check(
+        assert_type(left.rdiv(c), "pd.Series[complex]"), pd.Series, np.complexfloating
+    )
+
+
+def test_truediv_pd_series() -> None:
+    """Test pd.Series[bool] / pandas series"""
+    i = pd.Series([2, 3, 5])
```
Dr-Irv, line 177: I think this is the only case where you *can't* test `b=pd.Series([True, False, True])` because it will fail at runtime in all these operations.

Ideally, it would be cool to have a `if TYPE_CHECKING_INVALID_USAGE:` when `b` is involved to see if we can catch this usage with static typing.

tests/series/arithmetic/float/test_truediv.py
```diff
@@ -10,7 +10,7 @@
 
 def test_truediv_py_scalar() -> None:
     """Test pd.Series[float] / Python native scalars"""
-    i, f, c = 1, 1.0, 1j
+    b, i, f, c = True, 1, 1.0, 1j
```
Dr-Irv, line 13: in this function and other funcs in this file, you have to add the tests for `b`

tests/series/arithmetic/bool/test_truediv.py
cmp0xff:
```py
pd.Series([True, False]) / True  # NotImplementedError: operator 'truediv' not implemented for bool dtypes
```

tests/series/arithmetic/bool/test_truediv.py
cmp0xff:
```py
pd.Series([True, False]) / np.array([False, True], np.bool_)   # NotImplementedError: operator 'truediv' not implemented for bool dtypes
```

tests/series/arithmetic/bool/test_truediv.py
cmp0xff: effc25220e33589572763b4b6d702390d1cf54ae with unwanted ignores in tests, unfortunately

tests/series/arithmetic/bool/test_truediv.py
cmp0xff: effc25220e33589572763b4b6d702390d1cf54ae with unwanted ignores in tests, unfortunately

tests/series/arithmetic/float/test_truediv.py
cmp0xff: effc25220e33589572763b4b6d702390d1cf54ae with unwanted ignores in tests, unfortunately

tests/series/arithmetic/test_truediv.py
```diff
@@ -145,7 +145,7 @@ def test_truediv_pd_series() -> None:
     check(assert_type(left / f, pd.Series), pd.Series)
     check(assert_type(left / c, pd.Series), pd.Series)
 
-    check(assert_type(b / left, pd.Series), pd.Series)
+    check(assert_type(b / left, pd.Series), pd.Series)  # type: ignore[assert-type]
```
Dr-Irv: can you add a comment here as to what type mypy and pyright think this is?

tests/series/test_series.py
```diff
@@ -1607,7 +1607,7 @@ def test_series_min_max_sub_axis() -> None:
     check(assert_type(sa, pd.Series), pd.Series)
     check(assert_type(ss, pd.Series), pd.Series)
     check(assert_type(sm, pd.Series), pd.Series)
-    check(assert_type(sd, pd.Series), pd.Series)
+    check(assert_type(sd, pd.Series), pd.Series)  # type: ignore[assert-type]
```
Dr-Irv: I don't like the ignore here.   It seems to be a mypy issue, so can you add a comment here that mypy infers `Any` while `pyright` correctly infers `Series[Any]` ?

pandas-stubs/core/series.pyi
```diff
@@ -2160,28 +2190,51 @@ class Series(IndexOpsMixin[S1], NDFrame):
...
-        other: np_ndarray_complex,
+        other: _T_COMPLEX | Sequence[_T_COMPLEX] | Series[_T_COMPLEX],
         level: Level | None = None,
         fill_value: float | None = None,
         axis: AxisIndex = 0,
-    ) -> Series[complex]: ...
+    ) -> Series[_T_COMPLEX]: ...
     @overload
     def truediv(
         self: Series[float],
```
Dr-Irv, line 2240: I think you need to add `np_ndarray_bool` here as acceptable in `other`

tests/series/arithmetic/test_truediv.py
Dr-Irv: Oh, I see you fixed it in the last commit!

pandas-stubs/core/series.pyi
```diff
@@ -2101,22 +2101,25 @@ class Series(IndexOpsMixin[S1], NDFrame):
...
+        ),
     ) -> Series[float]: ...
     @overload
-    def __truediv__(  # type: ignore[overload-overlap]
-        self: Series[bool],
-        other: _T_NUM_NON_BOOL | Sequence[_T_NUM_NON_BOOL] | Series[_T_NUM_NON_BOOL],
-    ) -> Series[_T_NUM_NON_BOOL]: ...
+    def __truediv__(
+        self: Series[bool], other: complex | Sequence[complex] | Series[complex]
+    ) -> Series[complex]: ...
```
Dr-Irv: Shouldn't you make similar changes to `truediv()` since I would expect that to look the same with the `other` argument and the results of each function.

tests/series/arithmetic/bool/test_truediv.py
```diff
@@ -179,13 +179,13 @@ def test_truediv_pd_series() -> None:
     c = pd.Series([1.1j, 2.2j, 4.1j])
 
     if TYPE_CHECKING_INVALID_USAGE:
-        assert_type(left / b, Never)
+        _ = left / b  # type: ignore[operator] # pyright: ignore[reportOperatorIssue]
```
Dr-Irv, line 182: I like this!!! (and the one below)

Dr-Irv (changes requested):
I see that you were able to remove the 2 ignores.  Request a review when you've handled my comments from the previous review (2 of which you can now dismiss) and from this review.

pandas-stubs/core/series.pyi
cmp0xff: You are right, now added

tests/series/test_series.py
cmp0xff: Now removed

pandas-stubs/core/series.pyi
cmp0xff: Now they should look the same 424537503b732f88120ea8de87adb062a152808f

tests/series/arithmetic/bool/test_truediv.py
cmp0xff: Thanks 🙂 I managed to make more working like this.

pandas-stubs/core/series.pyi
```diff
@@ -2231,21 +2293,30 @@ class Series(IndexOpsMixin[S1], NDFrame):
     ) -> Series: ...
     div = truediv
     @overload
-    def __rtruediv__(self: Series[Never], other: Scalar | _ListLike) -> Series: ...
+    def __rtruediv__(self: Series[Never], other: complex | _ListLike) -> Series: ...
```
Dr-Irv: `rtruediv()` includes `Series` in `other` - please reconcile

pandas-stubs/core/series.pyi
```diff
@@ -2231,21 +2293,30 @@ class Series(IndexOpsMixin[S1], NDFrame):
...
+    @overload
+    def __rtruediv__(self: Series[bool], other: bool | np_ndarray_bool) -> Never: ...
```
Dr-Irv, line 2300: `rtruediv()` doesn't have this overload  - please reconcile

pandas-stubs/core/series.pyi
```diff
@@ -2231,21 +2293,30 @@ class Series(IndexOpsMixin[S1], NDFrame):
...
+    @overload
+    def __rtruediv__(  # pyright: ignore[reportOverlappingOverload]
+        self: Series[bool],
+        other: float | Sequence[float] | np_ndarray_anyint | np_ndarray_float,
```
Dr-Irv: `rtruediv() has `Series[int]` and `Series[float]` for `other` - please reconcile

pandas-stubs/core/series.pyi
```diff
@@ -2231,21 +2293,30 @@ class Series(IndexOpsMixin[S1], NDFrame):
...
+    ) -> Series[float]: ...
+    @overload
+    def __rtruediv__(
+        self: Series[bool], other: complex | Sequence[complex]
```
Dr-Irv: `rtruediv()` has `Series[complex]` for `other` - please reconcile

pandas-stubs/core/series.pyi
```diff
@@ -2231,21 +2293,30 @@ class Series(IndexOpsMixin[S1], NDFrame):
...
+    ) -> Series[complex]: ...
     @overload
     def __rtruediv__(
         self: Series[int],
-        other: int | Sequence[int] | np_ndarray_anyint | np_ndarray_float,
+        other: (
+            int | Sequence[int] | np_ndarray_bool | np_ndarray_anyint | np_ndarray_float
```
Dr-Irv: `rtruediv()` has `Series[_T_INT]` for `other` - please reconcile

pandas-stubs/core/series.pyi
```diff
@@ -2231,21 +2293,30 @@ class Series(IndexOpsMixin[S1], NDFrame):
...
+        ),
     ) -> Series[float]: ...
     @overload
     def __rtruediv__(
         self: Series[int], other: _T_COMPLEX | Sequence[_T_COMPLEX]
```
Dr-Irv: `rtrudediv()` has `Series[_T_COMPLEX]` for `other` - please reconcile

pandas-stubs/core/series.pyi
```diff
@@ -2231,21 +2293,30 @@ class Series(IndexOpsMixin[S1], NDFrame):
...
     ) -> Series[_T_COMPLEX]: ...
     @overload
-    def __rtruediv__(
-        self: Series[int], other: np_ndarray_complex
-    ) -> Series[complex]: ...
-    @overload
     def __rtruediv__(
         self: Series[float],
         other: int | Sequence[int] | np_ndarray_anyint | np_ndarray_float,
```
Dr-Irv: `rtruediv()` has `np_ndarray_bool` and `Series[_T_INT]` for `other` - please reconcile

pandas-stubs/core/series.pyi
```diff
@@ -2255,65 +2326,87 @@ class Series(IndexOpsMixin[S1], NDFrame):
         self: Series[float], other: _T_COMPLEX | Sequence[_T_COMPLEX]
```
Dr-Irv: `rtruediv()` has `Series[_T_COMPLEX]` for `other` - please reconcile

pandas-stubs/core/series.pyi
```diff
@@ -2255,65 +2326,87 @@ class Series(IndexOpsMixin[S1], NDFrame):
...
-    ) -> Series[complex]: ...
-    @overload
     def __rtruediv__(
         self: Series[complex],
         other: (
             _T_COMPLEX
             | Sequence[_T_COMPLEX]
+            | np_ndarray_bool
             | np_ndarray_anyint
             | np_ndarray_float
```
Dr-Irv, line 2358: `rtruediv()` has `Series[_T_COMPLEX]` for `other` - please reconcile

Dr-Irv (changes requested):
I checked and `__truediv__()` and `truediv()` have the same patterns.

Then checked `__rtruediv__()` and `rtruediv()` and found discrepancies.

Probably should do the same for `add()`  and `radd()` at some point, but that can be another PR

pandas-stubs/core/series.pyi
cmp0xff: In my plan, `Series[...] / Series[...]` is given by `__truediv__`, so that having `Series` in `others` for `__rtruediv__` is redundant, which may confuse the type checker. What would you say?

pandas-stubs/core/series.pyi
cmp0xff: But it does, https://github.com/pandas-dev/pandas-stubs/pull/1314/files#diff-6f90f15b437974706a3c0a75636efd6d163a42d4a786bc5513a5069b1040f031R2361-R2368

pandas-stubs/core/series.pyi
cmp0xff: - `np_ndarray_bool`: 6f39da6f3e133fe9f73a4d3c428bd3607136a904
- `Series[_T_INT]`: In my plan, `Series[...] / Series[...]` is given by `__truediv__`, so that having Series in others for `__rtruediv__` is redundant, which may confuse the type checker.

pandas-stubs/core/series.pyi
Dr-Irv: You're right, my mistake.

pandas-stubs/core/series.pyi
Dr-Irv: I think it will be OK.  Let's take the above example assuming you added `Series` to `other`.  If you had code with `x / y` where both `x` and `y` were `Series`, the type checker will first check `Series.__truediv__(y)` for a match.  If there is no match, then it will check `Series.__rtruediv__(x)` for a match.  So I don't think the redundancy will hurt at all.

pandas-stubs/core/series.pyi
Dr-Irv: see above comment - I think it will be fine.

pandas-stubs/core/series.pyi
cmp0xff: With 1853a8d9f5a823d5cd554bd764bfe3ee5e284cd9 all missing `Series` should have been added.

pandas-stubs/core/series.pyi
```diff
@@ -1830,45 +1831,51 @@ class Series(IndexOpsMixin[S1], NDFrame):
...
-            | Sequence[_T_COMPLEX]
+            | Series[_T_COMPLEX]
         ),
     ) -> Series[complex]: ...
     @overload
+    def __radd__(
+        self: Series[_T_COMPLEX], other: np_ndarray_complex
+    ) -> Series[complex]: ...
+    @overload
     def __radd__(self, other: S1) -> Self: ...
```
Dr-Irv: For `radd()`, this has `other: S1 | Series[S1]`

Dr-Irv (changes requested):
Thanks for also checking on `__radd__()`/`radd()` .  I've checked them all now, and found one that I commented on, and one that is missing but I couldn't write a comment.  This one is missing from `add()`:
```python
    @overload
    def __add__(self, other: Series[Never]) -> Series: ...
```

Otherwise, it's all good

pandas-stubs/core/series.pyi
cmp0xff: 7d585d915a8988f7165f50f1fa0d2c4d9b095473

PR #1293, opened 2025-07-27
Add defaults for parameters

yangdanny97:
Closes #1292

This PR uses a semi-automated/LLM-assisted method to add a few hundred parameter defaults to pandas-stubs, replacing what was previously `...`.

The codemod that made these transformations is here: https://gist.github.com/yangdanny97/0d81e16f6374582413dd5bcf90d902fd

How it works is:
1. parse pandas-stubs and find the module, method/function, & parameter name for any params w/ default `...`
2. parse the docstring of the method/function in pandas, and if the parameter's docstring contains "default" extract the part after that keyword
3. look up the type annotation in a mapping of the extracted snippet to type annotation
4. use libCST to apply the type annotation to the parameter

The main thing that needs careful reviewing is the `default_type_map` which maps the docstring snippet to the type annotation we replace the `...` with. That part was generated with chatGPT's help (I gave it every snippet I got after step 2), and I commented out the mappings that seemed wrong or unclear. So, even though this was "LLM-assisted" the script is deterministic and will always give you the same thing each time, and the LLM is only involved in a one-time manual step.

PR #1315, opened 2025-08-11
add some missing Optional type annotations

MarcoGorelli:
These are some cases where the overloads indicate that they are meant to match the parameter defaults, which in all these cases are `= None`. However, `None` is currently missing from the annotations for these parameters


pandas-stubs/core/indexes/interval.pyi
```diff
@@ -310,8 +310,8 @@ class IntervalIndex(ExtensionIndex[IntervalT], IntervalMixin):
 # int gets hit first and so the correct type is returned
 @overload
 def interval_range(  # pyright: ignore[reportOverlappingOverload]
-    start: int = ...,
-    end: int = ...,
+    start: int | None = ...,
+    end: int | None = ...,
```
MarcoGorelli: https://github.com/pandas-dev/pandas/blob/7863029c3084c79b226c0a1c3daeac82929534f1/pandas/core/indexes/interval.py#L1111-L1127

pandas-stubs/io/sql.pyi
```diff
@@ -115,7 +115,7 @@ def read_sql(
         | None
     ) = ...,
     parse_dates: list[str] | dict[str, str] | dict[str, dict[str, Any]] | None = ...,
-    columns: list[str] = ...,
+    columns: list[str] | None = ...,
```
MarcoGorelli, line 118: https://github.com/pandas-dev/pandas/blob/7863029c3084c79b226c0a1c3daeac82929534f1/pandas/io/sql.py#L542

https://github.com/pandas-dev/pandas/blob/7863029c3084c79b226c0a1c3daeac82929534f1/pandas/io/sql.py#L588-L590

Dr-Irv, line 118: Looks like #1293 missed the defaults for `read_sql()`.  So can you make this `columns: list[str] | None = None`  (and below as well)?

Dr-Irv (changes requested):
If you want to fix the other defaults for `read_sql()` that would be good as well, but not a problem if you don't.

pandas-stubs/core/indexes/interval.pyi
```diff
@@ -310,59 +310,17 @@ class IntervalIndex(ExtensionIndex[IntervalT], IntervalMixin):
...
-@overload
-def interval_range(  # pyright: ignore[reportOverlappingOverload]
-    start: int,
-    *,
-    end: None = ...,
-    periods: int | None = ...,
-    freq: int | None = ...,
-    name: Hashable = ...,
-    closed: IntervalClosedType = ...,
-) -> IntervalIndex[Interval[int]]: ...
```
MarcoGorelli, line 331: having corrected the first overload, this one would never be matched, so it can be removed

pandas-stubs/core/indexes/interval.pyi
```diff
@@ -310,59 +310,17 @@ class IntervalIndex(ExtensionIndex[IntervalT], IntervalMixin):
...
-@overload
-def interval_range(
-    start: float,
-    *,
-    end: None = ...,
-    periods: int | None = ...,
-    freq: int | None = ...,
-    name: Hashable = ...,
-    closed: IntervalClosedType = ...,
-) -> IntervalIndex[Interval[float]]: ...
```
MarcoGorelli, line 360: same

pandas-stubs/core/indexes/interval.pyi
```diff
@@ -388,19 +346,9 @@ def interval_range(
...
 @overload
-def interval_range(
-    start: _TimestampLike,
-    *,
-    end: None = ...,
-    periods: int | None = ...,
-    freq: str | BaseOffset | pd.Timedelta | dt.timedelta | None = ...,
-    name: Hashable = ...,
-    closed: IntervalClosedType = ...,
-) -> IntervalIndex[Interval[pd.Timestamp]]: ...
```
MarcoGorelli, line 399: same

pandas-stubs/core/indexes/interval.pyi
```diff
@@ -416,13 +364,3 @@ def interval_range(
...
-@overload
-def interval_range(
-    start: _TimedeltaLike,
-    *,
-    end: None = ...,
-    periods: int | None = ...,
-    freq: str | BaseOffset | pd.Timedelta | dt.timedelta | None = ...,
-    name: Hashable = ...,
-    closed: IntervalClosedType = ...,
-) -> IntervalIndex[Interval[pd.Timedelta]]: ...
```
MarcoGorelli, line 428: same

MarcoGorelli:
Thanks @Dr-Irv for your review!

I noticed these missing `| None`s while looking into a follow-up to #1293. There's quite a few more that are missing. For ease of review, OK if we start with fixing the missing `| None`s, and in a follow-up I'll address the defaults?

Dr-Irv:
> I noticed these missing `| None`s while looking into a follow-up to #1293. There's quite a few more that are missing. For ease of review, OK if we start with fixing the missing `| None`s, and in a follow-up I'll address the defaults?

Yes, that's fine.

PR #1318, opened 2025-08-13
Add more parameter defaults

MarcoGorelli:
Towards #1317 

Due to improvements in https://github.com/yangdanny97/docs2types, further places with missing `| None` have emerged, I'll fix those separately

- [ ] Closes #xxxx (Replace xxxx with the Github issue number)
- [ ] Tests added: Please use `assert_type()` to assert the type of any return value

PR #1319, opened 2025-08-13
add missing `| None` to `copy` argument in `concat`

MarcoGorelli:
Precursor to extending https://github.com/pandas-dev/pandas-stubs/pull/1318 to keyword-only arguments

🤔 though arguably, `copy` is deprecated on the pandas main branch, so perhaps shouldn't be done?


MarcoGorelli:
Actually, as per https://github.com/pandas-dev/pandas-stubs/pull/1293#issuecomment-3130128094

> I looked further into the copy argument issue. Eventually, that argument will be removed. So I think it would be best to not change copy: _bool = False or copy: _bool = True to be copy: _bool | None = None, because None won't be accepted in the future. So ignore those requests in my review.

this probably shouldn't be changed at all

Closing this then

PR #1320, opened 2025-08-13
remove (deprecated) `axis` from `group_by`

MarcoGorelli:


Dr-Irv (changes requested):
Should also remove `axis` argument in `pandas-stubs/core/series.pyi` for `groupby()` there.

PR #1321, opened 2025-08-13
fix overload defaults for `as_index` in `groupby`

MarcoGorelli:
These overloads currently show `as_index: Literal[False] = ...`, but that's not quite correct because the default is `True`

https://github.com/pandas-dev/pandas/blob/9597c0397962b228f00805e0750b91d0e5272ce9/pandas/core/frame.py#L9375

If `as_index: Literal[False]` were to be used instead, then mypy would complain that a non-default parameter (`as_index`) follows a default one (`level` / `axis`). Fixing this would require making an extra overload: one for when `as_index` is passed positionally and one for when it's passed by name

Given that:
- groupby already has 16 (!) overloads
- `as_index` is boolean, [so arguably it's a best practice to pass it by name anyway](https://docs.astral.sh/ruff/rules/boolean-positional-value-in-call/)

Is it OK to just type if it if it's passed by name? The alternative would be...to bring the number of overloads to 24 🤯 Certainly doable, my preference would be to keep them down and encourage passing `as_index` by name

### DIFF
--- a/pandas-stubs/core/frame.pyi
+++ b/pandas-stubs/core/frame.pyi
@@ -1145,7 +1145,7 @@ class DataFrame(NDFrame, OpsMixin, _GetItemHack):
         errors: IgnoreRaise = "ignore",
     ) -> None: ...
     @overload
-    def groupby(  # pyright: ignore reportOverlappingOverload
+    def groupby(
         self,
         by: Scalar,
         axis: AxisIndex | _NoDefaultDoNotUse = ...,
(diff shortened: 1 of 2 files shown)


MarcoGorelli:
The issue with that is that if `as_index` isn't specified, then it'll match the first overload, whereas it should match the second one

Perhaps 24 overloads isn't that bad? Or should I first deprecate non-keyword arguments except `level` in pandas itself?

Dr-Irv:
> The issue with that is that if `as_index` isn't specified, then it'll match the first overload, whereas it should match the second one

Right.  I see your point.  
> 
> Perhaps 24 overloads isn't that bad? Or should I first deprecate non-keyword arguments except `level` in pandas itself?

I'm OK with 24 overloads, although I think we should deprecate non-keyword arguments expect `level` in pandas itself first.

MarcoGorelli:
closing as this has been superseded by other prs 🥳

PR #1322, opened 2025-08-13
Clean up timefuncs tests, fix nightly due to new errors

loicdiridollou:

Some of the new error names from https://github.com/pandas-dev/pandas/pull/61468.
Nightly run is passing, see https://github.com/pandas-dev/pandas-stubs/actions/runs/16948579852.

### DIFF
--- a/tests/series/test_series.py
+++ b/tests/series/test_series.py
@@ -98,6 +98,10 @@ else:
     TimestampSeries: TypeAlias = pd.Series
     OffsetSeries: TypeAlias = pd.Series
 
+if not PD_LTE_23:
+    from pandas.errors import Pandas4Warning  # type: ignore[attr-defined]  # pyright: ignore  # isort: skip
+else:
+    Pandas4Warning: TypeAlias = FutureWarning  # type: ignore[no-redef]
 
 # Tests will use numpy 2.1 in python 3.10 or later
 # From Numpy 2.1 __init__.pyi
@@ -3863,7 +3867,7 @@ def test_series_reindex_like() -> None:
             upper="2.99",
         ),
         pytest_warns_bounded(
-            Warning,  # should be Pandas4Warning but only exposed starting pandas 3.0.0
+            Pandas4Warning,
             "the 'method' keyword is deprecated and will be removed in a future version. Please take steps to stop the use of 'method'",
             lower="2.99",
             upper="3.0.99",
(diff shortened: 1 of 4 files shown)


tests/series/test_series.py
```diff
@@ -3855,11 +3855,19 @@ def test_series_reindex() -> None:
...
+            "the 'method' keyword is deprecated and will be removed in a future version. Please take steps to stop the use of 'method'",
+            lower="2.3.99",
+            upper="2.99",
+        ),
+        pytest_warns_bounded(
+            Warning,  # should be Pandas4Warning but only exposed starting pandas 3.0.0
+            "the 'method' keyword is deprecated and will be removed in a future version. Please take steps to stop the use of 'method'",
+            lower="2.99",
+            upper="3.0.99",
+        ),
```
Dr-Irv: We can do this differently:
1. at the top of this file, have something like:
```python
if not PD_LTE_23:
    from pandas.errors import Pandas4Warning
else:
    Pandas4Warning: TypeAlias = FutureWarning
```
2. Then you do `pytest_warns_bounded(Pandas4Warning, "**put warning string here**, lower="2.3.99", upper="2.99")


I won't comment on the other places you should make similar changes.

tests/test_indexes.py
```diff
@@ -1372,7 +1372,9 @@ def test_datetimeindex_shift() -> None:
 
 def test_timedeltaindex_shift() -> None:
     ind = pd.date_range("1/1/2021", "1/5/2021") - pd.Timestamp("1/3/2019")
-    check(assert_type(ind.shift(1), pd.TimedeltaIndex), pd.TimedeltaIndex)
+    if PD_LTE_23:
+        # cannot shift with no freq starting in pandas 3.0.0
+        check(assert_type(ind.shift(1), pd.TimedeltaIndex), pd.TimedeltaIndex)
```
Dr-Irv: This is a bug in main right now, and there is a PR to fix it, so let's not make this change.  See https://github.com/pandas-dev/pandas/issues/62094 and the associated PR.  I'm fine to live with nightly broke on this for now given it will get fixed within a few days.

Dr-Irv (changes requested):

loicdiridollou:
I will keep an eye on the PR you mentioned, also quick note I had to add `# isort: skip` since it kept putting the `type: ignore` on a new line (because line was too long) but type ignore needs to be at the `from ...` level

Dr-Irv (approved):
thanks @loicdiridollou

PR #1323, opened 2025-08-15
Fix series creation from dict views

hamdanal:

`collections.dict_views` does not exist neither at runtime nor in the stubs, the correct type to use is `KeysView` from `collections.abc`. While fixing this I noticed that values views is not supported so I added it. Note that values views is more relevant  to the series class as it is NOT set-like but IS sequence-like.


Dr-Irv (approved):
thanks @hamdanal

PR #1325, opened 2025-08-15
Add numpy array shapes for return types

### LINKED ISSUE #1324: Shape types for numpy arrays in return types
Most of pandas functions that return a numpy array have a fixed and known shape. For example, `index.str.contains` always returns a 1D-array of type `np.bool` and `df.to_numpy` always returns a 2D-array. Currently, the codebase doesn't type the shape of numpy arrays. Having the shape types is super useful as it allows type checkers to better infer certain types and IDEs to offer better autocompletion. Consider this example:

```python
import numpy as np
import numpy.typing as npt

a_nd: npt.NDArray[np.float64] = np.array([1, 2], dtype=np.float64)
a_1d = a_nd.ravel()  # annotated as returning 1D array in numpy
for x in a_nd:
    reveal_type(x)
for y in a_1d:
    reveal_type(y)
```

mypy outputs the following:

```
ex.py:7: note: Revealed type is "Any"
ex.py:9: note: Revealed type is "numpy.float64"
```

It correctly identified the type of the loop variable as `float64` instead of `Any` because it knows the array is 1D.

I am exploring adding shape types to the numpy interface of pandas-stubs and modifying the test infrastructure to validate the shape as well as the dtype of arrays.

hamdanal:
- [x] Closes #1324

I made these changes:
- Added array shapes in several places where the shape is known statically
- Updated existing tests and added missing tests
- Modified the `check` function in the tests to validate the shape and dtype of numpy arrays


hamdanal:
The mypy test failures in CI (first couple commits) are weird. They appear on Python 3.10 only and only in CI. I couldn't reproduce them locally even with the same versions of python, mypy, pandas and numpy as in CI. I skip these test only for mypy and only on Python 3.10.

pandas-stubs/_libs/tslibs/timestamps.pyi
```diff
@@ -241,22 +251,26 @@ class Timestamp(datetime, SupportsIndex):
...
+    ) -> np_ndarray[ShapeT, np.datetime64]: ...
     @overload
     def __eq__(self, other: Timestamp | datetime | np.datetime64) -> bool: ...  # type: ignore[overload-overlap] # pyright: ignore[reportOverlappingOverload]
     @overload
     def __eq__(self, other: TimestampSeries) -> Series[bool]: ...  # type: ignore[overload-overlap]
     @overload
-    def __eq__(self, other: npt.NDArray[np.datetime64] | Index) -> np_ndarray_bool: ...  # type: ignore[overload-overlap]
+    def __eq__(self, other: Index) -> np_1darray[np.bool]: ...  # type: ignore[overload-overlap]
+    @overload  # TODO: using shape-aware arrays similar to other methods doesn't work in mypy
+    def __eq__(self, other: npt.NDArray[np.datetime64]) -> npt.NDArray[np.bool]: ...  # type: ignore[overload-overlap]
```
Dr-Irv: I got this to work.  See comments elsewhere.

pandas-stubs/core/dtypes/missing.pyi
```diff
@@ -41,7 +48,11 @@ def notna(obj: DataFrame) -> DataFrame: ...
 @overload
 def notna(obj: Series) -> Series[bool]: ...
 @overload
-def notna(obj: Index | list[Any] | ArrayLike) -> npt.NDArray[np.bool_]: ...
+def notna(obj: Index | ExtensionArray | list[ScalarT]) -> np_1darray[np.bool]: ...
+@overload
+def notna(obj: np_ndarray[ShapeT]) -> np_ndarray[ShapeT, np.bool]: ...
+@overload
+def notna(obj: Index | list[Any]) -> np_ndarray_bool: ...
```
Dr-Irv: same comment as above

tests/__init__.py
```diff
@@ -40,11 +43,16 @@
...
         np_ndarray_bool as np_ndarray_bool,
         np_ndarray_int as np_ndarray_int,
     )
 else:
+    _G = TypeVar("_G", bound=np.generic)
     # Separately define here so pytest works
+    np_1darray = np.ndarray[tuple[int], np.dtype[_G]]
+    np_2darray = np.ndarray[tuple[int, int], np.dtype[_G]]
     np_ndarray_bool = npt.NDArray[np.bool_]
     np_ndarray_int = npt.NDArray[np.signedinteger]
```
Dr-Irv: Declare these using `: TypeAlias`

tests/__init__.py
```diff
@@ -61,8 +69,39 @@ def check(
     attr: str = "left",
     index_to_check_for_type: Literal[0, -1] = 0,
 ) -> T:
-    if not isinstance(actual, klass):
+    __tracebackhide__ = True
```
Dr-Irv, line 76: what is this for???

tests/test_indexes.py
```diff
@@ -42,13 +43,13 @@ def test_index_duplicated() -> None:
     df = pd.DataFrame({"x": [1, 2, 3, 4]}, index=pd.Index([1, 2, 3, 2]))
     ind = df.index
     duplicated = ind.duplicated("first")
-    check(assert_type(duplicated, npt.NDArray[np.bool_]), np.ndarray, np.bool_)
+    check(assert_type(duplicated, np_1darray[np.bool]), np_1darray[np.bool])
```
Dr-Irv, line 46: With the previous test of `check(assert_type(duplicated, npt.NDArray[np.bool_]), np.ndarray, np.bool_)` the fcheck function would also be checking that the first element of the array would be of type `np.bool_`.  Now you have lost that in this particular test. 

I think here if you did `check(assert_type(duplicated, np_1darray[np.bool]), np_1darray[np.bool], np.bool_)` it would pass.

tests/test_indexes.py
```diff
@@ -42,13 +43,13 @@ def test_index_duplicated() -> None:
...
 
 
 def test_index_isin() -> None:
     ind = pd.Index([1, 2, 3, 4, 5])
     isin = ind.isin([2, 4])
-    check(assert_type(isin, npt.NDArray[np.bool_]), np.ndarray, np.bool_)
+    check(assert_type(isin, np_1darray[np.bool]), np_1darray[np.bool])
```
Dr-Irv, line 52: same coment

tests/test_indexes.py
```diff
@@ -202,13 +203,6 @@ def test_str_rsplit() -> None:
     )
 
 
-def test_str_match() -> None:
-    i = pd.Index(
-        ["applep", "bananap", "Cherryp", "DATEp", "eGGpLANTp", "123p", "23.45p"]
-    )
-    check(assert_type(i.str.match("pp"), npt.NDArray[np.bool_]), np.ndarray, np.bool_)
-
-
```
Dr-Irv, line 211: why is this test removed?

tests/test_indexes.py
```diff
@@ -289,10 +283,10 @@ def test_index_relops() -> None:
...
 
     ind = pd.Index([1, 2, 3])
-    check(assert_type(ind <= 2, npt.NDArray[np.bool_]), np.ndarray, np.bool_)
-    check(assert_type(ind >= 2, npt.NDArray[np.bool_]), np.ndarray, np.bool_)
-    check(assert_type(ind < 2, npt.NDArray[np.bool_]), np.ndarray, np.bool_)
-    check(assert_type(ind > 2, npt.NDArray[np.bool_]), np.ndarray, np.bool_)
+    check(assert_type(ind <= 2, np_1darray[np.bool]), np_1darray[np.bool])
+    check(assert_type(ind >= 2, np_1darray[np.bool]), np_1darray[np.bool])
+    check(assert_type(ind < 2, np_1darray[np.bool]), np_1darray[np.bool])
+    check(assert_type(ind > 2, np_1darray[np.bool]), np_1darray[np.bool])
```
Dr-Irv, line 305: same comment - keep the `np.bool_` argument.

tests/test_indexes.py
```diff
@@ -1299,27 +1293,40 @@ def test_datetime_operators_builtin() -> None:
...
+    )
+    overlap_interval_index = pd.IntervalIndex([i1, i2, i3])
+    check(
+        assert_type(
+            overlap_interval_index.get_loc(1), Union[int, slice, np_1darray[np.bool]]
         ),
-        np.ndarray,
-        np.bool_,
+        np_1darray[np.bool],
     )
```
Dr-Irv, line 1346: add back the `np.bool_` here

tests/test_pandas.py
```diff
@@ -428,20 +431,47 @@ def test_types_json_normalize() -> None:
...
+
+    # N-D numpy array
+    arr_nd = idx.to_numpy().reshape([2, 2])
+    check(assert_type(pd.isna(arr_nd), np_ndarray_bool), np_ndarray_bool)
+    check(assert_type(pd.notna(arr_nd), np_ndarray_bool), np_ndarray_bool)
+
+    # List of scalars
+    l_sca = [1, 2.5, float("nan")]
+    check(assert_type(pd.isna(l_sca), np_1darray[np.bool]), np_1darray[np.bool])
+    check(assert_type(pd.notna(l_sca), np_1darray[np.bool]), np_1darray[np.bool])
```
Dr-Irv, line 469: For all of these, add the `np.bool_` argument to `check()`

tests/test_scalars.py
```diff
@@ -314,73 +318,55 @@ def test_interval_cmp():
 
     interval_index_int = pd.IntervalIndex([interval_i])
     check(
-        assert_type(interval_index_int >= interval_i, np_ndarray_bool),
-        np.ndarray,
-        np.bool_,
+        assert_type(interval_index_int >= interval_i, np_1darray[np.bool]),
+        np_1darray[np.bool],
     )
```
Dr-Irv, line 320: for these tests, put back the `np.bool_` argument.  (This one and all the ones through line 384 of the previous version

tests/test_scalars.py
```diff
@@ -854,10 +840,10 @@ def test_timedelta_cmp() -> None:
     check(assert_type(td < c_dt_timedelta, bool), bool)
     check(assert_type(td < c_timedelta64, bool), bool)
     check(assert_type(td < c_ndarray_td64, np_ndarray_bool), np.ndarray, np.bool_)
-    check(assert_type(c_timedelta_index < td, np_ndarray_bool), np.ndarray, np.bool_)
+    check(assert_type(c_timedelta_index < td, np_1darray[np.bool]), np_1darray[np.bool])
```
Dr-Irv: add back the `np.bool_` argument

tests/test_scalars.py
```diff
@@ -854,10 +840,10 @@ def test_timedelta_cmp() -> None:
...
     check(assert_type(c_dt_timedelta < td, bool), bool)
     check(assert_type(c_ndarray_td64 < td, np_ndarray_bool), np.ndarray, np.bool_)
```
Dr-Irv: why isn't this `np.1darray[np.bool]` ?

tests/test_scalars.py
```diff
@@ -854,10 +840,10 @@ def test_timedelta_cmp() -> None:
...
-    check(assert_type(c_timedelta_index < td, np_ndarray_bool), np.ndarray, np.bool_)
+    check(assert_type(c_timedelta_index < td, np_1darray[np.bool]), np_1darray[np.bool])
```
Dr-Irv: add back the `np.bool_` argument

tests/test_scalars.py
```diff
@@ -912,10 +898,10 @@ def test_timedelta_cmp() -> None:
     assert (gt_a != le_a).all()
 
     gt_a = check(
-        assert_type(c_timedelta_index > td, np_ndarray_bool), np.ndarray, np.bool_
+        assert_type(c_timedelta_index > td, np_1darray[np.bool]), np_1darray[np.bool]
     )
     le_a = check(
-        assert_type(c_timedelta_index <= td, np_ndarray_bool), np.ndarray, np.bool_
+        assert_type(c_timedelta_index <= td, np_1darray[np.bool]), np_1darray[np.bool]
     )
```
Dr-Irv: add back the `np.bool_` arguments

tests/test_scalars.py
```diff
@@ -980,10 +966,10 @@ def test_timedelta_cmp() -> None:
     assert (lt_a != ge_a).all()
 
     lt_a = check(
-        assert_type(c_timedelta_index < td, np_ndarray_bool), np.ndarray, np.bool_
+        assert_type(c_timedelta_index < td, np_1darray[np.bool]), np_1darray[np.bool]
     )
     ge_a = check(
-        assert_type(c_timedelta_index >= td, np_ndarray_bool), np.ndarray, np.bool_
+        assert_type(c_timedelta_index >= td, np_1darray[np.bool]), np_1darray[np.bool]
```
Dr-Irv: add back the `np.bool_` arguments

tests/test_scalars.py
```diff
@@ -1066,10 +1052,10 @@ def test_timedelta_cmp_rhs() -> None:
     assert (eq_a != ne_a).all()
 
     eq_a = check(
-        assert_type(c_timedelta_index == td, np_ndarray_bool), np.ndarray, np.bool_
+        assert_type(c_timedelta_index == td, np_1darray[np.bool]), np_1darray[np.bool]
     )
     ne_a = check(
-        assert_type(c_timedelta_index != td, np_ndarray_bool), np.ndarray, np.bool_
+        assert_type(c_timedelta_index != td, np_1darray[np.bool]), np_1darray[np.bool]
```
Dr-Irv: add back the `np.bool_` arguments

tests/test_scalars.py
```diff
@@ -1281,13 +1271,17 @@ def test_timestamp_cmp() -> None:
...
     assert gt != lte
 
-    check(assert_type(ts > c_datetimeindex, np_ndarray_bool), np.ndarray, np.bool_)
-    check(assert_type(ts <= c_datetimeindex, np_ndarray_bool), np.ndarray, np.bool_)
-    check(assert_type(ts > c_unknown_index, np_ndarray_bool), np.ndarray, np.bool_)
-    check(assert_type(ts <= c_unknown_index, np_ndarray_bool), np.ndarray, np.bool_)
+    check(assert_type(ts > c_datetimeindex, np_1darray[np.bool]), np_1darray[np.bool])
+    check(assert_type(ts <= c_datetimeindex, np_1darray[np.bool]), np_1darray[np.bool])
+    check(assert_type(ts > c_unknown_index, np_1darray[np.bool]), np_1darray[np.bool])
+    check(assert_type(ts <= c_unknown_index, np_1darray[np.bool]), np_1darray[np.bool])
```
Dr-Irv: add back the `np.bool_` arguments

tests/test_scalars.py
```diff
@@ -1281,13 +1271,17 @@ def test_timestamp_cmp() -> None:
...
 
     check(assert_type(ts > c_np_ndarray_dt64, np_ndarray_bool), np.ndarray, np.bool_)
     check(assert_type(ts <= c_np_ndarray_dt64, np_ndarray_bool), np.ndarray, np.bool_)
```
Dr-Irv: why aren't these `np.1darray[np.bool]` ?

tests/test_scalars.py
```diff
@@ -1302,13 +1296,17 @@ def test_timestamp_cmp() -> None:
...
     assert gt != lte
 
-    check(assert_type(c_datetimeindex > ts, np_ndarray_bool), np.ndarray, np.bool_)
-    check(assert_type(c_datetimeindex <= ts, np_ndarray_bool), np.ndarray, np.bool_)
-    check(assert_type(c_unknown_index > ts, np_ndarray_bool), np.ndarray, np.bool_)
-    check(assert_type(c_unknown_index <= ts, np_ndarray_bool), np.ndarray, np.bool_)
+    check(assert_type(c_datetimeindex > ts, np_1darray[np.bool]), np_1darray[np.bool])
+    check(assert_type(c_datetimeindex <= ts, np_1darray[np.bool]), np_1darray[np.bool])
+    check(assert_type(c_unknown_index > ts, np_1darray[np.bool]), np_1darray[np.bool])
+    check(assert_type(c_unknown_index <= ts, np_1darray[np.bool]), np_1darray[np.bool])
```
Dr-Irv: add back the `np.bool_` arguments

tests/test_scalars.py
```diff
@@ -1302,13 +1296,17 @@ def test_timestamp_cmp() -> None:
...
 
     check(assert_type(c_np_ndarray_dt64 > ts, np_ndarray_bool), np.ndarray, np.bool_)
     check(assert_type(c_np_ndarray_dt64 <= ts, np_ndarray_bool), np.ndarray, np.bool_)
+    check(assert_type(c_np_2darray_dt64 > ts, np_2darray[np.bool]), np_2darray[np.bool])
+    check(
+        assert_type(c_np_2darray_dt64 <= ts, np_2darray[np.bool]), np_2darray[np.bool]
+    )
```
Dr-Irv: add back the `np.bool_` arguments

tests/test_scalars.py
```diff
@@ -1302,13 +1296,17 @@ def test_timestamp_cmp() -> None:
...
-    check(assert_type(c_datetimeindex <= ts, np_ndarray_bool), np.ndarray, np.bool_)
-    check(assert_type(c_unknown_index > ts, np_ndarray_bool), np.ndarray, np.bool_)
-    check(assert_type(c_unknown_index <= ts, np_ndarray_bool), np.ndarray, np.bool_)
+    check(assert_type(c_datetimeindex > ts, np_1darray[np.bool]), np_1darray[np.bool])
+    check(assert_type(c_datetimeindex <= ts, np_1darray[np.bool]), np_1darray[np.bool])
+    check(assert_type(c_unknown_index > ts, np_1darray[np.bool]), np_1darray[np.bool])
+    check(assert_type(c_unknown_index <= ts, np_1darray[np.bool]), np_1darray[np.bool])
 
     check(assert_type(c_np_ndarray_dt64 > ts, np_ndarray_bool), np.ndarray, np.bool_)
     check(assert_type(c_np_ndarray_dt64 <= ts, np_ndarray_bool), np.ndarray, np.bool_)
```
Dr-Irv: why aren't these `np.1darray[np.bool]` ?

tests/test_scalars.py
```diff
@@ -1325,13 +1323,17 @@ def test_timestamp_cmp() -> None:
...
+    check(assert_type(ts < c_datetimeindex, np_1darray[np.bool]), np_1darray[np.bool])
+    check(assert_type(ts >= c_unknown_index, np_1darray[np.bool]), np_1darray[np.bool])
+    check(assert_type(ts < c_unknown_index, np_1darray[np.bool]), np_1darray[np.bool])
 
     check(assert_type(ts >= c_np_ndarray_dt64, np_ndarray_bool), np.ndarray, np.bool_)
     check(assert_type(ts < c_np_ndarray_dt64, np_ndarray_bool), np.ndarray, np.bool_)
+    check(
+        assert_type(c_np_2darray_dt64 >= ts, np_2darray[np.bool]), np_2darray[np.bool]
+    )
+    check(assert_type(c_np_2darray_dt64 < ts, np_2darray[np.bool]), np_2darray[np.bool])
```
Dr-Irv: So I won't keep repeating here.  If the `np.bool_` argument was in `check()` before, it needs to be there now.

And if you still have `np.ndarray` in the `check` function, I think it should be changed to `np.1darray[np.bool]`

Have to do that throughout this test func

tests/test_scalars.py
```diff
@@ -1370,27 +1376,38 @@ def test_timestamp_cmp() -> None:
...
-    eq_arr = check(
-        assert_type(ts == c_np_ndarray_dt64, np_ndarray_bool), np.ndarray, np.bool_
-    )
-    ne_arr = check(
-        assert_type(ts != c_np_ndarray_dt64, np_ndarray_bool), np.ndarray, np.bool_
-    )
-    assert (eq_arr != ne_arr).all()
+    if sys.version_info >= (3, 11) or not MYPY:
+        # tests in this block fail with mypy on Python 3.10 in CI only
+        # I couldn't reproduce the failure locally so skip mypy on Python 3.10
```
Dr-Irv: I figured it out.  See summary comment.

tests/test_scalars.py
```diff
@@ -1370,27 +1376,38 @@ def test_timestamp_cmp() -> None:
...
+            assert_type(ts != c_np_ndarray_dt64, np_ndarray_bool), np.ndarray, np.bool_
+        )
+        assert (eq_arr != ne_arr).all()
+        # TODO: the following should be 2D-arrays but it doesn't work in mypy
+        eq_arr = check(
+            assert_type(ts == c_np_2darray_dt64, np_ndarray_bool), np_ndarray_bool
+        )
+        ne_arr = check(
+            assert_type(ts != c_np_2darray_dt64, np_ndarray_bool), np_ndarray_bool
+        )
```
Dr-Irv: The reason `mypy` was getting confused is twofold:

1. You are reusing `eq_arr` as the result, so it took the previous type from the tests a few lines above, and said it was a mismatch.  I renamed the variables here and that fixed the python 3.10 problem.
2. In the 2D tests, the type of `c_np_2darray_dt64` is explicitly a `np.ndarray[tuple[int, int], np.datetime64]` so I changed the test to test for `np.ndarray[tuple[int, int], np.bool]`

tests/test_string_accessors.py
```diff
@@ -71,40 +71,42 @@ def test_string_accessors_boolean_series():
 
 def test_string_accessors_boolean_index():
     idx = pd.Index(DATA)
-    _check = functools.partial(check, klass=np.ndarray, dtype=np.bool_)
-    _check(assert_type(idx.str.startswith("a"), np_ndarray_bool))
+    _check = functools.partial(check, klass=np_1darray[np.bool])
```
Dr-Irv, line 74:
```suggestion
    _check = functools.partial(check, klass=np_1darray[np.bool], dtype=np.bool_)
```
Then we will check the value in the array at runtime.

tests/test_timefuncs.py
```diff
@@ -618,22 +622,22 @@ def test_datetimeindex_accessors() -> None:
...
-    check(assert_type(i0.is_year_start, npt.NDArray[np.bool_]), np.ndarray, np.bool_)
-    check(assert_type(i0.is_year_end, npt.NDArray[np.bool_]), np.ndarray, np.bool_)
-    check(assert_type(i0.is_leap_year, npt.NDArray[np.bool_]), np.ndarray, np.bool_)
+    check(assert_type(i0.is_month_start, np_1darray[np.bool]), np_1darray[np.bool])
+    check(assert_type(i0.is_month_end, np_1darray[np.bool]), np_1darray[np.bool])
+    check(assert_type(i0.is_quarter_start, np_1darray[np.bool]), np_1darray[np.bool])
+    check(assert_type(i0.is_quarter_end, np_1darray[np.bool]), np_1darray[np.bool])
+    check(assert_type(i0.is_year_start, np_1darray[np.bool]), np_1darray[np.bool])
+    check(assert_type(i0.is_year_end, np_1darray[np.bool]), np_1darray[np.bool])
+    check(assert_type(i0.is_leap_year, np_1darray[np.bool]), np_1darray[np.bool])
```
Dr-Irv, line 635: put back the `np.bool_` in `check()`

Dr-Irv (changes requested):
Thanks for this PR.  Very interesting stuff.
I figured out your problem.  If you look at my changes in https://github.com/pandas-dev/pandas-stubs/pull/1328/commits/1db78e881a1de5c962164e6413ef7cf8cd4ca904 (and maybe  https://github.com/pandas-dev/pandas-stubs/pull/1328/commits/1db78e881a1de5c962164e6413ef7cf8cd4ca904, although I think the first commit I listed has everything), the changes I made will make it work for mypy with python 3.10, and allow you to use `ShapeT` in `Timestamp.__eq__()` and `Timestamp.__ne__()`

I was able to replicate the python 3.10 problem locally, making it easier to debug.

tests/test_scalars.py
```diff
@@ -48,7 +52,7 @@
     PeriodSeries: TypeAlias = pd.Series
     OffsetSeries: TypeAlias = pd.Series
 
-from tests import np_ndarray_bool
+MYPY = False
```
Dr-Irv: you'll need to remove this since I figured out how to make it work.

tests/test_indexes.py
hamdanal: @Dr-Irv I haven't looked at all your comments yet but I want to address this one. When you call `check(arr, np_1darray[np.bool])`, the following things are checked:
1. that `arr` if of type `np.ndarray`
2. that the shape of `arr` is 1D
3. that the dtype of `arr` is `np.bool`
4. that the elements of `arr` are **all** of type `np.bool`. This is guaranteed by numpy as this is what it means to have an array of dtype `np.bool`

The extra check you want is redundant in this case.

The only exception to point 4. above is with the `np.object_` dtype where the elements of the array are arbitrary Python objects. In the latter case, I keep the dtype argument passed to `check` to validate the exact object type we expect (see `test_datetimeindex_accessors` for example).

tests/test_indexes.py
hamdanal: @Dr-Irv This exact test is covered by `test_string_accessors_boolean_index` in `test_string_accessors.py`. I think this was a left over from when that file was created to test all string accessor related stuff.

tests/test_scalars.py
hamdanal: @Dr-Irv I don't know what `np.1darray[np.bool]` is as it doesn't exist!
Do you mean `np_1darray[np.bool]` (1D array of bool) or `np_ndarray_bool` (array of bool of unknown dimension)?

If you mean the former then it wouldn't be correct. When you compare a scalar timestamp to arbitrary shape array, you get arbitrary shape array of bool, not 1D array of bool.

If you mean the latter, then OK I can change it (but we already have `np.bool` passed to `check` so it is already accounted for in a sense)

tests/test_scalars.py
hamdanal: > And if you still have `np.ndarray` in the `check` function, I think it should be changed to `np.1darray[np.bool]`

Same answer as above. We shouldn't abruptly consider all results to be 1D arrays. If you compare the timestamp to an array with arbitrary shape, you get arbitrary shaped array of bool, not 1D array.

tests/__init__.py
hamdanal: https://docs.pytest.org/en/stable/example/simple.html#writing-well-integrated-assertion-helpers

This is a big QoL improvement as it allows pytest to show the line where the test fails directly instead of the internal traceback of the `check` function (which isn't useful when you get a failing test)

Compare before:
```
actual = False, klass = <class 'str'>, dtype = None, attr = 'left'
index_to_check_for_type = 0

    def check(
        actual: T,
        klass: type,
        dtype: type | None = None,
        attr: str = "left",
        index_to_check_for_type: Literal[0, -1] = 0,
    ) -> T:
        # __tracebackhide__ = True
        origin = get_origin(klass)
        if not isinstance(actual, origin or klass):
>           raise RuntimeError(f"Expected type '{klass}' but got '{type(actual)}'")
E           RuntimeError: Expected type '<class 'str'>' but got '<class 'bool'>'

tests/__init__.py:76: RuntimeError
```

to after (notice also that the function call that fail is underlined with `^^^^`):
```
>       gt = check(assert_type(ts > c_timestamp, bool), str)
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
E       RuntimeError: Expected type '<class 'str'>' but got '<class 'bool'>'
```

hamdanal:
@Dr-Irv this is ready for another round of review

tests/test_scalars.py
```diff
@@ -1931,128 +1825,144 @@ def test_period_add_subtract() -> None:
...
+    ne_rhs1 = check(assert_type(p_idx != p, np_1darray[np.bool]), np_1darray[np.bool])
+    assert (eq_rhs1 != ne_rhs1).all()
+
+
+def test_period_cmp_array() -> None:
+    p = pd.Period("2012-1-1", freq="D")
+    arr_nd = cast(  # cast to avoid pyright narrowing to a more precise shape
+        npt.NDArray[np.object_],
+        pd.period_range("2012-1-1", periods=4, freq="D").to_numpy(),
+    )
```
Dr-Irv: So the reason that is happening is because we know that `period_range()` returns a `PeriodIndex` and `PeriodIndex` returns a 1D numpy array.  

So I think you should add a test for the actual result of `to_numpy()`, i.e., you can test the non-precise shape version, and the version that is precise.

tests/test_scalars.py
```diff
@@ -1236,249 +1168,211 @@ def test_timestamp_add_sub() -> None:
...
+    ne_rhs_un1 = check(
+        assert_type(un_idx != ts, np_1darray[np.bool]), np_1darray[np.bool]
+    )
+    assert (eq_rhs_un1 != ne_rhs_un1).all()
+
+
+def test_timestamp_cmp_array() -> None:
+    ts = pd.Timestamp(year=2000, month=3, day=24, hour=12, minute=27)
+    arr_nd: npt.NDArray[np.datetime64] = np.array([1, 2, 3, 4], dtype="datetime64[ns]")
+    arr_2d = arr_nd.reshape(2, 2)
```
Dr-Irv: Should add test for `arr_1d = pd.DatetimeIndex(["2000-1-1", 2001-2-2"]).to_numpy()` since that is 1D

tests/test_scalars.py
```diff
@@ -824,248 +822,182 @@ def test_timedelta_mod_abs_unary() -> None:
...
+    # ==, != (td on the rhs, use == and != of lhs)
+    eq_rhs1 = check(assert_type(td_idx == td, np_1darray[np.bool]), np_1darray[np.bool])
+    ne_rhs1 = check(assert_type(td_idx != td, np_1darray[np.bool]), np_1darray[np.bool])
+    assert (eq_rhs1 != ne_rhs1).all()
+
+
+def test_timedelta_cmp_array() -> None:
+    td = pd.Timedelta("1 day")
+    arr_nd: npt.NDArray[np.timedelta64] = np.array([1, 2, 3, 4], dtype="timedelta64[D]")
+    arr_2d = arr_nd.reshape(2, 2)
```
Dr-Irv: Should add test for `arr_1d =  pd.to_timedelta([1, 2, 3], unit="D").to_numpy()`

tests/__init__.py
Dr-Irv: OK - thanks.  Very useful!

tests/test_indexes.py
Dr-Irv: OK - I now see why that is true.  The only reason I pointed it out is because I look at old code vs. new code.

tests/test_indexes.py
Dr-Irv: ok

tests/test_scalars.py
Dr-Irv: I mean `np_1darray[np.boll]`.  I suggested adding tests where we compare the result of `to_numpy()` which are 1D to make sure we get 1D results back.  That will cover this issue.

tests/test_scalars.py
Dr-Irv: Right, but if pandas returns a 1D array, then we need to test that as well.  Suggested elsewhere.

Dr-Irv (changes requested):
Thanks for the nice work on this and reorganizing the tests in `test_scalars.py`.

There is one set of tests to add which are testing that when we get `to_numpy()` on an `Index` or `Series`, we know that we have a 1D array, so we should test that comparing that 1D array preserves the 1D array of the boolean result.

tests/test_timefuncs.py
```diff
@@ -883,11 +887,223 @@ def test_timestampseries_offset() -> None:
...
+    check(
+        assert_type(o_s.to_numpy(dtype="bytes", copy=True), np_1darray),
+        np_1darray,
+        dtype=np.bytes_,
+    )
+    check(
+        assert_type(i_s.to_numpy(dtype="bytes", copy=True), np_1darray),
+        np_1darray,
+        dtype=np.bytes_,
+    )
```
Dr-Irv, line 982: might be better to use `"object"` here and below use `np.object` for both `OffsetSeries` and `IntervalSeries`

Dr-Irv (commented):
Very close.  One small thing in the tests that I think should be changed, but I could be convinced otherwise.

tests/test_timefuncs.py
hamdanal: `np.object_` is the default dtype for these two types. Using it here as you suggest doesn't really help test a different dtype like the test is meant to do

Dr-Irv: > `np.object_` is the default dtype for these two types. Using it here as you suggest doesn't really help test a different dtype like the test is meant to do

OK, that makes sense.

PR #1330, opened 2025-08-18
Add some missing types to BaseOffset and ExtensionArray

MarcoGorelli:


pandas-stubs/_libs/tslibs/offsets.pyi
```diff
@@ -67,7 +67,6 @@ class BaseOffset:
     def __rsub__(self, other: BaseOffset) -> Self: ...
     @overload
     def __rsub__(self, other: _TimedeltaT) -> _TimedeltaT: ...
-    def __call__(self, other): ...
```
MarcoGorelli, line 70: it's not callable, i don't think this should have been there to begin with

pandas-stubs/_libs/tslibs/offsets.pyi
```diff
@@ -87,8 +86,6 @@ class BaseOffset:
     def rollback(self, dt: datetime) -> datetime: ...
     def rollforward(self, dt: datetime) -> datetime: ...
     def is_on_offset(self, dt: datetime) -> bool: ...
-    def __setstate__(self, state) -> None: ...
-    def __getstate__(self): ...
```
MarcoGorelli, line 91: these are just used internally

pandas-stubs/core/arrays/base.pyi
```diff
@@ -23,16 +24,16 @@ class ExtensionArray:
     def __getitem__(self, item: ScalarIndexer) -> Any: ...
     @overload
     def __getitem__(self, item: SequenceIndexer) -> Self: ...
-    def __setitem__(self, key: int | slice | np.ndarray, value) -> None: ...
+    def __setitem__(self, key: int | slice | np.ndarray, value: Any) -> None: ...
     def __len__(self) -> int: ...
-    def __iter__(self): ...
+    def __iter__(self) -> Iterator[Any]: ...
```
MarcoGorelli, line 29: if `__getitem__` with `ScalarIndexer` returns `Any`, then so should these probably?

pandas-stubs/core/arrays/base.pyi
```diff
@@ -23,16 +24,16 @@ class ExtensionArray:
...
     def __contains__(self, item: object) -> bool | np.bool_: ...
     def to_numpy(
         self,
         dtype: npt.DTypeLike | None = ...,
         copy: bool = False,
         na_value: Scalar = ...,
-    ) -> np_1darray: ...
+    ) -> np_1darray[Any]: ...
```
MarcoGorelli, line 36: this is generic, but the inner type wasn't specified (leading pyright to report it as partially unknown)

pandas-stubs/_libs/tslibs/offsets.pyi
Dr-Irv: ok

pandas-stubs/_libs/tslibs/offsets.pyi
Dr-Irv: ok

pandas-stubs/core/arrays/base.pyi
Dr-Irv: sounds reasonable

pandas-stubs/core/arrays/base.pyi
Dr-Irv: It passed the tests, so I'm good with that.  

There might be a way to use the `dtype` argument to infer the type, but that's for someone else to figure out.

PR #1332, opened 2025-08-19
feat(series): arithmetic mul

cmp0xff:
This is inspired by #1274 and a follow-up for #1312.

### DIFF
--- a/tests/series/arithmetic/test_mul.py
+++ b/tests/series/arithmetic/test_mul.py
@@ -1,0 +1,127 @@
+import numpy as np
+from numpy import typing as npt  # noqa: F401
+import pandas as pd
+from typing_extensions import assert_type
+
+from tests import check
+
+left = pd.DataFrame({"a": [1, 2, 3]})["a"]  # left operand
+
+
+def test_mul_py_scalar() -> None:
+    """Test pd.Series[Any] * Python native scalars"""
+    b, i, f, c = True, 1, 1.0, 1j
+
+    check(assert_type(left * b, pd.Series), pd.Series)
+    check(assert_type(left * i, pd.Series), pd.Series)
+    check(assert_type(left * f, pd.Series), pd.Series)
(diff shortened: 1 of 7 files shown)


tests/series/arithmetic/test_mul.py
```diff
@@ -0,0 +1,127 @@
...
+    check(assert_type(left * b, pd.Series), pd.Series)
+    check(assert_type(left * i, pd.Series), pd.Series)
+    check(assert_type(left * f, pd.Series), pd.Series)
+    check(assert_type(left * c, pd.Series), pd.Series)
+
+    # `mypy` thinks the return types are `list[_T]`
+    check(assert_type(b * left, pd.Series), pd.Series)  # type: ignore[assert-type]
+    check(assert_type(i * left, pd.Series), pd.Series)  # type: ignore[assert-type]
+    check(assert_type(f * left, pd.Series), pd.Series)  # type: ignore[assert-type]
+    check(assert_type(c * left, pd.Series), pd.Series)  # type: ignore[assert-type]
```
Dr-Irv: Somehow `mypy` thinks that `Series` has the protocol `SupportsIndex`.  So here i a fix.  Add this line to the beginning of `class Series`:
```python
# Define __index__ because mypy thinks Series follows protocol `SupportsIndex`
__index__: ClassVar[None]
```

Then you can get rid of the ignores.

Dr-Irv (changes requested):
found a workarond for the mypy issue, otherwise looks good

tests/series/arithmetic/test_mul.py
cmp0xff: 93f0f0cb389f3b4d73a692f369c1ac796237df23

Dr-Irv: Looks like this is known:   https://github.com/python/mypy/issues/19341

PR #1333, opened 2025-08-20
GH1327 Add overloead for date_range and timedelta_range

### LINKED ISSUE #1327: Rewrite timedelta_range to use overloads in core/indexes/timedelta
https://github.com/pandas-dev/pandas/blob/e4a03b6e47a8ef9cd045902916289cbc976d3d33/pandas/core/indexes/timedeltas.py#L256-L259

followup from https://github.com/pandas-dev/pandas-stubs/pull/1293#discussion_r2233604589

loicdiridollou:
- [x] Closes #1327


tests/test_styler.py
```diff
@@ -141,7 +142,8 @@ def test_highlight_quantile() -> None:
 
 
 def test_loader() -> None:
-    check(assert_type(DF.style.loader, PackageLoader), PackageLoader)
+    if PD_LTE_23:
+        check(assert_type(DF.style.loader, PackageLoader), PackageLoader)
```
loicdiridollou, line 147: This is for the nightly, in 3.0.0 it will return a `FileSystemLoader` (see https://github.com/pandas-dev/pandas/pull/62123)

tests/test_timefuncs.py
```diff
@@ -1563,6 +1563,13 @@ def test_timedelta_range() -> None:
         pd.TimedeltaIndex,
     )
 
+    if TYPE_CHECKING_INVALID_USAGE:
+        day1 = pd.Timedelta(1, unit="D")
+        day10 = pd.Timedelta(10, unit="D")
+        pd.timedelta_range(
+            day1, day10, periods=10, freq="D"  # type: ignore[call-overload]  # pyright: ignore[reportArgumentType]
```
loicdiridollou: Passing all four should be banned

pandas-stubs/core/indexes/datetimes.pyi
```diff
@@ -102,11 +102,45 @@ class DatetimeIndex(
...
+    normalize: bool = False,
+    name: Hashable | None = None,
+    inclusive: IntervalClosedType = "both",
+    unit: TimeUnit | None = None,
+) -> DatetimeIndex: ...
+@overload
+def date_range(
+    start: str | DateAndDatetimeLike,
+    periods: int,
+    freq: str | timedelta | Timedelta | BaseOffset | None = None,
```
Dr-Irv: These would only work if the arguments are keyword arguments, so I think you need to put an asterisk (`*`) right after the opening parenthesis.

pandas-stubs/core/indexes/datetimes.pyi
```diff
@@ -102,11 +102,45 @@ class DatetimeIndex(
...
 def date_range(
-    start: str | DateAndDatetimeLike | None = None,
-    end: str | DateAndDatetimeLike | None = None,
-    periods: int | None = None,
-    freq: str | timedelta | Timedelta | BaseOffset = "D",
+    start: str | DateAndDatetimeLike,
+    end: str | DateAndDatetimeLike,
+    periods: int,
+    freq: None = None,
     tz: TimeZones = None,
```
Dr-Irv: I don't see the point here of having separate overloads.  Also, you need to put back the default values.

pandas-stubs/core/indexes/timedeltas.pyi
```diff
@@ -79,10 +79,51 @@ class TimedeltaIndex(
...
+    name: Hashable | None = None,
+    closed: Literal["left", "right"] | None = None,
+    *,
+    unit: None | str = ...,
+) -> TimedeltaIndex: ...
+@overload
+def timedelta_range(
+    start: TimedeltaConvertibleTypes,
+    periods: int,
+    freq: None = None,
```
Dr-Irv: I think you're going to have a similar issue in terms of mixing positional vs. keyword arguments.

tests/test_styler.py
Dr-Irv: Can you add a comment here?

Dr-Irv (changes requested):
You've added new overloads, but I don't see the benefit of them, without tests that don't work with the current stubs, but would work with your changes.

pandas-stubs/core/indexes/datetimes.pyi
loicdiridollou: So the original goal was to only allow three out of the four values among `start`, `end`, `periods` and `freq` according to the docs.
Although the `freq` argument is often not passed.

pandas-stubs/core/indexes/timedeltas.pyi
loicdiridollou: This is now fixed.

loicdiridollou:
I have added two tests for Type invalid because this is what we want to check (that passing four arguments is not allowed).

pandas-stubs/core/indexes/datetimes.pyi
```diff
@@ -102,11 +102,48 @@ class DatetimeIndex(
...
 
+@overload
 def date_range(
-    start: str | DateAndDatetimeLike | None = None,
-    end: str | DateAndDatetimeLike | None = None,
-    periods: int | None = None,
-    freq: str | timedelta | Timedelta | BaseOffset = "D",
+    *,
+    end: str | DateAndDatetimeLike = ...,
+    periods: int = ...,
```
Dr-Irv: I think this would allow `pd.date_range()` to be accepted.

The error message is " Of the four parameters: start, end, periods, and freq, exactly three must be specified".

So if you want this to work, you have to have overloads where 3 are required.  You can't have `= ...` on any of them (or default values either), because that means the argument is optional.

pandas-stubs/core/indexes/timedeltas.pyi
```diff
@@ -79,13 +79,53 @@ class TimedeltaIndex(
...
     def shift(self, periods: int = 1, freq=...) -> Self: ...
 
+@overload
 def timedelta_range(
-    start: TimedeltaConvertibleTypes | None = None,
-    end: TimedeltaConvertibleTypes | None = None,
-    periods: int | None = None,
+    start: TimedeltaConvertibleTypes = ...,
+    end: TimedeltaConvertibleTypes = ...,
+    periods: int = ...,
```
Dr-Irv: same issue here as with `date_range()`

tests/test_timefuncs.py
```diff
@@ -1563,6 +1563,13 @@ def test_timedelta_range() -> None:
         pd.TimedeltaIndex,
     )
 
+    if TYPE_CHECKING_INVALID_USAGE:
+        day1 = pd.Timedelta(1, unit="D")
+        day10 = pd.Timedelta(10, unit="D")
+        pd.timedelta_range(
+            day1, day10, 10, "D"  # type: ignore[call-overload]  # pyright: ignore[reportArgumentType]
+        )
```
Dr-Irv: should add checks where only 0, 1 or 2 of the 3 arguments are specified, as that should fail as well.

tests/test_timefuncs.py
```diff
@@ -1743,6 +1750,13 @@ def test_creating_date_range() -> None:
     dr = pd.date_range(start="2021-12-01", periods=24, freq="h")
     check(assert_type(dr.strftime("%H:%M:%S"), pd.Index), pd.Index, str)
 
+    if TYPE_CHECKING_INVALID_USAGE:
+        day1 = pd.Timestamp("2023-04-03")
+        day10 = pd.Timedelta("2023-04-08")
+        pd.date_range(
+            day1, day10, 10, "D"  # type: ignore[call-overload]  # pyright: ignore[reportCallIssue]
+        )
```
Dr-Irv: same comment - need to check that 0, 1 or 2 arguments specified causes a failure.

Dr-Irv (changes requested):

pandas-stubs/core/indexes/datetimes.pyi
loicdiridollou: The point where I am a bit unsure is that most of the usage I have seen it just about passing start and end, freq is often left alone so the usage does not reflect the docs.
Should we force the user to use three out of the four or allow only two?

Dr-Irv: So I think the docs and error message are a bit misleading.

You'll need to do a test to determine when 2 are allowed.  I think the rule is as follows:
- Whether or not `freq` is specified,  you need at least 2 out of the 3 of `start`, `end`, `periods`

loicdiridollou: If I look at the docs, they don't even use three exactly, when looking at the code they are using the fact that freq defaults to None but is then replaced by D for day.
https://pandas.pydata.org/docs/dev/reference/api/pandas.timedelta_range.html#pandas.timedelta_range

See https://github.com/pandas-dev/pandas/blob/78ec27668163808b3488da6ed2c64d0b8e9bfbc8/pandas/core/indexes/timedeltas.py#L330-L331 for the code where they change freq=None to freq="D"
Should we force the user to do 3 args even if the docs example only uses 2.
Happy to open a PR in the pandas repo to ask for clarification.

loicdiridollou: Sorry took a bit of time to find all the docs and you had answered before.
Let me setup a quick test to validate that we need two out of three of `start`, `end`, and `periods`.

loicdiridollou: Seems like the freq is indeed the argument that does not need to be passed (it is also the last one in the order).
I think the correct way to type hint is like you said forcing 2 out of the 3 with freq being an `str | None = ...` without the need to be passed when you pass two other. I have also tried passing four and that works as long as one of the three is None (yet it would be a strange use case of passing None since it is a default).
```
>>> pd.date_range("2023-04-05", "2023-04-07", freq="D")
DatetimeIndex(['2023-04-05', '2023-04-06', '2023-04-07'], dtype='datetime64[ns]', freq='D')
>>> pd.date_range("2023-04-05", "2023-04-07")
DatetimeIndex(['2023-04-05', '2023-04-06', '2023-04-07'], dtype='datetime64[ns]', freq='D')
```

loicdiridollou: https://github.com/pandas-dev/pandas/issues/62161

I have asked for clarifications.

loicdiridollou: I have read and reread the docs, it is a bit confusing at first glance but I will add all the tests for it.

loicdiridollou:
Rewrote the tests and the overloads properly with 2 (when omitting freq), 3 and forbidding 1 and 4 arguments among `start`, `end`, `periods` and `freq`.

Dr-Irv (approved):
thanks @loicdiridollou

PR #1335, opened 2025-08-20
Add test for DataFrame.from_records with list of tuples input

### LINKED ISSUE #1334: `DataFrame.from_records()` needs better annotations
**Describe the bug**
`DataFrame.from_records()` is missing argument annotations and also has `columns` and `exclude` incorrect.

**To Reproduce**
```python
df2 = pd.DataFrame.from_records([[1, 4], [2, 5], [3, 6]], columns=df.columns)
```
This fails with `pyright` because `columns` needs to be `ListLike | None` not `SequenceNotStr[str] | None`

 - OS: Windows 11
 - python version 3.11
 - version of type checker  1.1.403  (pyright)
 - version of installed `pandas-stubs`  -development version

Need to fix the stub and also create tests.

**Additional context**

iclectic:
-  Proposal to Close #1334
-  Tests added: Used assert_type() to assert the return type

This PR adds a test for  `DataFrame.from_records()` with tuple data to validate type annotations work correctly. The test ensures the method properly returns a DataFrame when called with a list of tuples and column names, addressing the type checking issues raised in the original issue.

### DIFF
--- a/pandas-stubs/core/frame.pyi
+++ b/pandas-stubs/core/frame.pyi
@@ -545,10 +545,15 @@ class DataFrame(NDFrame, OpsMixin, _GetItemHack):
     @classmethod
     def from_records(
         cls,
-        data,
-        index=...,
-        exclude: SequenceNotStr[str] | None = None,
+        data: (
+            np.ndarray
+            | Sequence[Any]
+            | Iterable[Mapping[str, Any]]
+            | Mapping[str, Sequence[Any]]
+        ),
+        index: str | Sequence[str] | None = None,
         columns: SequenceNotStr[str] | None = None,
+        exclude: SequenceNotStr[str] | None = None,
         coerce_float: bool = False,
         nrows: int | None = None,
     ) -> Self: ...
--- a/tests/test_frame.py
+++ b/tests/test_frame.py
@@ -4688,3 +4688,101 @@ def test_unstack() -> None:
         ),
         pd.DataFrame,
     )
+
+
+def test_from_records() -> None:
+    # testing with list of tuples
+    data_tuples = [(1, "a"), (2, "b"), (3, "c")]
+    check(
+        assert_type(
+            pd.DataFrame.from_records(data_tuples, columns=["id", "name"]),
+            pd.DataFrame,
+        ),
+        pd.DataFrame,
+    )
+
+    # Testing with numpy structured array
+    data_array = np.array(
+        [(1, "a"), (2, "b"), (3, "c")],
+        dtype=[("id", int), ("name", "U1")],
+    )
+    check(
+        assert_type(pd.DataFrame.from_records(data_array), pd.DataFrame),
+        pd.DataFrame,
+    )
+
+    # testing with list of dictionaries
+    data_dicts = [{"id": 1, "name": "a"}, {"id": 2, "name": "b"}]
+    check(
+        assert_type(pd.DataFrame.from_records(data_dicts), pd.DataFrame),
+        pd.DataFrame,
+    )
+
+    # Testing with mapping of sequences
+    data_mapping = {"id": [1, 2, 3], "name": ["a", "b", "c"]}
+    check(
+        assert_type(pd.DataFrame.from_records(data_mapping), pd.DataFrame),
+        pd.DataFrame,
+    )
+
+    # Testing with index parameter as string
+    check(
+        assert_type(
+            pd.DataFrame.from_records(
+                data_tuples, columns=["id", "name"], index="id"
+            ),
+            pd.DataFrame,
+        ),
+        pd.DataFrame,
+    )
+
+    #Testing with index parameter as sequence
+    check(
+        assert_type(
+            pd.DataFrame.from_records(
+                data_tuples, columns=["id", "name"], index=["id"]
+            ),
+            pd.DataFrame,
+        ),
+        pd.DataFrame,
+    )
+
+    # Testing  with exclude parameter
+    check(
+        assert_type(
+            pd.DataFrame.from_records(
+                [(1, "a", "extra"), (2, "b", "extra")],
+                columns=["id", "name", "extra"],
+                exclude=["extra"],
+            ),
+            pd.DataFrame,
+        ),
+        pd.DataFrame,
+    )
+
+    #Testing with all parameters
+    check(
+        assert_type(
+            pd.DataFrame.from_records(
+                data_tuples,
+                index=None,
+                columns=["id", "name"],
+                exclude=None,
+                coerce_float=True,
+                nrows=2,
+            ),
+            pd.DataFrame,
+        ),
+        pd.DataFrame,
+    )
+
+    # Testing parameter order
+    check(
+        assert_type(
+            pd.DataFrame.from_records(
+                data_tuples, columns=["id", "name"], exclude=None
+            ),
+            pd.DataFrame,
+        ),
+        pd.DataFrame,
+    )


pandas-stubs/core/frame.pyi
```diff
@@ -545,10 +545,15 @@ class DataFrame(NDFrame, OpsMixin, _GetItemHack):
     @classmethod
     def from_records(
         cls,
-        data,
-        index=...,
-        exclude: SequenceNotStr[str] | None = None,
+        data: (
+            np.ndarray
```
Dr-Irv: This should be `np_2d_array`

pandas-stubs/core/frame.pyi
```diff
@@ -545,10 +545,15 @@ class DataFrame(NDFrame, OpsMixin, _GetItemHack):
...
+            | Sequence[Any]
+            | Iterable[Mapping[str, Any]]
+            | Mapping[str, Sequence[Any]]
```
Dr-Irv: This should be `Sequence[Sequence[Scalar]] | Sequence[Mapping[str, Scalar]] | Mapping[str, Sequence[Scalar]]`

pandas-stubs/core/frame.pyi
```diff
@@ -545,10 +545,15 @@ class DataFrame(NDFrame, OpsMixin, _GetItemHack):
...
+        ),
+        index: str | Sequence[str] | None = None,
```
Dr-Irv: Should be `str | SequenceNotStr[str] | None = None`

pandas-stubs/core/frame.pyi
```diff
@@ -545,10 +545,15 @@ class DataFrame(NDFrame, OpsMixin, _GetItemHack):
...
         columns: SequenceNotStr[str] | None = None,
+        exclude: SequenceNotStr[str] | None = None,
```
Dr-Irv: Both of these should be `ListLike | None = None`

tests/test_frame.py
```diff
@@ -4688,3 +4688,101 @@ def test_unstack() -> None:
         ),
         pd.DataFrame,
     )
+
+
+def test_from_records() -> None:
```
Dr-Irv, line 4693: can you also add a test where `columns=pd.Index(["a", "b"])` (or something like that has an `Index` as the parameter.

pandas-stubs/core/frame.pyi
iclectic: Hello @Dr-Irv , in my recent commit, I used `np_2darray` because that is what is defined in `_typing.pyi` (line 834) and imported in `frame.pyi` (line 168). Using np_2d_array could cause a name error since that type does not exist. Does `np_2darray` work for what you were suggesting?

Dr-Irv: Yes, that was a typo

tests/test_frame.py
```diff
@@ -4688,3 +4688,115 @@ def test_unstack() -> None:
...
+    
+    #test with np.ndarray
+    arr = np.array([(1, "a"), (2, "b")], dtype=[("x", "i4"), ("y", "S1")])
```
Dr-Irv: CI failing here.  There are 2 options.  One is to accept any ND Array, but that is too wide.  In this test, if you do `arr = np.array([(1, "a"), (2, "b")], dtype=[("x", "i4"), ("y", "S1")]).reshape(2,2)` then numpy typing will infer this is a 2D array, which is what we want to allow.

pandas-stubs/core/frame.pyi
```diff
@@ -545,10 +545,15 @@ class DataFrame(NDFrame, OpsMixin, _GetItemHack):
...
         cls,
-        data,
-        index=...,
-        exclude: SequenceNotStr[str] | None = None,
-        columns: SequenceNotStr[str] | None = None,
+        data: (
+            np_2darray
+            | Sequence[SequenceNotStr]
+            | Sequence[Mapping[str, Scalar]]
+            | Mapping[str, Sequence[Scalar]]
```
Dr-Irv: CI failed, so I think the fix here is to change these 2 lines to
```suggestion
            | Sequence[Mapping[str, Any]]
            | Mapping[str, SequenceNotStr[Any]]
```

tests/test_frame.py
```diff
@@ -4688,3 +4688,112 @@ def test_unstack() -> None:
         ),
         pd.DataFrame,
     )
+
+
+def test_from_records() -> None:
+    
+    #test with np.ndarray
+    arr = np.array([[1, "a"], [2, "b"]], dtype=object)
```
Dr-Irv:
```suggestion
    arr = np.array([[1, "a"], [2, "b"]], dtype=object).reshape(2, 2)
```
CI failing here - has to do with differences of python 3.10 and later versions in terms of numpy support.

tests/test_frame.py
```diff
@@ -4688,3 +4688,112 @@ def test_unstack() -> None:
...
+        assert_type(pd.DataFrame.from_records(data_dict_tuples, columns=["id", "name"]), pd.DataFrame),
+        pd.DataFrame,
+    )
+
+    # Testing with mapping of sequences (use DataFrame constructor instead)
+    data_mapping_dict = {"id": [1, 2], "name": ["a", "b"]}
+    check(
+        assert_type(pd.DataFrame(data_mapping_dict), pd.DataFrame),
+        pd.DataFrame,
+    )
```
Dr-Irv: This should work with `DataFrame.from_records()`.  Please use that instead.

Dr-Irv (changes requested):
Make sure you are setting up the pre-commit, because your code needs to be formatted with `black`.

tests/test_frame.py
```diff
@@ -4688,3 +4688,116 @@ def test_unstack() -> None:
...
+            pd.DataFrame,
+        ),
+        pd.DataFrame,
+    )
+
+    # testing with list of dictionaries (convert to tuples for type compatibility)
+    data_dict_tuples = [(1, "a"), (2, "b")]
+    check(
+        assert_type(
+            pd.DataFrame.from_records(data_dict_tuples, columns=["id", "name"]),
```
Dr-Irv: You shouldn't convert the type here.  You need to test a list of dicts, and update the stub so that is accepted.

tests/test_frame.py
```diff
@@ -4688,3 +4688,116 @@ def test_unstack() -> None:
...
+            pd.DataFrame,
+        ),
+        pd.DataFrame,
+    )
+
+    # testing with mapping of sequences
+    data_mapping_dict = {"id": [1, 2], "name": ["a", "b"]}
+    check(
+        assert_type(pd.DataFrame(data_mapping_dict), pd.DataFrame),
```
Dr-Irv: Needs to be `DataFrame.from_records()`

Dr-Irv (changes requested):
Now the CI is passing, so that's good.

Still 2 issues in terms of testing `DataFrame.from_records()` with a list of dicts, and just a dict.

pandas-stubs/core/frame.pyi
```diff
@@ -545,10 +545,16 @@ class DataFrame(NDFrame, OpsMixin, _GetItemHack):
...
-        exclude: SequenceNotStr[str] | None = None,
-        columns: SequenceNotStr[str] | None = None,
+        data: (
+            np_2darray
+            | Sequence[SequenceNotStr]
+            | Sequence[Mapping[str, Any]]
+            | Mapping[str, Any]
+            | Mapping[str, SequenceNotStr[Any]]
+        ),
+        index: str | SequenceNotStr[str] | None = None,
```
Dr-Irv:
```suggestion
        index: str | SequenceNotStr[str] | None = None,
```
```suggestion
        index: str | SequenceNotStr[Hashable] | None = None,
```
This is why CI is failing

tests/test_frame.py
```diff
@@ -4688,3 +4688,125 @@ def test_unstack() -> None:
...
+        assert_type(
+            pd.DataFrame.from_records(data_single_dict, index=["0"]), pd.DataFrame
+        ),
+        pd.DataFrame,
+    )
+
+    # testing with mapping of sequences
+    data_mapping_dict = {"id": [1, 2], "name": ["a", "b"]}
+    check(
+        assert_type(pd.DataFrame(data_mapping_dict), pd.DataFrame),
```
Dr-Irv: this test should be `DataFrame.from_records()`

Dr-Irv (changes requested):
I resolved an issue by mistake.  One thing to fix in the tests.

Dr-Irv (approved):
thanks @iclectic

PR #1342, opened 2025-08-22
remove ``SingleConstructorOffset.__reduce__``

MarcoGorelli:
`__reduce__` isn't documented / part of the public API, but `SingleConstructorOffset` is used to group together some kinds of offsets, and type checkers can detect some erroneous usage, e.g.

https://github.com/pandas-dev/pandas-stubs/blob/f5fd35fdfb1812b346d5ec1aeca82a3a1fc56b61/tests/test_dtypes.py#L62-L68


Dr-Irv (approved):
thanks @MarcoGorelli

PR #1343, opened 2025-08-22
fix(series): arithmetics for Series[Any]

cmp0xff:
This PR implements the ideas from https://github.com/pandas-dev/pandas-stubs/pull/1274#discussion_r2291159392 and https://github.com/pandas-dev/pandas-stubs/pull/1274#discussion_r2291175165.

### DIFF
--- a/tests/series/arithmetic/str/test_add.py
+++ b/tests/series/arithmetic/str/test_add.py
@@ -18,7 +18,7 @@ left = pd.Series(["1", "23", "456"])  # left operand
 
 
 def test_add_py_scalar() -> None:
-    """Testpd.Series[str]+ Python native 'scalar's"""
+    """Test pd.Series[str] + Python native 'scalar's"""
     i = 4
     r0 = "right"
 
@@ -40,7 +40,7 @@ def test_add_py_scalar() -> None:
 
 
 def test_add_py_sequence() -> None:
-    """Testpd.Series[str]+ Python native sequence"""
+    """Test pd.Series[str] + Python native sequence"""
     i = [3, 5, 8]
     r0 = ["a", "bc", "def"]
     r1 = tuple(r0)
@@ -67,7 +67,7 @@ def test_add_py_sequence() -> None:
 
 
 def test_add_numpy_array() -> None:
-    """Testpd.Series[str]+ numpy array"""
+    """Test pd.Series[str] + numpy array"""
     i = np.array([3, 5, 8], np.int64)
     r0 = np.array(["a", "bc", "def"], np.str_)
 
@@ -101,7 +101,7 @@ def test_add_numpy_array() -> None:
 
 
 def test_add_pd_series() -> None:
-    """Testpd.Series[str]+ pandas series"""
+    """Test pd.Series[str] + pandas series"""
     i = pd.Series([3, 5, 8])
     r0 = pd.Series(["a", "bc", "def"])
 
--- a/tests/series/arithmetic/test_add.py
+++ b/tests/series/arithmetic/test_add.py
@@ -5,122 +5,127 @@ from typing_extensions import assert_type
 
 from tests import check
 
-left = pd.DataFrame({"a": [1, 2, 3]})["a"]  # left operand
+left_i = pd.DataFrame({"a": [1, 2, 3]})["a"]  # left operand
 
 
-def test_add_py_scalar() -> None:
-    """Test pd.Series[Any] + Python native scalars"""
+def test_add_i_py_scalar() -> None:
+    """Test pd.Series[Any] (int) + Python native scalars"""
     b, i, f, c = True, 1, 1.0, 1j
 
-    check(assert_type(left + b, pd.Series), pd.Series)
-    check(assert_type(left + i, pd.Series), pd.Series)
-    check(assert_type(left + f, pd.Series), pd.Series)
-    check(assert_type(left + c, pd.Series), pd.Series)
+    check(assert_type(left_i + b, pd.Series), pd.Series)
+    check(assert_type(left_i + i, pd.Series), pd.Series)
+    check(assert_type(left_i + f, pd.Series), pd.Series)
+    check(assert_type(left_i + c, pd.Series), pd.Series)
 
-    check(assert_type(b + left, pd.Series), pd.Series)
-    check(assert_type(i + left, pd.Series), pd.Series)
-    check(assert_type(f + left, pd.Series), pd.Series)
-    check(assert_type(c + left, pd.Series), pd.Series)
+    check(assert_type(b + left_i, pd.Series), pd.Series)
+    check(assert_type(i + left_i, pd.Series), pd.Series)
+    check(assert_type(f + left_i, pd.Series), pd.Series)
+    check(assert_type(c + left_i, pd.Series), pd.Series)
 
-    check(assert_type(left.add(b), pd.Series), pd.Series)
-    check(assert_type(left.add(i), pd.Series), pd.Series)
-    check(assert_type(left.add(f), pd.Series), pd.Series)
-    check(assert_type(left.add(c), pd.Series), pd.Series)
+    check(assert_type(left_i.add(b), pd.Series), pd.Series)
+    check(assert_type(left_i.add(i), pd.Series), pd.Series)
+    check(assert_type(left_i.add(f), pd.Series), pd.Series)
+    check(assert_type(left_i.add(c), pd.Series), pd.Series)
 
-    check(assert_type(left.radd(b), pd.Series), pd.Series)
-    check(assert_type(left.radd(i), pd.Series), pd.Series)
-    check(assert_type(left.radd(f), pd.Series), pd.Series)
-    check(assert_type(left.radd(c), pd.Series), pd.Series)
+    check(assert_type(left_i.radd(b), pd.Series), pd.Series)
+    check(assert_type(left_i.radd(i), pd.Series), pd.Series)
+    check(assert_type(left_i.radd(f), pd.Series), pd.Series)
+    check(assert_type(left_i.radd(c), pd.Series), pd.Series)
 
 
-def test_add_py_sequence() -> None:
-    """Test pd.Series[Any] + Python native sequence"""
+def test_add_i_py_sequence() -> None:
+    """Test pd.Series[Any] (int) + Python native sequence"""
     b, i, f, c = [True, False, True], [2, 3, 5], [1.0, 2.0, 3.0], [1j, 1j, 4j]
 
-    check(assert_type(left + b, pd.Series), pd.Series)
-    check(assert_type(left + i, pd.Series), pd.Series)
-    check(assert_type(left + f, pd.Series), pd.Series)
-    check(assert_type(left + c, pd.Series), pd.Series)
+    check(assert_type(left_i + b, pd.Series), pd.Series)
+    check(assert_type(left_i + i, pd.Series), pd.Series)
+    check(assert_type(left_i + f, pd.Series), pd.Series)
+    check(assert_type(left_i + c, pd.Series), pd.Series)
 
-    check(assert_type(b + left, pd.Series), pd.Series)
-    check(assert_type(i + left, pd.Series), pd.Series)
-    check(assert_type(f + left, pd.Series), pd.Series)
-    check(assert_type(c + left, pd.Series), pd.Series)
+    check(assert_type(b + left_i, pd.Series), pd.Series)
+    check(assert_type(i + left_i, pd.Series), pd.Series)
+    check(assert_type(f + left_i, pd.Series), pd.Series)
+    check(assert_type(c + left_i, pd.Series), pd.Series)
 
-    check(assert_type(left.add(b), pd.Series), pd.Series)
-    check(assert_type(left.add(i), pd.Series), pd.Series)
-    check(assert_type(left.add(f), pd.Series), pd.Series)
-    check(assert_type(left.add(c), pd.Series), pd.Series)
+    check(assert_type(left_i.add(b), pd.Series), pd.Series)
+    check(assert_type(left_i.add(i), pd.Series), pd.Series)
+    check(assert_type(left_i.add(f), pd.Series), pd.Series)
+    check(assert_type(left_i.add(c), pd.Series), pd.Series)
 
-    check(assert_type(left.radd(b), pd.Series), pd.Series)
-    check(assert_type(left.radd(i), pd.Series), pd.Series)
-    check(assert_type(left.radd(f), pd.Series), pd.Series)
-    check(assert_type(left.radd(c), pd.Series), pd.Series)
+    check(assert_type(left_i.radd(b), pd.Series), pd.Series)
+    check(assert_type(left_i.radd(i), pd.Series), pd.Series)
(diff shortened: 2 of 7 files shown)


tests/series/arithmetic/test_sub.py
```diff
@@ -108,85 +116,112 @@ def test_sub_numpy_array() -> None:
...
+    """Test pd.Series[Any] (Timestamp | Timedelta) - numpy datetime(s)"""
     s = np.datetime64(anchor)
     a = np.array([s + np.timedelta64(m, "m") for m in range(3)], np.datetime64)
 
     check(assert_type(left_ts - s, "TimedeltaSeries"), pd.Series, pd.Timedelta)
+    if TYPE_CHECKING_INVALID_USAGE:
+        _0 = left_td - s
     check(assert_type(left_ts - a, "TimedeltaSeries"), pd.Series, pd.Timedelta)
+    if TYPE_CHECKING_INVALID_USAGE:
+        _1 = left_td - a
```
Dr-Irv: When you have `TYPE_CHECKING_INVALID_USAGE`, that means we should have `# type: ignore` and `# pyright: ignore` statements that demonstrate the type checker can catch those errors.

tests/series/arithmetic/test_sub.py
```diff
@@ -108,85 +116,112 @@ def test_sub_numpy_array() -> None:
...
+    """Test pd.Series[Any] (Timestamp | Timedelta) - Pandas datetime(s)"""
     s = pd.Timestamp(anchor)
     a = pd.Series([s + pd.Timedelta(minutes=m) for m in range(3)])
 
     check(assert_type(left_ts - s, "TimedeltaSeries"), pd.Series, pd.Timedelta)
+    if TYPE_CHECKING_INVALID_USAGE:
+        _0 = left_td - s
     check(assert_type(left_ts - a, "TimedeltaSeries"), pd.Series, pd.Timedelta)
+    if TYPE_CHECKING_INVALID_USAGE:
+        _1 = left_ts - a
```
Dr-Irv: This is an example of valid code that should be accepted.

cmp0xff: Sorry, it was a typo. 1fb597b0c273f06882eb97e36e66dc41486801cb

However, I did not easily understand your comment here. 
> This is an example of valid code that should be accepted.

`left_ts` and `left_td` are `Series[Any]` at type checking. But you wrote
> I think we have to be selective here, and probably disallow subtraction with untyped Series when the other argument is known to be time related (Timestamp, Timedelta and associated Series)

It seems to me that in your plan, both `left_td - a` and `left_ts - a` should give an error / a `Never` at type checking. Am I right?

My proposed plan is more permissive and will not detect the problem of `left_ts - a`, because at runtime, `Series[Any] - TimedeltaSeries` can either be `TimestampSeries` or `TimedeltaSeries` or give an error. In my proposed plan, at type checking, it would give `Series[Any]`.

Dr-Irv:
> In my proposal, I would rather allow `Series[Any] - TimestampSeries` as well as `Series[Any] - TimedeltaSeries`, leaving the user to check the potential error at runtime. We can either give `Series[Any]` in both cases, or, using my plan, give `TimedeltaSeries` in the first case.

Or we can catch it in type checking??
> 
> We have already done the same to allow `Series[Any] - Series[int]`, giving `Series[Any]`, which could also fail at runtime if `Series[Any]` actually contains `str`.

Yes, this latter point is quite valid.  Here's a possible modification of my proposal: (with respect to `Series`) for binary operators `a X b`, where `X` is the operator:

1. If `a` and `b` are fully typed, we figure out the result, and if it is an invalid calculation, we catch it.
2. If a is `Series[Any]` and `b` is `Timestamp`, `Series[Timestamp]`, `Timedelta`, `Series[Timedelta]`, `str`, `Series[str]`, we catch that as a typing error.  
3. If `a` is `Timestamp`, `Series[Timestamp]`, `Timedelta`, `Series[Timedelta]`, `str`, `Series[str]`, and `b` is `Series[Any]`, we say that is an error.
4. If `a` and `b` are not fully typed (i.e., one is `Series[Any]` and the other is `Any` or `Series[Any]`, we accept the calculation in typing and don't report an error.

Is this possible?

cmp0xff:
I tried to implement the plan in e0b5b5903e9c694059d8a288cdcaa45b8b8379f1. The `str` part proves to be a bit tricky. I would like to leave out for now.

tests/series/arithmetic/test_sub.py
```diff
@@ -108,85 +113,132 @@ def test_sub_numpy_array() -> None:
...
+def test_sub_ts_numpy_datetime() -> None:
+    """Test pd.Series[Any] (Timestamp | Timedelta) - numpy datetime(s)"""
     s = np.datetime64(anchor)
     a = np.array([s + np.timedelta64(m, "m") for m in range(3)], np.datetime64)
 
-    check(assert_type(left_ts - s, "TimedeltaSeries"), pd.Series, pd.Timedelta)
-    check(assert_type(left_ts - a, "TimedeltaSeries"), pd.Series, pd.Timedelta)
+    if TYPE_CHECKING_INVALID_USAGE:
+        # `numpy` typing gives the corresponding `ndarray`s in the static type
+        # checking, where our `__rsub__` cannot override.
```
Dr-Irv: Given that we are catching this with the typing stubs, you can remove this comment.

tests/series/arithmetic/test_sub.py
```diff
@@ -108,85 +113,132 @@ def test_sub_numpy_array() -> None:
...
-    """Test pd.Series[Any] - Pandas datetime(s)"""
+def test_sub_ts_pd_datetime_1() -> None:
+    """Test pd.Series[Any] (Timestamp | Timedelta) - Pandas datetime(s)"""
     s = pd.Timestamp(anchor)
     a = pd.Series([s + pd.Timedelta(minutes=m) for m in range(3)])
 
-    check(assert_type(left_ts - s, "TimedeltaSeries"), pd.Series, pd.Timedelta)
-    check(assert_type(left_ts - a, "TimedeltaSeries"), pd.Series, pd.Timedelta)
+    if TYPE_CHECKING_INVALID_USAGE:
+        # When I merge this one to the previous one, mypy does not allow me to pass
```
Dr-Irv: I think when `mypy` sees the result of `Never` on a method (as opposed to an operator), it assumes the code following that won't be executed.

tests/series/arithmetic/test_sub.py
cmp0xff: Removed in 8d8691ed7a9bc3539736f15492dee55fd36390cc

tests/series/arithmetic/test_sub.py
cmp0xff: Caught with `pytest.raises` in 8d8691ed7a9bc3539736f15492dee55fd36390cc

cmp0xff (commented):
9c972d3d61b70687c8ca9f6b88c2cdfd5216e678 cherry picked as 5c23f687b091b04e57725d858bb8ac3704b18bdc

tests/series/arithmetic/str/test_add.py
```diff
@@ -72,7 +73,8 @@ def test_add_numpy_array() -> None:
     r0 = np.array(["a", "bc", "def"], np.str_)
 
     if TYPE_CHECKING_INVALID_USAGE:
-        assert_type(left + i, Never)
+        with pytest.raises(AssertionError):
+            assert_never(left + i)
```
Dr-Irv: `pytest` will never execute this part of the code because of the `if`.  I think this should remain as `assert_type(left + i, Never)`

tests/series/arithmetic/test_add.py
```diff
@@ -131,3 +138,18 @@ def test_add_i_pd_series() -> None:
...
+
+    if TYPE_CHECKING_INVALID_USAGE:
+        with pytest.raises(AssertionError):
+            assert_never(left_i + s)
+        with pytest.raises(AssertionError):
+            assert_never(s + left_i)
+        with pytest.raises(AssertionError):
+            assert_never(left_i.add(s))
+        with pytest.raises(AssertionError):
+            assert_never(left_i.radd(s))
```
Dr-Irv: sae issue - `pytest` will never run this code anyway.  So can't we just use `assert(left_i + s, Never)` ?

tests/series/arithmetic/test_sub.py
```diff
@@ -210,35 +210,39 @@ def test_sub_ts_pd_datetime() -> None:
...
         assert_type(a - left_td, Never)
 
         left_ts.sub(s)  # type: ignore[arg-type] # pyright: ignore[reportArgumentType,reportCallIssue]
-        assert_type(left_ts.sub(a), Never)
+        with pytest.raises(AssertionError):
+            assert_never(left_ts.sub(a))
 
+        left_td.sub(s)  # type: ignore[arg-type] # pyright: ignore[reportArgumentType,reportCallIssue]
+        with pytest.raises(AssertionError):
+            assert_never(left_td.sub(a))
```
Dr-Irv: same issue - remove the `pytest` part

tests/test_frame.py
```diff
@@ -4440,7 +4441,7 @@ def test_frame_setitem_na() -> None:
     df.loc[ind, :] = pd.NA
     df.iloc[[0, 2], :] = pd.NA
 
-    df["x"] = df["y"] + pd.Timedelta(days=3)
+    df["x"] = cast("TimestampSeries", df["y"]) + pd.Timedelta(days=3)
```
Dr-Irv: add a comment here indicating why the `cast` is necessary - it's because the type of `df["y"]` is unknown, so we have to cast to tell the type checker what kind of type it is when adding a `Timedelta`

Dr-Irv (changes requested):
Pretty close.  Just issue in using `pytest.raises`.  

I also think it would be worth adding a subsection to https://github.com/pandas-dev/pandas-stubs/blob/main/docs/philosophy.md#use-of-generic-types that explains the philosophy we've agreed to.  Namely, since `df["a"]` has unknown type, we restrict arithmetic when the the other operand is `Timedelta`, `Timestamp`, `str`, etc., and their `Series` variants.  Then we tell people they should use `cast` to tell the type checker "I know that this column of my `DataFrame` has the right type"

Dr-Irv:
There's been another issue with the overloads that has caused the dev version of mypy to fail on the current main.  I discovered the problem, and if you look at this set of diffs, it fixes the problem, working with both mypy 1.17.1 and the dev version.

Can you incorporate these changes into your PR?

https://github.com/pandas-dev/pandas-stubs/compare/9c972d3d61b70687c8ca9f6b88c2cdfd5216e678...Dr-Irv:pandas-stubs:d3b8f164b2dc469146d43d49fd73eabc7ebe643a

tests/series/arithmetic/str/test_add.py
cmp0xff: Reverted in b4235d087770e214835d915ada7742d75bc1f3fa

tests/series/arithmetic/test_add.py
cmp0xff: `+` reverted in b4235d087770e214835d915ada7742d75bc1f3fa.

`.add` and `.radd` not reverted, since they are functions giving `Never`, rendering `mypy` skipping everything after the first appearance.

`assert_never`, in contrast, converts `Never` to `AssertionError`. When I use `pyright.raises` to catch these Exceptions, `mypy` can continue.

Alternatively I can write individual functions for them, each of which ends with a function call giving `Never`. But I would say the current approach looks better.

tests/series/arithmetic/test_sub.py
cmp0xff: Not removed with arguments

tests/test_frame.py
cmp0xff: Added in 6a4692f83bca93d5d74a0bc30c1ec5fecfdd8295

tests/series/arithmetic/test_add.py
```diff
@@ -1,126 +1,154 @@
...
+    """Test pd.Series[Any] (int) + Python str"""
+    s = "abc"
+
+    if TYPE_CHECKING_INVALID_USAGE:
+        assert_type(left_i + s, Never)
+        assert_type(s + left_i, Never)
+        with pytest.raises(AssertionError):
+            assert_never(left_i.add(s))
+        with pytest.raises(AssertionError):
+            assert_never(left_i.radd(s))
```
Dr-Irv: I'm still concerned that this is not testing what we want to test.  You want to test that `left_i.add(s)` has the result `Never`.  But `assert_never()` means that the code is not reachable.  

If we have to split this into multiple test functions for mypy reasons, let's do that and document it with a comment.

tests/series/arithmetic/test_sub.py
```diff
@@ -108,85 +115,134 @@ def test_sub_numpy_array() -> None:
...
 
-    check(assert_type(s - left_ts, "TimedeltaSeries"), pd.Series, pd.Timedelta)
-    # `numpy` typing gives the corresponding `ndarray`s in the static type
-    # checking, where our `__rsub__` cannot override. At runtime, they return
-    # `Series`s.
-    check(assert_type(a - left_ts, "npt.NDArray[np.datetime64]"), pd.Series, pd.Timedelta)  # type: ignore[assert-type]
+        _4 = s - left_ts  # type: ignore[operator] # pyright: ignore[reportOperatorIssue]
+        # _5 = a - left_ts
+        _6 = s - left_td  # type: ignore[operator] # pyright: ignore[reportOperatorIssue]
+        # _7 = a - left_td
```
Dr-Irv, line 191: For the ones that are commented out here, can you add a comment that says we'd like these to be detected as invalid, but `numpy.ndarray.__rsub__()` will be invoked.

tests/series/arithmetic/test_add.py
Dr-Irv: > Alternatively I can write individual functions for them, each of which ends with a function call giving `Never`. But I would say the current approach looks better.

Still think it may not be testing what we want to test.  If we have to use separate functions, so be it.

cmp0xff: 3e9c06e8d8419c67b287e6aecc0ccb9d09981727

tests/series/arithmetic/test_add.py
cmp0xff: 3e9c06e8d8419c67b287e6aecc0ccb9d09981727

tests/series/arithmetic/test_sub.py
cmp0xff: 3e9c06e8d8419c67b287e6aecc0ccb9d09981727

Dr-Irv (approved):
thanks @cmp0xff .  Very nice contribution. 

Question for you - I could do a release with these latest changes, because we have all the arithmetic working, and only have the work to get rid of `TimestampSeries` and `TimedeltaSeries` left to do.  

So do  you think I should release this now, or wait?

cmp0xff:
Hi @Dr-Irv , thank you for asking me, I am honoured.

If I were in a the position to release a new version, I would do so now and save dropping `TimestampSeries` and `TimedeltaSeries` for the next release. The changes to arithmetic operations have already been significant, and it would be good to get some feedback.

PR #1344, opened 2025-08-23
add `scipy-stubs` to the `dev` dependency group

jorenham:
I figured that since scipy is a dependency, and we like static typing around here, it would probably help to add some stubs for it too :). See https://github.com/scipy/scipy-stubs for details.


Dr-Irv (approved):
thanks @jorenham

Issue #1347, opened 2025-08-26
Index arithmetic overloads show wrong types


MarcoGorelli:
**Describe the bug**

I think this is an issue:

https://github.com/pandas-dev/pandas-stubs/blob/fdda5a60b54a9b74dd1c5bf496c6eae5818512ba/pandas-stubs/core/arraylike.pyi#L24

as the dtype isn't preserved

**To Reproduce**
1. Provide a minimal runnable `pandas` example that is not properly checked by the stubs.
2. Indicate which type checker you are using (`mypy` or  `pyright`).
3. Show the error message received from that type checker while checking your example.

```python
from __future__ import annotations
import pandas as pd
from typing import reveal_type

idx = pd.Index([1,2,3])

reveal_type(idx - 1.5)
```
at runtime this gives
```
Index([-0.5, 0.5, 1.5], dtype='float64')
```
but it's revealed to be `Index[int]` by type checkers

**Please complete the following information:**
 - OS: [e.g. Windows, Linux, MacOS]
 - OS Version [e.g. 22]
 - python version
 - version of type checker
 - version of installed `pandas-stubs` 3033eea474b754f7deabfa25a3377ed1efb85c15

**Additional context**

MarcoGorelli:
Once this is addressed, it should be possible to uncomment some of the failing test cases in https://github.com/pandas-dev/pandas-stubs/pull/1348#pullrequestreview-3160418303

MarcoGorelli:
would that not require another pandas release?

Dr-Irv:
> would that not require another pandas release?

I'm guessing it's fixed in 3.0, so we could modify the tests that currently fail with 2.3.3 to not test with 2.3.3, but do test with 3.0+

PR #1348, opened 2025-08-26
Type `NAType`

MarcoGorelli:
This has uncovered a couple of bugs:
- https://github.com/pandas-dev/pandas/issues/62196
- https://github.com/pandas-dev/pandas-stubs/issues/1347

so some tests have their assertions ignored for now, but i've linked to them in those places


pandas-stubs/_libs/missing.pyi
```diff
@@ -1,46 +1,268 @@
...
+    @overload
+    def __rmul__(  # type: ignore[overload-overlap] #  pyright: ignore[reportOverlappingOverload]
+        self, other: Series, /
+    ) -> Series: ...
+    @overload
+    def __rmul__(self, other: Index, /) -> Index: ...  # type: ignore[overload-overlap]
+    @overload
+    def __rmul__(self, other: object, /) -> NAType: ...
+    def __matmul__(self, other: object, /) -> NAType: ...
+    def __rmatmul__(self, other: object, /) -> NAType: ...
```
MarcoGorelli, line 69: not sure I get this method, trying to use it with a Series just raises
```python
In [6]: pd.NA @ pd.Series([1,2,3], dtype='Int64')
---------------------------------------------------------------------------
IndexError                                Traceback (most recent call last)
Cell In[6], line 1
----> 1 pd.NA @ pd.Series([1,2,3], dtype='Int64')

File ~/pandas-dev/pandas/core/series.py:2976, in Series.__rmatmul__(self, other)
   2972 def __rmatmul__(self, other):
   2973     """
   2974     Matrix multiplication using binary `@` operator.
   2975     """
-> 2976     return self.dot(np.transpose(other))

File ~/pandas-dev/pandas/core/series.py:2949, in Series.dot(self, other)
   2947     lvals = self.values
   2948     rvals = np.asarray(other)
-> 2949     if lvals.shape[0] != rvals.shape[0]:
   2950         raise Exception(
   2951             f"Dot product shape mismatch, {lvals.shape} vs {rvals.shape}"
   2952         )
   2954 if isinstance(other, ABCDataFrame):

IndexError: tuple index out of range

In [7]: pd.NA @ pd.Series([1,2,3], dtype='Int64').to_frame()
---------------------------------------------------------------------------
IndexError                                Traceback (most recent call last)
<ipython-input-7-bced8200f82f> in ?()
----> 1 pd.NA @ pd.Series([1,2,3], dtype='Int64').to_frame()

~/pandas-dev/pandas/core/frame.py in ?(self, other)
   1754         Matrix multiplication using binary `@` operator.
   1755         """
   1756         try:
   1757             return self.T.dot(np.transpose(other)).T
-> 1758         except ValueError as err:
   1759             if "shape mismatch" not in str(err):
   1760                 raise
   1761             # GH#21581 give exception message for original shapes

~/pandas-dev/pandas/core/frame.py in ?(self, other)
   1708         else:
   1709             left = self
   1710             lvals = self.values
   1711             rvals = np.asarray(other)
-> 1712             if lvals.shape[1] != rvals.shape[0]:
   1713                 raise ValueError(
   1714                     f"Dot product shape mismatch, {lvals.shape} vs {rvals.shape}"
   1715                 )

IndexError: tuple index out of range
```

pandas-stubs/_libs/missing.pyi
```diff
@@ -1,46 +1,268 @@
...
-    def __lt__(self, other) -> bool: ...
-    def __gt__(self, other) -> bool: ...
-    def __ge__(self, other) -> bool: ...
-    def __neg__(self, other) -> NAType: ...
-    def __pos__(self, other) -> NAType: ...
-    def __abs__(self, other) -> NAType: ...
-    def __invert__(self, other) -> NAType: ...
-    def __pow__(self, other) -> NAType: ...
-    def __rpow__(self, other) -> NAType: ...
-    def __and__(self, other) -> NAType | None: ...
```
MarcoGorelli, line 37: not sure where the `| None` came from, as far as I can tell this can be either `NAType` or `False` if `other` is a scalar? And if it's a `Series` or `Index`, the output is `Index`

tests/test_natype.py
```diff
@@ -0,0 +1,197 @@
...
+from typing_extensions import assert_type
+
+from tests import check
+
+
+def test_arithmetic() -> None:
+    na = pd.NA
+
+    s_int: pd.Series[int] = pd.Series([1, 2, 3], dtype="Int64")
+    idx_int: pd.Index[int] = pd.Index([1, 2, 3], dtype="Int64")
```
MarcoGorelli: `pd.NA` originates with nullable types in pandas, so those are what I'm testing

I think using `pd.NA` with a non-nullable Series, or with a numpy-array, would be an anti-pattern anyway, right?

Dr-Irv:
@MarcoGorelli pine me when I should review

MarcoGorelli:
sure @Dr-Irv , happy for you to take a look, thanks!

pandas-stubs/_libs/missing.pyi
Dr-Irv: This method would get called in an operation such as `x @ pd.NA`, where `x.__matmul__()` is not defined.  E.g., `10 @ pd.NA` .

pandas-stubs/_libs/missing.pyi
Dr-Irv: leftover from 3 years ago when we copied the Microsoft version of the stubs.

tests/test_natype.py
Dr-Irv: yes

pandas-stubs/_libs/missing.pyi
```diff
@@ -1,46 +1,268 @@
...
+    Series,
+)
+from pandas.core.arrays.boolean import BooleanArray
 from typing_extensions import Self
 
 class NAType:
-    def __new__(cls, *args, **kwargs) -> Self: ...
+    def __new__(cls, *args: Any, **kwargs: Any) -> Self: ...
     def __format__(self, format_spec: str) -> str: ...
     def __bool__(self) -> None: ...
```
Dr-Irv: I think this should get removed.  `pd.NA.__bool__()` raises an exception.

pandas-stubs/_libs/missing.pyi
```diff
@@ -1,46 +1,268 @@
...
+        self, other: Series, /
+    ) -> tuple[Series, Series]: ...
+    @overload
+    def __rdivmod__(self, other: Index, /) -> tuple[Index, Index]: ...  # type: ignore[overload-overlap]
+    @overload
+    def __rdivmod__(self, other: object, /) -> tuple[NAType, NAType]: ...
+    @overload  # type: ignore[override]
+    def __eq__(  # type: ignore[overload-overlap] #  pyright: ignore[reportOverlappingOverload]
+        self, other: Series, /
+    ) -> Series: ...
```
Dr-Irv:
```suggestion
    ) -> Series:[bool] ...
```
and similar for the other comparison operators.

pandas-stubs/_libs/missing.pyi
```diff
@@ -1,46 +1,268 @@
...
+    @overload
+    def __rpow__(self, other: Index, /) -> Index: ...  # type: ignore[overload-overlap]
+    @overload
+    def __rpow__(self, other: object, /) -> NAType: ...
+    @overload
+    def __and__(  # type: ignore[overload-overlap] #  pyright: ignore[reportOverlappingOverload]
+        self, other: Series, /
+    ) -> Series: ...
```
Dr-Irv: If I do `pd.NA.__and__(pd.Series([1,2,3])` I get `NotImplemented`.  I think we should remove `__and__()` (and maybe others) from here if they don't produce a result

tests/test_natype.py
Dr-Irv: You shouldn't need the `: pd.Series[int]`   here, as the typing stubs should detect that because of the int values in the list.  Unfortunately, for the `Index` constructor, that isn't true, because we don't have `"Int64"` as a possible `dtype` value in `Index.__new__()` when passing integers.

tests/test_natype.py
```diff
@@ -0,0 +1,197 @@
...
+
+    # __add__
+    check(assert_type(na + s_int, pd.Series), pd.Series)
+    check(assert_type(na + idx_int, pd.Index), pd.Index)
+    check(assert_type(na + 1, NAType), NAType)
+
+    # __radd__
+    check(assert_type(s_int + na, pd.Series), pd.Series)
+    # https://github.com/pandas-dev/pandas-stubs/issues/1347
+    check(assert_type(idx_int + na, pd.Index), pd.Index)  # type: ignore[assert-type] # pyright: ignore[reportAssertTypeFailure]
```
Dr-Irv: That's indicative of a bug in the stubs.  Because `idx_int + na` works at runtime, which means you have to fix `Index.__add__()`

Similar comments on other operators.  Our basic rule in the tests is that we should only  have `ignore` in the tests when we are testing that the stubs should be picking up invalid code.

tests/test_natype.py
```diff
@@ -0,0 +1,197 @@
...
+    check(assert_type(na**2, NAType), NAType)
+
+    # __rpow__
+    check(assert_type(s_int**na, pd.Series), pd.Series)
+    # https://github.com/pandas-dev/pandas-stubs/issues/1347
+    check(assert_type(idx_int**na, pd.Index), pd.Index)  # type: ignore[assert-type]  # pyright: ignore[reportAssertTypeFailure]
+    check(assert_type(2**na, NAType), NAType)
+
+    # __and__
+    check(assert_type(na & s_int, pd.Series), pd.Series)
```
Dr-Irv: The reason this works is because `pd.NA.__and__()` returns `NotImplemented`, so then `pd.Series.__rand__(pd.NA)` is handled by the type checker.

I think we should be deleting the `__and__()`, `__or__()`, etc.

tests/test_natype.py
```diff
@@ -0,0 +1,197 @@
...
+            assert_type(
+                divmod(na, s_int),  # pyright: ignore[reportAssertTypeFailure]
+                tuple[pd.Series, pd.Series],
+            ),
+            tuple,
+        )
+    with pytest.raises(RuntimeError):
+        check(
+            assert_type(
+                divmod(na, idx_int),  # pyright: ignore[reportAssertTypeFailure]
```
Dr-Irv: something doesn't seem right here.  Why is pyright seeing this as a failure, but mypy doesn't?

Seems to be a `pyright` bug:  https://github.com/microsoft/pyright/issues/10849

Dr-Irv (changes requested):

tests/test_natype.py
MarcoGorelli: sure done, thanks!

tests/test_natype.py
MarcoGorelli: Yup I just wanted to indicate something like "this should work, but doesn't because of a different (orthogonal) issue"

perhaps i'll just leave it commented then, with a todo?

tests/test_natype.py
```diff
@@ -83,72 +82,68 @@ def test_arithmetic() -> None:
...
+    #     ),
+    #     tuple,
+    # )
+    # check(
+    #     assert_type(
+    #         divmod(na, idx_int),
+    #         tuple[pd.Index, pd.Index],
+    #     ),
+    #     tuple,
+    # )
```
Dr-Irv, line 100: I'm OK to leave this in, with the `#pyright: ignore`, because it works with `mypy`.  Then just make the comment about how `pyright` has the bug.

tests/test_natype.py
```diff
@@ -83,72 +82,68 @@ def test_arithmetic() -> None:
...
-        check(
-            assert_type(divmod(idx_int, na), tuple[pd.Index, pd.Index]),  # type: ignore[assert-type]  # pyright: ignore[reportAssertTypeFailure]
-            tuple,
-        )
+    # bug upstream: https://github.com/pandas-dev/pandas/issues/62196
+    # check(
+    #     assert_type(divmod(s_int, na), tuple[pd.Series, pd.Series]),
+    #     tuple,
+    # )
+    # https://github.com/pandas-dev/pandas-stubs/issues/1347
```
Dr-Irv, line 118: Isn't the bug upstream in `pandas`, not the orthogonal issue?

tests/test_natype.py
Dr-Irv: yes that's fine

tests/test_natype.py
MarcoGorelli: I'd say this is both, here

tests/test_natype.py
Dr-Irv: `pyright` 1.1.405 should fix this, so can you try changing `pyproject.toml` to use that version and see if you can put the test back?

MarcoGorelli: Sure, have updated, it does indeed allow us to remove the `# pyright: ignore`

tests/test_natype.py
```diff
@@ -0,0 +1,184 @@
...
+    #     tuple,
+    # )
+    # check(
+    #     assert_type(
+    #         divmod(na, idx_int),
+    #         tuple[pd.Index, pd.Index],
+    #     ),
+    #     tuple,
+    # )
+    check(assert_type(na.__divmod__(1), tuple[NAType, NAType]), tuple)
```
MarcoGorelli: i couldn't understand why, but this one fails if I write `divmod(na, 1)` instead, with
```python-traceback
  /home/marcogorelli/pandas-stubs-dev/tests/test_natype.py:105:23 - error: No overloads for "divmod" match the provided arguments (reportCallIssue)
  /home/marcogorelli/pandas-stubs-dev/tests/test_natype.py:105:23 - error: "assert_type" mismatch: expected "tuple[NAType, NAType]" but received "Unknown" (reportAssertTypeFailure)
  /home/marcogorelli/pandas-stubs-dev/tests/test_natype.py:105:34 - error: Argument of type "Literal[1]" cannot be assigned to parameter "y" of type "SupportsRDivMod[_T_contra@divmod, _T_co@divmod]" in function "divmod"
    "Literal[1]" is incompatible with protocol "SupportsRDivMod[NAType, _T_co@divmod]"
      "__rdivmod__" is an incompatible type
        Type "(value: int, /) -> tuple[int, int]" is not assignable to type "(other: _T_contra@SupportsRDivMod, /) -> _T_co@SupportsRDivMod"
          Parameter 1: type "_T_contra@SupportsRDivMod" is incompatible with type "int"
            "NAType" is not assignable to "int" (reportArgumentType)
```

Dr-Irv: New `pyright` bug.   https://github.com/microsoft/pyright/issues/10899

Let's wait to get that resolved.

MarcoGorelli: thanks for reporting to them

looks like they haven't responded - reckon it's ok to revisit this detail later and move forwards with the rest for now? there's the issue open in their repo to track it

Dr-Irv: Can you add a comment referring to the pyright issue?

Probably best to keep the test as `divmod(na, 1)` and put in a `#pyright :ignore` there

tests/test_natype.py
```diff
@@ -0,0 +1,187 @@
...
+
+    # __add__
+    check(assert_type(na + s_int, pd.Series), pd.Series)
+    check(assert_type(na + idx_int, pd.Index), pd.Index)
+    check(assert_type(na + 1, NAType), NAType)
+
+    # __radd__
+    check(assert_type(s_int + na, pd.Series), pd.Series)
+    # requires orthogonal fix: https://github.com/pandas-dev/pandas-stubs/issues/1347
+    # check(assert_type(idx_int + na, pd.Index), pd.Index)
```
Dr-Irv: This is now fixed for addition and subtraction.  So you can uncomment the test.

tests/test_natype.py
```diff
@@ -0,0 +1,187 @@
...
+
+    # __sub__
+    check(assert_type(na - s_int, pd.Series), pd.Series)
+    check(assert_type(na - idx_int, pd.Index), pd.Index)
+    check(assert_type(na - 1, NAType), NAType)
+
+    # __rsub__
+    check(assert_type(s_int - na, pd.Series), pd.Series)
+    # requires orthogonal fix: https://github.com/pandas-dev/pandas-stubs/issues/1347
+    # check(assert_type(idx_int - na, pd.Index), pd.Index)
```
Dr-Irv: This is now fixed for addition and subtraction.  So you can uncomment the test.

tests/test_natype.py
```diff
@@ -0,0 +1,187 @@
...
+    #     assert_type(
+    #         divmod(na, idx_int),
+    #         tuple[pd.Index, pd.Index],
+    #     ),
+    #     tuple,
+    # )
+    # https://github.com/microsoft/pyright/issues/10899.
+    check(
+        assert_type(divmod(na, 1), tuple[NAType, NAType]), tuple
+    )  # pyright: ignore[reportArgumentType, reportCallIssue, reportAssertTypeFailure]
```
Dr-Irv: This is failing in CI.  I got it to work by writing it this way:
```python
    check(
        assert_type(
            divmod(  # pyright: ignore[reportCallIssue, reportAssertTypeFailure]
                na, 1  # pyright: ignore[reportArgumentType]
            ),
            tuple[NAType, NAType],
        ),
        tuple,
    )
```

PR #1349, opened 2025-08-26
GH1336 Reconcile Series.index and DataFrame.index

### LINKED ISSUE #1336: Reconcile `Series.index` and `DataFrame.index`
- `Series.index`
  - ` def index(self) -> Index | MultiIndex: ...`
- `DataFrame.index`
  -  `def index(self) -> UnknownIndex: ...`

These are inconsistent.  I think they should both read `def index(self) -> Index: ...` 

I think this change will allow `UnknownIndex` to be removed.

loicdiridollou:
- [x] Closes #1336


tests/test_scalars.py
```diff
@@ -1322,9 +1322,6 @@ def test_timestamp_cmp_index() -> None:
     eq_dt1 = check(assert_type(ts == dt_idx, np_1darray[np.bool]), np_1darray[np.bool])
     ne_dt1 = check(assert_type(ts != dt_idx, np_1darray[np.bool]), np_1darray[np.bool])
     assert (eq_dt1 != ne_dt1).all()
-    eq_un1 = check(assert_type(ts == un_idx, np_1darray[np.bool]), np_1darray[np.bool])
-    ne_un1 = check(assert_type(ts != un_idx, np_1darray[np.bool]), np_1darray[np.bool])
-    assert (eq_un1 != ne_un1).all()
```
Dr-Irv: So I know I suggested removing these tests.  But what will happen now is that `ts == un_idx` will come out as `False`, and that's more misleading.

So let's go back to keeping this test in there, and in `timestamps.pyi`, change `DatetimeIndex` back to `Index`, and for the tests that `mypy` is getting wrong, just put a comment in there that this is some kind of mypy bug.

pandas-stubs/_libs/tslibs/timestamps.pyi
```diff
@@ -257,7 +256,7 @@ class Timestamp(datetime, SupportsIndex):
     @overload
     def __eq__(self, other: TimestampSeries) -> Series[bool]: ...  # type: ignore[overload-overlap]
     @overload
-    def __eq__(self, other: Index) -> np_1darray[np.bool]: ...  # type: ignore[overload-overlap]
+    def __eq__(self, other: DatetimeIndex) -> np_1darray[np.bool]: ...  # type: ignore[overload-overlap]
```
Dr-Irv: I think we have to change this to `Index`, see comment elsewhere.

And change `periods.pyi` and `timedeltas.pyi` to use `Index`, and for the tests that `mypy` fails on, just put in `#type: ignore` with comments.

tests/test_scalars.py
loicdiridollou: I think indeed the best solution was the intermediary one, a less common but still possible case is to have a mixed-type index which would result in a broken typing with the original proposal.
This is fixed.

PR #1350, opened 2025-08-27
feat(index): ➕ arithmetic addition

cmp0xff:
It seems that #1347 requires a series of changes, albeit not in the sense of a Pandas Series. This PR could be a good place to start.


tests/indexes/arithmetic/bool/test_add.py
```diff
@@ -0,0 +1,83 @@
...
+    check(assert_type(i + left, "npt.NDArray[np.int64]"), pd.Index, np.integer)
+    check(assert_type(f + left, "npt.NDArray[np.float64]"), pd.Index, np.floating)
+    check(
+        assert_type(c + left, "npt.NDArray[np.complex128]"),
+        pd.Index,
+        np.complexfloating,
+    )
+
+
+def test_add_pd_series() -> None:
```
Dr-Irv: change name to `test_add_pd_index` .  Do this here and in other test files.

Can you also add tests that test adding a `Series` to an `Index`, e.g.:
```python
s = pd.Series([2,3,5])
left + s   #  Should return a Series
```

Do above in other files as well.

Dr-Irv (changes requested):
Generally looks good.  

I think you can also remove `__add__()` and `__radd__()` from `IndexOpsMixin` because only `Index` and `Series` use `IndexOpsMixin` as a subclass, and now the definitions will reside in `class Series` and `class Index` .

I was thinking there might be a way to just put all the operators in `IndexOpsMixin`, but the challenge there is that `Series.__add__(Series)` returns a `Series` at runtime, but `Index.__add__(Series)` also returns a `Series` at runtime, so by not allowing `Series` as a parameter to `Index.__add__()`, the type checker will revert to the proper overload for `Series.__radd__()`

tests/indexes/arithmetic/bool/test_add.py
cmp0xff: 1a1ff50893585e0900d3906508210f25949495b4

cmp0xff:
- There is no `__add__()` or `__radd__()` in `IndexOpsMixin`
- There are `__add__()` and `__radd__()` in `OpsMixin`, which I removed in 8a0dcbdb1fa01316c9c7b0493cfad9b27911dda8. I feel like it is a good idea. There are arithmetic operations that break the pattern `-> Self`, e.g. `Series[Timestamp] - Series[Timestamp]`. Removing such definitions may help resolving the conflicts, which still persist in my effort to merge `main` into #1274.

tests/indexes/test_indexes.py
```diff
@@ -262,12 +262,10 @@ def test_types_to_numpy() -> None:
 def test_index_arithmetic() -> None:
     # GH 287
     idx = pd.Index([1, 2.2, 3], dtype=float)
-    check(assert_type(idx + 3, "pd.Index[float]"), pd.Index, np.float64)
```
Dr-Irv, line 265: why remove this test (and the one below for `3 + idx` ?

Dr-Irv:
> * There is no `__add__()` or `__radd__()` in `IndexOpsMixin`
> * There are `__add__()` and `__radd__()` in `OpsMixin`, which I removed in [8a0dcbd](https://github.com/pandas-dev/pandas-stubs/commit/8a0dcbdb1fa01316c9c7b0493cfad9b27911dda8). I feel like it is a good idea. There are arithmetic operations that break the pattern `-> Self`, e.g. `Series[Timestamp] - Series[Timestamp]`. Removing such definitions may help resolving the conflicts, which still persist in my effort to merge `main` into [refactor: #718 only drop TimestampSeries #1274](https://github.com/pandas-dev/pandas-stubs/pull/1274).

My mistake - I meant `OpsMixin`

tests/indexes/test_indexes.py
cmp0xff: I think it has now been covered by the more systematic tests

Dr-Irv: ok

PR #1353, opened 2025-08-28
GH1352 Add passing a dict with integer for pd.to_datetime

### LINKED ISSUE #1352: pd.to_datetime from a dict where one of the field is an int
**Describe the bug**

When the dictionary is passed to `to_datetime` there is a possibility of passing not a Series or list in one of the fields but a single int, like I want all the days to be 1 so one would pass:

```python
time_series = pd.date_range("2023-10-01", "2024-05-01", freq="1h")

time_series = pd.to_datetime({
    "year": time_series.year,
    "month": time_series.month,
    "day": 1,
})
```

**To Reproduce**

```python
time_series = pd.date_range("2023-10-01", "2024-05-01", freq="1h")

time_series = pd.to_datetime({
    "year": time_series.year,
    "month": time_series.month,
    "day": 1,
})
```

 - OS: MacOS
 - python version 1.13
 - version of type checker 1.17

**Additional context**

loicdiridollou:
- [x] Closes #1352 

+ fix for the nightly CI with pandas

### DIFF
--- a/pandas-stubs/core/tools/datetimes.pyi
+++ b/pandas-stubs/core/tools/datetimes.pyi
@@ -38,7 +38,7 @@ DatetimeScalar: TypeAlias = Scalar | datetime | np.datetime64 | date
 
 DatetimeScalarOrArrayConvertible: TypeAlias = DatetimeScalar | ArrayConvertible
 
-DatetimeDictArg: TypeAlias = list[Scalar] | tuple[Scalar, ...] | AnyArrayLike
+DatetimeDictArg: TypeAlias = list[Scalar] | tuple[Scalar, ...] | AnyArrayLike | int
 
 class YearMonthDayDict(TypedDict, total=True):
     year: DatetimeDictArg
--- a/tests/test_scalars.py
+++ b/tests/test_scalars.py
@@ -24,6 +24,7 @@ from pandas._libs.tslibs.timedeltas import Components
 from pandas._typing import TimeUnit
 
 from tests import (
+    PD_LTE_23,
     TYPE_CHECKING_INVALID_USAGE,
     check,
     np_1darray,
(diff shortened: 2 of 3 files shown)


pandas-stubs/core/tools/datetimes.pyi
```diff
@@ -38,7 +38,7 @@ DatetimeScalar: TypeAlias = Scalar | datetime | np.datetime64 | date
 
 DatetimeScalarOrArrayConvertible: TypeAlias = DatetimeScalar | ArrayConvertible
 
-DatetimeDictArg: TypeAlias = list[Scalar] | tuple[Scalar, ...] | AnyArrayLike
+DatetimeDictArg: TypeAlias = list[Scalar] | tuple[Scalar, ...] | AnyArrayLike | int
```
Dr-Irv: There are 2 ways of doing this:

1. Use `Scalar` instead of `int` so it mirrors the rest of the declaration.
2. Change `list[Scalar]`, `tuple[Scalar, ...]` and `AnyArrayLike` to be more restrictive.  For example, you can't pass `float` values, and maybe just `int` and `str` are accepted.  So right now the type is too wide.

I'm OK if you just want to do (1) for now.

Dr-Irv (changes requested):
thanks for also doing the `Pandas4Warning` thing to fix nightly.  I was going to do that when they fix that one other issue - still waiting for that.

pandas-stubs/core/tools/datetimes.pyi
loicdiridollou: So it turns out you can actually pass some float but it is a weird edge case and I would advise against it:
```python
time_series = pd.date_range("2023-10-01", "2024-05-01", freq="1h")
pd.to_datetime({"year": time_series.year, "month": 3, "day": 4.5})
```
To me this 4.5 at the end just gets converted to 4 so I would advise against allowing float, also no other tests fail with switching to `int | str` instead of `float | str`, please let me know if you agree with that.

loicdiridollou: Wrt `AnyArrayLike`, it can't be subscribed as a generic but indeed ideally it should be something like `AnyArrayLike[int | str]`, prob an issue for another day.

Dr-Irv: > Wrt `AnyArrayLike`, it can't be subscribed as a generic but indeed ideally it should be something like `AnyArrayLike[int | str]`, prob an issue for another day.

Agreed.  Could replace `AnyArrayLike` here with `ArrayLike | Index[str] | Index[int] | Series[int] | Series[str]`, but not worth it at this point.

Dr-Irv (approved):
thanks @loicdiridollou

PR #1354, opened 2025-08-29
with conda, use sys.executable to find python with ty

Dr-Irv:
I've been having trouble doing `poe ty` or `poe test` because `ty` is failing to find the python environment, and I use conda.  Fails for me on Windows and Linux. 

This fixes it for me.  I don't think it will break it for others.

@loicdiridollou can you review this by seeing if this change works on your system?

If you approve it, then I'll do the merge.

### DIFF
--- a/scripts/test/run.py
+++ b/scripts/test/run.py
@@ -154,7 +154,13 @@ def released_mypy():
 
 
 def ty():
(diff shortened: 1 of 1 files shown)


loicdiridollou (approved):
Works fine on my end so good to go

PR #1357, opened 2025-08-29
type IndexOpsMixin

### LINKED ISSUE #1356: `Index.searchsorted` return type incorrect
**Describe the bug**

`Index.searchsorted` returns `int | list[int]`, but at runtime it's a numpy array

**To Reproduce**

```python
import pandas as pd

idx = pd.Index([1,2,3])

reveal_type(idx.searchsorted([1]))
```

runtime: ndarray
type checking: int | list[int]

 - OS Version [e.g. 22]
 - python version
 - version of type checker
 - version of installed `pandas-stubs`  58d59f2390803c13be29773055996833dfb13f87

**Additional context**

MarcoGorelli:
closes #1356 

### DIFF
--- a/pandas-stubs/core/base.pyi
+++ b/pandas-stubs/core/base.pyi
@@ -6,6 +6,7 @@ from typing import (
     Any,
     Generic,
     Literal,
+    TypeAlias,
     final,
     overload,
 )
@@ -22,6 +23,7 @@ from typing_extensions import Self
 
 from pandas._typing import (
     S1,
+    ArrayLike,
     AxisIndex,
     DropKeep,
     DTypeLike,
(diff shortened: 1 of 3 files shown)


pandas-stubs/core/base.pyi
```diff
@@ -22,18 +23,22 @@ from typing_extensions import Self
...
     GenericT_co,
     NDFrameT,
     Scalar,
+    SequenceNotStr,
     SupportsDType,
     np_1darray,
 )
 from pandas.util._decorators import cache_readonly
 
+_ListLike: TypeAlias = ArrayLike | dict[str, np.ndarray] | SequenceNotStr[S1]
```
MarcoGorelli, line 40: not sure why this gets redefined in multiple files, but i've kept to the pattern

ok, or better to redefine in `_typing`?

Dr-Irv: If you do want to use it here, then I think moving the definition to `typing.pyi` makes sense, but exercise care on the naming, because there is a `ListLike` in `_typing` that is used elsewhere and I'm not sure whether they should be the same or not.   Maybe the name `_ListLike` is misleading.

Might be better to do the copying like you've done here, OR have `series.pyi` import `_ListLike` from `core/base.pyi`

Dr-Irv (commented):
ping when I should do the full review, but it looks OK for now

pandas-stubs/core/base.pyi
MarcoGorelli: > Might be better to do the copying like you've done here

sure, keeping like this for now then, thanks

MarcoGorelli:
sure, happy for you to take a look, thanks!

PR #1360, opened 2025-09-02
feat(index): ➖ arithmetic subtraction

cmp0xff:
It belongs to the series of changes suggested by pandas-dev/pandas-stubs#1347, although not in the sense of a Pandas Series.


pandas-stubs/core/indexes/base.pyi
```diff
@@ -77,12 +80,23 @@ from pandas._typing import (
...
+_NumListLike: TypeAlias = (
+    ExtensionArray
+    | np_ndarray_bool
+    | np_ndarray_anyint
+    | np_ndarray_float
+    | np_ndarray_complex
+    | dict[_str, np.ndarray]
+    | Sequence[complex]
+    | IndexOpsMixin[complex]
+)
```
Dr-Irv: Can we move these 2 types to `core/base.pyi` and import them from there, since they are the same for `Series` and `Index`

pandas-stubs/core/indexes/base.pyi
```diff
@@ -626,6 +640,156 @@ class Index(IndexOpsMixin[S1]):
...
+    @overload
+    def __sub__(
+        self: Index[Timestamp],
+        other: timedelta | np.timedelta64 | np_ndarray_td | TimedeltaIndex,
+    ) -> DatetimeIndex: ...
+    @overload
+    def __sub__(
+        self: Index[Timedelta],
+        other: timedelta | np.timedelta64 | np_ndarray_td | TimedeltaIndex,
+    ) -> TimedeltaIndex: ...
```
Dr-Irv: Ideally, we wouldn't have a `Index[Timestamp]` or `Index[Timedelta]`, because those should be `DatetimeIndex` and `TimedeltaIndex`.  Could we move those 2 overloads to those 2 classes, respectively?

tests/indexes/arithmetic/test_sub.py
```diff
@@ -0,0 +1,103 @@
...
+from typing_extensions import assert_type
+
+from tests import (
+    TYPE_CHECKING_INVALID_USAGE,
+    check,
+)
+
+# left operands
+left_i = pd.MultiIndex.from_tuples([(1,), (2,), (3,)]).levels[0]
+left_str = pd.MultiIndex.from_tuples([("1",), ("2",), ("3_",)]).levels[0]
```
Dr-Irv: you're not using `left_str` here, so remove this?  Or did you intend to add a test?

Dr-Irv (changes requested):
Looks pretty good

pandas-stubs/core/indexes/base.pyi
cmp0xff: Moved in 1877b98c701c36a80500ad587bb283f6de0d64e7. Unfortunately there is a `ListLike` in `_typing.pyi` which is different.

pandas-stubs/core/indexes/base.pyi
cmp0xff: Moved in 1877b98c701c36a80500ad587bb283f6de0d64e7.

tests/indexes/arithmetic/test_sub.py
cmp0xff: It was a mistake in copying. Removed in 1877b98c701c36a80500ad587bb283f6de0d64e7.

cmp0xff:
I understand that a set of systematic tests is desirable whenever the corresponding code is modified. In this case, however, I would personally prefer to develop these tests in a separate pull request or issue.

It has been quite exhausting to work out the timestamp and timedelta subtractions; the challenge may trace back to the fact that the subtraction here is not closed in the algebraic sense.

If needed, I can also revert the changes in `DatetimeIndex` and `TimedeltaIndex`, so that they are unmodified in this PR.

Dr-Irv (approved):
thanks @cmp0xff

Dr-Irv:
> I understand that a set of systematic tests is desirable whenever the corresponding code is modified. In this case, however, I would personally prefer to develop these tests in a separate pull request or issue.

That's fine.

PR #1362, opened 2025-09-08
Simplify and improve Series.sum

hamdanal:

`Series.sum` was annotated to return the type of the series elements except for `Series[bool]` that was special-cased to return `int`. This is conceptually wrong, although it works for the most common series types. The return value of `Series.sum` depends on the return value of its elements' `__add__` method. Fixing this simplifies the signature of `sum` and makes it work generically (for example if in the future `Series[np.bool]` becomes a thing).

Note that I didn't add a test case for this as it is already tested in `test_types_sum` in `test_series.py`. Note also that the `_SupportsAdd` type-check-only protocol is added in the series module instead of `_typing` as it is very specific to this use case.

### DIFF
--- a/pandas-stubs/core/series.pyi
+++ b/pandas-stubs/core/series.pyi
@@ -26,11 +26,13 @@ from typing import (
     Generic,
     Literal,
     NoReturn,
+    Protocol,
     final,
     overload,
     type_check_only,
 )
 
+from _typeshed import SupportsGetItem
 from matplotlib.axes import (
     Axes as PlotAxes,
     SubplotBase,
@@ -187,6 +189,7 @@ from pandas._typing import (
     ValueKeyFunc,
     VoidDtypeArg,
     WriteBuffer,
+    _T_co,
     np_1darray,
     np_ndarray_anyint,
     np_ndarray_bool,
@@ -202,6 +205,10 @@ from pandas.core.dtypes.base import ExtensionDtype
 from pandas.core.dtypes.dtypes import CategoricalDtype
 
 from pandas.plotting import PlotAccessor
+
+@type_check_only
+class _SupportsAdd(Protocol[_T_co]):
+    def __add__(self, value: Self, /) -> _T_co: ...
 
 class _iLocIndexerSeries(_iLocIndexer, Generic[S1]):
     # get item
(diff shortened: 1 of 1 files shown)


pandas-stubs/core/series.pyi
```diff
@@ -26,11 +26,13 @@ from typing import (
     Generic,
     Literal,
     NoReturn,
+    Protocol,
     final,
     overload,
     type_check_only,
 )
 
+from _typeshed import SupportsGetItem
```
Dr-Irv, line 35: I'm not familiar with this.  Is it documented somewhere that imports from `_typeshed` will work?

pandas-stubs/core/series.pyi
```diff
@@ -203,6 +206,10 @@ from pandas.core.dtypes.dtypes import CategoricalDtype
 
 from pandas.plotting import PlotAccessor
 
+@type_check_only
```
Dr-Irv, line 209: I just realized that you added the use of `type_check_only` in a previous PR, but now I'm wondering if it makes sense in a stub file, because if you're using a stub file, doesn't that mean that its contents are only for type checking?

pandas-stubs/core/series.pyi
hamdanal: The `_typeshed` namespace is distributed as part of the python standard library and is included with all type checkers similar to `typing_extensions`. It is documented here https://github.com/python/typeshed/tree/main/stdlib/_typeshed#utility-types-for-typeshed and is referenced in the mypy documentation and the Python typing documentation (for example here https://typing.python.org/en/latest/guides/writing_stubs.html#incomplete-stubs).

That said, we can always copy the definition of this protocol here but we will basically be duplicating the work already done and tested in typeshed. Let me know if you prefer this solution anyhow and I can do this.

pandas-stubs/core/series.pyi
hamdanal: This decorator is meant to be used with functions/classes that exist at type checking time but not at runtime. It is not strictly needed but it tells type checkers to flag any misuse of the decorated object at runtime. It is part of the typing module but it only exists at type checking time. The API documentation is here https://docs.python.org/3/library/typing.html#typing.type_check_only and the user guide is here https://typing.python.org/en/latest/guides/writing_stubs.html#stub-only-objects

pandas-stubs/core/series.pyi
Dr-Irv: > The `_typeshed` namespace is distributed as part of the python standard library and is included with all type checkers similar to `typing_extensions`. 

I don't think `_typeshed` is shipped as part of the python standard library.
If it is shipped with the type checkers, then I'm OK with it.

pandas-stubs/core/series.pyi
Dr-Irv: OK - thanks for the explanation.  We probably should be using that more in the pandas stubs

pandas-stubs/core/series.pyi
hamdanal: My bad, I meant to say the python standard library *stubs*, similar to how stubs are allowed to use `typing_extensions` without installing it.

PR #1374, opened 2025-09-17
fix(series): #1372 🧱✖️ cumprod

### LINKED ISSUE #1372: Series#cumprod overloads
**Describe the bug**

there are two overload for `Series#cumprod`:
https://github.com/pandas-dev/pandas-stubs/blob/e959fc933b024f7af6ad6efb5e8f07e8772ede61/pandas-stubs/core/series.pyi#L3372-L3386

When the `self` is `Series[Any]`, according to [current typing spec](https://github.com/erictraut/typing/blob/f0decdcb56d0dc9bd2d928cabda7e4462d098b1f/docs/spec/overload.rst), first one is selected and return type become `Never` unexpectedly.
(EDIT: I may be misunderstanding *Step 5* of the spec. but if the second one is not eliminated at that step, resulting `Any` is not convenient...)

**To Reproduce**

```py
from pandas import Series

def f(s: Series):
    reveal_type(s.cumprod())
```
```
$ mypy p.py 
p.py:5: note: Revealed type is "Never"
```

```
$ pyright p.py
p.py
  p.py:5:17 - information: Type of "s.cumprod()" is "Never"
0 errors, 0 warnings, 1 information
```

 - OS: Linux
 - OS Version: Ubuntu 24.04.2
 - Python: 3.12.3
 - mypy 1.18.1
 - pyright: 1.1.405
 - pandas-stubs: 2.3.2.250827

**Additional context**

originally reported on https://github.com/microsoft/pyright/issues/10935

cmp0xff:
- [x] Closes #1372

Thank you @hamdanal, the idea in #1362 really helps. Also grateful to @Dr-Irv for tagging me there.

### DIFF
--- a/tests/series/test_cumul.py
+++ b/tests/series/test_cumul.py
@@ -8,7 +8,7 @@ from tests import (
 )
 
 
-def test_cumul_any_int() -> None:
+def test_cumul_any_float() -> None:
     series = pd.DataFrame({"A": [1.0, float("nan"), 2.0]})["A"]
     check(assert_type(series.cumprod(), pd.Series), pd.Series, np.floating)
 
--- a/pandas-stubs/core/frame.pyi
+++ b/pandas-stubs/core/frame.pyi
@@ -549,9 +549,9 @@ class DataFrame(NDFrame, OpsMixin, _GetItemHack):
             | Mapping[str, Any]
             | Mapping[str, SequenceNotStr[Any]]
         ),
-        index: str | SequenceNotStr[Hashable] | None = None,
-        columns: ListLike | None = None,
+        index: str | Axes | None = None,
         exclude: ListLike | None = None,
+        columns: ListLike | None = None,
         coerce_float: bool = False,
         nrows: int | None = None,
     ) -> Self: ...
--- a/pandas-stubs/core/indexes/datetimes.pyi
+++ b/pandas-stubs/core/indexes/datetimes.pyi
@@ -20,6 +20,7 @@ from pandas import (
     TimedeltaIndex,
     Timestamp,
 )
+from pandas.core.arrays import DatetimeArray
 from pandas.core.indexes.accessors import DatetimeIndexProperties
 from pandas.core.indexes.datetimelike import DatetimeTimedeltaMixin
 from pandas.core.series import (
@@ -61,6 +62,11 @@ class DatetimeIndex(
         name: Hashable = ...,
     ) -> Self: ...
     def __reduce__(self): ...
+
+    # Override the array property to return DatetimeArray instead of ExtensionArray
+    @property
+    def array(self) -> DatetimeArray: ...
+
     # various ignores needed for mypy, as we do want to restrict what can be used in
     # arithmetic for these types
     @overload  # type: ignore[override]
--- a/tests/indexes/test_indexes.py
+++ b/tests/indexes/test_indexes.py
@@ -12,6 +12,7 @@ from typing import (
 import numpy as np
 from numpy import typing as npt
 import pandas as pd
+from pandas.core.arrays import DatetimeArray
 from pandas.core.arrays.categorical import Categorical
 from pandas.core.indexes.base import Index
 from typing_extensions import (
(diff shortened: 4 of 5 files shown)


tests/series/test_series.py
```diff
@@ -3948,24 +3948,18 @@ def test_timedelta_index_cumprod() -> None:
...
     if TYPE_CHECKING_INVALID_USAGE:
-        assert_type(pd.Series(["a", "b"]).cumprod(), Never)
+        offset_series.cumprod()  # type: ignore[misc] # pyright: ignore[reportAttributeAccessIssue]
 
     if TYPE_CHECKING_INVALID_USAGE:
-        assert_type(offset_series.cumprod(), Never)
+        pd.Series([pd.Timedelta(0), pd.Timedelta(1)]).cumprod()  # type: ignore[misc] # pyright: ignore[reportAttributeAccessIssue]
 
     if TYPE_CHECKING_INVALID_USAGE:
-        assert_type(pd.Series([pd.Timedelta(0), pd.Timedelta(1)]).cumprod(), Never)
```
Dr-Irv, line 3955: I think you dropped this test.  Can you put it back?

tests/series/window/__init__.py
Dr-Irv, line 1: I'm not sure why you called this "window".  Maybe it should be "cumul" because I guess you'd put `cumsum()` in here too eventually.

cmp0xff: The bigger plan is to test and fix all window functions, like `cumsum`. At the moment only `cumprod` is tested. I believe collectively they are called window functions.

tests/series/test_series.py
cmp0xff: I only removed `assert_type(pd.Series(["a", "b"]).cumprod(), Never)` because it is now covered in `test_window_str.py`. The rest are
- `offset_series.cumprod()`
- `pd.Series([pd.Timedelta(0), pd.Timedelta(1)]).cumprod()`
- `pd.Series(
                [pd.Timestamp("2024-04-29"), pd.Timestamp("2034-08-28")]
            ).cumprod(),
            Never,
        )`
- as_period_series.cumprod()

which are all kept.

Dr-Irv: OK.  I misread the change.

tests/series/window/__init__.py
Dr-Irv: Well, that's my objection.  I don't think they are window functions and it's potentially confusing because there is https://pandas.pydata.org/pandas-docs/stable/reference/window.html which are more like rolling windows, whereas the `cumXXX` functions are just cumulative over the whole series.  That's why I would prefer the name change.

cmp0xff: Naming fixed, also added tests for `Series[bool]` and `Series[Timestamp]`

tests/series/test_cumul.py
```diff
@@ -0,0 +1,51 @@
...
+import pandas as pd
+from typing_extensions import assert_type
+
+from tests import (
+    TYPE_CHECKING_INVALID_USAGE,
+    check,
+)
+
+
+def test_cumul_any_int() -> None:
```
Dr-Irv:
```suggestion
def test_cumul_any_float() -> None:
```

pandas-stubs/core/series.pyi
```diff
@@ -3821,23 +3825,23 @@ class Series(IndexOpsMixin[S1], NDFrame):
...
         axis: AxisIndex = ...,
         skipna: _bool = ...,
         *args: Any,
         **kwargs: Any,
-    ) -> Never: ...
+    ) -> Series[int]: ...
     @overload
     def cumprod(
-        self,
+        self: SupportsGetItem[Scalar, _SupportsMul[S1]],
```
twoertwein, line 3844: Does that work with classes that have overloads for `__mul__`? My understanding from https://github.com/microsoft/pyright/discussions/6549#discussioncomment-7669091 is that type checkers will then pick the first overload. It might be okay in this case (and I wish we could use this pattern!) - just something to be aware of when using this patterns in more places.

Dr-Irv: Not sure that it would matter, because `S1` are types we define in `pandas` as well as fundamental types (`int`, `float`, etc.), so I think we will be fine.  It's a bit different than the case you wrote up, because of the use of `_SupportsMul[S1]` requiring the type that supports `__mul__()` to be in `S1`

tests/series/test_cumul.py
cmp0xff: a383fbee5abd0b450a4a689ea8dc4b880ab95f3c

pandas-stubs/core/series.pyi
cmp0xff: I with this pattern can help us saving efforts, too.
- Note that there are already `SupportsAdd` and `SupportsMul` in `_typeshed`. They are however different somehow from our `_SupportsAdd` and `_SupportsMul`. We probably need a renaming in the near future.
- It would be great if we could work out a pattern that deal with typing changes like `A.__operate__(B) ->B` and `A.__operate__(A) -> B`, for example `bool + int -> int`, `int / int -> float`.

PR #1377, opened 2025-09-23
fix: fix Series construction with data as an iterator

terencehonles:

### DIFF
--- a/pandas-stubs/core/series.pyi
+++ b/pandas-stubs/core/series.pyi
@@ -294,7 +294,7 @@ class Series(IndexOpsMixin[S1], NDFrame):
     @overload
     def __new__(  # type: ignore[overload-overlap] # pyright: ignore[reportOverlappingOverload]
         cls,
-        data: Sequence[Never],
+        data: Iterator[Never],
         index: AxesData | None = ...,
         dtype: Dtype = ...,
         name: Hashable = ...,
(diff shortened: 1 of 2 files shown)


tests/series/test_series.py
```diff
@@ -155,6 +156,11 @@ def test_types_init() -> None:
         pd.Series,
         float,
     )
+    check(
+        assert_type(pd.Series(iter([1.0])), "pd.Series[float]"),
+        pd.Series,
+        float,
+    )
```
cmp0xff, line 163: Several `__new__`'s have been retyped, but only one test has been added. Please complete the new tests. Alternatively, you can split the old test function into two.

pandas-stubs/core/series.pyi
```diff
@@ -312,7 +312,7 @@ class Series(IndexOpsMixin[S1], NDFrame):
     @overload
     def __new__(
         cls,
-        data: Sequence[_str],
+        data: Iterator[_str],
```
cmp0xff, line 315: `Iterator[_str]` can be ambiguous, because `_str` is an `Iterable` in its own right. However, I don't have more insights how to improve this.

Dr-Irv: You can use `Iterator[SequenceNotStr[_str]]`

PR #1382, opened 2025-09-26
refactor(arithmetic): ➕ simplify addition

cmp0xff:
- Using `_typeshed.SupportsRAdd` and `SupportsAdd` in `other`, I am able to reduce 110 lines of code
- Unfortunately I am not able to use `SupportsAdd` or `SupportsRAdd` in the typing of `self`
   - One can imagine something like `def __add__(self: SupportsGetItem[Scalar, SupportsAdd[S1_CT, S1_CO]], S1_CT | Sequence[S1_CT]) -> S1_CO`. However it does not work.


pandas-stubs/core/indexes/base.pyi
```diff
@@ -471,7 +475,6 @@ class Index(IndexOpsMixin[S1]):
     def shape(self) -> tuple[int, ...]: ...
     # Extra methods from old stubs
     def __eq__(self, other: object) -> np_1darray[np.bool]: ...  # type: ignore[override] # pyright: ignore[reportIncompatibleMethodOverride]
-    def __iter__(self) -> Iterator[S1]: ...
```
Dr-Irv, line 474: not sure why you removed this?

pandas-stubs/core/series.pyi
```diff
@@ -1656,70 +1662,83 @@ class Series(IndexOpsMixin[S1], NDFrame):
...
+    ) -> Series[float]: ...
+    @overload
+    def __add__(
+        self: Series[T_COMPLEX], other: np_ndarray_float | Index[float] | Series[float]
+    ) -> Series[T_COMPLEX]: ...
+    @overload
+    def __add__(
+        self: Series[T_COMPLEX],
+        other: np_ndarray_complex | Index[complex] | Series[complex],
     ) -> Series[complex]: ...
```
Dr-Irv, line 1743: A bit surprised that `Sequence[bool]`, `Sequence[int]`, etc. are not there when `np_ndarray_bool`, `np_ndarray_anyint`, etc. are there.

Dr-Irv (commented):
A couple of comments that are questions.

pandas-stubs/core/indexes/base.pyi
cmp0xff: It is already defined in the base class `IndexOpsMixin`

pandas-stubs/core/series.pyi
cmp0xff: > `Sequence[bool]`, `Sequence[int]`, etc. are not there

They are handled by `Sequence[SupportsRAdd[S1_CT_NDT, S1_CO]]`.

For example, how does type checker decide `Series[int] + Sequence[complex] -> Series[complex]`? Here `S1_CT_NDT` is `int`, and the `complex` inside `Series` is `S1_CO`.

I figured out that `complex().__radd__(int())` is defined, giving a result type of `complex` (or it fits. Hence the `complex` in `Sequence` is `SupportsRAdd[int, complex]`.

> `np_ndarray_bool`, `np_ndarray_anyint`, etc. are there.

I cannot make them work. Operated with python types, numpy types give numpy types, so that our `S1_CO` in `Series[S1_CO]` becomes a numpy type, which is not valid. That's why they need to be given explicitly.

`Index[S1]` is also there, partially probably because `S1` is not covariant, so that unlike `Sequence`, `Index[SupportsRAdd[S1_CT_NDT, S1_CO]]` does not work well.

`Series[S1]` is also there, because I cannot define `Series + complex` and then using `Protocol` to make `Series + Series[complex]` work, in the same class `Series`. Type checkers do not seem to work iteratively / recursively with respect to the class itself.

---

These are just my reasoning as a human. When type checkers work, I can try to explain. There are more cases where type checkers don't work and I don't know why.

Dr-Irv: > They are handled by `Sequence[SupportsRAdd[S1_CT_NDT, S1_CO]]`.

OK, that makes sense.

pandas-stubs/core/indexes/base.pyi
Dr-Irv: Then I think you should also delete it from `core/series.pyi`

cmp0xff: 69fefaa0d2fc7a399cfeb498c6c5b7298d1ad163

PR #1385, opened 2025-09-28
refactor(series)!: drop IntervalSeries

### LINKED ISSUE #1379: Remove PeriodSeries, OffsetSeries, IntervalSeries

cmp0xff:
- [x] Closes #1379
- [x] Tests added


tests/test_extension.py
```diff
@@ -24,7 +24,7 @@ def test_tolist() -> None:
     s = pd.Series(data)
     data1 = [1, 2, 3]
     s1 = pd.Series(data1)
-    check(assert_type(s.array.tolist(), list), list)
+    check(assert_type(s.array.tolist(), list), list)  # type: ignore[assert-type]
```
cmp0xff, line 29: It is mysterious that `mypy` does not like it

pandas-stubs/core/indexes/accessors.pyi
```diff
@@ -434,49 +443,37 @@ class TimedeltaIndexProperties(
...
+    ) -> IntervalArray: ...
+    @overload
+    def __get__(
+        self, instance: IndexOpsMixin[Timestamp], owner: type[IndexOpsMixin]
+    ) -> DatetimeArray: ...
+    @overload
+    def __get__(
+        self, instance: IndexOpsMixin[Timedelta], owner: type[IndexOpsMixin]
+    ) -> TimedeltaArray: ...
+    # should be NumpyExtensionArray
```
Dr-Irv, line 482: `NumpyExtensionArray` is a subclass of `ExtensionArray`, so comment not needed.

Issue here is that `pd.Series([1,2,3], dtype=int).array` is `NumpyExtensionArray` while `pd.Series([1,2,3],dtype="Int64").array` is `IntegerArray`, but the `Series` for typing purposes in both cases is `Series[int]`

pandas-stubs/core/series.pyi
```diff
@@ -1234,8 +1229,7 @@ class Series(IndexOpsMixin[S1], NDFrame):
         Series[_str],
         Series,
     ]: ...
-    @property
-    def dt(self) -> _dtDescriptor[S1]: ...
+    dt: ClassVar = DtDescriptor()  # noqa
```
Dr-Irv: Ha - this is similar to what I suggested in #1379

pandas-stubs/core/series.pyi
```diff
@@ -4623,6 +4614,22 @@ class Series(IndexOpsMixin[S1], NDFrame):
...
+        **kwargs,
+    ) -> np_1darray[np.object_]: ...
+    @overload
+    def to_numpy(
+        self: Series[Interval],
+        dtype: np.dtype[GenericT] | SupportsDType[GenericT] | type[GenericT],
+        copy: bool = False,
+        na_value: Scalar = ...,
+        **kwargs,
+    ) -> np_1darray[GenericT]: ...
```
Dr-Irv: This is probably too wide.  E.g., `pd.Series[Interval].to_numpy(dtype=int)` won't work.  I think only a few of the numpy types will work.  Might be better to just leave this out?

tests/indexes/test_indexes.py
```diff
@@ -1519,12 +1524,77 @@ def test_period_index_asof_locs() -> None:
...
     dt_index = pd.DatetimeIndex(["2020-01-01", "2020-01-02"])
-    check(assert_type(dt_index.array, DatetimeArray), DatetimeArray)
+    check(assert_type(dt_index.array, DatetimeArray), DatetimeArray, pd.Timestamp)
+
+    check(
+        assert_type(pd.to_timedelta(["1s"]).array, TimedeltaArray),
+        TimedeltaArray,
+        pd.Timedelta,
+    )
+    # Should be NumpyExtensionArray
```
Dr-Irv: Not necessarily - see comment above, so remove the comment

tests/test_extension.py
Dr-Irv: I'm wondering if we can come up with a simple example that illustrates the issue with `mypy`.

Dr-Irv (changes requested):

pandas-stubs/core/series.pyi
cmp0xff: `class DtDescriptor` is now also greatly simplified. It is not a subclass anymore, nor is it generic. It also gives errors when trying to be accessed from `Series[int]`.

pandas-stubs/core/series.pyi
cmp0xff: When I remove, the following type asserting in `test_series_types_to_numpy` fails:
```py
    check(
        assert_type(i_s.to_numpy(dtype=np.bytes_), np_1darray[np.bytes_]),
        np_1darray,
        dtype=np.bytes_,
    )
```

tests/indexes/test_indexes.py
cmp0xff: The issue is that I cannot import `NumpyExtensionArray` in my python scripts. Removed in 7cd514e6338c74e0f8c42669425daaf3eed4007e.

tests/test_extension.py
cmp0xff:
```py
assert_type(pd.Series([1, "s"]), "pd.Series[Any]")  # only pyright complains that it should be `Series[int | str]`
assert_type(pd.Series([1, "s"]).array, ExtensionArray)  # only mypy complains that it should be `Any`
```

Also I cannot make `Series[Any].dt` giving error or `Never`. It may be worthy asking the `mypy` community. See 7cd514e6338c74e0f8c42669425daaf3eed4007e.

pandas-stubs/core/indexes/accessors.pyi
cmp0xff: Removed in 7cd514e6338c74e0f8c42669425daaf3eed4007e for now.

tests/series/test_properties.py
```diff
@@ -36,7 +33,9 @@ def test_dt_property() -> None:
     )
 
     if TYPE_CHECKING_INVALID_USAGE:
-        _ = pd.Series([1]).dt  # type: ignore[arg-type] # pyright: ignore[reportAttributeAccessIssue]
+        # mypy gives Any
+        _0 = pd.Series([1, "s"]).dt  # pyright: ignore[reportAttributeAccessIssue]
```
Dr-Irv: I think this might be a case of where `pyright` is incorrect and `mypy` is correct. 

Because `pyright` interprets `pd.Series([1,"s"])` as `Series[int | str]`, and then it matches the `Series[Never]`.  But `mypy` is interpreting `pd.Series([1,"s"])` as `Series[Any]`, so it can match multiple `__get__()` methods, so it comes up with `Any`.

pandas-stubs/core/indexes/accessors.pyi
```diff
@@ -434,49 +442,43 @@ class TimedeltaIndexProperties(
     _DatetimeRoundingMethods[TimedeltaIndex],
 ): ...
 
-class _dtDescriptor(CombinedDatetimelikeProperties, Generic[S1]):
+@type_check_only
+class DtDescriptor:
     @overload
-    def __get__(self, instance: Series[Never], owner: Any) -> Never: ...
+    def __get__(self, instance: Series[Never], owner: type[Series]) -> Never: ...
```
Dr-Irv: I think we should remove this.  If someone has an untyped `Series`, and they want to use a `dt` method, then they should cast the untyped `Series` to `Series[Timestamp]` or `Series[Timedelta]` to tell the type checker that they know what they are doing.

I know this is different than what we did with the arithmetic, but I think `.dt` (and `.str` for that matter) are different because they will fail if you apply them to a Series with the wrong dtype.

pandas-stubs/core/indexes/accessors.pyi
```diff
@@ -434,49 +442,43 @@ class TimedeltaIndexProperties(
...
-        ambiguous: Literal["raise", "infer", "NaT"] | bool | np_ndarray_bool = ...,
-        nonexistent: (
-            Literal["shift_forward", "shift_backward", "NaT", "raise"]
-            | timedelta
-            | Timedelta
-        ) = ...,
-    ) -> Series[S1]: ...
-    def as_unit(self, unit: TimeUnit) -> Series[S1]: ...
+        self, instance: IndexOpsMixin[Never], owner: type[IndexOpsMixin]
+    ) -> ExtensionArray: ...
```
Dr-Irv, line 465: In contrast to `.dt`, I think keeping this in makes sense.  That's because `.array` works no matter what the underlying dtype of the Series is.

pandas-stubs/core/series.pyi
Dr-Irv: So we could just add an overload to handle that specific case, because only certain numpy dtypes (`np.bytes_`, `np.str_`) would work with `Series[Interval[...]]`

tests/test_extension.py
Dr-Irv: As stated elsewhere, I actually think `mypy` may be correct here with `.array`, given the way overloads are supposed to work.  Also, we are using a subset of "strict" type checking with `pyright`, so `pd.Series([1,"s"])` becomes `Series[str | int]` with `pyright`, but `Series[Any]` with `mypy`.

But for `.dt`, I think we should not allow a `Series[Any].dt` to be successful.

Dr-Irv (changes requested):

Dr-Irv:
Looks like CI is failing because `beautifulsoup4` got upgraded.  Can you modify `pyproject.toml` to pin the version to 4.13.5 ? xref https://github.com/pandas-dev/pandas/issues/62492

pandas-stubs/core/indexes/accessors.pyi
cmp0xff: I am a bit confused. The code here was intended to disallow `df["timestamp"].dt`, because `df["timestamp"] is untyped.

- Is it a good idea to disallow such a usage? I have changed my mind, and think it can be annoying.
- The code did not work, because
   - `def __get__(self, instance: Series[Timestamp], owner: type[Series]) -> TimestampProperties: ...` also makes `Series[Any].dt` giving `TimestampProperties`
   - `def __get__(self, instance: Series[Timedelta], owner: type[Series]) -> TimedeltaProperties: ...` also makes `Series[Any].dt` giving `TimedeltaProperties`
   - `mypy` is already confused enough and gives `Any`
- It might be fine to have `.dt` giving `Any`, because you cannot chain it further, like `.dt.floor` will give you an attribute warning from the type checkers 

9eb2ed8addf9abe2eb5394243bfb179c6b071502

tests/test_extension.py
cmp0xff: - With `.array`: I have defined `.array` for all cases, and the types of all returned values are subclasses of `ExtensionArray`. I do not understand why `mypy` is confused and givies `Any`.
- With `.dt`: replied in another thread

pandas-stubs/core/series.pyi
cmp0xff: 6100ccf97d72502df309836baedb85f3cab6f039

tests/series/test_properties.py
cmp0xff: 9eb2ed8addf9abe2eb5394243bfb179c6b071502 what do you think?

tests/test_extension.py
Dr-Irv: I think with `.array`, `mypy` is *not* confused, because `Series[Any]` can match any of the `__get__()` methods in `ArrayDescriptor`, so that's why it is returning `Any`.  `pyright` may be the one that is wrong here.

pandas-stubs/core/series.pyi
Dr-Irv: Can you also add an overload for `np.str_` ?

tests/series/test_properties.py
Dr-Irv: I think I prefer the `Never`, because that forces the user to tell us which type `Series[Any]` is.

pandas-stubs/core/indexes/accessors.pyi
Dr-Irv: > mypy is already confused enough and gives Any

I actually think `mypy` could be correct, because the ambiguous result is giving `Any`.  It relates to that `pyright` discussion we had elsewhere.

> * It might be fine to have `.dt` giving `Any`, because you cannot chain it further, like `.dt.floor` will give you an attribute warning from the type checkers

If that's the case, can we add a test that shows that?  I think you can chain it further.

tests/test_extension.py
cmp0xff: What I wanted to argue was that all results are subclasses of ExtensionArray, so that they are compatible. Mypy should hence pick the first one, namely `ExtensionArray`. In reality, it seems that `mypy` does not think so.

How should we proceed from here?

Dr-Irv: This is all not clear to me how it should work from the typing spec perspective.  On the one hand, the different overloads for `__get__()` are returning different types, which all happen to be a subtype of `ExtensionArray`, so it should default to `ExtensionArray` if nothing matches.  On the other hand, everything matches with `Series[Any]` is passed, so there is a conflict and `mypy` returns `Any` (really "Unknown").  

I am fine for now to have the `#type: ignore` in the test for `mypy` .

cmp0xff: I created python/mypy#19952 and added a comment here

pandas-stubs/core/series.pyi
cmp0xff: aef9fb701098cf1592fb1f048e39e7abd4e7d196

tests/series/test_properties.py
cmp0xff: 69cb0cbfd6fda18654ea68dd9ff84d5a454dcbf9

Since `Never` is incompatible with `Properties` and its subclasses, it is not easy to make `Series[Any].dt` gives `Never` or an error, while keeping `Series[Timestamp].dt` giving a subclass of `Properties`.

If python/mypy#19952 can be resolved, I can imagine that `Series[Any].dt` can return an empty `Properties`, which contains no properties.

pandas-stubs/core/indexes/accessors.pyi
cmp0xff: 69cb0cbfd6fda18654ea68dd9ff84d5a454dcbf9

You are right, I am able to chain further. With `Any` I can chain anything. This is not too good.

pandas-stubs/core/series.pyi
```diff
@@ -834,23 +839,19 @@ class Series(IndexOpsMixin[S1], NDFrame):
...
-    def diff(self: Series[Timedelta], periods: int = ...) -> Series[Timedelta]: ...  # type: ignore[overload-overlap]
+    def diff(self: Series[Period], periods: int = ...) -> OffsetSeries: ...
     @overload
-    def diff(self: Series[Period], periods: int = ...) -> OffsetSeries: ...  # type: ignore[overload-overlap]
+    def diff(self: Series[Interval], periods: int = ...) -> Never: ...
     @overload
-    def diff(self, periods: int = ...) -> Series[float]: ...
+    def diff(
+        self: SupportsGetItem[Scalar, SupportsSelfSub[S1_CO]], periods: int = ...
+    ) -> Series[S1_CO]: ...
```
cmp0xff, line 864: Implementing the idea in https://github.com/pandas-dev/pandas-stubs/pull/1386#discussion_r2386277295

Unfortunately `mypy` thinks `Interval` can subtract with itself, and without R849-R850, `Series[Interval[int]].diff()` would give `Series[Never]` by `mypy`

pandas-stubs/core/series.pyi
```diff
@@ -834,23 +839,19 @@ class Series(IndexOpsMixin[S1], NDFrame):
...
     def autocorr(self, lag: int = 1) -> float: ...
     @overload
```
cmp0xff, line 866: Implementing the idea in https://github.com/pandas-dev/pandas-stubs/pull/1386#discussion_r2386277295

Unfortunately `mypy` thinks `Interval` can subtract with itself, and without R849-R850, `Series[Interval[int]].diff()` would give `Series[Never]` by `mypy`

tests/series/test_series.py
```diff
@@ -3503,47 +3502,39 @@ def test_diff() -> None:
...
-    # bool -> object
+    # bool -> Any
     check(
         assert_type(
-            pd.Series([True, True, False, False, True]).diff(),
-            "pd.Series[type[object]]",
+            pd.Series([True, True, False, False, True]).diff(), "pd.Series[Any]"
         ),
         pd.Series,
         object,
```
cmp0xff: - We have nowhere else `assert_type(..., "pd.Series[type[object]]")`. I believe it's better to give `pd.Series[Any]`.
- `pd.Series[bool].diff()` gives `pd.Series([float("nan"), bool, bool...])` at runtime, with dtype `object`

tests/series/test_series.py
```diff
@@ -3503,47 +3502,39 @@ def test_diff() -> None:
...
-    # object -> object
-    check(
-        assert_type(s.astype(object).diff(), "pd.Series[type[object]]"),
-        pd.Series,
-        object,
-    )
+    # Any -> float
+    s_o = s.astype(object)
+    assert_type(s_o, "pd.Series[Any]")
+    check(assert_type(s_o.diff(), "pd.Series[float]"), pd.Series, float)
```
cmp0xff, line 3529: Same argument, We have nowhere else `assert_type(..., "pd.Series[type[object]]")`

tests/series/test_properties.py
```diff
@@ -0,0 +1,75 @@
...
+            pd.period_range(start="2022-06-01", periods=10).to_series().dt,
+            PeriodProperties,
+        ),
+        PeriodProperties,
+    )
+
+    if TYPE_CHECKING_INVALID_USAGE:
+        s = pd.DataFrame({"a": [1]})["a"]
+        assert_type(s.dt, Any)
+        assert_type(s.dt.year, Any)
```
cmp0xff: Unfortunately `Any` can always chain further

pandas-stubs/core/indexes/accessors.pyi
```diff
@@ -446,7 +444,7 @@ class TimedeltaIndexProperties(
 @type_check_only
 class DtDescriptor:
     @overload
-    def __get__(self, instance: Series[Period], owner: Any) -> PeriodProperties: ...
+    def __get__(self, instance: Series[Never], owner: type[Series]) -> Any: ...
```
Dr-Irv: I thought that we'd change this to `Never` ?

pandas-stubs/core/series.pyi
Dr-Irv: Do we have a test that causes `Series[Interval[int]].__sub__(Series[Interval[int]])` to be rejected by `mypy`?  (without using `Never`

tests/series/test_properties.py
Dr-Irv: So can we change it to `Never` ?

tests/series/test_series.py
```diff
@@ -3503,47 +3502,38 @@ def test_diff() -> None:
             BaseOffset,
             index_to_check_for_type=-1,
         )
-    # bool -> object
+    # bool -> Any
     check(
         assert_type(
-            pd.Series([True, True, False, False, True]).diff(),
-            "pd.Series[type[object]]",
+            pd.Series([True, True, False, False, True]).diff(), "pd.Series[Any]"
```
Dr-Irv, line 3506: If you do
```python
pd.Series([True, True, False, False, True], dtype=pd.BooleanDtype()).diff()
```
you get a `Series` with `pd.BooleanDtype`.  (displays as `"boolean"`)

I'd rather support the nullable dtype version, so can we make this return `pd.Series[bool]` instead?

tests/series/test_series.py
Dr-Irv: see other comment - with the `BooleanDtype`, it gives dtype `boolean`, so let's support that instead.

tests/series/test_series.py
Dr-Irv: OK, I'm fine with that here

tests/test_extension.py
Dr-Irv: Thanks for coming up with a simple example

pandas-stubs/core/indexes/accessors.pyi
Dr-Irv: Yes, so we should figure out how to deal with that.  Maybe return `Never` or an empty properties class.

tests/series/test_properties.py
Dr-Irv: I guess the other option is that we make it work for pyright, and know that it is incorrect for mypy.  Using either `Never` or an empty `Properties` .

cmp0xff: I am using `Properties` now in 0a9f5a2dd48d67bbdc71df5a14af1fa65dc5543c. Unfortunately python/mypy#19952 is closed, I'll try to raise my concern again.

pandas-stubs/core/indexes/accessors.pyi
cmp0xff: I am using `Properties` now in 0a9f5a2dd48d67bbdc71df5a14af1fa65dc5543c. Unfortunately python/mypy#19952 is closed, I'll try to raise my concern again.

pandas-stubs/core/series.pyi
cmp0xff: Added in 0a9f5a2dd48d67bbdc71df5a14af1fa65dc5543c

tests/series/test_series.py
cmp0xff: Can we deal with the nullable and non-nullable cases separately? 0a841c77f43c119ee868d0c2e9bebf0afbc4258f pandas-dev/pandas-stubs#1395

tests/series/test_series.py
cmp0xff: Can we deal with the nullable and non-nullable cases separately? 0a841c77f43c119ee868d0c2e9bebf0afbc4258f pandas-dev/pandas-stubs#1395

tests/series/test_properties.py
cmp0xff: I am using `Properties` now in 0a9f5a2dd48d67bbdc71df5a14af1fa65dc5543c. Unfortunately python/mypy#19952 is closed, I'll try to raise my concern again.

tests/series/test_properties.py
Dr-Irv: I think `mypy` is right - the overload rules mean that `Any` is the correct resolution.  This relates to our discussion here:  https://github.com/pandas-dev/pandas-stubs/pull/1274#issuecomment-3282517624

Dr-Irv (commented):
This is all looking good, but you need to resolve conflicts  - probably with `to_series()`, since that was updated in another PR.  And now `OffsetSeries` is gone, so have to reconcile that as well.

PR #1386, opened 2025-09-28
GH1379 Remove PeriodSeries

### LINKED ISSUE #1379: Remove PeriodSeries, OffsetSeries, IntervalSeries

loicdiridollou:
- [x] Closes #1379 partially

### DIFF
--- a/pandas-stubs/core/series.pyi
+++ b/pandas-stubs/core/series.pyi
@@ -70,10 +70,7 @@ from pandas.core.generic import NDFrame
 from pandas.core.groupby.generic import SeriesGroupBy
 from pandas.core.groupby.groupby import BaseGroupBy
 from pandas.core.indexers import BaseIndexer
-from pandas.core.indexes.accessors import (
-    PeriodProperties,
-    _dtDescriptor,
-)
+from pandas.core.indexes.accessors import _dtDescriptor
 from pandas.core.indexes.category import CategoricalIndex
 from pandas.core.indexes.datetimes import DatetimeIndex
 from pandas.core.indexes.interval import IntervalIndex
@@ -363,7 +360,7 @@ class Series(IndexOpsMixin[S1], NDFrame):
         dtype: PeriodDtype = ...,
         name: Hashable = ...,
         copy: bool = ...,
-    ) -> PeriodSeries: ...
+    ) -> Series[Period]: ...
     @overload
     def __new__(
         cls,
@@ -850,6 +847,8 @@ class Series(IndexOpsMixin[S1], NDFrame):
     def diff(self: Series[Timestamp], periods: int = ...) -> Series[Timedelta]: ...  # type: ignore[overload-overlap]
     @overload
     def diff(self: Series[Timedelta], periods: int = ...) -> Series[Timedelta]: ...  # type: ignore[overload-overlap]
+    @overload
+    def diff(self: Series[Period], periods: int = ...) -> OffsetSeries: ...  # type: ignore[overload-overlap]
     @overload
     def diff(self, periods: int = ...) -> Series[float]: ...
     def autocorr(self, lag: int = 1) -> float: ...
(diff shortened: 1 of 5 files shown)


pandas-stubs/core/series.pyi
```diff
@@ -851,6 +848,8 @@ class Series(IndexOpsMixin[S1], NDFrame):
     @overload
     def diff(self: Series[Timedelta], periods: int = ...) -> Series[Timedelta]: ...  # type: ignore[overload-overlap]
     @overload
+    def diff(self: Series[Period], periods: int = ...) -> OffsetSeries: ...  # type: ignore[overload-overlap]
+    @overload
```
cmp0xff, line 852: Could try the following idea to simplify, if you want.
```py
class SupportsSelfSub(Protocol[_T_co]):
    def __sub__(self, x: Self, /) -> _T_co: ...
```
And
```py
def diff(self: SupportsGetItem[Scalar, SupportsSelfSub[S1_CO]], period: int = ...) -> Series[S1_CO]: ...
```
I've first seen the idea in pandas-dev/pandas-stubs#1362 and it worked again in pandas-dev/pandas-stubs#1374

pandas-stubs/core/series.pyi
```diff
@@ -3201,6 +3200,8 @@ class Series(IndexOpsMixin[S1], NDFrame):
         ),
     ) -> Series[Timedelta]: ...
     @overload
+    def __sub__(self: Series[Period], other: Series[Period]) -> OffsetSeries: ...
```
cmp0xff: TLDR:

- Use `-> Series[Offset]`
- Please also add `def __rsub__(self: Series[Period], other: Series[Period]) -> OffsetSeries: ...` and the corresponding `sub` and `rsub`.
- Would be great if `tests/series/period/test_sub.py` can be added

---

This one is the cause of https://github.com/pandas-dev/pandas-stubs/pull/1386#issuecomment-3343676915. I'll try to give my reasoning.

Note that `self: Series[Period]` implies it also accepts `self: Series[Any]`. The same applies to `other: Series[Period]`. `mypy` checks if the existing rules are "compatible" with the current result. Existing rules must have given `Series[...]`, whereas the current result is `OffsetSeries`. `mypy` thinks this is _in_compatible, hence gives `Any`.

That `pyright` does not give `Any`, could be a `pyright` bug. See microsoft/pyright#10924

Dr-Irv: Definitely changing the return type to `Series[BaseOffset]` fixes that `mypy` problem.  But there is a problem with `test_timefuncs.py` and `PeriodProperties` and `pyright` that I am trying to figure out.

```suggestion
    def __sub__(self: Series[Period], other: Series[Period]) -> Series[BaseOffset]: ...
```

Dr-Irv (changes requested):
I got the `test_timefuncs.py` to work by doing this change in `series.pyi`:

```python
    dt = _dtDescriptor[S1]()  # type: ignore[misc]
```
and removing the `@property` and `def dt` for `Series`

Also, pre-commit is failing, so make sure to update your local env before pushing.

cmp0xff:
> I got the `test_timefuncs.py` to work by doing this change in `series.pyi`:
> 
> ```python
>     dt = _dtDescriptor[S1]()  # type: ignore[misc]
> ```
> 
> and removing the `@property` and `def dt` for `Series`

so this is the same changes as I have been doing in pandas-dev/pandas-stubs#1385. Maybe @loicdiridollou can manually cherry-pick 🍒⛏  the relevant parts to reduce rebasing.

In the end, the `S1` in the descriptor does not seem to be necessary, either.

loicdiridollou:
Thanks for the suggestion, I tried to add some tests into a new file according to @cmp0xff suggestion, we may be able to consolidate better in the future but this is a start.

pandas-stubs/core/series.pyi
```diff
@@ -1233,8 +1233,7 @@ class Series(IndexOpsMixin[S1], NDFrame):
         Series[_str],
         Series,
     ]: ...
-    @property
-    def dt(self) -> _dtDescriptor[S1]: ...
+    dt = _dtDescriptor[S1]()  # type: ignore[misc]
```
Dr-Irv: @cmp0xff indicated that the `[S1]` here may not be necessary (and then the `#type: ignore` isn't needed either).   Can you test that?

loicdiridollou: Yes that is correct it works, I had missed the comment

Dr-Irv (approved):
thanks @loicdiridollou

PR #1390, opened 2025-09-30
GH1379 Drop OffsetSeries replacing it with Series[BaseOffset]

### LINKED ISSUE #1379: Remove PeriodSeries, OffsetSeries, IntervalSeries

loicdiridollou:
- [x] Closes #1379

Also dropped the locking of the version of `beautifulsoup`, it seems like they shipped another version that fixed the issue.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -65,7 +65,7 @@ scipy = { version = ">=1.9.1", python = "<3.14" }
 scipy-stubs = ">=1.15.3.0"
 SQLAlchemy = ">=2.0.39"
 types-python-dateutil = ">=2.8.19"
-beautifulsoup4 = "<=4.13.5"
+beautifulsoup4 = ">=4.13.5"
 html5lib = ">=1.1"
 python-calamine = ">=0.2.0"
 
--- a/tests/test_timefuncs.py
+++ b/tests/test_timefuncs.py
@@ -52,7 +52,6 @@ from pandas.tseries.offsets import (
 if TYPE_CHECKING:
     from pandas.core.series import (
         IntervalSeries,
-        OffsetSeries,
     )
 
 if not PD_LTE_23:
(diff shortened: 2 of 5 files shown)


pyproject.toml
```diff
@@ -65,7 +65,7 @@ scipy = { version = ">=1.9.1", python = "<3.14" }
 scipy-stubs = ">=1.15.3.0"
 SQLAlchemy = ">=2.0.39"
 types-python-dateutil = ">=2.8.19"
-beautifulsoup4 = "<=4.13.5"
+beautifulsoup4 = ">=4.13.5"
```
Dr-Irv:
```suggestion
beautifulsoup4 = ">=4.14.2"
```

tests/test_timefuncs.py
```diff
@@ -947,7 +946,8 @@ def test_series_types_to_numpy() -> None:
     ts_s = pd.to_datetime(pd.Series(["2020-01-01", "2020-01-02"]))
     p_s = pd.Series(pd.period_range("2012-1-1", periods=10, freq="D"))
     o_s = cast(
-        "OffsetSeries", pd.Series([pd.DateOffset(days=1), pd.DateOffset(days=2)])
+        "pd.Series[BaseOffset]",
+        pd.Series([pd.DateOffset(days=1), pd.DateOffset(days=2)]),
     )
```
Dr-Irv: Can you see if we can create a `Series.__new__()` overload that would accept a sequence of `BaseOffset` so that we don't need the cast here?

Dr-Irv (changes requested):

tests/test_timefuncs.py
loicdiridollou: So turns out this creates a `Series[DateOffset]` and not `Series[BaseOffset]` that is why this is an issue, not sure why though it is not working since it is supposed to inherit from BaseOffset.

loicdiridollou: So I checked and seems like `Series[DateOffset]` does not match with `Series[BaseOffset]` even though `DateOffset` is always a `BaseOffset`:
`assert_type(o_s, "pd.Series[BaseOffset]") ■ "assert_type" mismatch: expected "Series[BaseOffset]" but received "Series[DateOffset]"`

Dr-Irv: I think you need to add an overload for `Series.__new__()` that accepts a sequence of `BaseOffset` and then it should work.   Similar to how there are overloads that produce `Series[Timestamp]`, `Series[Timedelta]`, etc.

loicdiridollou: Removed the to_numpy overloads and added the __new__ overloads and it is looking good!

PR #1391, opened 2025-09-30
GH1383 Add support for creating Series/Index with dtype='category'

### LINKED ISSUE #1383: `Index([1], dtype="category")` is typed as `Index[Any]`
## Describe the bug
`Index([1], dtype="category")` gives `CategoricalIndex([1], categories=[1], ordered=False, dtype='category')` at runtime, but is typed as `Index[Any]`

### minimal runnable example

```py

import pandas as pd

reveal_type(pd.Index([1], dtype="category"))
```
 
### Which type checker

both

#### `pyright`

```text
Type of "Index([1], dtype="category")" is "Index[Any]"
```

#### `mypy`

```text
Revealed type is "pandas.core.indexes.base.Index[Any]"
```

## Information
 - OS: Linux
 - OS Version: Debian GNU/Linux 12 (bookworm)
 - python version: 3.12.11
 - version of type checker: mypy 1.18.2, pyright 1.1.405
 - version of installed `pandas-stubs`: `main` branch

**Additional context**

loicdiridollou:
- [x] Closes #1383


tests/series/test_properties.py
```diff
@@ -57,9 +56,7 @@ def test_array_property() -> None:
     # casting due to pandas-dev/pandas-stubs#1383
```
cmp0xff: Please remove thie comment

cmp0xff (changes requested):
> `s.astype("dictionary[pyarrow]")` is `Unknown`, not sure if we have the correct overload for it. Also when I try to do it at run time I get a `TypeError` so wondering what is the goal of having that only for type checking and not run time. Open to suggestions

I see, those commented-out tests do not make sense to me, either.

One small thing, otherwise good to go.

cmp0xff (approved):
It looks good to me now, but you have not yet requested me to review again. Would you agree to merge @loicdiridollou ?

PR #1394, opened 2025-09-30
type `freq` in `shift`, consistently use `Frequency` alias

MarcoGorelli:

### DIFF
--- a/pandas-stubs/_libs/tslibs/timedeltas.pyi
+++ b/pandas-stubs/_libs/tslibs/timedeltas.pyi
@@ -23,12 +23,12 @@ from typing_extensions import (
 )
 
 from pandas._libs.tslibs import (
-    BaseOffset,
     NaTType,
 )
 from pandas._libs.tslibs.period import Period
 from pandas._libs.tslibs.timestamps import Timestamp
 from pandas._typing import (
+    Frequency,
     ShapeT,
     TimeUnit,
     np_1darray,
@@ -130,9 +130,9 @@ class Timedelta(timedelta):
     @property
     def asm8(self) -> np.timedelta64: ...
     # TODO: round/floor/ceil could return NaT?
-    def round(self, freq: str | BaseOffset) -> Self: ...
-    def floor(self, freq: str | BaseOffset) -> Self: ...
-    def ceil(self, freq: str | BaseOffset) -> Self: ...
+    def round(self, freq: Frequency) -> Self: ...
+    def floor(self, freq: Frequency) -> Self: ...
+    def ceil(self, freq: Frequency) -> Self: ...
     @property
     def resolution_string(self) -> str: ...
     # Override due to more types supported than dt.timedelta
--- a/pandas-stubs/core/frame.pyi
+++ b/pandas-stubs/core/frame.pyi
@@ -106,6 +106,7 @@ from pandas._typing import (
     FillnaOptions,
     FloatFormatType,
     FormattersType,
+    Frequency,
     GroupByObjectNonScalar,
     HashableT,
     HashableT1,
@@ -2341,7 +2342,7 @@ class DataFrame(NDFrame, OpsMixin, _GetItemHack):
     @overload
     def rolling(
         self,
-        window: int | str | dt.timedelta | BaseOffset | BaseIndexer,
+        window: int | Frequency | dt.timedelta | BaseIndexer,
         min_periods: int | None = ...,
         center: _bool = ...,
         on: Hashable | None = ...,
@@ -2355,7 +2356,7 @@ class DataFrame(NDFrame, OpsMixin, _GetItemHack):
     @overload
     def rolling(
         self,
-        window: int | str | dt.timedelta | BaseOffset | BaseIndexer,
+        window: int | Frequency | dt.timedelta | BaseIndexer,
         min_periods: int | None = ...,
         center: _bool = ...,
         on: Hashable | None = ...,
--- a/pandas-stubs/core/indexes/accessors.pyi
+++ b/pandas-stubs/core/indexes/accessors.pyi
@@ -33,10 +33,10 @@ from pandas.core.series import (
 from typing_extensions import Never
 
 from pandas._libs.tslibs import BaseOffset
-from pandas._libs.tslibs.offsets import DateOffset
 from pandas._libs.tslibs.period import Period
 from pandas._typing import (
     S1,
+    Frequency,
     TimeAmbiguous,
     TimeNonexistent,
     TimestampConvention,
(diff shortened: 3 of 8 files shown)


pandas-stubs/core/indexes/accessors.pyi
```diff
@@ -357,12 +355,12 @@ class _PeriodProperties(
     def strftime(self, date_format: str) -> _PeriodStrReturnTypes: ...
     def to_timestamp(
         self,
-        freq: str | DateOffset | None = ...,
+        freq: Frequency | None = ...,
```
loicdiridollou: Okay so here is a good example:
```
idx = pd.PeriodIndex(["2023", "2024", "2025"], freq="D")
s1 = pd.Series([1, 2, 3], index=idx)
s1.to_timestamp(freq=pd.tseries.offsets.BDay())
```

This will raise a `FutureWarning`: 
```
<stdin>:1: FutureWarning: PeriodDtype[B] is deprecated and will be removed in a future version. Use a DatetimeIndex with freq='B' instead
```

The issue is that `BDay` is not a DateOffset. I think it is a bit of an edge case and this situation works at runtime for now but good to think about the future, I am good with letting the types a bit looser here, @Dr-Irv what do you think?

MarcoGorelli:
thanks for your review

there's quite a few places where `PeriodDtype[B]` can't be used, so they should probably all use `DateOffset | str`

MarcoGorelli:
Conversely, there are some parts which currently are annotated with `BaseOffset` but should probably be `DateOffset`

```python
In [39]: pd.Period('2020').asfreq(pd.offsets.BusinessDay())
<ipython-input-39-1953d30c79b5>:1: FutureWarning: Period with BDay freq is deprecated and will be removed in a future version. Use a DatetimeIndex with BDay freq instead.
  pd.Period('2020').asfreq(pd.offsets.BusinessDay())
Out[39]: Period('2020-12-31', 'B')
```

If BusinessDay is the only offset that this catches, then for what it's worth I don't think it's worth distinguishing them

pandas-stubs/core/frame.pyi
```diff
@@ -1683,14 +1684,14 @@ class DataFrame(NDFrame, OpsMixin, _GetItemHack):
     ) -> Self: ...
     def to_timestamp(
         self,
-        freq=...,
+        freq: PeriodFrequency | None = None,
```
MarcoGorelli, line 1685: i've always found bizarre that `to_timestamp` only accepts period frequencies
```python
In [4]: pd.Period('2020').to_timestamp('MS')
---------------------------------------------------------------------------
AttributeError                            Traceback (most recent call last)
Cell In[4], line 1
----> 1 pd.Period('2020').to_timestamp('MS')

File ~/scratch/.39venv/lib/python3.9/site-packages/pandas/_libs/tslibs/period.pyx:1868, in pandas._libs.tslibs.period._Period.to_timestamp()

AttributeError: 'pandas._libs.tslibs.offsets.MonthBegin' object has no attribute '_period_dtype_code'
```
(and, on newer versions of pandas):
```python
In [2]: pd.Period('2020').to_timestamp('MS')
---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
File pandas/_libs/tslibs/offsets.pyx:5363, in pandas._libs.tslibs.offsets.to_offset()

File pandas/_libs/tslibs/offsets.pyx:5246, in pandas._libs.tslibs.offsets._validate_to_offset_alias()

ValueError: for Period, please use 'M' instead of 'MS'

The above exception was the direct cause of the following exception:

ValueError                                Traceback (most recent call last)
Cell In[2], line 1
----> 1 pd.Period('2020').to_timestamp('MS')

File pandas/_libs/tslibs/period.pyx:2041, in pandas._libs.tslibs.period._Period.to_timestamp()

File pandas/_libs/tslibs/period.pyx:1771, in pandas._libs.tslibs.period._Period._maybe_convert_freq()

File pandas/_libs/tslibs/offsets.pyx:5402, in pandas._libs.tslibs.offsets.to_offset()

ValueError: Invalid frequency: MS, failed to parse with error message: ValueError("for Period, please use 'M' instead of 'MS'")
```

but whatever, if that's how it is then it should at least be typed to respect that 🤷

pandas-stubs/core/frame.pyi
```diff
@@ -2224,7 +2225,7 @@ class DataFrame(NDFrame, OpsMixin, _GetItemHack):
         self,
         periods: int = 1,
         fill_method: None = None,
-        freq: DateOffset | dt.timedelta | _str | None = ...,
+        freq: Frequency | dt.timedelta | None = ...,
```
MarcoGorelli, line 2226: I think the typing here was too tight to begin with
```
In [6]: s = pd.Series([1,2,3], index=pd.date_range('2020', freq='MS', periods=3))

In [7]: s.pct_change(freq='MS')
Out[7]:
2020-01-01    NaN
2020-02-01    1.0
2020-03-01    0.5
Freq: MS, dtype: float64
```

pandas-stubs/tseries/frequencies.pyi
```diff
@@ -8,10 +8,10 @@ from pandas import (
...
 from pandas._typing import Frequency
 
-from pandas.tseries.offsets import DateOffset
+from pandas.tseries.offsets import BaseOffset
 
 @overload
 def to_offset(freq: None, is_period: bool = ...) -> None: ...
 @overload
-def to_offset(freq: Frequency, is_period: bool = ...) -> DateOffset: ...
+def to_offset(freq: Frequency, is_period: bool = ...) -> BaseOffset: ...
```
MarcoGorelli, line 16: this is capable of returning frequencies beyond just the `DateOffset` ones
```
In [8]: pd.tseries.frequencies.to_offset('B')
Out[8]: <BusinessDay>
```

pandas-stubs/_libs/tslibs/timedeltas.pyi
```diff
@@ -130,9 +130,9 @@ class Timedelta(timedelta):
     @property
     def asm8(self) -> np.timedelta64: ...
     # TODO: round/floor/ceil could return NaT?
-    def round(self, freq: str | BaseOffset) -> Self: ...
-    def floor(self, freq: str | BaseOffset) -> Self: ...
-    def ceil(self, freq: str | BaseOffset) -> Self: ...
+    def round(self, freq: Frequency) -> Self: ...
+    def floor(self, freq: Frequency) -> Self: ...
+    def ceil(self, freq: Frequency) -> Self: ...
```
MarcoGorelli, line 137: docs say `str` but offset objects are also accepted, so I think it's ok to keep the current annotation
```python
In [11]: pd.Timedelta(minutes=37).round(pd.tseries.offsets.Hour())
Out[11]: Timedelta('0 days 01:00:00')
```

loicdiridollou: Maybe worth a question in the pandas repo so we can make sure we are not allowing a behavior that is not fully supported. Otherwise the rest looks good, @Dr-Irv do you see anything else prior to merging?

Dr-Irv: @loicdiridollou I'm going to trust your review process rather than having me look at it!

PR #1401, opened 2025-10-03
remove FrozenList, type Grouper

MarcoGorelli:
`FrozenlList` isn't public, and doesn't appear in any other types. OK to just remove it?


pandas-stubs/core/groupby/grouper.pyi
```diff
@@ -47,7 +48,7 @@ class Grouper:
         dropna: bool = ...,
     ) -> Self: ...
     @overload
-    def __new__(cls, *args, freq: Frequency, **kwargs) -> TimeGrouper: ...
+    def __new__(cls, *args: Any, freq: Frequency, **kwargs: Any) -> TimeGrouper: ...
```
cmp0xff, line 51: I don't see that we always add `Any` throughout this repo - shall we do so, @Dr-Irv ?

It looks better for me, though.

pandas-stubs/api/typing/__init__.pyi
```diff
@@ -2,7 +2,6 @@ from pandas.core.groupby import (
     DataFrameGroupBy as DataFrameGroupBy,
     SeriesGroupBy as SeriesGroupBy,
 )
-from pandas.core.indexes.frozen import FrozenList as FrozenList
```
cmp0xff, line 5: From https://github.com/pandas-dev/pandas-stubs?tab=readme-ov-file#differences-between-type-declarations-in-pandas-and-pandas-stubs:

> The <https://github.com/pandas-dev/pandas-stubs/> project provides type declarations for the pandas _public_ API.

Hence we __should__ remove `FrozenList` which belongs to the [Internals](https://pandas.pydata.org/docs/development/internals.html).

pandas-stubs/core/groupby/grouper.pyi
Dr-Irv: > I don't see that we always add `Any` throughout this repo - shall we do so, @Dr-Irv ?

For `**kwargs` and `*args`, it's OK.

PR #1403, opened 2025-10-03
refactor: comply to more ruff and pyright rules

### LINKED ISSUE #950: `df.drop` signature does not capture all constraints

cmp0xff:
- This also removes `reportUnusedVariable` errors in the pyright strict test
- Closes #950: fixes a remaining issue in `test_arguments_drop` from there


pyproject.toml
```diff
@@ -138,7 +138,7 @@ args = [
 
 
 [tool.black]
-target-version = ['py310']
+target-version = ['py39']
```
Dr-Irv: let's leave this at 3.10.  pandas is dropping support for 3.9 soon

pyproject.toml
```diff
@@ -168,6 +168,25 @@ force_grid_wrap = 2
 force_sort_within_sections = true
 skip_glob = "env"
 
+
+[tool.ruff]
+target-version = "py39"
```
Dr-Irv: should be 310

Dr-Irv (changes requested):
If it's the case that this makes `reportUnusedVariable` go away in `strict`, then you can delete the line in `pyproject.toml` that refers to it.

pyproject.toml
cmp0xff: 313f07d6f90d1475d8e1b2a3f3fbbaafd0a1a7ab

pyproject.toml
cmp0xff: 313f07d6f90d1475d8e1b2a3f3fbbaafd0a1a7ab

cmp0xff (commented):
The changes are bigger now because two pyright rules are dropped

Dr-Irv (approved):
thanks @cmp0xff 

Seems like most of the changes are moving from 3.9 compliant code to 3.10.   Did you use a tool to make that easier to do?

cmp0xff:
> Seems like most of the changes are moving from 3.9 compliant code to 3.10. Did you use a tool to make that easier to do?

Hi @Dr-Irv, I used regex to replace in one case, otherwise I used a brute-force approach.

Issue #1171, opened 2025-03-13
Strict report as of 3/13/2025


Dr-Irv:
When we do a merge, we run pyright using the `strict` option.  Current statistics on the types of errors found:
```text
reportUnknownMemberType       2387
reportUnknownParameterType    2112
reportMissingTypeArgument     1560 
reportMissingParameterType    1247
reportUnknownArgumentType      613
reportUnknownVariableType      428
reportUnusedVariable            88
reportUnknownLambdaType         58
reportUnusedExpression          41
reportPrivateUsage              15
reportMissingTypeStubs           7
```
Can be output by running `poe test pyright_strict`

There are 8556 errors, and 3541 of them are in the stubs themselves.  Rest are reported from the tests (but may be due to the stubs).

FYI for @MarcoGorelli 

Some simple things that would help:

- Change references to `npt.NDArray` to be `npt.NDArray[Any, Any]`
- Change `args` and `kwargs`  (see #1168) in all stubs file
- Maybe change references from plain `Series` to `UnknownSeries` and change from plain `Index` to `UnknownIndex` in all stubs file

PR #1405, opened 2025-10-04
chore(pyright): #1171 return None when possible; disable reportUnknownLambdaType

cmp0xff:
- [x] Addresses #1171.
   - `reportPrivateUsage` 42. Can we disable this forever? Importing `_PrivateObjects` is widely used in Pandas. We cannot remove them.
   - `reportUnknownLambdaType`: 50. Can we disable this forever? They are only in tests at places where one can insert a callable.
   - `reportUnknownArgumentType`: 118.
   - `reportUnknownVariableType`: 405.
   - `reportMissingTypeArgument`: 430.
   - `reportMissingParameterType`: 799.
   - `reportUnknownParameterType`: 1235.
   - `reportUnknownMemberType`: 1411.

If one wants to investigate the errors individually, here's the JSON snippet that can be inserted to `pyrightconfig-strict.json`. Simply enable the specific categories of error of interest.

```json
{
  "reportUnknownArgumentType": false,
  "reportUnknownVariableType": false,
  "reportMissingTypeArgument": false,
  "reportMissingParameterType": false,
  "reportUnknownParameterType": false,
  "reportUnknownMemberType": false
}
```


pyrightconfig-strict.json
```diff
@@ -5,5 +5,7 @@
   "enableTypeIgnoreComments": false,
   "reportUnnecessaryTypeIgnoreComment": true,
   "reportMissingModuleSource": true,
-  "useLibraryCodeForTypes": false
+  "useLibraryCodeForTypes": false,
+  "reportPrivateUsage": false,
```
Dr-Irv: I think we need to keep `reportPrivateUsage` in there.  I think pyright is wrong for reporting them in stub files.  Created an issue:  https://github.com/microsoft/pyright/issues/10992

From the test files, we will be able to fix this once pandas 3.0 is released, where there will be a `pandas.api.typing.aliases` which will be the supported way for people to import instead of using `pandas._typing`.  One reason I introduced the aliases was to get around this particular warning.

cmp0xff: Added back in 7d3c15267921ad9661a38250de5a360becf62741

cmp0xff (commented):
- Also added `-> None` to tests and other possible places
- `pyright_strict` errors reduced from 3732 to 3682.

Dr-Irv (approved):
thanks @cmp0xff .  With respect to the python functions in `.py` files returning `None`, is there some test we can include for new PRs (in pre-commit, or in the type checkers) that would pick that up?

cmp0xff:
We are very far from making all ANN rules work.

Alternatively I can propose that we `select` `"ALL"` ruff rules and exclude not working rules, which we gradually fix and remove (or decide not to support)

PR #1412, opened 2025-10-07
refactor(ruff): switch from `select` to `ignore`

cmp0xff:
`ruff` rules usually improve the code quality. By specifying `ignore` rather than `select`, we get better hints how to comply to more rules.

In many cases this reduces errors in `pyright` strict, hence contributing to pandas-dev/pandas-stubs#1171.

For example, if we comply to [`ANN201`](https://docs.astral.sh/ruff/rules/missing-return-type-undocumented-public-function/), test functions will be foreced to have `-> None`, which reduces `pyright` strict errors. This was discussed in https://github.com/pandas-dev/pandas-stubs/pull/1405#issuecomment-3368485010.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -176,16 +176,48 @@ fix = true
 
 
 [tool.ruff.lint]
-extend-select = ["B007", "B018", "PYI", "UP", "RUF100", "RUF059"]
+extend-select = ["ALL"]
 ignore = [
-  "E501",   # https://docs.astral.sh/ruff/rules/line-too-long/
-  "E731",   # https://docs.astral.sh/ruff/rules/lambda-assignment/
-  "PYI042", # https://docs.astral.sh/ruff/rules/snake-case-type-alias/
+  # The following rules are ignored permanently for good reasons
+  "COM",     # https://docs.astral.sh/ruff/rules/#flake8-commas-com
+  "D",       # https://docs.astral.sh/ruff/rules/#pydocstyle-d
+  "E501",    # handled by black https://docs.astral.sh/ruff/rules/line-too-long/
+  "EM",      # https://docs.astral.sh/ruff/rules/#flake8-errmsg-em
+  "FBT",     # https://docs.astral.sh/ruff/rules/#flake8-boolean-trap-fbt
+  "I",       # handled by isort
 ]
 
 
 [tool.ruff.lint.per-file-ignores]
-"_*.pyi" = ["PYI001"]
+"*.pyi" = [
+  # The following rules are ignored permanently for good reasons
+  "A001",    # https://docs.astral.sh/ruff/rules/builtin-variable-shadowing/
+  "A002",    # https://docs.astral.sh/ruff/rules/builtin-argument-shadowing/
+  "A004",    # https://docs.astral.sh/ruff/rules/builtin-import-shadowing/
+  "N",       # https://docs.astral.sh/ruff/rules/#pep8-naming-n
+  "PYI001",  # https://docs.astral.sh/ruff/rules/unprefixed-type-param/
+  "PYI042",  # https://docs.astral.sh/ruff/rules/snake-case-type-alias/
+  "TD002",   # https://docs.astral.sh/ruff/rules/missing-todo-author/
+  # The following rules are ignored temporarily. Either fix them or move to the permanent list above.
+  "ERA", "S", "ANN", "FIX002", "TD003", "TD004", "PLR0402", "PLC0105"
+  ]
+"scripts/*" = [
+  # The following rules are ignored permanently for good reasons
+  "S603",    # https://docs.astral.sh/ruff/rules/subprocess-without-shell-equals-true/
+  # The following rules are ignored temporarily. Either fix them or move to the permanent list above.
+  "TRY", "PT", "BLE"
+]
+"tests/*" = [
+  # The following rules are ignored permanently for good reasons
+  "E731",    # https://docs.astral.sh/ruff/rules/lambda-assignment/
+  "PD",      # Pandas is here
+  "PLC0415", # https://docs.astral.sh/ruff/rules/import-outside-top-level/
+  "S101",    # https://docs.astral.sh/ruff/rules/assert/
+  "S301",    # https://docs.astral.sh/ruff/rules/suspicious-pickle-usage/
+  "SLF001",  # https://docs.astral.sh/ruff/rules/private-member-access/
+  # The following rules are ignored temporarily. Either fix them or move to the permanent list above.
+  "ANN", "ARG", "ERA", "RUF", "SIM", "TRY", "PT", "NPY", "N", "DTZ", "PLR", "TC", "PGH", "PTH", "S311", "C901", "FIX", "TD", "A001", "PYI042", "ANN201"
+]
 
 
 [tool.mypy]
--- a/tests/__init__.py
+++ b/tests/__init__.py
@@ -87,8 +87,7 @@ def check(
             and get_origin(shape_type) is tuple
             and (tuple_args := get_args(shape_type))
             and ... not in tuple_args  # fixed-length tuple
-            and (arr_ndim := getattr(actual, "ndim"))
-            != (expected_ndim := len(tuple_args))
+            and (arr_ndim := actual.ndim) != (expected_ndim := len(tuple_args))
         ):
             raise RuntimeError(
                 f"Array has wrong dimension {arr_ndim}, expected {expected_ndim}"
@@ -100,7 +99,7 @@ def check(
             and (dtype_args := get_args(dtype_type))
             and isinstance((expected_dtype := dtype_args[0]), type)
             and issubclass(expected_dtype, np.generic)
-            and (arr_dtype := getattr(actual, "dtype")) != expected_dtype
+            and (arr_dtype := actual.dtype) != expected_dtype
         ):
             raise RuntimeError(
                 f"Array has wrong dtype {arr_dtype}, expected {expected_dtype.__name__}"
@@ -208,8 +207,6 @@ def pytest_warns_bounded(
         current = Version(version_str)
     if lb < current < ub:
         return pytest.warns(warning, match=match)
-    else:
-        if upper_exception is None:
-            return nullcontext()
-        else:
-            return suppress(upper_exception)
+    if upper_exception is None:
+        return nullcontext()
+    return suppress(upper_exception)
--- a/tests/__init__.py
+++ b/tests/__init__.py
@@ -87,7 +87,7 @@ def check(
             and get_origin(shape_type) is tuple
             and (tuple_args := get_args(shape_type))
             and ... not in tuple_args  # fixed-length tuple
-            and (arr_ndim := actual.ndim) != (expected_ndim := len(tuple_args))
+            and (arr_ndim := actual.ndim) != (expected_ndim := len(tuple_args))  # type: ignore[attr-defined]  # pyright: ignore[reportAttributeAccessIssue]
         ):
             raise RuntimeError(
                 f"Array has wrong dimension {arr_ndim}, expected {expected_ndim}"
@@ -99,7 +99,7 @@ def check(
             and (dtype_args := get_args(dtype_type))
             and isinstance((expected_dtype := dtype_args[0]), type)
             and issubclass(expected_dtype, np.generic)
-            and (arr_dtype := actual.dtype) != expected_dtype
+            and (arr_dtype := actual.dtype) != expected_dtype  # type: ignore[attr-defined]  # pyright: ignore[reportAttributeAccessIssue]
         ):
             raise RuntimeError(
                 f"Array has wrong dtype {arr_dtype}, expected {expected_dtype.__name__}"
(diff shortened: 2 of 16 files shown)


tests/extension/decimal/array.py
```diff
@@ -270,7 +266,7 @@ def convert_values(param):
 
         # If the operator is not defined for the underlying objects,
         # a TypeError should be raised
-        res = [op(a, b) for (a, b) in zip(lvalues, rvalues)]
+        res = [op(a, b) for (a, b) in zip(lvalues, rvalues, strict=False)]
```
Dr-Irv: This is annoying.  `strict=False` is the default, so we shouldn't have to specify it.  Can you ignore that rule with ruff?

tests/indexes/test_indexes.py
```diff
@@ -120,7 +120,9 @@ def test_multiindex_constructors() -> None:
         pd.MultiIndex,
     )
     check(
-        assert_type(pd.MultiIndex.from_tuples(zip([1, 2], [3, 4])), pd.MultiIndex),
+        assert_type(
+            pd.MultiIndex.from_tuples(zip([1, 2], [3, 4], strict=False)), pd.MultiIndex
```
Dr-Irv: same comment about zip

tests/series/test_series.py
```diff
@@ -2141,7 +2141,7 @@ def func() -> MySeries[float]:
 def test_change_to_dict_return_type() -> None:
     id = [1, 2, 3]
     value = ["a", "b", "c"]
-    df = pd.DataFrame(zip(id, value), columns=["id", "value"])
+    df = pd.DataFrame(zip(id, value, strict=False), columns=["id", "value"])
```
Dr-Irv: same comment about zip

tests/test_pandas.py
```diff
@@ -241,9 +241,7 @@ def test_types_concat() -> None:
     check(assert_type(pd.concat({1: df, None: df2}), pd.DataFrame), pd.DataFrame)
 
     check(
-        assert_type(
-            pd.concat(map(lambda _: s2, ["some_value", 3]), axis=1), pd.DataFrame
-        ),
+        assert_type(pd.concat((s2 for _ in ["some_value", 3]), axis=1), pd.DataFrame),
```
Dr-Irv: Not sure of this change.  I think we explicitly wanted to use `map` and `lambda` here

tests/test_pandas.py
```diff
@@ -828,7 +826,7 @@ def test_wide_to_long() -> None:
             "A1980": {0: "d", 1: "e", 2: "f"},
             "B1970": {0: 2.5, 1: 1.2, 2: 0.7},
             "B1980": {0: 3.2, 1: 1.3, 2: 0.1},
-            "X": dict(zip(range(3), np.random.randn(3))),
+            "X": dict(zip(range(3), np.random.randn(3), strict=False)),
```
Dr-Irv: same zip comment

pyproject.toml
```diff
@@ -176,16 +176,48 @@ fix = true
...
+extend-select = ["ALL"]
 ignore = [
-  "E501",   # https://docs.astral.sh/ruff/rules/line-too-long/
-  "E731",   # https://docs.astral.sh/ruff/rules/lambda-assignment/
-  "PYI042", # https://docs.astral.sh/ruff/rules/snake-case-type-alias/
+  # The following rules are ignored permanently for good reasons
+  "COM",     # https://docs.astral.sh/ruff/rules/#flake8-commas-com
+  "D",       # https://docs.astral.sh/ruff/rules/#pydocstyle-d
+  "E501",    # handled by black https://docs.astral.sh/ruff/rules/line-too-long/
+  "EM",      # https://docs.astral.sh/ruff/rules/#flake8-errmsg-em
```
Dr-Irv: curious as to why we can't include this?

pyproject.toml
```diff
@@ -176,16 +176,48 @@ fix = true
...
+  "FBT",     # https://docs.astral.sh/ruff/rules/#flake8-boolean-trap-fbt
+  "I",       # handled by isort
 ]
 
 
 [tool.ruff.lint.per-file-ignores]
-"_*.pyi" = ["PYI001"]
+"*.pyi" = [
+  # The following rules are ignored permanently for good reasons
+  "A001",    # https://docs.astral.sh/ruff/rules/builtin-variable-shadowing/
```
Dr-Irv: would be nice if we could make this work.  Might require converting some `str` to `_str`

pyproject.toml
```diff
@@ -176,16 +176,48 @@ fix = true
...
+  "A002",    # https://docs.astral.sh/ruff/rules/builtin-argument-shadowing/
+  "A004",    # https://docs.astral.sh/ruff/rules/builtin-import-shadowing/
```
Dr-Irv: Do we do this?

pyproject.toml
```diff
@@ -176,16 +176,48 @@ fix = true
...
+  "N",       # https://docs.astral.sh/ruff/rules/#pep8-naming-n
+  "PYI001",  # https://docs.astral.sh/ruff/rules/unprefixed-type-param/
```
Dr-Irv: We should try to fix this eventually

pyproject.toml
```diff
@@ -176,16 +176,48 @@ fix = true
...
+  "PYI042",  # https://docs.astral.sh/ruff/rules/snake-case-type-alias/
```
Dr-Irv: Another one we should hope to fix

pyproject.toml
```diff
@@ -176,16 +176,48 @@ fix = true
...
+  "TD002",   # https://docs.astral.sh/ruff/rules/missing-todo-author/
```
Dr-Irv, line 196: TODO is in one file, so can we make this a single file ignore?

cmp0xff:
Maybe we can freeze versions of all type checkers and upgrade manually to prevent [this](https://github.com/pandas-dev/pandas-stubs/actions/runs/18315183564)

pyproject.toml
```diff
@@ -176,16 +176,48 @@ fix = true
...
+  # The following rules are ignored temporarily. Either fix them or move to the permanent list above.
+  "ERA", "S", "ANN", "FIX002", "TD003", "TD004", "PLR0402", "PLC0105"
```
Dr-Irv: - Surprised that "S" would get picked up in our code in a PYI file
- FIX002 and TD002 are related
- TD003 is surprising because are single TODO does have a link
- TD004 is easy to fix since we have one TODO

pyproject.toml
```diff
@@ -176,16 +176,48 @@ fix = true
...
+  # The following rules are ignored temporarily. Either fix them or move to the permanent list above.
+  "TRY", "PT", "BLE"
+]
+"tests/*" = [
+  # The following rules are ignored permanently for good reasons
+  "E731",    # https://docs.astral.sh/ruff/rules/lambda-assignment/
+  "PD",      # Pandas is here
+  "PLC0415", # https://docs.astral.sh/ruff/rules/import-outside-top-level/
+  "S101",    # https://docs.astral.sh/ruff/rules/assert/
+  "S301",    # https://docs.astral.sh/ruff/rules/suspicious-pickle-usage/
```
Dr-Irv: I think we can limit this to one file  `tests/test_io.py`

pyproject.toml
```diff
@@ -176,16 +176,48 @@ fix = true
...
+  "SLF001",  # https://docs.astral.sh/ruff/rules/private-member-access/
```
Dr-Irv: I think we should aim to fix this one so put it in the second list below

pyproject.toml
```diff
@@ -176,16 +176,48 @@ fix = true
...
+  # The following rules are ignored temporarily. Either fix them or move to the permanent list above.
+  "ANN", "ARG", "ERA", "RUF", "SIM", "TRY", "PT", "NPY", "N", "DTZ", "PLR", "TC", "PGH", "PTH", "S311", "C901", "FIX", "TD", "A001", "PYI042", "ANN201"
```
Dr-Irv: - `FIX` and `TD` relates to `TODO`, so we can do a per-file ignore
- Surprised that ANN201 isn't already taken care of because that's the one that requires return types on functions

tests/__init__.py
```diff
@@ -87,8 +87,7 @@ def check(
             and get_origin(shape_type) is tuple
             and (tuple_args := get_args(shape_type))
             and ... not in tuple_args  # fixed-length tuple
-            and (arr_ndim := getattr(actual, "ndim"))
-            != (expected_ndim := len(tuple_args))
+            and (arr_ndim := actual.ndim) != (expected_ndim := len(tuple_args))  # type: ignore[attr-defined]  # pyright: ignore[reportAttributeAccessIssue]
```
Dr-Irv: This is annoying, because `getattr()` is a good usage here.  If there is a RUFF rule that is causing this fix, let's ignore it on a per-file basis.

tests/__init__.py
```diff
@@ -100,7 +99,7 @@ def check(
             and (dtype_args := get_args(dtype_type))
             and isinstance((expected_dtype := dtype_args[0]), type)
             and issubclass(expected_dtype, np.generic)
-            and (arr_dtype := getattr(actual, "dtype")) != expected_dtype
+            and (arr_dtype := actual.dtype) != expected_dtype  # type: ignore[attr-defined]  # pyright: ignore[reportAttributeAccessIssue]
```
Dr-Irv: same comment about doing a per-file ignore

Dr-Irv:
> Maybe we can freeze versions of all type checkers and upgrade manually to prevent [this](https://github.com/pandas-dev/pandas-stubs/actions/runs/18315183564)

We can do it on a per-type-checker basis, and if we see that a new version has these kinds of failures, we can pin it.  And report it to them so they are aware.

We have pinned `mypy` because when they released changes, it tended to be disruptive, and we definitely want to fix those issues.  But they are better now, we probably don't need to pin `mypy` any more.  They test `mypy` with `pandas-stubs`.
We don't pin `pyright`, because they they release changes, there are typically only a few things we might have to deal with, or they have introduced a bug.  They also test `pyright` with `pandas-stubs`.
`pyrefly` and `ty` are supposed to be tested with `pandas-stubs` (just the stub file), so if a new version of either of those report a bunch of errors like the above, we should report it to them, and *then* pin the version until they fix it.

Dr-Irv (commented):
I don't think we should do those `pyrefly` ignores, because they are in disagreement with both `mypy` and `pyright`.  

I also disagree with them reporting the `bad-param-name-override` on methods like `__getitem__()` and `__iter__()`.  

We should report this to them, and if we need to, add something in our config to ignore the rules you inserted in the commit #6847e8657878201653b49f401c18b72e680b9eb9 and then put the ignore (for now) in `pyproject.toml`

tests/extension/decimal/array.py
cmp0xff: 06f1136a85239969d731d2eb4b06b8edb0de6040

tests/indexes/test_indexes.py
cmp0xff: 06f1136a85239969d731d2eb4b06b8edb0de6040

tests/series/test_series.py
cmp0xff: 06f1136a85239969d731d2eb4b06b8edb0de6040

tests/__init__.py
cmp0xff: `getattr` is no better than `.`, it also raises an error when the attribute is not there. Nevertheless, f251900ab6399e72b11aff1244bd212180afb6d2

tests/__init__.py
cmp0xff: `getattr` is no better than `.`, it also raises an error when the attribute is not there. Nevertheless, f251900ab6399e72b11aff1244bd212180afb6d2

tests/test_pandas.py
cmp0xff: 3e7bb89e87b860d259985677dce630ccdbe226d9

tests/test_pandas.py
cmp0xff: 06f1136a85239969d731d2eb4b06b8edb0de6040

pyproject.toml
cmp0xff: They enforce

```py
msg = "something"
raise ValueError(msg)
```

instead of 

```py
raise ValueError("something")
```

I find them annoying, especially that we are raising only for tests and CI here in `pandas-stubs`.

pyproject.toml
cmp0xff: Moved to the list below

pyproject.toml
cmp0xff: Moved to the list below

pyproject.toml
cmp0xff: Moved to the list below

pyproject.toml
cmp0xff: Moved to the list below

pyproject.toml
cmp0xff: Fixed

pyproject.toml
cmp0xff: - `S` and most of `ERA` removed
- I don't think we'll comply to `FIX002`. It is fine for me to leave `TODO`s.
- There are a few TODOs without link. We can fix them by adding issues.
- `TD004` fixed

pyproject.toml
cmp0xff: Done

pyproject.toml
cmp0xff: Moved below

pyproject.toml
cmp0xff: - `FIX` and `TD` now more detailed
- `ANN201` will be fixed in a separate PR

PR #1414, opened 2025-10-09
GH1398 Remove cast from Index(..., dtype='categorical').array

### LINKED ISSUE #1398: `Index([1], dtype="category").array` gives `ExtensionArray` instead of `pd.Categorical`
After pandas-dev/pandas-stubs#1391,

- `Index([1], dtype="category").array` gives `ExtensionArray` instead of `pd.Categorical`. We need a fix here.
- `cast("Index[pd.CategoricalDtype]", Index([1], dtype="category")).array`, which exists in the current tests, gives `pd.Categorical`, which is okay-ish.

loicdiridollou:
- [x] Closes #1398

### DIFF
--- a/pandas-stubs/core/indexes/category.pyi
+++ b/pandas-stubs/core/indexes/category.pyi
@@ -2,15 +2,11 @@ from collections.abc import (
     Hashable,
     Iterable,
 )
-from typing import (
-    final,
-    type_check_only,
-)
+from typing import final
 
 import numpy as np
 from pandas.core import accessor
 from pandas.core.arrays.categorical import Categorical
-from pandas.core.base import IndexOpsMixin
 from pandas.core.indexes.base import Index
 from pandas.core.indexes.extension import ExtensionIndex
 from typing_extensions import Self
@@ -20,14 +16,11 @@ from pandas._typing import (
     DtypeArg,
 )
 
-@type_check_only
-class CategoricalArrayDescriptor:
-    def __get__(self, instance, owner: type[IndexOpsMixin]) -> Categorical: ...
-
 class CategoricalIndex(ExtensionIndex[S1], accessor.PandasDelegate):
     codes: np.ndarray = ...
     categories: Index = ...
-    array = CategoricalArrayDescriptor()  # pyrefly: ignore[bad-override]
+    @property
+    def array(self) -> Categorical: ...  # type: ignore[override] # pyrefly: ignore[bad-override]
     def __new__(
         cls,
         data: Iterable[S1] = ...,


pandas-stubs/core/indexes/accessors.pyi
```diff
@@ -468,6 +468,7 @@ class ArrayDescriptor:
         self, instance: IndexOpsMixin[CategoricalDtype], owner: type[IndexOpsMixin]
     ) -> Categorical: ...
     @overload
+    @overload
```
cmp0xff: Duplicated?

pandas-stubs/core/indexes/category.pyi
```diff
@@ -15,9 +20,14 @@ from pandas._typing import (
     DtypeArg,
 )
 
+@type_check_only
+class CategoricalArrayDescriptor:
+    def __get__(self, instance, owner: type[IndexOpsMixin]) -> Categorical: ...
```
cmp0xff: Would it work if you add the following to the existing `ArrayDescriptor`, instead of adding a new descriptor?

```py
    @overload
    def __get__(
        self, instance: CategoricalIndex, owner: type[CategoricalIndex]
    ) -> Categorical: ...
```

tests/indexes/test_indexes.py
```diff
@@ -1532,7 +1532,7 @@ def test_array_property() -> None:
     # casting due to pandas-dev/pandas-stubs#1383
     check(
         assert_type(
-            cast("Index[pd.CategoricalDtype]", Index([1], dtype="category")).array,
+            Index([1], dtype="category").array,
             pd.Categorical,
```
cmp0xff:
```suggestion
            Index([1], dtype="category").array, pd.Categorical
```

pandas-stubs/core/indexes/category.pyi
loicdiridollou: I had tried that originally and although pyright is fine with it, mypy does not see it as the right type and I get the following error:
`tests/indexes/test_indexes.py:1534: error: Expression is of type "Any", not "Categorical"  [assert-type]`

pandas-stubs/core/indexes/accessors.pyi
loicdiridollou: Yes my bad I had tried your idea and left some crumbs this is corrected.

pandas-stubs/core/indexes/category.pyi
```diff
@@ -15,9 +20,14 @@ from pandas._typing import (
...
+
 class CategoricalIndex(ExtensionIndex[S1], accessor.PandasDelegate):
     codes: np.ndarray = ...
     categories: Index = ...
+    array = CategoricalArrayDescriptor()  # pyrefly: ignore[bad-override]
```
cmp0xff: There is only one return type, so we can use a static `@property` instead.

```suggestion
class CategoricalIndex(ExtensionIndex[S1], accessor.PandasDelegate):
    codes: np.ndarray = ...
    categories: Index = ...
    @property
    def array(self) -> Categorical: ...  # pyrefly: ignore[bad-override]
```

PR #1417, opened 2025-10-09
feat(ruff): comply to ANN202, 002, TRY, PT, BLE

cmp0xff:
- We now comply to `ANN002` and `ANN202` everywhere (or ignored per line)
- `"scripts/*"` is also fixed so that `TRY`, `PT` and `BLE` are complied to (or ignored per line)

### DIFF
--- a/pandas-stubs/core/indexes/category.pyi
+++ b/pandas-stubs/core/indexes/category.pyi
@@ -6,6 +6,7 @@ from typing import final
 
 import numpy as np
 from pandas.core import accessor
+from pandas.core.arrays.categorical import Categorical
 from pandas.core.indexes.base import Index
 from pandas.core.indexes.extension import ExtensionIndex
 from typing_extensions import Self
@@ -15,6 +16,8 @@ from pandas._typing import S1
 class CategoricalIndex(ExtensionIndex[S1], accessor.PandasDelegate):
     codes: np.ndarray = ...
     categories: Index = ...
+    @property
+    def array(self) -> Categorical: ...  # type: ignore[override] # pyrefly: ignore[bad-override]
     def __new__(
         cls,
         data: Iterable[S1] = ...,
--- a/tests/indexes/test_indexes.py
+++ b/tests/indexes/test_indexes.py
@@ -1531,10 +1531,7 @@ def test_array_property() -> None:
     """Test that Index.array and semantic Index.array return ExtensionArray and its subclasses"""
     # casting due to pandas-dev/pandas-stubs#1383
     check(
-        assert_type(
-            cast("Index[pd.CategoricalDtype]", Index([1], dtype="category")).array,
-            pd.Categorical,
-        ),
+        assert_type(Index([1], dtype="category").array, pd.Categorical),
         pd.Categorical,
         int,
     )


pandas-stubs/core/arrays/base.pyi
```diff
@@ -21,6 +21,8 @@ from pandas._typing import (
 from pandas.core.dtypes.dtypes import ExtensionDtype as ExtensionDtype
 
 class ExtensionArray:
+    def __array__(self, dtype=...) -> np_1darray: ...
```
Dr-Irv: I don't see this in the implementation

pandas-stubs/core/arrays/categorical.pyi
```diff
@@ -89,8 +89,6 @@ class Categorical(ExtensionArray):
     @property
     def shape(self): ...
     def shift(self, periods=1, fill_value=...): ...
-    def __array__(self, dtype=...) -> np_1darray: ...
```
Dr-Irv, line 92: I think you have to put these back because the base class does not define it.  I won't comment on it in the other array classes.

scripts/test/__init__.py
```diff
@@ -22,9 +22,9 @@
 
 
 def test(
-    src: bool = False,
-    dist: bool = False,
-    type_checker: Literal["", "mypy", "pyright"] = "",
+    src: bool = False,  # noqa: PT028
+    dist: bool = False,  # noqa: PT028
+    type_checker: Literal["", "mypy", "pyright"] = "",  # noqa: PT028
```
Dr-Irv: If we rename this to `test_all` and change the reference in `pyproject.toml`, then you shouldn't need the `noqa`

ruff thinks this is pytest code because the function is called `test`

tests/series/test_series.py
```diff
@@ -3178,7 +3182,7 @@ def cond(x: int) -> bool:
     check(assert_type(s.mask(cond, 10), "pd.Series[int]"), pd.Series, np.integer)
 
     # Test case with a boolean condition and a callable
-    def double(x):
+    def double(x: SupportsMul[int, T_co]) -> T_co:
```
Dr-Irv: This is overkill.  We want to test "typical" code.  Can't you just use `def double(x: int) -> int:` ?

tests/test_frame.py
```diff
@@ -1712,7 +1713,12 @@ def test_types_groupby_agg() -> None:
     agg_dict1 = {"col2": "min", "col3": "max", 0: "sum"}
     check(assert_type(df.groupby("col1").agg(agg_dict1), pd.DataFrame), pd.DataFrame)
 
-    def wrapped_min(x: Any) -> Any:
+    T_co = TypeVar("T_co", covariant=True)
+
+    class SupportsMin(Protocol[T_co]):
+        def min(self) -> T_co: ...
+
+    def wrapped_min(x: SupportsMin[T_co]) -> T_co:
```
Dr-Irv: Too much complexity.  We want to test "typical" code.  So can use `def wrapped_min(x: int) -> int`

scripts/test/run.py
```diff
@@ -62,7 +62,9 @@ def rename_src() -> None:
     if Path(r"pandas-stubs").exists():
         Path(r"pandas-stubs").rename("_pandas-stubs")
     else:
-        raise FileNotFoundError("'pandas-stubs' folder does not exists.")
+        raise FileNotFoundError(  # noqa: TRY003
```
Dr-Irv: Looking at this rule, I disagree with it. I'd rather disable the rule in `pyproject` and not use the `noqa`.  You should be able to use a string with a built-in exception class.

tests/test_frame.py
cmp0xff: `int(1).min()` doesn't make sense. I don't think it's too complicated. We just need something supporting `min`.

pandas-stubs/core/arrays/base.pyi
cmp0xff: Thanks, I'll double check next time. 91a1bf14a41fcb4648a79894a18a243288e04faa

pandas-stubs/core/arrays/categorical.pyi
cmp0xff: 91a1bf14a41fcb4648a79894a18a243288e04faa

scripts/test/__init__.py
cmp0xff: It cannot start with `test`. I used `run_tests` instead in e2a87d2958129515c8f7ed4516836055c01a88f1.

scripts/test/run.py
cmp0xff: 927b1bbeb742de7508890bf976884e64fc46349c

tests/series/test_series.py
cmp0xff: 1c98286da2a263b108a1b6b70b99b62bee82d1f6

pandas-stubs/core/arrays/datetimes.pyi
```diff
@@ -29,7 +29,6 @@ class DatetimeArray(DatetimeLikeArrayMixin, TimelikeOps, DatelikeOps):
     def tzinfo(self) -> _tzinfo | None: ...
     @property
     def is_normalized(self): ...
-    def __array__(self, dtype=...) -> np.ndarray: ...
```
Dr-Irv, line 32: I think either this declaration of `__array__()` or the one in `datetimelike.pyi` needs to be in there.

pandas-stubs/core/arrays/masked.pyi
```diff
@@ -32,7 +31,6 @@ class BaseMaskedArray(ExtensionArray, ExtensionOpsMixin):
         na_value: Scalar = ...,
     ) -> np.ndarray: ...
     __array_priority__: int = ...
-    def __array__(self, dtype=...) -> np.ndarray: ...
```
Dr-Irv, line 35: I think this needs to be in there.

tests/test_frame.py
Dr-Irv: I think `def wrapped_min(x: pd.Series) -> int:` is the right thing then.

pandas-stubs/core/arrays/datetimes.pyi
cmp0xff: Added in `datetimelike.pyi` 5b684a3c09e938d5dcd5eba9b8a0db5a2e93be0a

pandas-stubs/core/arrays/masked.pyi
cmp0xff: 5b684a3c09e938d5dcd5eba9b8a0db5a2e93be0a

tests/test_frame.py
cmp0xff: 6f8217877e7ec6fa633d516a5a99b74a8bd5e1a8

PR #1420, opened 2025-10-09
GH1419 Allow Series and Index for other in Index.where(..., other)

### LINKED ISSUE #1419: `Index.where(..., other)` should support `other` with more than array-like
**Describe the bug**
Right now `Index.where(..., other=...)` only supports the strict array-like (ExtensionArray and np.array), it seems that having Series or Index works at runtime and makes sense.

**To Reproduce**

```python
import pandas as pd
import numpy as np

datetime_index = pd.DatetimeIndex(pd.date_range(start="2025-01-01", freq="h", periods=48))
mask = np.ones(48, dtype=bool)

condition = datetime_index.where(mask,datetime_index - pd.Timedelta(days=1))
```

It yields the following error: 
```
Argument 2 to "where" of "Index" has incompatible type "DatetimeIndex"; expected "str | bytes | date | datetime | timedelta | <13 more items> | None"  [arg-type]
```

 - OS: MacOS
 - OS Version 26
 - python version 3.13
 - version of type checker 1.18.1

**Additional context**

loicdiridollou:
- [x] Closes #1419


tests/indexes/test_indexes.py
```diff
@@ -1608,3 +1609,18 @@ def test_to_series() -> None:
         np.complexfloating,
     )
     check(assert_type(Index(["1"]).to_series(), "pd.Series[str]"), pd.Series, str)
+
+
+def test_index_where() -> None:
+    """Test Index.where with multiple types of other GH1419."""
+    datetime_index = pd.DatetimeIndex(
+        pd.date_range(start="2025-01-01", freq="h", periods=48)
+    )
```
cmp0xff: `pd.date_range` already gives a `DatetimeIndex` at run time. You can do
```suggestion
    datetime_index = pd.date_range(start="2025-01-01", freq="h", periods=48)
```
unless you need to `cast` (which can be fixed separately)

pandas-stubs/core/indexes/base.pyi
```diff
@@ -440,7 +441,7 @@ class Index(IndexOpsMixin[S1], ElementOpsMixin[S1]):
     @property
     def values(self) -> np_1darray: ...
     def memory_usage(self, deep: bool = False): ...
-    def where(self, cond, other: Scalar | ArrayLike | None = None): ...
+    def where(self, cond, other: Scalar | AnyArrayLike | None = None) -> Self: ...
```
cmp0xff: - I understand that it wasn't intended to give a full typing for `where` here, but it's already so close. Could you do something like `cond: Sequence[bool] | np_ndarray_bool | BooleanArray | IndexOpsMixin[bool]` as well?
- It turns out the following also works:
  ```py
  >>> pd.date_range(start="2025-01-01", freq="h", periods=2).where(pd.Series([True, False]), pd.Index([0, 1]))
  Index([2025-01-01 00:00:00, 1], dtype='object')
  ```
  I would say `-> Self` is incorrect. I think we can do two overloads here in `class Index`. If `other` is `S1 | Self`, we could say `-> Self`. Then if `other` is `Scalar | AnyArrayLike | None`, we could only say `-> Index`. What do you think?
  - `RangeIndex.where` will probably break, but we could leave it in another effort.

loicdiridollou: Okay I will take care of that so we are fully typed there, I had originally taken the docs and it did not uncover thsoe edge cases, nice job digging into it!

pandas-stubs/core/indexes/range.pyi
```diff
@@ -94,3 +99,8 @@ class RangeIndex(_IndexSubclassBase[int, np.int64]):
     def __getitem__(  # pyright: ignore[reportIncompatibleMethodOverride]
         self, idx: int
     ) -> int: ...
+    def where(
+        self,
+        cond: Sequence[bool] | np_ndarray_bool | BooleanArray | IndexOpsMixin[bool],
+        other: Scalar | AnyArrayLike | None = None,
+    ) -> Self: ...
```
cmp0xff: Originally I suggested skipping this part. Consider the following cases:
```py
>>> pd.RangeIndex(2).where([False, True], 3) 
RangeIndex(start=3, stop=-1, step=-2)  # fine
>>> pd.RangeIndex(3).where([False, True, True], 3)
Index([3, 1, 2], dtype='int64')  # need to fix
pd.RangeIndex(2).where([False, True])         
Index([nan, 1.0], dtype='float64')  # need to fix
```

For simplicity I'd like to propose always returning `Index` here.

```suggestion
    def where(
        self,
        cond: Sequence[bool] | np_ndarray_bool | BooleanArray | IndexOpsMixin[bool],
        other: Scalar | AnyArrayLike | None = None,
    ) -> Index: ...
```

loicdiridollou: Okay good idea, we can refine it in the future!

MarcoGorelli: I don't think it's possible to determine statically when the return value is `RangeIndex`, so i'm not sure it can be refined later unfortunately 😩 
```py
In [22]: pd.RangeIndex(3).where([True, True, True], 3)
Out[22]: RangeIndex(start=0, stop=3, step=1)

In [23]: pd.RangeIndex(3).where([True, True, False], 3)
Out[23]: Index([0, 1, 3], dtype='int64')
```

cmp0xff (approved):
Thank you @loicdiridollou !

PR #1421, opened 2025-10-09
GH1418 Disallow pd.Timedelta * bool

### LINKED ISSUE #1418: Disallow `pd.Timedelta` * `bool`
Following pandas-dev/pandas#62316 and pandas-dev/pandas-stubs#1413, `pd.Timedelta` * `bool` will be disallowed in 3.0.

In the stubs, we need to find everywhere with `pandas-dev/pandas#62316` and invalidate the usages.

loicdiridollou:
- [ ] Closes #1418

### DIFF
--- a/tests/scalars/timedelta/test_arithmetic.py
+++ b/tests/scalars/timedelta/test_arithmetic.py
@@ -15,11 +15,16 @@ def test_mul() -> None:
     b = True
     c = 1.0
     d = 5
+    e = 1 + 3.0j
 
     check(assert_type(a * c, pd.Timedelta), pd.Timedelta)
+    check(assert_type(c * a, pd.Timedelta), pd.Timedelta)
     check(assert_type(a * d, pd.Timedelta), pd.Timedelta)
+    check(assert_type(d * a, pd.Timedelta), pd.Timedelta)
 
     if TYPE_CHECKING_INVALID_USAGE:
         # pd.Timedelta * bool is not allowed, see GH1418
         _0 = a * b  # type: ignore[operator] # pyright: ignore[reportOperatorIssue]
         _1 = b * a  # type: ignore[operator] # pyright: ignore[reportOperatorIssue]
+        _2 = a * e  # type: ignore[operator] # pyright: ignore[reportOperatorIssue]
+        _3 = e * a  # type: ignore[operator] # pyright: ignore[reportOperatorIssue]


pandas-stubs/_libs/tslibs/timedeltas.pyi
```diff
@@ -209,12 +212,16 @@ class Timedelta(timedelta):
...
+    @overload
     def __mul__(self, other: float) -> Self: ...
     @overload
     def __mul__(
         self, other: np_ndarray[ShapeT, np.bool_ | np.integer | np.floating]
     ) -> np_ndarray[ShapeT, np.timedelta64]: ...
     @overload
+    def __rmul__(self, other: bool) -> Never: ...
+    @overload
     def __rmul__(self, other: float) -> Self: ...
```
cmp0xff: I would say compared to `Never`, "unsupported" is more in favour, which works in this case. You can `from pandas._typing import Just`, and then
```suggestion
    def __mul__(self, other: Just[int] | Just[float]) -> Self: ...
    @overload
    def __mul__(
        self, other: np_ndarray[ShapeT, np.bool_ | np.integer | np.floating]
    ) -> np_ndarray[ShapeT, np.timedelta64]: ...
    @overload
    def __rmul__(self, other: Just[int] | Just[float]) -> Self: ...
```

tests/indexes/arithmetic/timedelta/test_mul.py
```diff
@@ -113,3 +113,13 @@ def test_mul_pd_index(left: "pd.Index[pd.Timedelta]") -> None:
...
+
+
+def test_mul_bool() -> None:
+    """Test checking that pd.Timedelta * bool is not allowed GH1418."""
+    a = pd.Timedelta("1 day")
+    b = True
+
+    if TYPE_CHECKING_INVALID_USAGE:
+        assert_type(a * b, Never)
+        assert_type(b * a, Never)
```
cmp0xff: This file is for tests for `Index`es.

- Please move the test to somewhere like `‎tests/scalars/timedelta/test_arithmetic.py` - For now I guess we just need one file per scalar type for arithemtic.
- Please also test that multiplying `int` and `float` is fine, and multiplying `complex` is forbidden. This will completely test the two overloads you changed, in one place.

pandas-stubs/_libs/tslibs/timedeltas.pyi
loicdiridollou: This is fixed!

tests/indexes/arithmetic/timedelta/test_mul.py
loicdiridollou: Not sure about the path of the testing, seems quite a complicated way to organize tests, may be easier if we just replicate the architecture of the repo but this is an issue for later.

tests/scalars/timedelta/test_arithmetic.py
```diff
@@ -0,0 +1,24 @@
...
+
+def test_mul_bool() -> None:
+    """Test checking that pd.Timedelta * bool is not allowed GH1418."""
+    a = pd.Timedelta("1 day")
+    b = True
+    c = 1.0
+    d = 5
+
+    check(assert_type(a * c, pd.Timedelta), pd.Timedelta)
+    check(assert_type(a * d, pd.Timedelta), pd.Timedelta)
```
cmp0xff, line 22: - Also `c * a` and `d * a`, to be complete.
- Also multiplying and r-multiplying with a complex number, to be complete.

tests/scalars/timedelta/test_arithmetic.py
```diff
@@ -0,0 +1,24 @@
...
+from typing_extensions import assert_type
+
+from tests import (
+    TYPE_CHECKING_INVALID_USAGE,
+    check,
+)
+
+
+def test_mul_bool() -> None:
+    """Test checking that pd.Timedelta * bool is not allowed GH1418."""
```
cmp0xff: - `test_mul` or `test_mul_num` would be better, since not just boolean variables are tested.
- Also the docstring.

tests/scalars/timedelta/test_arithmetic.py
loicdiridollou: Adjusting, thanks for the feedback

cmp0xff: Hi, did you push your commit corresponding to the two comments here?

loicdiridollou: Sorry forgot to add, should be good now (complex multiplication is invalid so it is in the INVALID_TYPE_USAGE)

PR #1423, opened 2025-10-13
Type ``ExtensionArray.{dropna,unique,repeat,take,ravel}``

MarcoGorelli:

### DIFF
--- a/pandas-stubs/core/arrays/base.pyi
+++ b/pandas-stubs/core/arrays/base.pyi
@@ -1,6 +1,4 @@
-from collections.abc import (
-    Iterator,
-)
+from collections.abc import Iterator
 from typing import (
     Any,
     Literal,
--- a/tests/arrays/test_extension_array.py
+++ b/tests/arrays/test_extension_array.py
@@ -9,6 +9,7 @@ from tests import check
 
 
 def test_ea_common() -> None:
+    # Note: `ExtensionArray` is abstract, so we use `IntegerArray` for the tests.
     arr = array([1, 2, 3])
 
     check(assert_type(arr.repeat(1), IntegerArray), IntegerArray)
@@ -18,6 +19,7 @@ def test_ea_common() -> None:
         IntegerArray,
     )
     check(assert_type(arr.repeat(pd.Index([1, 2, 3])), IntegerArray), IntegerArray)
+    check(assert_type(arr.repeat([1, 2, 3]), IntegerArray), IntegerArray)
 
     check(assert_type(arr.unique(), IntegerArray), IntegerArray)
     check(assert_type(arr.dropna(), IntegerArray), IntegerArray)
--- a/pandas-stubs/_typing.pyi
+++ b/pandas-stubs/_typing.pyi
@@ -865,7 +865,7 @@ np_1darray: TypeAlias = np.ndarray[tuple[int], np.dtype[GenericT]]
 np_2darray: TypeAlias = np.ndarray[tuple[int, int], np.dtype[GenericT]]
 
 AnyArrayLikeInt: TypeAlias = (
-    IntegerArray | Index[int] | Series[int] | np_1darray[np.integer]
+    IntegerArray | Index[int] | Series[int] | np_1darray[np.integer] | Sequence[int]
 )
 
 class SupportsDType(Protocol[GenericT_co]):


pandas-stubs/core/arrays/base.pyi
```diff
@@ -51,23 +54,23 @@ class ExtensionArray:
...
-    def dropna(self): ...
+    def fillna(self, value=..., method=None, limit=None) -> Self: ...
+    def dropna(self) -> Self: ...
     def shift(self, periods: int = 1, fill_value: object = ...) -> Self: ...
-    def unique(self): ...
+    def unique(self) -> Self: ...
     def searchsorted(self, value, side: str = ..., sorter=...): ...
     def factorize(self, use_na_sentinel: bool = True) -> tuple[np_1darray, Self]: ...
-    def repeat(self, repeats, axis=...): ...
+    def repeat(self, repeats: int | Sequence[int], axis: None = None) -> Self: ...
```
MarcoGorelli: https://pandas.pydata.org/docs/reference/api/pandas.api.extensions.ExtensionArray.repeat.html documents:

```
        axis : None
            Must be ``None``. Has no effect but is accepted for compatibility
            with numpy.
```

given that `None` is the only accepted type, that's what I've typed here

pandas-stubs/core/arrays/base.pyi
```diff
@@ -51,23 +54,23 @@ class ExtensionArray:
...
     def take(
         self,
         indexer: TakeIndexer,
         *,
-        allow_fill: bool = ...,
-        fill_value=...,
+        allow_fill: bool = False,
+        fill_value: Any = None,
```
MarcoGorelli, line 67: `fill_value` here I think could really be anything, as people can make `ExtensionArray` that hold whatever scalars they like

pandas types it as `Any`

https://github.com/pandas-dev/pandas/blob/7263a95efeaec865eb4174871f01f081dfd6a0cf/pandas/core/arrays/base.py#L1744

pandas-stubs/core/arrays/base.pyi
```diff
@@ -52,22 +55,22 @@ class ExtensionArray:
...
     ) -> Self: ...
     def copy(self) -> Self: ...
     def view(self, dtype=...) -> Self | np_1darray: ...
-    def ravel(self, order="C") -> Self: ...
+    def ravel(self, order: str = "C") -> Self: ...
```
cmp0xff:
```suggestion
    def ravel(self, order: Literal["C", "F", "A", "K"] | None = "C") -> Self: ...
```
https://github.com/pandas-dev/pandas/blob/9c8bc3e55188c8aff37207a74f1dd144980b8874/pandas/core/arrays/base.py#L1808-L1832

pandas-stubs/core/arrays/base.pyi
cmp0xff: If it holds "scalar", would it make sense to use `S1` or `Scalar`?

pandas-stubs/core/arrays/base.pyi
cmp0xff: Would it work if `repeats` is a numpy array, pandas array or pandas `Series`, etc.? These are not `Sequence`s.

pandas-stubs/core/arrays/base.pyi
MarcoGorelli: `S1` is a `TypeVar` which doesn't appear in `ExtensionArray`, so I don't think it can be used here

I think given that for `ExtensionArray` EA authors can define their own scalars, `Scalar` would be too narrow

pandas-stubs/core/arrays/base.pyi
MarcoGorelli: yup that's allowed, thanks - is there an alias in pandas-stubs for "array-like of integers"?

cmp0xff: > is there an alias in pandas-stubs for "array-like of integers"?

Not that I am aware of. Maybe you can create one and put it below `AnyArrayLike` in `_typing.pyi`.

pandas-stubs/core/arrays/base.pyi
```diff
@@ -52,22 +55,22 @@ class ExtensionArray:
...
     def fillna(self, value=..., method=None, limit=None): ...
-    def dropna(self): ...
+    def dropna(self) -> Self: ...
     def shift(self, periods: int = 1, fill_value: object = ...) -> Self: ...
-    def unique(self): ...
+    def unique(self) -> Self: ...
     def searchsorted(self, value, side: str = ..., sorter=...): ...
     def factorize(self, use_na_sentinel: bool = True) -> tuple[np_1darray, Self]: ...
-    def repeat(self, repeats, axis=...): ...
+    def repeat(self, repeats: int | AnyArrayLikeInt, axis: None = None) -> Self: ...
```
cmp0xff, line 61: `Sequence[int]` now missing

tests/arrays/test_extension_array.py
```diff
@@ -0,0 +1,29 @@
...
+    check(assert_type(arr.repeat(pd.Index([1, 2, 3])), IntegerArray), IntegerArray)
+
+    check(assert_type(arr.unique(), IntegerArray), IntegerArray)
+    check(assert_type(arr.dropna(), IntegerArray), IntegerArray)
+    check(assert_type(arr.take([1, 0, 2]), IntegerArray), IntegerArray)
+    check(
+        assert_type(arr.take([1, 0, 2], allow_fill=True, fill_value=-1), IntegerArray),
+        IntegerArray,
+    )
+    check(assert_type(arr.ravel(), IntegerArray), IntegerArray)
```
cmp0xff, line 31: You are actually testing `IntegerArray`. `ExtensionArray` is abstract, anyway.

pandas-stubs/core/arrays/base.pyi
```diff
@@ -1,4 +1,6 @@
-from collections.abc import Iterator
+from collections.abc import (
+    Iterator,
+)
```
cmp0xff:
```suggestion
from collections.abc import Iterator
```

tests/arrays/test_extension_array.py
MarcoGorelli: sure, thanks, i've added a comment about that

PR #1424, opened 2025-10-14
feat: introduce ElementOpsMixin

cmp0xff:
It is an attempt to reduce duplicated script for arithmetic operations in `Index` and `Series`.

### DIFF
--- a/pandas-stubs/core/base.pyi
+++ b/pandas-stubs/core/base.pyi
@@ -189,15 +189,15 @@ class ElementOpsMixin(Generic[S2]):
     ) -> ElementOpsMixin[bool]: ...
     @overload
     def _proto_add(
-        self: ElementOpsMixin[int], other: bool | np.bool_
+        self: ElementOpsMixin[int], other: int | np.integer
     ) -> ElementOpsMixin[int]: ...
     @overload
     def _proto_add(
-        self: ElementOpsMixin[float], other: int | np.integer
+        self: ElementOpsMixin[float], other: float | np.floating
     ) -> ElementOpsMixin[float]: ...
     @overload
     def _proto_add(
-        self: ElementOpsMixin[complex], other: float | np.floating
+        self: ElementOpsMixin[complex], other: complex | np.complexfloating
     ) -> ElementOpsMixin[complex]: ...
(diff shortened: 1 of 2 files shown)


pandas-stubs/core/base.pyi
```diff
@@ -176,3 +180,108 @@ NumListLike: TypeAlias = (
...
+
+@type_check_only
+class ElementOpsMixin(Generic[S2]):
+    @overload
+    def _add(
+        self: ElementOpsMixin[bool], other: bool | np.bool_
+    ) -> ElementOpsMixin[bool]: ...
+    @overload
+    def _add(
+        self: ElementOpsMixin[int], other: bool | np.bool_
```
Dr-Irv: I'm a bit puzzled here as to why these methods are called `_add` vs `__add__` and `_mul` vs `__mul__`

cmp0xff: If they were called `__add__`, they would be overridden (instead of be overloaded) by the subclass overloads. That's the reason for the complication.

Dr-Irv: Then I don't understand how the operator `Series[bool] + bool` will get matched.  Is `_add` equivalent to `__add__()` in some way?

cmp0xff: I defined a protocol `Supports_ElementAdd`, which matches `ElementOpsMixin`. Note that `Series` is now also a subclass of `ElementOpsMixin`, so that in `Series`, `__add__(self: Supports_ElementAdd[_T_contra, S2], other: _T_contra | Sequence[_T_contra])` can match.

Dr-Irv: OK - now I get it.

I think it would be easier to understand if we called the methods something like `_proto_add` and `_proto_mul` ,etc.  Then it's clear it connects to the protocol.

cmp0xff: cb645f1e96fa37b1e74bee4e3b2263f9865c47a3

Dr-Irv: Thanks for those changes.  Makes it easier to understand

PR #1425, opened 2025-10-15
maint: split `Sequence[int]` out from `AnyArrayLikeInt`

MarcoGorelli:
#1423 introduced `AnyArrayLikeInt`, but on reflection i'm not too happy about how it's diverged from `AnyArrayLike`

I'd like to suggest taking `Sequence[int]` out of it, and just using `| Sequence[int]` where necessary

### DIFF
--- a/pandas-stubs/_typing.pyi
+++ b/pandas-stubs/_typing.pyi
@@ -86,6 +86,9 @@ HashableT5 = TypeVar("HashableT5", bound=Hashable)
 
 ArrayLike: TypeAlias = ExtensionArray | np.ndarray
 AnyArrayLike: TypeAlias = ArrayLike | Index | Series
+AnyArrayLikeInt: TypeAlias = (
+    IntegerArray | Index[int] | Series[int] | npt.NDArray[np.integer]
+)
 
 # list-like
 
@@ -863,11 +866,6 @@ np_ndarray: TypeAlias = np.ndarray[ShapeT, np.dtype[GenericT]]
 # Numpy arrays with known shape (Do not use as argument types, only as return types)
 np_1darray: TypeAlias = np.ndarray[tuple[int], np.dtype[GenericT]]
 np_2darray: TypeAlias = np.ndarray[tuple[int, int], np.dtype[GenericT]]
-
-# Keep in-sync with `AnyArrayLike`.
-AnyArrayLikeInt: TypeAlias = (
-    IntegerArray | Index[int] | Series[int] | np_1darray[np.integer]
-)
 
 class SupportsDType(Protocol[GenericT_co]):
     @property


pandas-stubs/_typing.pyi
```diff
@@ -864,8 +864,9 @@ np_ndarray: TypeAlias = np.ndarray[ShapeT, np.dtype[GenericT]]
 np_1darray: TypeAlias = np.ndarray[tuple[int], np.dtype[GenericT]]
 np_2darray: TypeAlias = np.ndarray[tuple[int, int], np.dtype[GenericT]]
 
+# Keep in-sync with `AnyArrayLike`.
 AnyArrayLikeInt: TypeAlias = (
-    IntegerArray | Index[int] | Series[int] | np_1darray[np.integer] | Sequence[int]
+    IntegerArray | Index[int] | Series[int] | np_1darray[np.integer]
 )
```
cmp0xff: Would it make more sense to put it near `AnyArrayLike`, i.e. L88?

cmp0xff: - In the test, `np.ndarray` is not tested. Could you add it?
- I am thinking requiring `np_1darray` as an _input_ might be too restrictive for users.

  At the moment, `np.array([1, 2])` gives an `np.ndarray[tuple[Any, ...], dtype[Any]`.

  The 1-dimentional part in the requirement for an input needs to be manually specified by the user, which may not be very friendly.

MarcoGorelli: yeah i tried that but that's before `np_1darray` is defined, and moving everything around  so the `AnyArrayLike` were close together proved trickier than i expected

MarcoGorelli: > The 1-dimentional part in the requirement for an input needs to be manually specified by the user

I don't think it needs to be specified:
- `np.array([1, 2])` gives `tuple[Any, ...]` as the shape
- `Any` turns off the type checker for that argument
- `np_1darray` expects `tuple[int]` as the shape, and it's fine to assign an `Any` to that

I've added a test, thanks for the suggestion 👍 the type checkers seem fine with it

MarcoGorelli: > the type checkers seem fine with it

oh maybe not 🤔 not sure why it passed locally, taking a look

MarcoGorelli: Ah I see, it fails on older versions of NumPy (prior to 2.3.0, which isn't available on Python 3.10)

MarcoGorelli: i've switched to just using `npt.NDArray[np.integer]`, so like this there's no issues for older versions of NumPy, and we can keep the aliases close together

thanks for your review!

PR #1426, opened 2025-10-15
Fix erroneous `_ListLike` type in `Series` / `Index` methods

### LINKED ISSUE #1427: `_ListLike` incorrect annotation for `Series.searchsorted` and `Series.quantile`
**Describe the bug**

**To Reproduce**

```python
import pandas as pd
import numpy as np

s = pd.Series([1,2,3])

s.searchsorted({'a': np.array([1])})
s.quantile(q={'a': np.array([1])})
```

This is fine statically but raises at runtime
```
TypeError: '>=' not supported between instances of 'dict' and 'dict'
TypeError: '>=' not supported between instances of 'dict' and 'int'
```

I don't think `_ListLike` is the correct annotation here

 - OS Version [e.g. 22]
 - python version
 - version of type checker
 - version of installed `pandas-stubs` 9105941ee5522b9137f2626eb36489a700e6e405

**Additional context**

MarcoGorelli:

closes #1427 

`_ListLike` is used in the `Series` constructor for the `data` argument, but I think it may be misnamed, as it's then been reused in other places that don't accept the same types (in particular, that don't accept `dict`)

So I've:
- renamed `_ListLike` to `_DataLike`
- used `ListLike` in other places which currently use `_ListLike` but don't actually accept `dict` input. In cases where we end up with `ListLike | Index | Series`, I've removed `Index | Series` as `ListLike` already contains them
- For Index arithmetic methods, I've used `ArrayLike | SequenceNotStr[S1]`, which is the same as what `_ListLike` was but without the extraneous `dict[str, np.ndarray]`

### DIFF
--- a/pandas-stubs/core/series.pyi
+++ b/pandas-stubs/core/series.pyi
@@ -2470,7 +2470,7 @@ class Series(IndexOpsMixin[S1], ElementOpsMixin[S1], NDFrame):
     ) -> Series[_bool]: ...
     @overload
     def __mul__(  # type: ignore[overload-overlap]
-        self: Series[Never], other: complex | NumListLike
+        self: Series[Never], other: complex | NumListLike | Index | Series
     ) -> Series: ...
     @overload
     def __mul__(self, other: Index[Never] | Series[Never]) -> Series: ...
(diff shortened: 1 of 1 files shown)


cmp0xff (approved):
I like the changes. But there is a lot of them. I would like to wait for a second pair of eyes. cc @Dr-Irv @loicdiridollou

PR #1433, opened 2025-10-16
refactor(ruff): comply to `PTH` and `TD003`

cmp0xff:
- `PTH` is just modernising the style
- `TD003` can prevent hidden `TODO`s, forcing an issue

PR #1434, opened 2025-10-19
chore: 🐍 pytest with py314

cmp0xff:
This is an attempt to run pytest also with py314.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -3,7 +3,7 @@ ci:
     autofix_prs: false
 repos:
 -   repo: https://github.com/astral-sh/ruff-pre-commit
-    rev: v0.14.1
+    rev: v0.14.2
     hooks:
     -   id: ruff-check
         args: [--exit-non-zero-on-fix]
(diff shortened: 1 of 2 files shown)


cmp0xff:
> looks like it is failing because numpy got upgraded to version 2.3.4 on October 15

Trying to fix in pandas-dev/pandas-stubs#1433

cmp0xff:
[`pyarrow` 21.0.0 cannot be built with python3.14](https://github.com/pandas-dev/pandas-stubs/actions/runs/18635953336/job/53127130040). We'll need to wait for apache/arrow#47438.

loicdiridollou:
Correct @cmp0xff `pyarrow` is not yet compatible with 314 so will have to wait for that, I tried a few days ago and fell into the same trap.

Dr-Irv:
> It only raises on 3.14, I dug into it and turns out it will alert if `sys.getrefcount(self)` in https://github.com/pandas-dev/pandas/blob/f4e1c73159cc59a5aff076604f8b43a08c4dbf5d/pandas/core/frame.py#L4291 is under 3, it is exactly 3 in 3.14 but 4 in 3.13. Maybe a bit painful to wrap all the tests, but the warning is at least raised inside the functions and not in our tests so open to options

Can you report this on the pandas repo?

PR #1435, opened 2025-10-20
Type ``ExtensionArray.view``, ``ExtensionArray.astype``, ``ExtensionArray.fillna`` and ``ExtensionArray.searchsorted``

MarcoGorelli:

### DIFF
--- a/tests/extension/decimal/array.py
+++ b/tests/extension/decimal/array.py
@@ -32,7 +32,6 @@ from pandas._typing import (
     ArrayLike,
     AstypeArg,
     TakeIndexer,
-    np_1darray,
 )
 
 from pandas.core.dtypes.base import ExtensionDtype
@@ -41,6 +40,8 @@ from pandas.core.dtypes.common import (
     is_float,
     pandas_dtype,
 )
+
+from tests import np_1darray
 
 
 @register_extension_dtype


pandas-stubs/core/arrays/base.pyi
```diff
@@ -50,16 +53,36 @@ class ExtensionArray:
     def ndim(self) -> int: ...
     @property
     def nbytes(self) -> int: ...
-    def astype(self, dtype, copy: bool = True): ...
+    @overload
+    def astype(self, dtype: np.dtype, copy: bool = True) -> np.ndarray: ...
```
cmp0xff: This could probably be improved
```suggestion
    def astype(self, dtype: np.dtype, copy: bool = True) -> np_1darray: ...
```

pandas-stubs/core/arrays/base.pyi
```diff
@@ -50,16 +53,36 @@ class ExtensionArray:
...
+    @overload
+    def astype(self, dtype: ExtensionDtype, copy: bool = True) -> ExtensionArray: ...
```
cmp0xff, line 59: I would try
```suggestion
    def astype(self, dtype: ExtensionDtype, copy: bool = True) -> Self: ...
```
because I guess you can't convert an `IntervalArray` to an `IntegerArray` by `.astype`. But I understand that this is unspecified in the docs, so keeping `-> ExtensionArray` is also fine.

MarcoGorelli: I think it possible though?
```python
In [34]: arr = pd.Series([1.]).array

In [35]: arr
Out[35]:
<NumpyExtensionArray>
[1.0]
Length: 1, dtype: float64

In [36]: arr.astype('Int64')
Out[36]:
<IntegerArray>
[1]
Length: 1, dtype: Int64
```

cmp0xff: I see, my assumption was not correct.

cmp0xff (approved):
Looks good to me. I hope one day we can get rid of `-> ExtensionArray`, but we can leave it for another project. Thank you @MarcoGorelli !

PR #1437, opened 2025-10-21
Fix: proper return types for MultiIndex.swaplevel & MultiIndex.union

### LINKED ISSUE #1436: MultiIndex method corrections
**Describe the bug**
The methods `MultiIndex.swaplevel()` and `MultiIndex.union()` produce incorrect results:

- `swaplevel()` is missing the return type of `Self`
- `union` should be overridden, as the union of 2 `MultiIndex` should be a `MultiIndex`, not an `Index`

**To Reproduce**
```python
import pandas as pd

mi = pd.MultiIndex.from_product([["a", "b"], [1, 2]], names=["let", "num"])
miswap = mi.swaplevel(0, 1)
reveal_type(miswap)
mi2= pd.MultiIndex.from_product([["a", "b"], [3, 4]], names=["let", "num"])
miu = mi.union(mi2)
reveal_type(miu)
```

At runtime, this produces:
```text
Runtime type is 'MultiIndex'
Runtime type is 'MultiIndex'
```
With `pyright` and the code currently in `master`, I get:
```text
  mitest.py:6:13 - information: Type of "miswap" is "Unknown"
  mitest.py:9:13 - information: Type of "miu" is "Index[Any]"
```

 - OS:  Windows 11
 - OS Version:  ??
 - python version   3.11
 - version of type checker   pyright 1.1.406
 - version of installed `pandas-stubs`   From commit #71b53992643d62c41d60c9903417ecc631d3e05e

realzachbrownjohn:
- [x] Closes #1436
- [x] Tests added: None yet

This is my first contribution so I would appreciate your guidance. I think I have fixed the issue, but I'm confused on where to add test(s), or if I need to at all. Would it be acceptable to use Dr-Irv's reproduction script, or perhaps a modified copy?

### DIFF
--- a/tests/indexes/test_indexes.py
+++ b/tests/indexes/test_indexes.py
@@ -334,15 +334,17 @@ def test_index_union_sort() -> None:
     """Test sort argument in pd.Index.union GH1264."""
     check(
         assert_type(
-            pd.Index(["e", "f"]).union(["a", "b", "c"], sort=True), pd.Index[str]
+            pd.Index(["e", "f"]).union(["a", "b", "c"], sort=True), "pd.Index[str]"
         ),
-        pd.Index[str],
+        pd.Index,
+        str,
     )
     check(
         assert_type(
-            pd.Index(["e", "f"]).union(["a", "b", "c"], sort=False), pd.Index[str]
+            pd.Index(["e", "f"]).union(["a", "b", "c"], sort=False), "pd.Index[str]"
         ),
-        pd.Index[str],
+        pd.Index,
+        str,
     )
 
 
@@ -1614,7 +1616,6 @@ def test_multiindex_union() -> None:
 
     check(assert_type(mi.union(mi2), "pd.MultiIndex"), pd.MultiIndex)
     check(assert_type(mi.union([("c", 3), ("d", 4)]), "pd.MultiIndex"), pd.MultiIndex)
-    check(assert_type(mi.union([1, 2, 3]), "pd.MultiIndex"), pd.MultiIndex)
 
 
 def test_multiindex_swaplevel() -> None:


pandas-stubs/core/indexes/multi.pyi
```diff
@@ -163,3 +164,7 @@ class MultiIndex(Index):
     def insert(self, loc, item): ...
     def delete(self, loc): ...
     def isin(self, values, level=...) -> np_1darray[np.bool]: ...
+    @overload
+    def union(self, other: MultiIndex, sort: bool | None = ...) -> MultiIndex: ...
+    @overload
+    def union(self, other: Index | list[HashableT], sort: bool | None = ...) -> Index[Any]: ...
```
cmp0xff: I think this should be fixed in `def union` in `class Index`, where the return type should be `Self` intead of `Index`. Then you can remove these two overloads.

What you added here indicates that `MultiIndex.union(list)` can give an `Index`, which looks quite suspcious to me.

Dr-Irv: Well, I think the overloads can be combined, but should be specific for `MultiIndex`:
```python
>>> mi=pd.MultiIndex.from_product([["a", "b"], [1,2]], names=["ab", "12"])
>>> mi.union(pd.Index([1,2,3])
... )
  File "<stdin>", line 1, in <module>
  File "C:\Condadirs\envs\pandasstubs311\Lib\site-packages\pandas\core\indexes\base.py", line 3338, in union
    raise NotImplementedError(
NotImplementedError: Can only union MultiIndex with MultiIndex or Index of tuples, try mi.to_flat_index().union(other) instead.
>>> mi.union(pd.Index([("c", 3), ("d", 4)]))
MultiIndex([('a', 1),
            ('a', 2),
            ('b', 1),
            ('b', 2),
            ('c', 3),
            ('d', 4)],
           )
>>> mi.union([("c", 3), ("d", 4)])
MultiIndex([('a', 1),
            ('a', 2),
            ('b', 1),
            ('b', 2),
            ('c', 3),
            ('d', 4)],
           names=['ab', '12'])
```

So you can union with `Index` or `list[tuple[Hashable...]]`

pandas-stubs/_typing.pyi
```diff
@@ -904,7 +904,10 @@ SeriesDType: TypeAlias = (
     | datetime.datetime  # includes pd.Timestamp
     | datetime.timedelta  # includes pd.Timedelta
 )
-S1 = TypeVar("S1", bound=SeriesDType, default=Any)
+
+IndexKey: TypeAlias = SeriesDType | tuple[Hashable, ...] # to support Indexes of tuples for MultiIndex unions
+
+S1 = TypeVar("S1", bound=IndexKey, default=Any)
```
cmp0xff: Please leave `S1` intact, it is too fundamental to change easily.

At least `Series` also uses `S1`, where you probably don't want to change.

pandas-stubs/core/indexes/multi.pyi
```diff
@@ -135,7 +135,7 @@ class MultiIndex(Index):
     def append(self, other): ...
     def repeat(self, repeats, axis=...): ...
     def drop(self, codes, level: Level | None = None, errors: str = "raise") -> Self: ...  # type: ignore[override] # pyright: ignore[reportIncompatibleMethodOverride]
-    def swaplevel(self, i: int = -2, j: int = -1): ...
+    def swaplevel(self, i: int = -2, j: int = -1) -> Self: ...
```
cmp0xff, line 138: Please add a test for `swaplevel`

realzachbrownjohn:
I can see now that all checks have failed, looks related to my extension of S1 (in adding support for `tuple[Hashable, ...]` I seem to have violated some sort of constraint). Any guidance would be greatly appreciated.

pandas-stubs/core/indexes/multi.pyi
```diff
@@ -163,3 +163,7 @@ class MultiIndex(Index):
     def insert(self, loc, item): ...
     def delete(self, loc): ...
     def isin(self, values, level=...) -> np_1darray[np.bool]: ...
+    @overload
+    def union(self, other: MultiIndex | Index[tuple[Hashable, ...]] | list[tuple[Hashable, ...]], sort: bool | None = None,) -> Self: ...
+    @overload
+    def union(self, other: list[HashableT] | Index, sort: bool | None = None,) -> Index: ...
```
cmp0xff: > I've tried my best to spend a bit of time unpacking this. Ultimately, I didn't think it was best to include all types of Index in the overload because not all types of Index are valid to union with MultiIndex (only those with tuples). Referring to tuples within Indexes wasn't supported so I had to add that support in `_typing.pyi`.

Maybe `Index[tuple[Hashable, ...]]` is eligible to be unioned?


```suggestion
    def union(self, other: Sequence[tuple[Hashable, ...]] | Index[tuple[Hashable, ...]] | Self, sort: bool | None = None) -> Self: ...
```

zacharybrownjohn: I don't think `Index[tuple[Hashable, ...]` is valid in this context because `tuple[Hashable, ...]` isn't apart of `SeriesDType` (used by `S1`), which was my original motivation behind messing with S1. 

I can't just do `def union(self, other: Index | list[tuple[Hashable, ...]], sort: bool | None = ...,) -> Self: ...` either because the overloads (seemingly) need to cover as much as the base definition.

cmp0xff: > I don't think `Index[tuple[Hashable, ...]` is valid in this context because `tuple[Hashable, ...]` isn't apart of `SeriesDType` (used by `S1`), which was my original motivation behind messing with S1.

If you check `def union` in `class Index`, you will see that it uses `other: list[HashableT] | Self`, which does not have to do with `S1`. I guess we can follow the example.

Unlike `class Index`, which is generic, `MultiIndex` is not a generic class, so `S1` does not play a role here, and we are allowed to keep things simpler.

> I can't just do `def union(self, other: Index | list[tuple[Hashable, ...]], sort: bool | None = ...,) -> Self: ...` either because the overloads (seemingly) need to cover as much as the base definition.

What is missing?

Dr-Irv: > I can't just do `def union(self, other: Index | list[tuple[Hashable, ...]], sort: bool | None = ...,) -> Self: ...` either because the overloads (seemingly) need to cover as much as the base definition.

This is a case where we can't be specific on the types within the `Index`, so just using `Index` is fine.  It is true that a runtime `Index` with dtype `int` would fail when doing a union with a `MultiIndex` - we just can't catch it at runtime.

I'd also be OK with just leaving out `Index` as a possible type in `MultiIndex.union()`, because people really shouldn't be using an `Index` containing tuples.  The workaround would be to use `Index.to_list()` to convert the index into a list, and then do a `cast` to tell the type checker that you know it is an index of tuples.

This is similar to the issue with having people use `DataFrame.index`, which could be an `Index` or `MultiIndex`.  We can't track it with static typing, so if you know it is a `MultiIndex`, you have to cast it as a user.

zacharybrownjohn: Thank you for that clarification.

zacharybrownjohn: > > I can't just do `def union(self, other: Index | list[tuple[Hashable, ...]], sort: bool | None = ...,) -> Self: ...` either because the overloads (seemingly) need to cover as much as the base definition.
> 
> What is missing?

Other types of list I think (i.e: non-tuple lists).

pandas-stubs/core/indexes/multi.pyi
```diff
@@ -163,3 +163,7 @@ class MultiIndex(Index):
     def insert(self, loc, item): ...
     def delete(self, loc): ...
     def isin(self, values, level=...) -> np_1darray[np.bool]: ...
+    @overload
+    def union(self, other: Self | list[tuple[Hashable, ...]], sort: bool | None = ...,) -> Self: ...
+    @overload
+    def union(self, other: Index | list[HashableT], sort: bool | None = ...,) -> Index: ...
```
cmp0xff: If `MultiIndex.union` does not produce an `Index`, please remove this overload

pandas-stubs/_typing.pyi
```diff
@@ -904,6 +904,7 @@ SeriesDType: TypeAlias = (
     | datetime.datetime  # includes pd.Timestamp
     | datetime.timedelta  # includes pd.Timedelta
 )
+
```
cmp0xff: This extra empty line can also be removed

tests/indexes/test_indexes.py
```diff
@@ -1613,3 +1613,31 @@ def test_to_series() -> None:
         np.complexfloating,
     )
     check(assert_type(Index(["1"]).to_series(), "pd.Series[str]"), pd.Series, str)
+
+
+def test_swaplevel_rettype() -> None:
```
cmp0xff:
```suggestion
def test_multiindex_swaplevel_rettype() -> None:
```

The advantage is, one can run `pytest -k "test_multiindex"` to run `MultiIndex`-related tests only.

tests/indexes/test_indexes.py
```diff
@@ -1613,3 +1613,31 @@ def test_to_series() -> None:
...
+    mi2 = pd.MultiIndex.from_product([["a", "b"], [3, 4]], names=["let", "num"])
+
+    check(
+        assert_type(mi.union(mi2), "pd.MultiIndex"),
+        pd.MultiIndex,
+    )
+    check(
+        assert_type(mi.union(pd.Index([("c", 3), ("d", 4)])), "pd.Index"),
+        pd.Index,
+    )
```
cmp0xff: But the result at runtime is actually a `MultiIndex`, not an `Index`.

If it cannot easily be fixed, you can also remove the test, as said in https://github.com/pandas-dev/pandas-stubs/pull/1437#discussion_r2452100510.

```python
>>> import pandas as pd
>>> mi = pd.MultiIndex.from_product([["a", "b"], [1, 2]], names=["let", "num"])
>>> mi.union(pd.Index([("c", 3), ("d", 4)]))
MultiIndex([('a', 1),
            ('a', 2),
            ('b', 1),
            ('b', 2),
            ('c', 3),
            ('d', 4)],
           )
```

cmp0xff: I see, `pd.Index([("c", 3), ("d", 4)])` actually produces a `MultiIndex`.

```python
>>> import pandas as pd
>>> pd.Index([("c", 3), ("d", 4)])
MultiIndex([('c', 3),
            ('d', 4)],
           )
```

so we should probably discourage `Index[tuple[Hashable, ...]]` and remove it from the overloads.

zacharybrownjohn: > I see, `pd.Index([("c", 3), ("d", 4)])` actually produces a `MultiIndex`.
> so we should probably discourage `Index[tuple[Hashable, ...]]` and remove it from the overloads.

`Index[tuple[Hashable, ...]]` hasn't been supported since I reverted my changes to S1, but that's a helpful find thank you.

realzachbrownjohn:
> Please also run `pre-commit` locally.

My mistake. I had ran a pre-commit after committing, and didn't commit the refactoring it made.

pandas-stubs/core/indexes/multi.pyi
```diff
@@ -163,3 +163,8 @@ class MultiIndex(Index):
     def insert(self, loc, item): ...
     def delete(self, loc): ...
     def isin(self, values, level=...) -> np_1darray[np.bool]: ...
+    def union(
+        self,
+        other: Self | Index | list[HashableT],
+        sort: bool | None = ...,
```
cmp0xff: I suppose `Index` should be removed. It is not tested, either.

```suggestion
        self, other: list[HashableT] | Self, sort: bool | None = ...
```

pandas-stubs/_typing.pyi
cmp0xff: Would you?

tests/indexes/test_indexes.py
```diff
@@ -1601,3 +1601,22 @@ def test_to_series() -> None:
...
+        pd.MultiIndex,
+    )
+    check(
+        assert_type(mi.union(mi2), "pd.MultiIndex"),
+        pd.MultiIndex,
+    )
+    check(
+        assert_type(mi.union([("c", 3), ("d", 4)]), "pd.MultiIndex"),
+        pd.MultiIndex,
+    )
```
cmp0xff: Please make two tests, since `swaplevel` and `union` are two independent functionalities:


```suggestion
def test_multiindex_union() -> None:
    """Test that MultiIndex.union returns MultiIndex"""
    mi = pd.MultiIndex.from_product([["a", "b"], [1, 2]], names=["let", "num"])
    mi2 = pd.MultiIndex.from_product([["a", "b"], [3, 4]], names=["let", "num"])

    check(
        assert_type(mi.union(mi2), "pd.MultiIndex"), pd.MultiIndex
    )
    check(
        assert_type(mi.union([("c", 3), ("d", 4)]), "pd.MultiIndex"), pd.MultiIndex
    )

    
def test_multiindex_swaplevel() -> None:
    """Test that MultiIndex.swaplevel returns MultiIndex"""
    mi = pd.MultiIndex.from_product([["a", "b"], [1, 2]], names=["let", "num"])
    check(
        assert_type(mi.swaplevel(0, 1), "pd.MultiIndex"), pd.MultiIndex
    )
```

pandas-stubs/core/indexes/multi.pyi
zacharybrownjohn: Index can't be removed; the combined scope of the overloaded union definitions in MultiIndex need to at least match the scope of the base definition, and removing Index breaks that. I will add a test for it instead.

pandas-stubs/_typing.pyi
zacharybrownjohn: > Would you?

That extra newline disappeared once I sync'd with the up-to-date master. It shouldn't be there anymore, or at least it isn't on my end.

pandas-stubs/core/indexes/multi.pyi
Dr-Irv: If the type checkers are complaining about mismatched signatures between the parent class declaration `Index.union()` and the subclass `MultiIndex.union()`, you can use `@override`  where you import `override` from `typing_extensions`, since we know we want to override the signature from the base class

zacharybrownjohn: It might be complaining about a signature mismatch?

> Method "union" overrides class "Index" in an incompatible manner
  Parameter 2 type mismatch: base parameter is type "list[HashableT@union] | Index[Any]", override parameter is type "MultiIndex | list[HashableT@union]"
    Type "list[HashableT@union] | Index[Any]" is not assignable to type "MultiIndex | list[HashableT@union]"
      Type "Index[Any]" is not assignable to type "MultiIndex | list[HashableT@union]"
        "Index[Any]" is not assignable to "MultiIndex"
        "Index[Any]" is not assignable to "list[HashableT@union]"

In any case, I tried `@override` above the code to no avail. Gives the same error.

Dr-Irv: OK, but I think there is a better way to do this whole thing.  Don't add `union` to `MultiIndex`.  Rather, change the definition of `union` in `core/indexes/base.pyi` to:
```python
    def union(
        self, other: list[HashableT] | Self, sort: bool | None = None
    ) -> Self: ...
```
I think that will work.

zacharybrownjohn: I agree. Will do that now.

pandas-stubs/_typing.pyi
cmp0xff: Could you try to apply the suggestion?

```suggestion
```

pandas-stubs/py.typed
cmp0xff, line 1: Please restore the file

cmp0xff (changes requested):
Just two small things.

- In addition please also fix `mypy` type checking.

realzachbrownjohn:
> * In addition please also fix `mypy` type checking and further `poe` checks.

Removing the overload of union from the MultiIndex class and instead setting MultiIndex to return Self caused a few existing tests to fail. I would prefer to reinstate the overload to maintain the existing test behaviour, but I could alternatively just modify the existing tests.

Which would you prefer?

cmp0xff:
I believe your changes actually improve the typing in the exising tests, if you take a look. So you can update the static and runtime typing checks in the tests.

In the stubs file for `RangeIndex`, you can try to ignore the error, or use `@override`.

tests/indexes/test_indexes.py
```diff
@@ -1601,3 +1605,19 @@ def test_to_series() -> None:
...
+
+
+def test_multiindex_union() -> None:
+    """Test that MultiIndex.union returns MultiIndex"""
+    mi = pd.MultiIndex.from_product([["a", "b"], [1, 2]], names=["let", "num"])
+    mi2 = pd.MultiIndex.from_product([["a", "b"], [3, 4]], names=["let", "num"])
+
+    check(assert_type(mi.union(mi2), "pd.MultiIndex"), pd.MultiIndex)
+    check(assert_type(mi.union([("c", 3), ("d", 4)]), "pd.MultiIndex"), pd.MultiIndex)
+    check(assert_type(mi.union([1, 2, 3]), "pd.MultiIndex"), pd.MultiIndex)
```
cmp0xff: I think `pd.MultiIndex.from_product([["a", "b"], [1, 2]]).union([1, 2, 3])` does not work, and this line fails at runtime.

zacharybrownjohn: > I think `pd.MultiIndex.from_product([["a", "b"], [1, 2]]).union([1, 2, 3])` does not work, and this line fails at runtime.

I agree, so I assume I just remove that test.

Though there's another test failing:

```
def test_index_union_sort() -> None:
    """Test sort argument in pd.Index.union GH1264."""
    check(
        assert_type(
            pd.Index(["e", "f"]).union(["a", "b", "c"], sort=True), pd.Index[str]
        ),
        pd.Index[str],
    )
    check(
        assert_type(
            pd.Index(["e", "f"]).union(["a", "b", "c"], sort=False), pd.Index[str]
        ),
        pd.Index[str],
    )
```

I had to change mentions of Index to Index[str] for compatibility with the new union return type (Self), but at runtime it errors saying they I'm not allowed to specify the kind of Index.

`FAILED tests/indexes/test_indexes.py::test_index_union_sort - TypeError: 'type' object is not subscriptable`

tests/indexes/test_indexes.py
```diff
@@ -333,12 +333,16 @@ def test_range_index_union() -> None:
...
+        ),
+        pd.Index[str],
     )
     check(
-        assert_type(pd.Index(["e", "f"]).union(["a", "b", "c"], sort=False), pd.Index),
-        pd.Index,
+        assert_type(
+            pd.Index(["e", "f"]).union(["a", "b", "c"], sort=False), pd.Index[str]
+        ),
+        pd.Index[str],
```
cmp0xff: I would write
```suggestion
        assert_type(
            pd.Index(["e", "f"]).union(["a", "b", "c"], sort=True), "pd.Index[str]"
        ),
        pd.Index,
        str,
    )
    check(
        assert_type(
            pd.Index(["e", "f"]).union(["a", "b", "c"], sort=False), "pd.Index[str]"
        ),
        pd.Index,
        str,
```

PR #1439, opened 2025-10-22
complete typing in `BaseGroupBy`

MarcoGorelli:
really minor, looks like this was the only thing missing here


cmp0xff (approved):
Thank you @MarcoGorelli ! However [`.get_group` seems to be deprecated](https://pandas.pydata.org/docs/reference/api/pandas.core.groupby.DataFrameGroupBy.get_group.html). Should we keep it? @Dr-Irv

PR #1445, opened 2025-10-23
refactor(ruff): comply to ANN206, PLC0105, etc.

cmp0xff:
## `*.pyi`

- Now we comply to `ANN206`, `ANN401`, `PLC0105`
- [`ANN401`](https://docs.astral.sh/ruff/rules/any-type/) is permanently ignored

## `tests/*`

- Now we comply to all `ANN` except for `ANN401`
- [`ANN401`](https://docs.astral.sh/ruff/rules/any-type/) and [`ARG`](https://docs.astral.sh/ruff/rules/#flake8-unused-arguments-arg) are permanently ignored

## Known issue

[tests/extension/decimal/array.py](https://github.com/cmp0xff/pandas-stubs/pull/15/files#diff-5ec7af4ad156aaf7b07a217469d004ac340d810e92bdfc98d02f43e8814fe8df) is a bit ugly now, hopefully it's okay. We can improve when we do typing for the arrays.

### DIFF
--- a/tests/extension/decimal/array.py
+++ b/tests/extension/decimal/array.py
@@ -9,6 +9,7 @@ import decimal
 import numbers
 import sys
 from typing import (
+    TYPE_CHECKING,
     Any,
     cast,
     overload,
@@ -40,7 +41,6 @@ from pandas._typing import (
     SequenceIndexer,
     SequenceNotStr,
     TakeIndexer,
-    np_1darray,
 )
 
 from pandas.core.dtypes.base import ExtensionDtype
@@ -49,6 +49,9 @@ from pandas.core.dtypes.common import (
     is_float,
     pandas_dtype,
 )
+
+if TYPE_CHECKING:
+    from pandas._typing import np_1darray
 
 
 @register_extension_dtype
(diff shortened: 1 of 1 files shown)


cmp0xff:
> Not really convinced about the value of forcing ANN on the tests files

The idea originated from https://github.com/pandas-dev/pandas-stubs/pull/1405#pullrequestreview-3302030008. `ANN` on the test files contributes to complying to pyright strict.

loicdiridollou:
Thanks @cmp0xff

tests/extension/decimal/array.py
```diff
@@ -208,21 +240,26 @@ def astype(self, dtype, copy=True):
 
         return super().astype(dtype, copy=copy)
 
-    def __setitem__(self, key, value) -> None:
+    def __setitem__(self, key: object, value: decimal._DecimalNew) -> None:
```
Dr-Irv, line 243: I think this is incorrect.  It could be  `decimal._DecimalNew` or `list[decimal._DecimalNew]`

If you do that, I think the  ignores can be removed, and, if not, you can add `assert isinstance(value, list)`

Or it might need to be another type of `Sequence`

cmp0xff: ec57e2171444b06fb88c5cba361d27a6a3bc718e it is hard, because `Decimal` can be constructed from both a scalar and from `tuple[int, Sequence[int], int]`, which is an iterable itself.

PR #1449, opened 2025-10-24
Fix nightly with pd.DataFrame.rename calling for MultiIndex.fillna

loicdiridollou:

PR replaces the `DataFrame.rename` when using a multi index in the replacer while the dataframe only has `Index`.

### DIFF
--- a/tests/test_frame.py
+++ b/tests/test_frame.py
@@ -2555,8 +2555,10 @@ def test_types_rename() -> None:
     )
 
     df_multiindex = pd.DataFrame(columns=[("a", 1), ("a", 2)])
-    check(assert_type(df_multiindex.rename(columns={(1, 2): ("b", "a")}), pd.DataFrame), pd.DataFrame)
-
+    check(
+        assert_type(df_multiindex.rename(columns={(1, 2): ("b", "a")}), pd.DataFrame),
+        pd.DataFrame,
+    )
 
 
 def test_types_rename_axis() -> None:


tests/test_frame.py
```diff
@@ -2550,11 +2550,16 @@ def test_types_rename() -> None:
     # Apparently all of these calls are accepted by pandas
     check(assert_type(df.rename(columns={None: "b"}), pd.DataFrame), pd.DataFrame)
     check(assert_type(df.rename(columns={"": "b"}), pd.DataFrame), pd.DataFrame)
-    check(assert_type(df.rename(columns={(2, 1): "b"}), pd.DataFrame), pd.DataFrame)
```
cmp0xff, line 2553: Could you change the stub files and invalidate this usage?

loicdiridollou: That would actually break, the issue is at run time when you try to rename a dataframe with an Index when the mapper is designed for a multiindex. Since we do not have the capacity to differenciate 'dataframe with index' and 'dataframe with multiindex' you can't force that behavior in the stubs. Do you see what I mean?

cmp0xff: One possible solution would be to only allow `rename(column=d)`, where `d` is a dictionary containing tuples as keys and values of equal length, so that `columns={(2, 1): "b"}` will be forbidden.

However I couldn't work out how to implement this. For now, I'll leave it as it is.

cmp0xff (approved):
Thank you @loicdiridollou !

PR #1450, opened 2025-10-25
GH1441 Allow for axis argument in DataFrame.pct_change

### LINKED ISSUE #1441: pandas-stubs: Missing `axis` parameter in DataFrame.pct_change() type annotations
**Describe the bug**
The pandas-stubs type annotations for `DataFrame.pct_change()` method do not include the `axis` parameter, even though the actual pandas implementation accepts it via `**kwargs`. According to the pandas documentation, `pct_change()` accepts additional keyword arguments that are passed to `DataFrame.shift` or `Series.shift`, and since `shift()` accepts an `axis` parameter, `pct_change()` should also accept it.

**To Reproduce**
1. Minimal pandas example that should work but fails type checking:
```python
import pandas as pd

# Create sample data
df = pd.DataFrame({
    'A': [100, 110, 120],
    'B': [200, 180, 160]
})

# This works at runtime but fails type checking
result = df[['A', 'B']].pct_change(axis=1)
print(result)
```
2. Type checker: `mypy`

3. Error message:
```error: Unexpected keyword argument "axis" for "pct_change" of "DataFrame"  [call-arg]```

Please complete the following information:
OS: Linux
OS Version: 6.8.0-45-generic
python version: 3.12
version of type checker: mypy 1.18.2
version of installed pandas-stubs: 2.3.2.250926

**Additional context**
- The pandas documentation at https://pandas.pydata.org/docs/reference/api/pandas.DataFrame.pct_change.html clearly states that `pct_change()` accepts `**kwargs` which are passed to `DataFrame.shift` or `Series.shift`
- The `shift()` method accepts an `axis` parameter, so `pct_change()` should also accept it through `**kwargs`
- This is a common pattern in pandas where methods accept additional parameters via `**kwargs` for flexibility
- The current pandas-stubs signature only includes: `periods`, `fill_method`, `freq`, and `fill_value` parameters, but is missing the `axis` parameter that should be available via `**kwargs`

### PR DESCRIPTION
**Pull request by loicdiridollou**
- [x] Closes #1441

### DIFF
--- a/tests/test_frame.py
+++ b/tests/test_frame.py
@@ -2826,7 +2826,6 @@ def test_dataframe_pct_change() -> None:
     check(assert_type(df.pct_change(axis="index"), pd.DataFrame), pd.DataFrame)
 
 
-
 def test_indexslice_setitem() -> None:
     df = pd.DataFrame(
         {"x": [1, 2, 2, 3], "y": [1, 2, 3, 4], "z": [10, 20, 30, 40]}

### CODE REVIEW

File: pandas-stubs/core/frame.pyi [hunk 1]
```diff
@@ -2227,6 +2225,7 @@ class DataFrame(NDFrame, OpsMixin, _GetItemHack):
         fill_method: None = None,
         freq: Frequency | dt.timedelta | None = ...,
         fill_value: Scalar | NAType | None = ...,
+        axis: Axis | None = ...,
```
cmp0xff, line 2228: What about [`**kwargs`](https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.DataFrame.pct_change.html)? Can we just add `**kwargs: Any` or copy-paste all additional arguments from `DataFrame.shift` or `Series.shift`?

loicdiridollou: So using the actual arguments and not the kwargs was something Marco started a while ago - since it helps with readability - to know what needs to be passed, you ensure fuller typing. I have added the axis, the other argument in `shift` method is `suffix` but it does not apply in our case (does not work when period is an int).
So these are all the args we can pass and they are fully typed.
Let me know what you think about the approach.

cmp0xff: I agree that this is better than `**kwargs`.

cmp0xff (approved):
Thank you @loicdiridollou !

Issue #1378, opened 2025-09-25
Create tests for arithmetic for `TimedeltaIndex` and `DatetimeIndex`


cmp0xff:
For `Series[Timedelta]` and `Series[Timestamp]`, comprehensive tests for arithmetic have been added and passed. We would like to have a parallel set of tests for `TimedeltaIndex` and `DatetimeIndex`.

_Originally posted by @Dr-Irv in https://github.com/pandas-dev/pandas-stubs/pull/1273#pullrequestreview-3268015843_

PR #1431, opened 2025-10-16
feat(arithmetic): #1347 #1378 ➗ truediv

cmp0xff:
- [x] Addresses #1347 and #1378
- [x] Tests added

PR #1452, opened 2025-10-28
feat(arithmetic): 🪜 floordiv

cmp0xff:
This PR adds `floordiv` for `Index` and `Series`.
- [x] Addresses #1347 and #1378
- [x] Tests added

### DIFF
--- a/tests/series/arithmetic/test_truediv.py
+++ b/tests/series/arithmetic/test_truediv.py
@@ -175,7 +175,7 @@ def test_truediv_numpy_array(left_i: pd.Series) -> None:
         pd.Series,
     )
     if TYPE_CHECKING_INVALID_USAGE:
-        assert_type(s / left_i, Any)
+        assert_type(s / left_i, Any)  # TODO: numpy/numpy#30173 numpy should reject this
         assert_type(d / left_i, Any)  # pyright: ignore[reportAssertTypeFailure]
 
     check(assert_type(left_i.truediv(b), pd.Series), pd.Series)


tests/indexes/arithmetic/float/test_floordiv.py
```diff
@@ -0,0 +1,131 @@
...
+        _04 = left // s  # type: ignore[operator] # pyright: ignore[reportOperatorIssue]
+        _05 = left // d  # type: ignore[operator] # pyright: ignore[reportOperatorIssue]
+
+    check(assert_type(b // left, "pd.Index[float]"), pd.Index, np.floating)
+    check(assert_type(i // left, "pd.Index[float]"), pd.Index, np.floating)
+    check(assert_type(f // left, "pd.Index[float]"), pd.Index, np.floating)
+    if TYPE_CHECKING_INVALID_USAGE:
+        _13 = c // left  # type: ignore[operator] # pyright: ignore[reportOperatorIssue]
+        _14 = s // left  # type: ignore[operator] # pyright: ignore[reportOperatorIssue]
+        _15 = d // left  # type: ignore[operator] # pyright: ignore[reportOperatorIssue]
```
Dr-Irv: Bit surprised here that `timedelta // left` works but a `list[timedelta] // left` does not..  Shouldn't it have picked up `Index[float].__rfloordiv__(list[timedelta])` ?

tests/indexes/arithmetic/float/test_truediv.py
```diff
@@ -20,31 +29,46 @@ def left() -> "pd.Index[float]":
...
+        _04 = left / s  # type: ignore[operator] # pyright: ignore[reportOperatorIssue]
+        _05 = left / d  # type: ignore[operator] # pyright: ignore[reportOperatorIssue]
 
     check(assert_type(b / left, "pd.Index[float]"), pd.Index, np.floating)
     check(assert_type(i / left, "pd.Index[float]"), pd.Index, np.floating)
     check(assert_type(f / left, "pd.Index[float]"), pd.Index, np.floating)
     check(assert_type(c / left, "pd.Index[complex]"), pd.Index, np.complexfloating)
+    if TYPE_CHECKING_INVALID_USAGE:
+        _14 = s / left  # type: ignore[operator] # pyright: ignore[reportOperatorIssue]
+        _15 = d / left  # type: ignore[operator] # pyright: ignore[reportOperatorIssue]
```
Dr-Irv: same comment as in `indexes/test_floordiv.py` .  This does work:
```python
>>> left = pd.Index([1.0, 2.0, 3.0])
>>> from datetime import timedelta
>>> d = [timedelta(seconds=s) for s in (1, 2, 3)]
>>> d / left
Index([0:00:01, 0:00:01, 0:00:01], dtype='object')
```

tests/indexes/arithmetic/int/test_floordiv.py
```diff
@@ -0,0 +1,127 @@
...
+        _04 = left // s  # type: ignore[operator] # pyright: ignore[reportOperatorIssue]
+        _05 = left // d  # type: ignore[operator] # pyright: ignore[reportOperatorIssue]
+
+    check(assert_type(b // left, "pd.Index[int]"), pd.Index, np.integer)
+    check(assert_type(i // left, "pd.Index[int]"), pd.Index, np.integer)
+    check(assert_type(f // left, "pd.Index[float]"), pd.Index, np.floating)
+    if TYPE_CHECKING_INVALID_USAGE:
+        _13 = c // left  # type: ignore[operator] # pyright: ignore[reportOperatorIssue]
+        _14 = s // left  # type: ignore[operator] # pyright: ignore[reportOperatorIssue]
+        _15 = d // left  # type: ignore[operator] # pyright: ignore[reportOperatorIssue]
```
Dr-Irv: same comment - this should work

tests/indexes/arithmetic/int/test_floordiv.py
```diff
@@ -0,0 +1,127 @@
...
+    # `numpy` typing gives the corresponding `ndarray`s in the static type
+    # checking, where our `__rfloordiv__` cannot override. At runtime, they lead to
+    # errors or pd.Index.
+    check(b // left, pd.Index, np.integer)
+    check(i // left, pd.Index, np.integer)
+    check(f // left, pd.Index, np.floating)
+    if TYPE_CHECKING_INVALID_USAGE:
+        assert_type(c // left, Any)
+        assert_type(s // left, Any)
+        assert_type(d // left, "np.typing.NDArray[np.int64]")
```
Dr-Irv: this works

tests/indexes/arithmetic/int/test_truediv.py
```diff
@@ -20,31 +29,46 @@ def left() -> "pd.Index[int]":
...
+        _04 = left / s  # type: ignore[operator] # pyright: ignore[reportOperatorIssue]
+        _05 = left / d  # type: ignore[operator] # pyright: ignore[reportOperatorIssue]
 
     check(assert_type(b / left, "pd.Index[float]"), pd.Index, np.floating)
     check(assert_type(i / left, "pd.Index[float]"), pd.Index, np.floating)
     check(assert_type(f / left, "pd.Index[float]"), pd.Index, np.floating)
     check(assert_type(c / left, "pd.Index[complex]"), pd.Index, np.complexfloating)
+    if TYPE_CHECKING_INVALID_USAGE:
+        _14 = s / left  # type: ignore[operator] # pyright: ignore[reportOperatorIssue]
+        _15 = d / left  # type: ignore[operator] # pyright: ignore[reportOperatorIssue]
```
Dr-Irv: this works

tests/indexes/arithmetic/test_floordiv.py
```diff
@@ -0,0 +1,127 @@
...
+        _04 = left_i // s  # type: ignore[operator] # pyright: ignore[reportOperatorIssue]
+        _05 = left_i // d  # type: ignore[operator] # pyright: ignore[reportOperatorIssue]
+
+    check(assert_type(b // left_i, pd.Index), pd.Index, np.integer)
+    check(assert_type(i // left_i, pd.Index), pd.Index, np.integer)
+    check(assert_type(f // left_i, pd.Index), pd.Index, np.floating)
+    if TYPE_CHECKING_INVALID_USAGE:
+        _13 = c // left_i  # type: ignore[operator] # pyright: ignore[reportOperatorIssue]
+        _14 = s // left_i  # type: ignore[operator] # pyright: ignore[reportOperatorIssue]
+        _15 = d // left_i  # type: ignore[operator] # pyright: ignore[reportOperatorIssue]
```
Dr-Irv: this works

tests/indexes/arithmetic/test_floordiv.py
```diff
@@ -0,0 +1,127 @@
...
+    # `numpy` typing gives the corresponding `ndarray`s in the static type
+    # checking, where our `__rfloordiv__` cannot override. At runtime, they lead to
+    # errors or pd.Index.
+    check(b // left_i, pd.Index, np.integer)
+    check(i // left_i, pd.Index, np.integer)
+    check(f // left_i, pd.Index, np.floating)
+    if TYPE_CHECKING_INVALID_USAGE:
+        assert_type(c // left_i, Any)
+        assert_type(s // left_i, Any)
+        assert_type(d // left_i, Any)  # pyright: ignore[reportAssertTypeFailure]
```
Dr-Irv: this works

tests/series/arithmetic/float/test_floordiv.py
```diff
@@ -13,166 +17,239 @@
...
 
-        _2 = b // left  # type: ignore[operator] # pyright: ignore[reportOperatorIssue]
+    check(assert_type(b // left, "pd.Series[float]"), pd.Series, np.floating)
     check(assert_type(i // left, "pd.Series[float]"), pd.Series, np.floating)
     check(assert_type(f // left, "pd.Series[float]"), pd.Series, np.floating)
     if TYPE_CHECKING_INVALID_USAGE:
-        _3 = c // left  # type: ignore[operator] # pyright: ignore[reportOperatorIssue]
+        _13 = c // left  # type: ignore[operator] # pyright: ignore[reportOperatorIssue]
+        _14 = s // left  # type: ignore[operator] # pyright: ignore[reportOperatorIssue]
+        _15 = d // left  # type: ignore[operator] # pyright: ignore[reportOperatorIssue]
```
Dr-Irv: this works

tests/series/arithmetic/int/test_floordiv.py
```diff
@@ -13,166 +17,237 @@
...
 
-        _2 = b // left  # type: ignore[operator] # pyright: ignore[reportOperatorIssue]
+    check(assert_type(b // left, "pd.Series[int]"), pd.Series, np.integer)
     check(assert_type(i // left, "pd.Series[int]"), pd.Series, np.integer)
     check(assert_type(f // left, "pd.Series[float]"), pd.Series, np.floating)
     if TYPE_CHECKING_INVALID_USAGE:
-        _3 = c // left  # type: ignore[operator] # pyright: ignore[reportOperatorIssue]
+        _13 = c // left  # type: ignore[operator] # pyright: ignore[reportOperatorIssue]
+        _14 = s // left  # type: ignore[operator] # pyright: ignore[reportOperatorIssue]
+        _15 = d // left  # type: ignore[operator] # pyright: ignore[reportOperatorIssue]
```
Dr-Irv: this works

tests/series/arithmetic/int/test_floordiv.py
```diff
@@ -13,166 +17,237 @@
...
+    # `numpy` typing gives the corresponding `ndarray`s in the static type
+    # checking, where our `__rfloordiv__` cannot override. At runtime, they lead to
+    # errors or pd.Series.
+    check(b // left, pd.Series, np.integer)
     check(i // left, pd.Series, np.integer)
     check(f // left, pd.Series, np.floating)
     if TYPE_CHECKING_INVALID_USAGE:
         assert_type(c // left, Any)
+        assert_type(s // left, Any)
+        assert_type(d // left, "np.typing.NDArray[np.int64]")
```
Dr-Irv: this works

tests/series/arithmetic/test_floordiv.py
```diff
@@ -0,0 +1,253 @@
...
+        _04 = left_i // s  # type: ignore[operator] # pyright: ignore[reportOperatorIssue]
+        _05 = left_i // d  # type: ignore[operator] # pyright: ignore[reportOperatorIssue]
+
+    check(assert_type(b // left_i, pd.Series), pd.Series, np.integer)
+    check(assert_type(i // left_i, pd.Series), pd.Series, np.integer)
+    check(assert_type(f // left_i, pd.Series), pd.Series, np.floating)
+    if TYPE_CHECKING_INVALID_USAGE:
+        _13 = c // left_i  # type: ignore[operator] # pyright: ignore[reportOperatorIssue]
+        _14 = s // left_i  # type: ignore[operator] # pyright: ignore[reportOperatorIssue]
+        _15 = d // left_i  # type: ignore[operator] # pyright: ignore[reportOperatorIssue]
```
Dr-Irv: this works

tests/series/arithmetic/test_floordiv.py
```diff
@@ -0,0 +1,253 @@
...
+    # `numpy` typing gives the corresponding `ndarray`s in the static type
+    # checking, where our `__rfloordiv__` cannot override. At runtime, they lead to
+    # errors or pd.Series.
+    check(b // left_i, pd.Series, np.integer)
+    check(i // left_i, pd.Series, np.integer)
+    check(f // left_i, pd.Series, np.floating)
+    if TYPE_CHECKING_INVALID_USAGE:
+        assert_type(c // left_i, Any)
+        assert_type(s // left_i, Any)
+        assert_type(d // left_i, Any)  # pyright: ignore[reportAssertTypeFailure]
```
Dr-Irv: this works

tests/series/arithmetic/test_floordiv.py
```diff
@@ -0,0 +1,253 @@
...
+
+    check(assert_type(left_i.floordiv(b), pd.Series), pd.Series, np.integer)
+    check(assert_type(left_i.floordiv(i), pd.Series), pd.Series, np.integer)
+    check(assert_type(left_i.floordiv(f), pd.Series), pd.Series, np.floating)
+    if TYPE_CHECKING_INVALID_USAGE:
+        left_i.floordiv(c)  # type: ignore[arg-type] # pyright: ignore[reportArgumentType,reportCallIssue]
+        left_i.floordiv(s)  # type: ignore[arg-type] # pyright: ignore[reportArgumentType,reportCallIssue]
+        # left_i.floordiv(d)  # This invalid one cannot be detected by static type checking
```
Dr-Irv: This is very surprising.  You were able to detect that `left_i // d` is invalid, so you should be able to do this one.

tests/series/arithmetic/test_floordiv.py
```diff
@@ -0,0 +1,253 @@
...
+    s = pd.Index([datetime(2025, 10, d) for d in (27, 28, 29)])
+    d = pd.Index([timedelta(seconds=s + 1) for s in range(3)])
+
+    check(assert_type(left_i // b, pd.Series), pd.Series, np.integer)
+    check(assert_type(left_i // i, pd.Series), pd.Series, np.integer)
+    check(assert_type(left_i // f, pd.Series), pd.Series, np.floating)
+    if TYPE_CHECKING_INVALID_USAGE:
+        _03 = left_i // c  # type: ignore[operator] # pyright: ignore[reportOperatorIssue]
+        _04 = left_i // s  # type: ignore[operator] # pyright: ignore[reportOperatorIssue]
+        # _05 = left_i // d  # This invalid one cannot be detected by static type checking
```
Dr-Irv: A bit surprising, but then I see that you didn't add `s` and `d` in the tests in `tests/series/arithmetic/test_truediv.py`, so you should add that and things should be the same between floordiv and truediv

tests/series/arithmetic/test_floordiv.py
```diff
@@ -0,0 +1,253 @@
...
+        _14 = s // left_i  # type: ignore[operator] # pyright: ignore[reportOperatorIssue]
+    check(assert_type(d // left_i, "pd.Series[pd.Timedelta]"), pd.Series, pd.Timedelta)
+
+    check(assert_type(left_i.floordiv(b), pd.Series), pd.Series, np.integer)
+    check(assert_type(left_i.floordiv(i), pd.Series), pd.Series, np.integer)
+    check(assert_type(left_i.floordiv(f), pd.Series), pd.Series, np.floating)
+    if TYPE_CHECKING_INVALID_USAGE:
+        left_i.floordiv(c)  # type: ignore[arg-type] # pyright: ignore[reportArgumentType,reportCallIssue]
+        left_i.floordiv(s)  # type: ignore[arg-type] # pyright: ignore[reportArgumentType,reportCallIssue]
+        # left_i.floordiv(d)  # This invalid one cannot be detected by static type checking
```
Dr-Irv: surprised at this

tests/series/arithmetic/test_floordiv.py
```diff
@@ -0,0 +1,253 @@
...
+    s = pd.Series([datetime(2025, 10, d) for d in (27, 28, 29)])
+    d = pd.Series([timedelta(seconds=s + 1) for s in range(3)])
+
+    check(assert_type(left_i // b, pd.Series), pd.Series, np.integer)
+    check(assert_type(left_i // i, pd.Series), pd.Series, np.integer)
+    check(assert_type(left_i // f, pd.Series), pd.Series, np.floating)
+    if TYPE_CHECKING_INVALID_USAGE:
+        _03 = left_i // c  # type: ignore[operator] # pyright: ignore[reportOperatorIssue]
+        _04 = left_i // s  # type: ignore[operator] # pyright: ignore[reportOperatorIssue]
+        # _05 = left_i // d  # This invalid one cannot be detected by static type checking
```
Dr-Irv: same comment

tests/series/arithmetic/test_floordiv.py
```diff
@@ -0,0 +1,253 @@
...
+        _14 = s // left_i  # type: ignore[operator] # pyright: ignore[reportOperatorIssue]
+    check(assert_type(d // left_i, pd.Series), pd.Series, pd.Timedelta)
+
+    check(assert_type(left_i.floordiv(b), pd.Series), pd.Series, np.integer)
+    check(assert_type(left_i.floordiv(i), pd.Series), pd.Series, np.integer)
+    check(assert_type(left_i.floordiv(f), pd.Series), pd.Series, np.floating)
+    if TYPE_CHECKING_INVALID_USAGE:
+        left_i.floordiv(c)  # type: ignore[arg-type] # pyright: ignore[reportArgumentType,reportCallIssue]
+        left_i.floordiv(s)  # type: ignore[arg-type] # pyright: ignore[reportArgumentType,reportCallIssue]
+        # left_i.floordiv(d)  # This invalid one cannot be detected by static type checking
```
Dr-Irv, line 265: same comment

tests/indexes/arithmetic/float/test_floordiv.py
cmp0xff: This has been discussed in https://github.com/pandas-dev/pandas-stubs/pull/1431#pullrequestreview-3356679891. The problem is that the result is `Index[Timestamp]`, with element type `Timestamp`, but numpy dtype `object`. In #1431 we've decided to ban such usages.

tests/indexes/arithmetic/float/test_truediv.py
cmp0xff: This has been discussed in https://github.com/pandas-dev/pandas-stubs/pull/1431#pullrequestreview-3356679891. The problem is that the result is `Index[Timestamp]`, with element type `Timestamp`, but numpy dtype `object`. In #1431 we've decided to ban such usages.

tests/indexes/arithmetic/int/test_floordiv.py
cmp0xff: This has been discussed in https://github.com/pandas-dev/pandas-stubs/pull/1431#pullrequestreview-3356679891. The problem is that the result is `Index[Timestamp]`, with element type `Timestamp`, but numpy dtype `object`. In #1431 we've decided to ban such usages.

tests/indexes/arithmetic/int/test_floordiv.py
cmp0xff: c95aa395feaf6ec1f0da7e61f39469a2e0288508

tests/indexes/arithmetic/int/test_truediv.py
cmp0xff: This has been discussed in https://github.com/pandas-dev/pandas-stubs/pull/1431#pullrequestreview-3356679891. The problem is that the result is `Index[Timestamp]`, with element type `Timestamp`, but numpy dtype `object`. In #1431 we've decided to ban such usages.

tests/series/arithmetic/test_floordiv.py
cmp0xff: The result is `Series` with `dtype` `object`, not `timedelta64` as expected. From pandas-dev/pandas-stubs#1431 I guess we also ban it.

tests/series/arithmetic/float/test_floordiv.py
cmp0xff: The result is `Series` with `dtype` `object`, not `timedelta64` as expected. From pandas-dev/pandas-stubs#1431 I guess we also ban it.

tests/series/arithmetic/test_floordiv.py
cmp0xff: The result is `Series` with `dtype` `object`, not `timedelta64` as expected. From pandas-dev/pandas-stubs#1431 I guess we also ban it.

cmp0xff (commented):
>     1. You're inconsistent when having `timedelta` divided by `int` where "divided" is either regular division or floor division, and `timedelta` is a single element or a sequence of some type.  You have that as being not allowed in the stubs, but they work at runtime.  Sometimes you do allow it, so you have to fix that.

When the runtime result has the correct `dtype`, this PR should give type hints. When the operation crashes at runtime or gives `object` as `dtype`, this PR should show error. This was discussed in e.g. https://github.com/pandas-dev/pandas-stubs/pull/1431#discussion_r2445222090, where `Index[Timedelta]` (which has `dtype` `object`) is banned.

>     2. There are a few places where you say you can't detect via static typing, and I'm surprised at that.

Still working on them. Will ping when finished.


>     1. You changed the `pyrefly` override syntax

Will do separately

>     2. You changed some imports from `from pandas import Series` to `from pandas.core.series import Series`

Will do separately.

tests/indexes/arithmetic/float/test_floordiv.py
Dr-Irv: actually, I was wrong here.  It does fail so your change is correct here.  I thought I tested it yesterday...

tests/indexes/arithmetic/float/test_truediv.py
Dr-Irv: In this case, can we just return `Index` without specifying what is inside?  i.e., `Index[Any]` or just `Index` ?

tests/indexes/arithmetic/int/test_floordiv.py
Dr-Irv: In this case, can we just return `Index` without specifying what is inside? i.e., `Index[Any]` or just `Index` ?

tests/indexes/arithmetic/int/test_truediv.py
Dr-Irv: In this case, can we just return `Index` without specifying what is inside? i.e., `Index[Any]` or just `Index` ?

tests/series/arithmetic/float/test_floordiv.py
Dr-Irv: actually, I was mistaken.  I just tried it and it doesn't work

tests/series/arithmetic/int/test_floordiv.py
Dr-Irv: same comment - I think we should just return `Index` or `Index[Any]` in the stubs

tests/indexes/arithmetic/test_floordiv.py
Dr-Irv: In this case, can we just return `Index` without specifying what is inside? i.e., `Index[Any]` or just `Index` ?

tests/indexes/arithmetic/test_floordiv.py
Dr-Irv: and it returns `TimedeltaIndex`

tests/series/arithmetic/int/test_floordiv.py
Dr-Irv: and returns `Series[Timedelta]`

tests/series/arithmetic/test_floordiv.py
Dr-Irv: So can we just return `Series[Any]` or `Series` in this case?

tests/series/arithmetic/test_floordiv.py
Dr-Irv:
```python
>>> left = pd.DataFrame({"a": [1, 2, 3]})["a"]
>>> d = np.array([np.timedelta64(s + 1, "s") for s in range(3)], np.timedelta64)
>>> d // left
0   0 days 00:00:01
1   0 days 00:00:01
2   0 days 00:00:01
Name: a, dtype: timedelta64[s]
```
So it is returning `Series[Timedelta]`

Dr-Irv:
> When the runtime result has the correct `dtype`, this PR should give type hints. When the operation crashes at runtime or gives `object` as `dtype`, this PR should show error. This was discussed in e.g. [#1431 (comment)](https://github.com/pandas-dev/pandas-stubs/pull/1431#discussion_r2445222090), where `Index[Timedelta]` (which has `dtype` `object`) is banned.

I think this situation is different.  When we banned `Index[Timedelta]` when the `dtype` was `object`, my concern was that you were doing tests using that type, and we don't want to assume it could be created.

Here, this is about the results of the computation.  If we know the result has `object` dtype, I think the stubs can return a result without a specified value of `S1` (i.e., either `Index` or `Index[Any]`; `Series` or `Series[Any]`)

tests/series/arithmetic/test_floordiv.py
cmp0xff: So `Series[Timedelta] // Series[Any]` gives `Series`, and `Series[Timedelta] / Series[Any]` also gives `Series`?

Dr-Irv: yes

tests/indexes/arithmetic/float/test_truediv.py
cmp0xff: Now fixed to be `Index`

tests/indexes/arithmetic/int/test_floordiv.py
cmp0xff: Now fixed to be `Index`, but blocked by pandas-dev/pandas#63007

tests/series/arithmetic/int/test_floordiv.py
cmp0xff: Now fixed to be `Index`

tests/series/arithmetic/int/test_floordiv.py
cmp0xff: Now fixed to be `Series[Timedelta]`

tests/series/arithmetic/test_floordiv.py
cmp0xff: Now fixed to be `Series[Any]`

tests/series/arithmetic/test_floordiv.py
cmp0xff: ef55cacc2463122113df6429e2d706a3d47dd4c2

tests/series/arithmetic/test_floordiv.py
cmp0xff: Now fixed to be `Never`

tests/series/arithmetic/test_floordiv.py
cmp0xff: Now fixed to be `Never`

tests/series/arithmetic/test_floordiv.py
cmp0xff: Now fixed to be `Never`

tests/series/arithmetic/test_floordiv.py
cmp0xff: I cannot fix this one. We need `Series[Timedelta].__rfloordiv__(Series[Timedelta])`, giving `Series[int]`. This disables us from forbidding `Series[Any].__rfloordiv__(Series[Timedelta])`, either giving error or `Never` (`mypy` thinks they are incompatible, thus giving `Any` in both cases)

tests/series/arithmetic/test_floordiv.py
cmp0xff: I cannot fix this one. We need `Series[Timedelta].rfloordiv(Series[Timedelta])`, giving `Series[int]`. This disables us from forbidding `Series[Any].rfloordiv(Series[Timedelta])`, either giving error or `Never` (`mypy` thinks they are incompatible, thus giving `Any` in both cases)

tests/indexes/arithmetic/int/test_truediv.py
cmp0xff: Now fixed to be `Index[Any]`

tests/indexes/arithmetic/test_floordiv.py
cmp0xff: Now fixed to be `Index[Any]`

tests/indexes/arithmetic/test_floordiv.py
cmp0xff: Now fixed to be `TimedeltaIndex` (`numpy`'s `__floordiv__` has higher prio than ours)

tests/series/arithmetic/test_floordiv.py
Dr-Irv: OK - I see.  With `Series[Any]`, we can't detect everything in a static context.

pandas-stubs/core/base.pyi
```diff
@@ -168,7 +172,84 @@ class IndexOpsMixin(OpsMixin, Generic[S1, GenericT_co]):
...
+    | np.timedelta64
+    | Sequence[timedelta | np.timedelta64]
+    | ArrayIndexTimedeltaNoSeq
+)
+ArrayIndexSeriesTimedeltaNoSeq: TypeAlias = ArrayIndexTimedeltaNoSeq | Series[Timedelta]
+ScalarArrayIndexSeriesTimedelta: TypeAlias = (
+    ScalarArrayIndexTimedelta | Series[Timedelta]
+)
+
+NumListLike: TypeAlias = (  # deprecated, do not use
```
Dr-Irv: can you create an issue to track removing this?

tests/series/arithmetic/test_floordiv.py
```diff
@@ -138,22 +143,35 @@ def test_floordiv_numpy_array(left_i: pd.Series) -> None:
...
     check(assert_type(left_i.floordiv(b), pd.Series), pd.Series, np.integer)
     check(assert_type(left_i.floordiv(i), pd.Series), pd.Series, np.integer)
     check(assert_type(left_i.floordiv(f), pd.Series), pd.Series, np.floating)
-    if TYPE_CHECKING_INVALID_USAGE:
+
+    def _23() -> None:  # pyright: ignore[reportUnusedFunction]
         left_i.floordiv(c)  # type: ignore[arg-type] # pyright: ignore[reportArgumentType,reportCallIssue]
+
+    def _24() -> None:  # pyright: ignore[reportUnusedFunction]
         left_i.floordiv(s)  # type: ignore[arg-type] # pyright: ignore[reportArgumentType,reportCallIssue]
```
Dr-Irv: If these 2 are returning `Never`, then you can do an `assert_type()` as you did in `_25()`.  If we are detecting it at typing time as being invalid (which I think you are doing here), then it should be in a `if TYPE_CHECKING_INVALID_USAGE` block

tests/series/arithmetic/test_floordiv.py
```diff
@@ -138,22 +143,35 @@ def test_floordiv_numpy_array(left_i: pd.Series) -> None:
...
     check(assert_type(left_i.rfloordiv(b), pd.Series), pd.Series, np.integer)
     check(assert_type(left_i.rfloordiv(i), pd.Series), pd.Series, np.integer)
     check(assert_type(left_i.rfloordiv(f), pd.Series), pd.Series, np.floating)
-    if TYPE_CHECKING_INVALID_USAGE:
+
+    def _33() -> None:  # pyright: ignore[reportUnusedFunction]
         left_i.rfloordiv(c)  # type: ignore[arg-type] # pyright: ignore[reportArgumentType,reportCallIssue]
+
+    def _34() -> None:  # pyright: ignore[reportUnusedFunction]
         left_i.rfloordiv(s)  # type: ignore[arg-type] # pyright: ignore[reportArgumentType,reportCallIssue]
```
Dr-Irv: same comment as above with respect to `_23()` and `_24()`

tests/series/arithmetic/test_floordiv.py
```diff
@@ -235,8 +257,11 @@ def test_floordiv_pd_series(left_i: pd.Series) -> None:
     check(assert_type(left_i.floordiv(b), pd.Series), pd.Series, np.integer)
     check(assert_type(left_i.floordiv(i), pd.Series), pd.Series, np.integer)
     check(assert_type(left_i.floordiv(f), pd.Series), pd.Series, np.floating)
-    if TYPE_CHECKING_INVALID_USAGE:
+
+    def _23() -> None:  # pyright: ignore[reportUnusedFunction]
         left_i.floordiv(c)  # type: ignore[arg-type] # pyright: ignore[reportArgumentType,reportCallIssue]
```
Dr-Irv: should be surrounded by `if TYPE_CHECKING_INVALID_USAGE:`

tests/series/arithmetic/test_truediv.py
```diff
@@ -121,26 +174,45 @@ def test_truediv_numpy_array(left_i: pd.Series) -> None:
         assert_type(c / left_i, Any),  # pyright: ignore[reportAssertTypeFailure]
         pd.Series,
     )
+    if TYPE_CHECKING_INVALID_USAGE:
+        assert_type(s / left_i, Any)
```
Dr-Irv: surprised that `pyright` isn't detecting this one

tests/series/arithmetic/test_truediv.py
```diff
@@ -150,42 +222,66 @@ def test_truediv_pd_index(left_i: pd.Series) -> None:
...
+    d = pd.Index([timedelta(seconds=s) for s in range(3)])
 
     check(assert_type(left_i / a, pd.Series), pd.Series)
     check(assert_type(left_i / b, pd.Series), pd.Series)
     check(assert_type(left_i / i, pd.Series), pd.Series)
     check(assert_type(left_i / f, pd.Series), pd.Series)
     check(assert_type(left_i / c, pd.Series), pd.Series)
+    if TYPE_CHECKING_INVALID_USAGE:
+        _05 = left_i / s  # type: ignore[operator] # pyright: ignore[reportOperatorIssue]
+        assert_type(left_i / d, Never)
```
Dr-Irv, line 235: wouldn't you need to put this in a function?

tests/series/arithmetic/test_truediv.py
```diff
@@ -271,8 +387,8 @@ def test_truediv_str_py_str(left_i: pd.Series) -> None:
     s = "abc"
 
     if TYPE_CHECKING_INVALID_USAGE:
-        _0 = left_i / s  # type: ignore[operator] # pyright:ignore[reportOperatorIssue]
-        _1 = s / left_i  # type: ignore[operator] # pyright:ignore[reportOperatorIssue]
+        _00 = left_i / s  # type: ignore[operator] # pyright:ignore[reportOperatorIssue]
+        _01: pd.Series = s / left_i  # type: ignore[operator] # pyright:ignore[reportOperatorIssue]
```
Dr-Irv: why include the `pd.Series` annotation here?

tests/series/arithmetic/int/test_floordiv.py
Dr-Irv: We both meant `Series`  :-)

tests/series/arithmetic/test_floordiv.py
Dr-Irv: OK, that makes sense because of the `Series[Any]` issue

Dr-Irv (changes requested):
pretty close.  Only a few things left

tests/series/arithmetic/test_floordiv.py
cmp0xff: `mypy` thinks they are both `Never` and invalid. `pyright` thinks they are simply invalid. bddadcbcb1b5d5dc12aff4460b1987d34c2e9168

tests/series/arithmetic/test_floordiv.py
cmp0xff: Moved to `TYPE_CHECKING_INVALID_USAGE`

tests/series/arithmetic/test_floordiv.py
cmp0xff: Moved to `TYPE_CHECKING_INVALID_USAGE`

tests/series/arithmetic/test_truediv.py
cmp0xff: I don't have further insights 😔

tests/series/arithmetic/test_truediv.py
cmp0xff: It is arithmetic, not function call, so it's fine

tests/series/arithmetic/test_truediv.py
cmp0xff: I was struggling with `mypy`. Now removed bddadcbcb1b5d5dc12aff4460b1987d34c2e9168

pandas-stubs/core/base.pyi
cmp0xff: pandas-dev/pandas-stubs#1474

tests/series/arithmetic/test_truediv.py
Dr-Irv: It's on the `numpy` side.  pyright sees the type of `s` as  `"ndarray[tuple[Any, ...], dtype[datetime64[date | int | None]]]"` .  `numpy` has this overload that matches:
```python
    @overload
    def __truediv__(self: NDArray[Any], other: _ArrayLikeObject_co, /) -> Any: ...
```
Can you add a comment that indicates that `numpy` is not picking up that dividing a numpy array by anything and link to this issue I just created?  https://github.com/numpy/numpy/issues/30173

cmp0xff: b85a68a99ece4ce072c4ca4c602991861ffab2bd

PR #1453, opened 2025-10-28
Type `RangeIndex`

### LINKED ISSUE #1442: Typing RangeIndex

- [x] `__new__`
- [x] max
- [x] min
- [x] from_range
- [x] `__reduce__` (undocumented, can probably be removed?)
- [x] get_indexer
- [x] tolist
- [x] argsort
- [x] factorize
- [x] equals
- [x] join
- [x] `__floordiv__`

Some of these might be removable in favour of the parent implementation

---

PR incoming

MarcoGorelli:

closes #1442

### DIFF
--- a/pandas-stubs/core/indexes/base.pyi
+++ b/pandas-stubs/core/indexes/base.pyi
@@ -451,7 +451,7 @@ class Index(IndexOpsMixin[S1], ElementOpsMixin[S1]):
         *,
         how: JoinHow = "left",
         level: Level | None = None,
-        return_indexers: Literal[False] = ...,
+        return_indexers: Literal[False] = False,
         sort: bool = False,
     ) -> Index: ...
     @property


pandas-stubs/core/indexes/base.pyi
```diff
@@ -428,15 +434,26 @@ class Index(IndexOpsMixin[S1], ElementOpsMixin[S1]):
...
+        sort: bool = False,
+    ) -> tuple[Index, np_1darray[np.intp] | None, np_1darray[np.intp] | None]: ...
+    @overload
+    def join(
+        self,
+        other: Index,
+        *,
+        how: JoinHow = "left",
+        level: Level | None = None,
+        return_indexers: Literal[False] = ...,
```
cmp0xff: Maybe just add the default value
```suggestion
        return_indexers: Literal[False] = False,
```

pandas-stubs/core/indexes/base.pyi
```diff
@@ -428,15 +434,26 @@ class Index(IndexOpsMixin[S1], ElementOpsMixin[S1]):
...
-    ): ...
+        how: JoinHow = "left",
+        level: Level | None = None,
+        return_indexers: Literal[True],
+        sort: bool = False,
+    ) -> tuple[Index, np_1darray[np.intp] | None, np_1darray[np.intp] | None]: ...
+    @overload
+    def join(
+        self,
+        other: Index,
```
cmp0xff, line 450: Just a comment, it also works with a `Series`, which is unspecified. Reported in pandas-dev/pandas#62897

pandas-stubs/core/indexes/base.pyi
```diff
@@ -418,8 +420,12 @@ class Index(IndexOpsMixin[S1], ElementOpsMixin[S1]):
     ) -> Self: ...
     def get_loc(self, key: Label) -> int | slice | np_1darray[np.bool]: ...
     def get_indexer(
-        self, target, method: ReindexMethod | None = ..., limit=..., tolerance=...
-    ): ...
+        self,
+        target: Index,
+        method: ReindexMethod | None = None,
+        limit: int | None = None,
+        tolerance: Scalar | AnyArrayLike | Sequence[Scalar] | None = None,
```
MarcoGorelli, line 427: taken from 

https://github.com/pandas-dev/pandas-stubs/blob/5e16b863eaec6f34203035a470722111ee1a1729/pandas-stubs/core/series.pyi#L1165

pandas-stubs/core/indexes/base.pyi
```diff
@@ -428,15 +434,26 @@ class Index(IndexOpsMixin[S1], ElementOpsMixin[S1]):
...
-        how: _str = ...,
-        level=...,
-        return_indexers: bool = ...,
-        sort: bool = ...,
-    ): ...
+        how: JoinHow = "left",
+        level: Level | None = None,
+        return_indexers: Literal[True],
+        sort: bool = False,
+    ) -> tuple[Index, np_1darray[np.intp] | None, np_1darray[np.intp] | None]: ...
```
cmp0xff, line 446: It is interesting to understand when the `indexer`s can be `None`. But we can do it in a separate PR.

pandas-stubs/core/indexes/range.pyi
```diff
@@ -55,29 +57,15 @@ class RangeIndex(_IndexSubclassBase[int, np.int64]):
     @property
     def has_duplicates(self) -> bool: ...
     def __contains__(self, key: int | np.integer) -> bool: ...
-    @final
-    def get_indexer(self, target, method=..., limit=..., tolerance=...): ...
-    def tolist(self): ...
-    def min(self, axis=..., skipna: bool = ..., *args: Any, **kwargs: Any): ...
-    def max(self, axis=..., skipna: bool = ..., *args: Any, **kwargs: Any): ...
-    def argsort(self, *args: Any, **kwargs: Any): ...
```
MarcoGorelli, line 63: i'm removing these so the parent definition is used, there's no need to override them right?

the only one, AFAICT, that needs overriding is `__floordiv__`, which I've kept

pandas-stubs/core/indexes/range.pyi
```diff
@@ -25,16 +26,17 @@ from pandas._typing import (
...
-        start: int | RangeIndex | range = ...,
-        stop: int = ...,
-        step: int = ...,
-        dtype=...,
-        copy: bool = ...,
-        name: Hashable = ...,
+        start: int | RangeIndex | range | None = None,
+        stop: int | None = None,
+        step: int | None = None,
+        dtype: Dtype | None = None,
```
cmp0xff, line 32: https://pandas.pydata.org/docs/dev/reference/api/pandas.RangeIndex.html

```suggestion
        start: int | range | Self = 0,
        stop: int = 1,
        step: int = 1,
        dtype: Literal[np.int64] = np.int64,
```

pandas-stubs/core/indexes/range.pyi
cmp0xff: pandas-dev/pandas-stubs#1452 can help with `__floordiv__`, but I guess this PR will be merged first

pandas-stubs/core/indexes/base.pyi
MarcoGorelli: not sure this can be determined statically, e.g.
```py
In [16]: s = pd.Index([1,2,3])

In [17]: s.join(s, return_indexers=True)
Out[17]: (Index([1, 2, 3], dtype='int64'), None, None)

In [18]: s.join(s+1, return_indexers=True)
Out[18]: (Index([1, 2, 3], dtype='int64'), None, array([-1,  0,  1]))
```

pandas-stubs/core/indexes/range.pyi
MarcoGorelli: true, the docs say that, but the code has `None` as the default

https://github.com/pandas-dev/pandas/blob/e11bf72296ed77c7de3183277f04f6693d19cb54/pandas/core/indexes/range.py#L154-L162

do you have a policy of which to pick?

cmp0xff: Reported in pandas-dev/pandas#62899. How do we proceed from here, @Dr-Irv ?

cmp0xff: OK, it was replied that the docs will be fixed. I think we can stick to the source code.

PR #1456, opened 2025-10-29
Index: ensure Index(DatetimeIndex) returns DatetimeIndex (fixes #1440)

### LINKED ISSUE #1440: `Index(DatetimeIndex)` erroneously gives `Index[Timestamp]`
On 01d597ff0622a055abdfc5d3a834c3f6dd744c84, the stubs give `Index[Timestamp]` when providing a `DatetimeIndex` as `data`. The result should be `DatetimeIndex`.

```python
import pandas as pd
from tests import check
from typing import assert_type

data = check(
    assert_type(pd.date_range("2022-01-01", "2022-01-31", freq="D"), pd.DatetimeIndex),
    pd.DatetimeIndex,  # fine
)
check(
    assert_type(pd.Index(data, name="date"), "pd.Index[pd.Timestamp]"), pd.DatetimeIndex  # static typing is wrong
)
```

This affects `test_index_relops`.

yusufcank34:
- [ ] Closes #1440 
- [ ] Tests added: Please use `assert_type()` to assert the type of any return value.

### DIFF
--- a/tests/indexes/test_datetime_index_1440.py
+++ b/tests/indexes/test_datetime_index_1440.py
@@ -1,8 +1,0 @@
-from typing import assert_type
-import pandas as pd
-
-# GH#1440 – constructing Index from DatetimeIndex should return DatetimeIndex
-data = pd.date_range("2022-01-01", "2022-01-03", freq="D")
-idx = pd.Index(data, name="date")
-
-assert_type(idx, pd.DatetimeIndex)


pandas-stubs/core/indexes/base.pyi
```diff
@@ -275,6 +274,16 @@ class Index(IndexOpsMixin[S1], ElementOpsMixin[S1]):
         name: Hashable = ...,
         tupleize_cols: bool = ...,
     ) -> IntervalIndex[Interval[Any]]: ...
+    @overload
+    def __new__(
```
loicdiridollou, line 280: This was added too low, as you can see the tests are failing meaning that your changes have not changed how the type checker identifies the types of an index of a DatetimeIndex.
What you will likely need is to drop this and add `DatetimeIndex` at line 190 along the sequence and IndexOpsMixIn of datetime.

tests/indexes/test_datetime_index_1440.py
```diff
@@ -0,0 +1,8 @@
+from typing import assert_type
```
loicdiridollou: Since there is already a test file for datetime index it is probably cleaner to consolidate your tests there (if we made one test file by issue it would quickly become overwhelming).
Another point is that you’d need to add `check(assert_type(…), pd.DatetimeIndex)` in top of the assertion you have already added here.

loicdiridollou (approved):
Thanks @yusufcank34, nice job on your first PR!

tests/indexes/test_datetime_index_1440.py
loicdiridollou: @yusufcank34 Can you remove that file since it is already covered by the other test you modified? And run pre-commit before you push again, I will merge after it.

loicdiridollou (approved):

PR #1459, opened 2025-10-30
Type CategoricalIndex

### LINKED ISSUE #1443: Typing CategoricalIndex

- [ ] `__new__`
- [ ] equals
- [ ] values
- [ ] `__contains__`
- [ ] unique
- [ ] reindex
- [ ] get_indexer
- [ ] get_indexer_non_unique
- [ ] delete
- [ ] insert

MarcoGorelli:
closes #1443 

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -204,10 +204,6 @@ ignore = [
   "PYI042",  # https://docs.astral.sh/ruff/rules/snake-case-type-alias/
   "ERA001", "PLR0402", "PLC0105"
 ]
-"*category.pyi" = [
-  # TODO: remove when pandas-dev/pandas-stubs#1443 is resolved
-  "ANN001", "ANN201", "ANN204", "ANN206",
-]
 "*series.pyi" = [
   # TODO: remove when pandas-dev/pandas-stubs#1444 is resolved
   "ANN001", "ANN201", "ANN204", "ANN206",


pandas-stubs/core/indexes/base.pyi
```diff
@@ -430,12 +431,12 @@ class Index(IndexOpsMixin[S1], ElementOpsMixin[S1]):
     ) -> np_1darray[np.intp]: ...
     def reindex(
         self,
-        target,
-        method: ReindexMethod | None = ...,
-        level=...,
-        limit=...,
-        tolerance=...,
-    ): ...
+        target: Iterable[Any],
```
MarcoGorelli, line 434: the docs just say "iterable"

https://github.com/pandas-dev/pandas/blob/b015a3bfb49ec98eec2635d93ce3ebb8d5dad0c2/pandas/core/indexes/base.py#L4116

pandas-stubs/core/indexes/base.pyi
```diff
@@ -430,12 +431,12 @@ class Index(IndexOpsMixin[S1], ElementOpsMixin[S1]):
...
+        method: ReindexMethod | None = None,
+        level: int | None = None,
+        limit: int | None = None,
+        tolerance: Scalar | AnyArrayLike | Sequence[Scalar] | None = None,
```
MarcoGorelli, line 438: the docs are self-contradictory, they say

https://github.com/pandas-dev/pandas/blob/b015a3bfb49ec98eec2635d93ce3ebb8d5dad0c2/pandas/core/indexes/base.py#L4129-L4138

but i've verified that sequences are indeed fine
```py
In [56]: pd.Index([1, 0]).reindex([0, 1.1], method='nearest', tolerance=[0, .2])
Out[56]: (Index([0.0, 1.1], dtype='float64'), array([1, 0]))
```
not just int or float as they say

pandas-stubs/core/indexes/base.pyi
```diff
@@ -471,7 +472,7 @@ class Index(IndexOpsMixin[S1], ElementOpsMixin[S1]):
         cond: Sequence[bool] | np_ndarray_bool | BooleanArray | IndexOpsMixin[bool],
         other: Scalar | AnyArrayLike | None = None,
     ) -> Index: ...
-    def __contains__(self, key) -> bool: ...
+    def __contains__(self, key: Hashable) -> bool: ...
```
MarcoGorelli, line 475: https://github.com/pandas-dev/pandas/blob/b015a3bfb49ec98eec2635d93ce3ebb8d5dad0c2/pandas/core/indexes/base.py#L5182-L5185

pandas-stubs/core/indexes/category.pyi
```diff
@@ -21,28 +24,20 @@ class CategoricalIndex(ExtensionIndex[S1], accessor.PandasDelegate):
     def __new__(
         cls,
         data: Iterable[S1] = ...,
-        categories=...,
-        ordered=...,
-        dtype=...,
-        copy: bool = ...,
-        name: Hashable = ...,
+        categories: ListLike | None = None,
```
MarcoGorelli, line 27: i've mirrored

https://github.com/pandas-dev/pandas-stubs/blob/5365bfc232e4305daddb9acf20ec7d59d9a9a7ab/pandas-stubs/core/arrays/categorical.pyi#L149-L154

here

pandas-stubs/core/indexes/base.pyi
cmp0xff: It works for a non-Sequence generator. Fun fact:
```py
>>> import pandas as pd
>>> pd.Index([1, 2, 3]).reindex(range(1, 3))
(RangeIndex(start=1, stop=3, step=1), array([0, 1]))
```

Maybe for another PR.

pandas-stubs/core/indexes/base.pyi
cmp0xff: pandas-dev/pandas#62923

pandas-stubs/core/indexes/base.pyi
```diff
@@ -488,7 +489,7 @@ class Index(IndexOpsMixin[S1], ElementOpsMixin[S1]):
     @overload
     def append(self, other: Index | Sequence[Index]) -> Index: ...
     def putmask(self, mask, value): ...
-    def equals(self, other) -> bool: ...
+    def equals(self, other: object) -> bool: ...
```
cmp0xff, line 492: Question: what is the difference between an `object` and making it `Any`?

pandas-stubs/core/indexes/range.pyi
```diff
@@ -61,7 +61,6 @@ class RangeIndex(_IndexSubclassBase[int, np.int64]):
     def is_monotonic_decreasing(self) -> bool: ...
     @property
     def has_duplicates(self) -> bool: ...
-    def __contains__(self, key: int | np.integer) -> bool: ...
```
cmp0xff, line 64: It can make sense to make it different from the base class here

pandas-stubs/core/indexes/base.pyi
MarcoGorelli: If it's just in the stubs, then I don't think it makes any difference

There seems to be a mix in pandas-stubs

Do you have a preference?

pandas-stubs/core/indexes/range.pyi
MarcoGorelli: I think the current one is too tight, `key` doesn't have to be `int / np.integer`, it'll just return `False` if it's an incompatible type
```
In [2]: 'foo' in pd.RangeIndex(3)
Out[2]: False
```

pandas-stubs/core/indexes/base.pyi
cmp0xff: Personally I understand `Any` and don't understand `object` at the moment. I'd prefer `Any`. But if you say they are interchangable in the stubs, I can live with it.

pandas-stubs/core/indexes/range.pyi
cmp0xff: if `-> Literal[True]` and `-> Literal[False]` works, we can add two overloads. Otherwise I agree to use the version from the base class.

MarcoGorelli: I think even then it's not failsafe to return `Literal[False]`, as a `float` input (which is not `int` / `np.integer`) could still return `True`
```py
In [3]: 1. in pd.RangeIndex(3)
Out[3]: True
```

pandas-stubs/core/indexes/base.pyi
MarcoGorelli: As far as I understand it, `object` means "don't make any assumptions about this object", so for example it would reject:
```py
from typing import Any

def foo(a: object) -> None:
    return a.bar
```
whereas using `Any` would pass the type-checks (arguably, it would be better if the type checker rejected this, as the implementation is assuming that `.bar` is present)

But from the caller's perspective, they should be interchangeable, and in the stubs we don't have any implementations so here I don't think it matters

pandas-stubs/core/indexes/range.pyi
cmp0xff: Okay, let's leave it as it is. Please consider https://github.com/pandas-dev/pandas-stubs/pull/1459#pullrequestreview-3401583681 and then we can merge.

MarcoGorelli: sure, thanks for your review!

PR #1463, opened 2025-10-31
Fix issue with using .`assign()` with result from `df.get()`

Dr-Irv:
Following code was failing:
```python
df.assign(foo=lambda df: df.get("abc", None))
```
saying that a callable returning `Series | None` could not work.  This PR fixes that.


cmp0xff (approved):
Apparently related to pandas-dev/pandas-stubs#1238 and pandas-dev/pandas-stubs#1239. Thank you @Dr-Irv !

PR #1466, opened 2025-11-02
fix: `pd.NAType`

cmp0xff:
This PR fixes `pd.NAType`. In particular, `pd.NA.__add__(other: object)` is too wide. For example, `pd.NA + [1]` does not work.


cmp0xff:
Hi @loicdiridollou , unfortunately I don't have much expertise on `__new__` myself. Can we resolve the nightly issue separately?

PR #1467, opened 2025-11-03
Type `Series`

### LINKED ISSUE #1444: Typing Series

- [ ] view
- [ ] array_ufunc
- [ ] `__setitem__`
- [ ] mode
- [ ] first
- [ ] last
- [ ] where
- [ ] set_axis

MarcoGorelli:
closes #1444 

### DIFF
--- a/pandas-stubs/core/series.pyi
+++ b/pandas-stubs/core/series.pyi
@@ -1600,7 +1600,7 @@ class Series(IndexOpsMixin[S1], ElementOpsMixin[S1], NDFrame):
             | Callable[[Series[S1]], Series[bool]]
             | Callable[[S1], bool]
         ),
-        other: S1 | Self | Callable[..., S1 | Self] | _NoDefaultDoNotUse = ...,
+        other: S1 | Self | Callable[..., S1 | Self] = ...,
         *,
         inplace: Literal[True],
         axis: AxisIndex | None = 0,
@@ -1616,7 +1616,7 @@ class Series(IndexOpsMixin[S1], ElementOpsMixin[S1], NDFrame):
             | Callable[[Series[S1]], Series[bool]]
             | Callable[[S1], bool]
         ),
-        other: Scalar | Self | Callable[..., Scalar | Self] | _NoDefaultDoNotUse = ...,
+        other: Scalar | Self | Callable[..., Scalar | Self] = ...,
         *,
         inplace: Literal[False] = False,
         axis: AxisIndex | None = 0,


pandas-stubs/core/series.pyi
```diff
@@ -1527,10 +1581,6 @@ class Series(IndexOpsMixin[S1], ElementOpsMixin[S1], NDFrame):
         axis: AxisIndex | None = 0,
     ) -> Series[S1]: ...
     @final
-    def first(self, offset) -> Series[S1]: ...
-    @final
-    def last(self, offset) -> Series[S1]: ...
-    @final
```
MarcoGorelli, line 1533: these are deprecated

pandas-stubs/core/series.pyi
```diff
@@ -513,33 +522,78 @@ class Series(IndexOpsMixin[S1], ElementOpsMixin[S1], NDFrame):
...
+    # Keep in sync with `_iLocIndexerSeries.__setitem__`
     @overload
-    def __getitem__(self, idx: Scalar) -> S1: ...
-    def __setitem__(self, key, value) -> None: ...
+    def __setitem__(self, idx: int, value: S1 | None) -> None: ...
+    @overload
+    def __setitem__(
+        self,
+        idx: Index | slice | np_ndarray_anyint | list[int],
+        value: S1 | Series[S1] | None,
```
cmp0xff: - Can the values not be `Index[S1]`? What about using `IndexOpsMixin[S1]` here?
- Can we maybe make `Series` a subclass of `_LocIndexerSeries` and `_iLocIndexerSeries` to reduce duplication?

MarcoGorelli: - sure, will try
- the issue with subclassing is that then `Series` would inherit `__init__` https://github.com/pandas-dev/pandas-stubs/blob/5aeb6cd6c5fabdf2c12026c116374f358cd19eec/pandas-stubs/_libs/indexing.pyi#L2, which causes tonnes of other type-checking issues

pandas-stubs/core/series.pyi
```diff
@@ -1550,7 +1600,7 @@ class Series(IndexOpsMixin[S1], ElementOpsMixin[S1], NDFrame):
             | Callable[[Series[S1]], Series[bool]]
             | Callable[[S1], bool]
         ),
-        other=...,
+        other: S1 | Self | Callable[..., S1 | Self] | _NoDefaultDoNotUse = ...,
         *,
         inplace: Literal[True],
```
MarcoGorelli, line 1605: the inplace case requires `other` to be of type `S1` if it's a scalar https://pandas.pydata.org/pdeps/0006-ban-upcasting.html

pandas-stubs/core/series.pyi
cmp0xff: In this case, a `type_check_only` class of `LocMixin` could help. But I won't insist on a potential improvement.

pandas-stubs/core/series.pyi
cmp0xff: What is `_NoDefaultDoNotUse` here?

MarcoGorelli: it's the default, so I think it needs to be one of the possible types

https://github.com/pandas-dev/pandas/blob/22e251201035833cb2f037b6cb541292db63747f/pandas/core/generic.py#L10080

as is done in other parts of pandas-stubs

https://github.com/pandas-dev/pandas-stubs/blob/4168b12e9e687ed1c265f0d0540f499b59b480a8/pandas-stubs/core/frame.pyi#L1202

cmp0xff: I cannot easily verify. This is different from what I can check in `pandas`. @Dr-Irv could you take a look? Thank you.

Dr-Irv: So the way this works is that the default value is either `np.nan` or `pd.NA`, dependent on the dtype of the underlying `Series`.  Chasing down the code, there is this:
```python
        if other is lib.no_default:
            other = self.fill_value
```
So I don't think we should include `_NoDefaultDoNotUse` in this case, unless we want to make it the default by saying  `other: S1 | Self | Callable[..., S1 | Self] | _NoDefaultDoNotUse = _NoDefaultDoNotUse`

MarcoGorelli: sure, thanks all - have removed `_NoDefaultDoNotUse`

PR #1468, opened 2025-11-03
wip: add backend as a generic variable

cmp0xff:
There are a lot of tests broken. Let's see if we can sort them out.


cmp0xff:
It is too difficult to make `mypy` happy. I give up for now.

Issue #1446, opened 2025-10-23
Typing DataFrame


MarcoGorelli:
Here's the methods in Dataframe which are missing some type annotation or return type

- [ ] `__new__`
- [ ] `__dataframe__` (but this should probably be deprecated so perhaps hold off on this one for now)
- [ ] `__rmatmul__`
- [ ] `__setitem__`
- [ ] `pivot_table`
- [ ] `hist`
- [ ] `boxplot`
- [ ] `radd`
- [ ] `rsub`
- [ ] `asof`
- [ ] `asfreq`
- [ ] `eq`
- [ ] `first` (deprecated)
- [ ] `last` (deprecated)
- [ ] `ge`
- [ ] `gt`
- [ ] `le`
- [ ] `lt`
- [ ] `rdiv`
- [ ] `rfloordiv`
- [ ] `rmod`
- [ ] `rmul`
- [ ] `rpow`
- [ ] `rtruediv`
- [ ] `set_axis`
- [ ] `where`

PR #1470, opened 2025-11-05
Type DataFrame methods: __dataframe__, first, last, asfreq, `where`

MarcoGorelli:
addresses part of #1446 , just splitting it into smaller chunks as there's quite a bit to do


pandas-stubs/api/interchange/__init__.pyi
```diff
@@ -1,2 +1 @@
-from pandas.core.interchange.dataframe_protocol import DataFrame as DataFrame
```
MarcoGorelli, line 1: this is now deprecated

cmp0xff: Could you also remove it from `dataframe_protocol.pyi`?

cmp0xff (approved):
This #1470 partially addresses pandas-dev/pandas-stubs#1446. Thank you @MarcoGorelli !

PR #1471, opened 2025-11-06
GH1447 Example

### LINKED ISSUE #1447: CLEAN:  Use `typing_extensions.override`
Throughout the stubs, we have ignores for overrides that we know are valid. We can use `@override` when we know that the override is correct. 

Have to look for:
```python
# type: ignore[override] # pyright: ignore[reportIncompatibleMethodOverride]
```

What I'm not sure of is that for places where `mypy` complains but `pyright` does not, if we include `@override`, whether pyright will complain saying something like it's an unnecessary override.

Should import `override` from `typing_extensions`, since it was introduced in python 3.12

loicdiridollou:
- [ ] Closes #1447

### DIFF
--- a/pandas-stubs/core/series.pyi
+++ b/pandas-stubs/core/series.pyi
@@ -106,6 +106,7 @@ from pandas.core.window.rolling import (
 from typing_extensions import (
     Never,
     Self,
+    override,
 )
 import xarray as xr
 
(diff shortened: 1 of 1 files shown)


pandas-stubs/core/series.pyi
```diff
@@ -2190,13 +2191,17 @@ class Series(IndexOpsMixin[S1], ElementOpsMixin[S1], NDFrame):
         axis: int = 0,
     ) -> Series[_str]: ...
     # ignore needed for mypy as we want different results based on the arguments
-    @overload  # type: ignore[override]
+    @overload
+    @override
```
loicdiridollou, line 2195: @Dr-Irv here is the example with `@override`, not sure I am using it the right way since it is failing any pipeline.

PR #1477, opened 2025-11-10
stubs: add pandas.io.sql.get_schema to fix Pylance reportAttributeAcc…

sadegh-pz:
### Summary
Add a stub for `pandas.io.sql.get_schema`. The function exists at runtime but
is missing from the `pandas-stubs`, causing Pylance/Pyright to report:

> "get_schema" is unknown import symbol (reportAttributeAccessIssue)

### Rationale
- Improves developer experience in VS Code / Pyright without changing pandas runtime.
- Keeps pandas’ API policy unchanged (no need to modify pandas or `__all__`).

### Signature
```py
def get_schema(
    frame: DataFrame,
    name: str,
    keys: str | Sequence[str] | None = ...,
    con: Any = ...,
    dtype: DtypeArg | None = ...,
    schema: str | None = ...,
) -> str: ...

### DIFF
--- a/pandas-stubs/io/sql.pyi
+++ b/pandas-stubs/io/sql.pyi
@@ -10,6 +10,7 @@ from typing import (
     Literal,
     TypeAlias,
     overload,
+    Sequence,
 )
 
 from pandas.core.frame import DataFrame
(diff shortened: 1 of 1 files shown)


cmp0xff:
Thank you for following up. I understand that you use `pandas.io.sql.get_schema` and that adding it to stubs could be helpful.

I would like to refer to [README.md](https://github.com/pandas-dev/pandas-stubs?tab=readme-ov-file#differences-between-type-declarations-in-pandas-and-pandas-stubs), where it states

> The https://github.com/pandas-dev/pandas-stubs/ project provides type declarations for the pandas _public_ API.

SadeghPouriyanZadeh:
Thanks. I haven't read that part. I get it now. I thought internal APIs are included too.

PR #1479, opened 2025-11-11
fix(nightly): #1478

### LINKED ISSUE #1478: nightly failing after latest PR for fixing numpy types
See https://github.com/pandas-dev/pandas-stubs/actions/runs/19245526290/job/55018489783

PR #1462 caused this

cmp0xff:
- [x] Closes #1478
- [x] Tests added
- [x] Propagates pandas-dev/pandas#58690 pandas-dev/pandas#61942 pandas-dev/pandas#61952

/pandas_nightly

### DIFF
--- a/tests/test_resampler.py
+++ b/tests/test_resampler.py
@@ -156,7 +156,7 @@ def test_interpolate() -> None:
 def test_interpolate_inplace() -> None:
     with pytest_warns_bounded(
         Pandas4Warning,
-        "The 'inplace' keyword in DatetimeIndexResampler.interpolate is deprecated and will be removed in a future version. resample(...).interpolate is never inplace.",
+        r"The 'inplace' keyword in DatetimeIndexResampler.interpolate is deprecated and will be removed in a future version. resample\(...\).interpolate is never inplace.",
         lower="2.99",
     ):
         check(


tests/series/test_series.py
```diff
@@ -1415,23 +1414,17 @@ def test_types_rename_axis() -> None:
...
-    else:
-        valresult_type = StringArray
     check(
         assert_type(
-            pd.Series(list("aabc")).values,
-            np_1darray | ExtensionArray | pd.Categorical,
+            pd.Series(list("aabc")).values, np_1darray | ExtensionArray | pd.Categorical
         ),
-        valresult_type,
+        # TODO: ArrowStringArray after pandas-dev/pandas-stubs#1469
```
Dr-Irv: It will need to be `BaseStringArray`, because that will take care of whether or not arrow is present.

tests/test_frame.py
```diff
@@ -838,7 +843,7 @@ def test_dataframe_clip() -> None:
         df.clip(lower=pd.Series([1, 2]), upper=pd.Series([4, 5]), axis=None)  # type: ignore[call-overload] # pyright: ignore[reportCallIssue, reportArgumentType]
         df.copy().clip(lower=pd.Series([1, 2]), upper=None, axis=None, inplace=True)  # type: ignore[call-overload] # pyright: ignore[reportCallIssue, reportArgumentType]
         df.copy().clip(lower=None, upper=pd.Series([1, 2]), axis=None, inplace=True)  # type: ignore[call-overload] # pyright: ignore[reportCallIssue, reportArgumentType]
-        df.copy().clip(lower=pd.Series([4, 5]), upper=pd.Series([1, 2]), axis=None, inplace=True)  # type: ignore[call-overload] # pyright: ignore[reportCallIssue, reportArgumentType]
+        df.copy().clip(lower=pd.Series([4, 5]), upper=pd.Series([1, 2]), axis=None, inplace=True)  # type: ignore[call-overload]# pyright: ignore[reportCallIssue, reportArgumentType]
```
Dr-Irv: Not sure why the space is removed between `]# pyright`  Was `] # pyright`

tests/test_frame.py
```diff
@@ -3771,18 +3770,31 @@ def test_to_dict_simple() -> None:
     # orient param accepting "tight" added in 1.4.0 https://pandas.pydata.org/docs/whatsnew/v1.4.0.html
     check(assert_type(data.to_dict("tight"), dict[str, list]), dict, str)
 
-    if TYPE_CHECKING_INVALID_USAGE:
-
-        def test(mapping: Mapping) -> None:  # pyright: ignore[reportUnusedFunction]
-            data.to_dict(into=mapping)  # type: ignore[call-overload] # pyright: ignore[reportArgumentType,reportCallIssue]
+    def _0(mapping: Mapping) -> None:  # pyright: ignore[reportUnusedFunction]
+        data.to_dict(into=mapping)  # type: ignore[call-overload] # pyright: ignore[reportArgumentType,reportCallIssue]
```
Dr-Irv: This should remain in a `if TYPE_CHECKING_INVALID_USAGE` block, even though it is never executed

tests/test_resampler.py
```diff
@@ -147,8 +147,8 @@ def test_interpolate() -> None:
 
 
 def test_interpolate_inplace() -> None:
+    # TODO: ValueError: Cannot interpolate inplace on a resampled object. Disable in Pandas 3.0  pandas-dev/pandas#57064
     if PD_LTE_23:
-        # Bug in main see https://github.com/pandas-dev/pandas/issues/58690
         check(
             assert_type(DF.resample("ME").interpolate(inplace=True), None), type(None)
         )
```
Dr-Irv: Given the nature of the discussion on this issue in `pandas`, can we do the following:

- disallow the `inplace=True` argument in `resample().interpolate()`
- remove the 2 tests using that

tests/series/test_series.py
cmp0xff: 3a69444115a38e47523cc78abd33aad62697cf65

tests/test_frame.py
cmp0xff: e089fc4664f84be039ffc9d498af8604b44b97a6

tests/test_frame.py
cmp0xff: bf8ecb4e2dc27999259f29f7cff8917746508976

tests/test_resampler.py
cmp0xff: I am testing `if TYPE_CHECKING_INVALID_USAGE` 036d326882a2fe0306c848c1c2d0f64989b3f2b6

tests/test_groupby.py
```diff
@@ -427,9 +417,6 @@ def f(val: Series) -> float:
             Series,
             float,
         )
```
Dr-Irv: We can do these tests for 2.3 and 3.0, so can you remove the `else` ?

tests/test_groupby.py
```diff
@@ -259,22 +259,12 @@ def df2scalar(val: DataFrame) -> float:
                 ),
                 DataFrame,
             )
-            check(
-                assert_type(GB_DF.resample("ME").interpolate(inplace=True), None),
-                type(None),
-            )
         else:
```
Dr-Irv: we can do these tests that are in the `else` part for 2.3 and 3.0, so can you remove the `else`?

tests/test_resampler.py
Dr-Irv: I looked in that commit and don't see the test (although I think having the test is a good idea).  i.e., we should make sure the `df.resample("ME").interpolate(inplace=True)` is rejected.

tests/test_groupby.py
cmp0xff: 62ceaee958d09b2750d0cb451f7b40a94dbc6527

tests/test_resampler.py
cmp0xff: So I checked the implementation in https://github.com/pandas-dev/pandas/pull/62847/files#diff-345e833a2b0533a66aee6d2440b4522f06f789321519218d88ff3add50894cf0R990-R1002
- `interpolate` accepts `**kwargs` and anything can go into the interpolation function, except for `inplace`
- If `inplace` is present, a warning is issued
- If it is `False`, it is simply dropped and does not go into the interpolation function
- If it is `True`, an Error is raised

I have added tests for both cases b783f02caf8258260bb38e943e8c54274059846f

cmp0xff (commented):
> The challenge I'm having is that I look for commits since last review and it picks up other changes that you may have done (or maybe was done due to a merge?)

Sorry, there had been indeed another PR merged and I was merging `main` into my branch in this PR.

loicdiridollou:
If you want to pass the full string, this should be the one, I think the pytest warn checker (it uses `re.search` under the hood) is confused with the parenthesis so you would pass the following string:
`r"The 'inplace' keyword in DatetimeIndexResampler.interpolate is deprecated and will be removed in a future version. resample\(...\).interpolate is never inplace."`

Dr-Irv:
> Yes usually I will truncate the message and avoid the parts where there are those "problematic" characters, I had to debug the code to fully understand what was going on (and in general it does not make too much sense to have the full message, the start usually suffices).

Yes it does

Dr-Irv (approved):
Thanks @cmp0xff

PR #1481, opened 2025-11-12
Type `DataFrame`

MarcoGorelli:

closes https://github.com/pandas-dev/pandas-stubs/issues/1446

### DIFF
--- a/pandas-stubs/core/frame.pyi
+++ b/pandas-stubs/core/frame.pyi
@@ -2429,13 +2429,6 @@ class DataFrame(NDFrame, OpsMixin, _GetItemHack):
         ascending: _bool = True,
         pct: _bool = False,
     ) -> Self: ...
-    def rdiv(
-        self,
-        other: complex | ListLike | DataFrame,
-        axis: Axis = "columns",
-        level: Level | None = None,
-        fill_value: float | None = None,
-    ) -> Self: ...
     @final
     def reindex_like(
         self,
@@ -2548,6 +2541,7 @@ class DataFrame(NDFrame, OpsMixin, _GetItemHack):
         level: Level | None = None,
         fill_value: float | None = None,
     ) -> Self: ...
+    rdiv = rtruediv
     @final
     def sample(
         self,


pandas-stubs/core/interchange/from_dataframe.pyi
```diff
@@ -1,3 +1,5 @@
+from typing import Any
+
 import pandas as pd
 
-def from_dataframe(df, allow_copy: bool = True) -> pd.DataFrame: ...
+def from_dataframe(df: Any, allow_copy: bool = True) -> pd.DataFrame: ...
```
MarcoGorelli, line 6: technically this could be made more precise, but given that this interchange protocol is mostly deprecated i don't think it's worth bothering with. `df` represents an external library's dataframe anyway, and not something part of pandas

cmp0xff: pandas-dev/pandas#63111

pandas-stubs/core/frame.pyi
```diff
@@ -1128,7 +1128,7 @@ class DataFrame(NDFrame, OpsMixin, _GetItemHack):
     def dropna(
         self,
         *,
-        axis: Axis = ...,
+        axis: Axis = "columns",
```
cmp0xff: Also for other `dropna`s
```suggestion
        axis: Axis = 0,
```

pandas-stubs/core/frame.pyi
```diff
@@ -1174,7 +1174,7 @@ class DataFrame(NDFrame, OpsMixin, _GetItemHack):
         self,
         by: _str | Sequence[_str],
         *,
-        axis: Axis = ...,
+        axis: Axis = "columns",
```
cmp0xff: Also for other `sort_values`
```suggestion
        axis: Axis = 0,
```

pandas-stubs/core/frame.pyi
```diff
@@ -1199,8 +1199,8 @@ class DataFrame(NDFrame, OpsMixin, _GetItemHack):
     def sort_index(
         self,
         *,
-        axis: Axis = ...,
-        level: Level | None = ...,
+        axis: Axis = "columns",
```
cmp0xff: I found that among the _public_ APIs, only in rich comparison methods, like `DataFrame.eq`, are `"columns"` the default value for `axis`. All other cases have `0` as the default value. (Please double check)
```suggestion
        axis: Axis = 0,
```

pandas-stubs/core/frame.pyi
```diff
@@ -2296,8 +2311,18 @@ class DataFrame(NDFrame, OpsMixin, _GetItemHack):
     ) -> Series: ...
     @final
     def last_valid_index(self) -> Scalar: ...
-    def le(self, other, axis: Axis = "columns", level: Level | None = ...) -> Self: ...
-    def lt(self, other, axis: Axis = "columns", level: Level | None = ...) -> Self: ...
+    def le(
+        self,
+        other: complex | ListLike | DataFrame,
+        axis: Axis = "columns",
```
cmp0xff, line 2306: `"columns"` is correct, do not change.

pandas-stubs/core/frame.pyi
```diff
@@ -2296,8 +2311,18 @@ class DataFrame(NDFrame, OpsMixin, _GetItemHack):
...
+        level: Level | None = None,
+    ) -> Self: ...
+    def lt(
+        self,
+        other: complex | ListLike | DataFrame,
+        axis: Axis = "columns",
```
cmp0xff, line 2312: `"columns"` is correct, do not change.

pandas-stubs/core/frame.pyi
```diff
@@ -2362,17 +2387,22 @@ class DataFrame(NDFrame, OpsMixin, _GetItemHack):
...
         axis: Axis | None = ...,
-        level: Level | None = ...,
+        level: Level | None = None,
         fill_value: float | None = None,
     ) -> Self: ...
-    def ne(self, other, axis: Axis = "columns", level: Level | None = ...) -> Self: ...
+    def ne(
+        self,
+        other: complex | ListLike | DataFrame,
+        axis: Axis = "columns",
```
cmp0xff, line 2393: `"columns"` is correct, do not change.

pandas-stubs/core/frame.pyi
```diff
@@ -2179,13 +2179,18 @@ class DataFrame(NDFrame, OpsMixin, _GetItemHack):
         self,
         other: complex | ListLike | DataFrame,
         axis: Axis | None = ...,
```
cmp0xff:
```suggestion
        axis: Axis | None = "columns",
```

pandas-stubs/core/frame.pyi
```diff
@@ -2418,9 +2448,9 @@ class DataFrame(NDFrame, OpsMixin, _GetItemHack):
...
     def rdiv(
         self,
-        other,
-        axis: Axis = ...,
-        level: Level | None = ...,
+        other: complex | ListLike | DataFrame,
+        axis: Axis = "columns",
+        level: Level | None = None,
         fill_value: float | None = None,
     ) -> Self: ...
```
cmp0xff:
```suggestion
    rdiv = rtruediv
```

pandas-stubs/core/frame.pyi
```diff
@@ -1515,28 +1515,28 @@ class DataFrame(NDFrame, OpsMixin, _GetItemHack):
...
         **kwargs: Any,
     ) -> Series: ...
     @overload
     def agg(
         self,
         func: list[AggFuncTypeBase] | AggFuncTypeDictFrame = ...,
-        axis: Axis = ...,
+        axis: Axis = 0,
         **kwargs: Any,
     ) -> Self: ...
```
cmp0xff: In `pandas` we have `agg = aggregate`

pandas-stubs/core/frame.pyi
```diff
@@ -2781,7 +2811,7 @@ class DataFrame(NDFrame, OpsMixin, _GetItemHack):
         self,
         key: Hashable,
         axis: Axis = 0,
-        level: Level | None = ...,
+        level: Level | None = None,
         drop_level: _bool = True,
     ) -> Self | Series: ...
```
cmp0xff: In `pandas` this is in `class NDFrame`. I think we should move this one to `class NDFrame`.
```py
    @final
    def xs(
        self,
        key: IndexLabel,
        axis: Axis = 0,  # in Series maybe only AxesIndex is applicable, so the override in Series may be necessary
        level: IndexLabel | None = None,
        drop_level: _bool = True,
    ) -> Self: ...
````

MarcoGorelli: sure, but if it gets overwritten in `Series.xs`, then does this make a difference?

cmp0xff: I would say it prevents potential problems when NDFrame is inherited by other subclasses

cmp0xff: Also the `key` has typing `IndexLabel` instead of just `Hashable`, from `pandas`.

cmp0xff:
Please set the PR to ready (and request a review) when it is ready. Thanks.

pandas-stubs/core/frame.pyi
cmp0xff: I think this one is missing?

PR #1487, opened 2025-11-13
fix: 🌃 nightly

cmp0xff:


tests/series/test_series.py
```diff
@@ -2222,7 +2222,7 @@ def test_pandera_generic() -> None:
 
     class MySeries(pd.Series, Generic[T]):
         def __new__(cls, *args: Any, **kwargs: Any) -> Self:
-            return object.__new__(cls)
+            return super().__new__(cls)  # type: ignore[return-value] # pyright: ignore[reportReturnType]
```
Dr-Irv, line 2225: I really don't want `# type: ignore` in the test files, except in the following cases:
1. the type checkers disagree - one is OK, but the other flags it, so we have to ignore for one of the checkers
2. When we are doing `TYPE_CHECKING_INVALID_USAGE`

I think the following will make this work:
```python
        def __new__(cls, *args: Any, **kwargs: Any) -> MySeries[T]:
            return cast(MySeries[T], super().__new__(cls))
```
In this case, you have to do the `cast` because of how `__new()` works.

@cmp0xff or @loicdiridollou can you try this and get it to work?

cmp0xff: Sorry I missed this comment. @loicdiridollou has created pandas-dev/pandas-stubs#1489.

My understanding used to be that ignores are not allowed in tests, and these new ignores are not in tests themselves.

PR #1489, opened 2025-11-14
Fix nightly

loicdiridollou:

Follow up from @Dr-Irv comment in https://github.com/pandas-dev/pandas-stubs/pull/1487


cmp0xff (approved):
It works and removes `ignore`s in tests. Thank you @Dr-Irv @loicdiridollou!

PR #1490, opened 2025-11-14
GH1488 Add overload for DataFrame.max when axis is None

### LINKED ISSUE #1488: `DataFrame.max` needs the same overloads as `DataFrame.any` has
**Describe the bug**
`max` and other aggregations are missing some overloads, like `any` has

https://github.com/pandas-dev/pandas-stubs/blob/232b4a16a75ebb98648123c7a30e5fd413151895/pandas-stubs/core/frame.pyi#L1998-L2015

https://github.com/pandas-dev/pandas-stubs/blob/232b4a16a75ebb98648123c7a30e5fd413151895/pandas-stubs/core/frame.pyi#L2333-L2339

**To Reproduce**
```py
df = pd.DataFrame({'a': [1,2,3], 'b': [4,5,6]})
print(df.max(axis=None))
```

The stubs say this always returns `Series`, but actually, for `axis=None`, it returns a scalar

 - OS Version [e.g. 22]
 - python version
 - version of type checker

**Additional context**

loicdiridollou:
- [x] Closes #1488

### DIFF
--- a/tests/test_frame.py
+++ b/tests/test_frame.py
@@ -713,7 +713,7 @@ def test_types_mean() -> None:
     df = pd.DataFrame(data={"col1": [2, 1], "col2": [3, 4]})
     check(assert_type(df.mean(), pd.Series), pd.Series)
     check(assert_type(df.mean(axis=0), pd.Series), pd.Series)
-    check(assert_type(df.mean(axis=None), Scalar), np.float64)
+    check(assert_type(df.mean(axis=None), Scalar), np.floating)
     check(assert_type(df.groupby(level=0).mean(), pd.DataFrame), pd.DataFrame)
     check(
         assert_type(df.mean(axis=1, skipna=True, numeric_only=False), pd.Series),
@@ -725,7 +725,7 @@ def test_types_median() -> None:
     df = pd.DataFrame(data={"col1": [2, 1], "col2": [3, 4]})
     check(assert_type(df.median(), pd.Series), pd.Series)
     check(assert_type(df.median(axis=0), pd.Series), pd.Series)
-    check(assert_type(df.median(axis=None), Scalar), np.float64)
+    check(assert_type(df.median(axis=None), Scalar), np.floating)
     check(assert_type(df.groupby(level=0).median(), pd.DataFrame), pd.DataFrame)
     check(
         assert_type(df.median(axis=1, skipna=True, numeric_only=False), pd.Series),
@@ -822,14 +822,14 @@ def test_types_min() -> None:
     df = pd.DataFrame(data={"col1": [2, 1], "col2": [3, 4]})
     check(assert_type(df.min(), pd.Series), pd.Series)
     check(assert_type(df.min(axis=0), pd.Series), pd.Series)
-    check(assert_type(df.min(axis=None), Scalar), np.int64)
+    check(assert_type(df.min(axis=None), Scalar), np.integer)
 
 
 def test_types_max() -> None:
     df = pd.DataFrame(data={"col1": [2, 1], "col2": [3, 4]})
     check(assert_type(df.max(), pd.Series), pd.Series)
     check(assert_type(df.max(axis=0), pd.Series), pd.Series)
-    check(assert_type(df.max(axis=None), Scalar), np.int64)
+    check(assert_type(df.max(axis=None), Scalar), np.integer)
 
 
 def test_types_quantile() -> None:


pandas-stubs/core/frame.pyi
cmp0xff: It seems that the same overload can also be added to `mean`, `min`, `median`. Could you do that and add tests?

cmp0xff (changes requested):

pandas-stubs/core/frame.pyi
loicdiridollou: Good idea, this is addressed!

pandas-stubs/core/frame.pyi
```diff
@@ -2344,30 +2344,66 @@ class DataFrame(NDFrame, OpsMixin, _GetItemHack):
...
     ) -> Self: ...
+    @overload
     def max(
         self,
-        axis: Axis | None = 0,
+        axis: None,
+        skipna: _bool | None = True,
+        numeric_only: _bool = False,
+        **kwargs: Any,
+    ) -> Scalar: ...
```
cmp0xff, line 2354: We can probably further restrict to `_OrderableT` from `pandas-stubs/_libs/intervals.pyi`, instead of just `Scalar`. But this can be done in a separate PR.

tests/test_frame.py
```diff
@@ -713,6 +713,7 @@ def test_types_mean() -> None:
     df = pd.DataFrame(data={"col1": [2, 1], "col2": [3, 4]})
     check(assert_type(df.mean(), pd.Series), pd.Series)
     check(assert_type(df.mean(axis=0), pd.Series), pd.Series)
+    check(assert_type(df.mean(axis=None), Scalar), np.float64)
```
cmp0xff: I think we usually just test for the more generic `np.floating` unless there is a reason to test for the specific `np.float64`.
```suggestion
    check(assert_type(df.mean(axis=None), Scalar), np.floating)
```

tests/test_frame.py
```diff
@@ -724,6 +725,7 @@ def test_types_median() -> None:
     df = pd.DataFrame(data={"col1": [2, 1], "col2": [3, 4]})
     check(assert_type(df.median(), pd.Series), pd.Series)
     check(assert_type(df.median(axis=0), pd.Series), pd.Series)
+    check(assert_type(df.median(axis=None), Scalar), np.float64)
```
cmp0xff:
```suggestion
    check(assert_type(df.median(axis=None), Scalar), np.floating)
```

tests/test_frame.py
```diff
@@ -820,12 +822,14 @@ def test_types_min() -> None:
     df = pd.DataFrame(data={"col1": [2, 1], "col2": [3, 4]})
     check(assert_type(df.min(), pd.Series), pd.Series)
     check(assert_type(df.min(axis=0), pd.Series), pd.Series)
+    check(assert_type(df.min(axis=None), Scalar), np.int64)
```
cmp0xff:
```suggestion
    check(assert_type(df.min(axis=None), Scalar), np.integer)
```

tests/test_frame.py
```diff
@@ -820,12 +822,14 @@ def test_types_min() -> None:
...
 
 
 def test_types_max() -> None:
     df = pd.DataFrame(data={"col1": [2, 1], "col2": [3, 4]})
     check(assert_type(df.max(), pd.Series), pd.Series)
     check(assert_type(df.max(axis=0), pd.Series), pd.Series)
+    check(assert_type(df.max(axis=None), Scalar), np.int64)
```
cmp0xff:
```suggestion
    check(assert_type(df.max(axis=None), Scalar), np.integer)
```

cmp0xff (changes requested):
Thank you for adding `mean`, `min`, `median`. I think we just need to relax the type testing.

pandas-stubs/core/frame.pyi
loicdiridollou: Is it a type that can be returned? I thought it was more of a generic for Interval or something like that. I have left it for later.

tests/test_frame.py
loicdiridollou: All addressed!

pandas-stubs/core/frame.pyi
cmp0xff: I'm not sure, either. @Dr-Irv can we return `_OrderableT`?

Dr-Irv: > I'm not sure, either. @Dr-Irv can we return `_OrderableT`?

We can't return a `TypeVar` when the `TypeVar` is not part of the `Generic` in the class.

But we can create a new type that is a union of what is in `_OrderableT` , and that might be a better result here.

Although be careful.  Because `_OrderableT` doesn't include `str`, which could be the result of `max` and `min`.  Would need to check the other scalar types as well.

cmp0xff: Thanks, I think it may be too much hair-splitting to refine further here.

PR #1492, opened 2025-11-15
GH1432 Partial resolution for some TODOs

### LINKED ISSUE #1432: Issues from `TODO`s in `timestamps.pyi`, `timedeltas.pyi` and `orc.pyi`
This issue is created so that [TD003](https://docs.astral.sh/ruff/rules/missing-todo-link/) can be complied to.

## `orc.pyi`

- [ ] In `read_orc`, we need to type the filesystem with the correct `pyarrow` types.

## `timestamps.pyi`

- [ ] We need to test `Timestamp - np.timedelta64`.
- [ ] We need to check if `tz_localize` could return `NaT` or not.
- [ ] We need to check if round/floor/ceil could return `NaT` or not.

## `timedeltas.pyi`

- [ ] We need to check if round/floor/ceil could return `NaT` or not.

loicdiridollou:
- [ ] Closes #1432

Resolution for:
- read_orc (although there is no docs or tests in pandas repo to properly test that)
- Timestamp - datetime64 (There is the typing to the rsub but mypy sees the wrong type)
- tz_localize returning NaT (edge case of changing time with DST)

### DIFF
--- a/tests/test_timefuncs.py
+++ b/tests/test_timefuncs.py
@@ -105,9 +105,13 @@ def test_types_arithmetic() -> None:
     check(assert_type(ts - dt.datetime(2021, 1, 3), pd.Timedelta), pd.Timedelta)
 
     if TYPE_CHECKING_INVALID_USAGE:
-        # TODO: pandas-dev/pandas-stubs#1432 mypy sees datetime.timedelta but pyright is correct
-        assert_type(ts_np - ts, pd.Timedelta)  # type: ignore[assert-type]
-        assert_type(ts_np_time - ts, pd.Timedelta)  # type: ignore[assert-type]
+        # TODO: pandas-dev/pandas-stubs#1511 numpy.datetime64.__sub__ gives datetime.timedelta, which has higher priority
+        assert_type(
+            ts_np - ts, dt.timedelta  # pyright: ignore[reportAssertTypeFailure]
+        )
+        assert_type(
+            ts_np_time - ts, dt.timedelta  # pyright: ignore[reportAssertTypeFailure]
+        )
 
 
 def test_types_comparison() -> None:


pandas-stubs/io/orc.pyi
```diff
@@ -14,8 +16,6 @@ def read_orc(
     path: FilePath | ReadBuffer[bytes],
     columns: list[HashableT] | None = None,
     dtype_backend: DtypeBackend | _NoDefaultDoNotUse = "numpy_nullable",
```
cmp0xff, line 18: In 3.0 the default value will be switched from `"numpy_nullable"` to `no_default`. I think this can be confusing to users using the stable releases 2.x. What should we do @Dr-Irv ? Anyway, we can address this in a separate PR.

tests/test_timefuncs.py
```diff
@@ -95,9 +95,13 @@ def test_types_init() -> None:
 def test_types_arithmetic() -> None:
     ts = pd.to_datetime("2021-03-01")
     ts2 = pd.to_datetime("2021-01-01")
+    ts_np = np.datetime64("2021-01-01")
     delta = pd.to_timedelta("1 day")
 
     check(assert_type(ts - ts2, pd.Timedelta), pd.Timedelta)
+    check(assert_type(ts - ts_np, pd.Timedelta), pd.Timedelta)
+    # TODO: pandas-dev/pandas-stubs#1432 mypy sees datetime.timedelta but pyright is correct
+    # check(assert_type(ts_np - ts, pd.Timedelta), pd.Timedelta)
```
cmp0xff: I see our `__rsub__` is correct but `mypy` takes `ts_np.__sub__` as a higher priority. This has happened a lot in `Series` and `Index` arithmetic. I think neither `mypy` nor `pyright` is correct, we should see `Any` here because of the discrepency. Do you have any suggestion to improve here, @Dr-Irv ?

pandas-stubs/io/orc.pyi
Dr-Irv: Use the 2.3 default for now, and create an issue that we need to fix it when 3.0 is released.

I still think there are things in the release notes for 2.0, 2.1, 2.2 and 2.3 that we never put in the stubs.  I created a 2.0 tracker, and I think many of those are still open.  Never had the time to create trackers for 2.1, 2.2 and 2.3.  And when 3.0 is released, we should create a 3.0 tracker.  So we could wait until then instead of creating the issue now

tests/test_timefuncs.py
Dr-Irv: This is a bit of a mess, because I don't think that `numpy` can know whether it supports subtraction or not with a datetime:
```python
>>> import numpy as np
>>> import datetime as dt
>>> ts_np = np.datetime64("2021-01-01")
>>> ts_dt = dt.datetime(2020,12,1)
>>> ts_np - ts_dt
  File "<stdin>", line 1, in <module>
TypeError: unsupported operand type(s) for -: 'datetime.date' and 'datetime.datetime'
>>> ts_dt = dt.datetime(2020,12,1,3,24)
>>> ts_np - ts_dt
  File "<stdin>", line 1, in <module>
TypeError: unsupported operand type(s) for -: 'datetime.date' and 'datetime.datetime'
>>> ts_np = np.datetime64("2021-01-01T03:24")
>>>
>>> ts_np - ts_dt
datetime.timedelta(days=31)
>>> ts_dt = dt.datetime(2021,12,1)
>>> ts_np - ts_dt
datetime.timedelta(days=-334, seconds=12240)
```
So if the numpy `datetime64` does not specify hour/minute, then you can't do subtraction.  But if it does have the hour/minute, you can.  And their typing can't see the difference, which makes sense.

I suggest doing tests with a `numpy` type that has no hour/minute resolution (as in this PR) AND with hour/minute resolution.

Without the resolution, it will fail at runtime, but we can't detect it, so put it in ` if TYPE_CHECKING_INVALID_USAGE` with the ignores that detect it (in this case, need a pyright ignore, but not a mypy one)

With the resolution, it will work at runtime, but we can't get the right type with either type checker (pyright will reject, mypy will fail in assert type), so put in appropriate ignores.

In both cases, include comments as to what is going on.

Note - I think `pyright` is wrong here in not having `numpy` type declarations figure out that `np.datetime64.__sub__(pd.Timestamp)` *is* valid from a type perspective, because `pd.Timestamp` is a subclass of `datetime.datetime`  But the `__sub__()` in `numpy` will take precedence here because it has to allow `__sub__(datetime.datetime)`

Created a pyright issue:  https://github.com/microsoft/pyright/issues/11135

pandas-stubs/io/orc.pyi
cmp0xff: pandas-dev/pandas-stubs#1502

tests/test_timefuncs.py
loicdiridollou: Cleaned up the tests, the issue is really between datetime and tatetime64 but does not really impact when using Timestamps, interestingly enough mypy sees `dt.timedelta` and not `pd.Timedelta` so I have added ignores there.

tests/test_timefuncs.py
```diff
@@ -95,13 +95,22 @@ def test_types_init() -> None:
...
+    check(assert_type(ts - ts_np, pd.Timedelta), pd.Timedelta)
+    check(assert_type(ts - ts_np_time, pd.Timedelta), pd.Timedelta)
     check(assert_type(ts + delta, pd.Timestamp), pd.Timestamp)
     check(assert_type(ts - delta, pd.Timestamp), pd.Timestamp)
     check(assert_type(ts - dt.datetime(2021, 1, 3), pd.Timedelta), pd.Timedelta)
 
+    if TYPE_CHECKING_INVALID_USAGE:
+        # TODO: pandas-dev/pandas-stubs#1432 mypy sees datetime.timedelta but pyright is correct
+        assert_type(ts_np - ts, pd.Timedelta)  # type: ignore[assert-type]
+        assert_type(ts_np_time - ts, pd.Timedelta)  # type: ignore[assert-type]
```
cmp0xff: In pandas-dev/pandas-stubs#1432, neither `mypy` nor `pyright` is mentioned, and I am confused without reading the comment again. Also I think we agreed on the reverse statement, that `mypy` is correct and `pyright` is wrong here, so maybe something like the following?

```suggestion
        # TODO: pandas-dev/pandas-stubs#1511 numpy.datetime64.__sub__ gives datetime.timedelta, which has higher priority
        assert_type(ts_np - ts, dt.timedelta)  # pyright: ignore[reportAssertTypeFailure]
        assert_type(ts_np_time - ts, dt.timedelta)  # pyright: ignore[reportAssertTypeFailure
```

loicdiridollou: Wait, I am getting confused because at runtime we are getting `pd.Timedelta`. I think @Dr-Irv's point was about numpy - datetime, nothing to do with `pd.Timestamp`.
Here is what I see at runtime:
```
>>> import pandas as pd
>>> import numpy as np
>>> ts_np = np.datetime64("2021-01-01")
>>> ts = pd.to_datetime("2021-03-01")
>>> ts - ts_np
Timedelta('59 days 00:00:00')
>>> ts_np - ts
Timedelta('-59 days +00:00:00')
```
So np.datetime64 - pd.Timestamp gives a pd.Timedelta

cmp0xff: > So np.datetime64 - pd.Timestamp gives a pd.Timedelta

This is correct at runtime. For type checkers, however, I think `np.datetime64.__sub__(datetime.datetime)` should have taken over, giving `timedelta`. This is true for `mypy` but not true for `pyright`.

PR #1469, opened 2025-11-04
TYP: complete constructions for `ExtensionArray`s

cmp0xff:
- [x] Tests added

PR #1493, opened 2025-11-18
feat: `FloatingArray`, `Series.__new__` + `astype` and also for `Index`

### LINKED ISSUE #1428: `FloatingArray` documented but missing from stubs
https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.arrays.FloatingArray.html

This is documented, but not in the stubs

cmp0xff:
This is a PR simpler than #1469 which focuses more on adding `FloatingArray`.

I have made many more categories for the `DtypeArg`s, because of the following reasons:

- Only `Pandas` dtypes are relevant for making a `FloatingArray`. Native and `Numpy` dtypes lead to a `NumpyExtensionArray`, and `PyArrow` dtypes lead to an `ArrowExtensionArray`, if not `ArrowStringArray`
- `Pandas` supports more dtypes in `astype` than in constructors
- `Index` does not support `np.float16`

- [x] Closes #1428
- [x] Tests added

### DIFF
--- a/tests/indexes/test_index_float.py
+++ b/tests/indexes/test_index_float.py
@@ -89,6 +89,7 @@ def test_constructor_dtype(
         assert_type(pd.Index([1.0], dtype="float64[pyarrow]"), "pd.Index[float]")
         assert_type(pd.Index([1.0], dtype="double[pyarrow]"), "pd.Index[float]")
 
+    # TODO: pandas-dev/pandas-stubs#1501
     # if TYPE_CHECKING_INVALID_USAGE:
     #     # numpy float16
     #     pd.Index([1.0], dtype=np.half)


tests/series/test_series.py
```diff
@@ -2364,50 +2367,10 @@ def test_change_to_dict_return_type() -> None:
...
-    ("f16", np.longdouble),
-    # ("float96", np.longdouble),  # NOTE: unsupported
-    ("float128", np.longdouble),  # NOTE: UNIX ONLY
-    # pyarrow float32
-    ("float32[pyarrow]", float),
-    ("float[pyarrow]", float),
-    # pyarrow float64
-    ("float64[pyarrow]", float),
-    ("double[pyarrow]", float),
-]
```
cmp0xff, line 2410: Moved to `tests/__init__.py`

tests/series/test_series.py
```diff
@@ -2738,68 +2705,6 @@ def test_astype_uint(cast_arg: IntDtypeArg, target_type: type) -> None:
         assert_type(s.astype("uint64[pyarrow]"), "pd.Series[int]")
 
 
-@pytest.mark.parametrize("cast_arg, target_type", ASTYPE_FLOAT_ARGS, ids=repr)
```
cmp0xff, line 2744: Moved to `test_series_float.py`

tests/indexes/test_index_float.py
```diff
@@ -0,0 +1,164 @@
...
+        assert_type(pd.Index([1.0], dtype="float64[pyarrow]"), "pd.Index[float]")
+        assert_type(pd.Index([1.0], dtype="double[pyarrow]"), "pd.Index[float]")
+
+    # if TYPE_CHECKING_INVALID_USAGE:
+    #     # numpy float16
+    #     pd.Index([1.0], dtype=np.half)
+    #     pd.Index([1.0], dtype="half")
+    #     pd.Index([1.0], dtype="float16")
+    #     pd.Index([1.0], dtype="e")
+    #     pd.Index([1.0], dtype="f2")
```
cmp0xff, line 99: At the moment, our `__new__` overloads in `Index` cover everything including these invalid ones. Maybe we will clean them up in the future and recognise them as invalid.

tests/indexes/test_index_float.py
```diff
@@ -0,0 +1,164 @@
...
+        assert_type(s.astype("float64[pyarrow]"), "pd.Index[float]")
+        assert_type(s.astype("double[pyarrow]"), "pd.Index[float]")
+
+    # if TYPE_CHECKING_INVALID_USAGE:
+    #     # numpy float16
+    #     s.astype(np.half)
+    #     s.astype("half")
+    #     s.astype("float16")
+    #     s.astype("e")
+    #     s.astype("f2")
```
cmp0xff, line 158: At the moment, our `__new__` overloads in `Index` cover everything including these invalid ones. Maybe we will clean them up in the future and recognise them as invalid.

pandas-stubs/_typing.pyi
```diff
@@ -296,76 +305,81 @@ UIntDtypeArg: TypeAlias = (
...
-        "clongdouble",
-        "clongfloat",
-        "longcomplex",
-    ]
+    | Literal["G", "clongdouble"]
 )
-# Refer to https://numpy.org/doc/stable/reference/arrays.datetime.html#datetime-units
-TimedeltaDtypeArg: TypeAlias = Literal[
+ComplexDtypeArg: TypeAlias = BuiltinComplexDtypeArg | NumpyComplexDtypeArg
+PandasAstypeTimedeltaDtypeArg: TypeAlias = Literal[
```
cmp0xff, line 375:
```py
import pandas as pd

pd.Series(["2025-01-01"], dtype="datetime64[D]")  # TypeError: dtype=datetime64[D] is not supported. Supported resolutions are 's', 'ms', 'us', and 'ns'
```

These only work upon `astype`.

pandas-stubs/_typing.pyi
```diff
@@ -296,76 +305,81 @@ UIntDtypeArg: TypeAlias = (
...
+NumpyComplexDtypeArg: TypeAlias = (
     # https://numpy.org/doc/stable/reference/arrays.scalars.html#numpy.csingle
-    | type[np.csingle]
-    | Literal["F", "c8", "complex64", "csingle", "singlecomplex"]
+    type[np.csingle]  # noqa: PYI030
+    | Literal["F", "c8", "complex64", "csingle"]
     # https://numpy.org/doc/stable/reference/arrays.scalars.html#numpy.cdouble
     | type[np.cdouble]
-    | Literal["D", "c16", "complex128", "cdouble", "cfloat", "complex_"]
+    | Literal["D", "c16", "complex128", "cdouble"]
```
cmp0xff, line 368: The removed values do not work for me in pytests at runtime. [I think they are `NON_NUMPY20_ALIASES`](https://github.com/pandas-dev/pandas-stubs/commit/56ddb6fb4e09e63a778c401929d516f8610ae8f0).

tests/__init__.py
```diff
@@ -67,6 +79,364 @@
         np_ndarray_td as np_ndarray_td,
     )
 else:
+    # Builtin bool type and its string alias
+    BuiltinBooleanDtypeArg: TypeAlias = type[bool] | Literal["bool"]
+    # Pandas nullable boolean type and its string alias
+    PandasBooleanDtypeArg: TypeAlias = pd.BooleanDtype | Literal["boolean"]
```
cmp0xff, line 83: It is a bit annoying that all these has to be duplicated in `tests/__init__.py`.

pandas-stubs/core/indexes/base.pyi
```diff
@@ -161,9 +175,8 @@ class Index(IndexOpsMixin[S1], ElementOpsMixin[S1]):
     @overload
     def __new__(
         cls,
-        data: Sequence[float | np.floating] | IndexOpsMixin[float] | np_ndarray_float,
```
Dr-Irv, line 164: I don't see how you can drop `IndexOpsMixin[float]` here

pandas-stubs/_typing.pyi
Dr-Irv: they were removed in numpy 2.0 .  I think you have to leave them in case people are still not using numpy 2.x

tests/arrays/test_floating_array.py
```diff
@@ -0,0 +1,48 @@
...
+def test_constructor() -> None:
+    check(assert_type(pd.array([1.0]), FloatingArray), FloatingArray)
+    check(assert_type(pd.array([1.0, np.float64(1)]), FloatingArray), FloatingArray)
+    check(assert_type(pd.array([1.0, None]), FloatingArray), FloatingArray)
+    check(assert_type(pd.array([1.0, pd.NA, None]), FloatingArray), FloatingArray)
+
+    check(
+        assert_type(  # type: ignore[assert-type] # I do not understand
+            pd.array(np.array([1.0], np.float64)), FloatingArray
+        ),
```
Dr-Irv: This has to do with bi-directional type matching.

If you do:
```python
    nparray = np.array([1.0], np.float64)
    check(assert_type(pd.array(nparray), FloatingArray), FloatingArray)
```
it works.

Probably an odd mypy bug, but at least we have a workaround

tests/indexes/test_index_float.py
```diff
@@ -0,0 +1,162 @@
...
+
+def test_constructor() -> None:
+    check(assert_type(pd.Index([1.0]), "pd.Index[float]"), pd.Index, np.floating)
+    check(
+        assert_type(pd.Index([1.0, np.float64(1)]), "pd.Index[float]"),
+        pd.Index,
+        np.floating,
+    )
+    check(
+        assert_type(  # type: ignore[assert-type] # I do not understand
```
Dr-Irv: see above.  put the `np.array()` in a variable, then pass it in

tests/indexes/test_index_float.py
Dr-Irv: create an issue or a TODO ?

tests/__init__.py
Dr-Irv: Here's a possible way we can handle this.

What if we put all these declarations in a file `_typing_dtypes.pyi`, and `_typing.pyi` imports that file.

Then when we run the tests, we copy `_typing_dtypes.pyi` to a file `tests/typing_dtypes.py` and then import that file here.

So then we have exactly one copy.   We just have to engineer the tests to do the copying first.

cmp0xff: I tried to move but it caused unrelated errors. Let's deal with it in a separate PR.

tests/arrays/test_floating_array.py
cmp0xff: d810f8d1ddf2a712302934d2228e1c8533429b0e

pandas-stubs/core/indexes/base.pyi
cmp0xff: We have this overload a few lines below:

```py
    @overload
    def __new__(
        cls,
        data: (
            S1 | _DataLikeS1[S1] | dict[HashableT1, S1] | KeysView[S1] | ValuesView[S1]
        ),
        index: AxesData | None = None,
        dtype: Dtype | None = None,
        name: Hashable = None,
        copy: bool | None = None,
    ) -> Self: ...
```

pandas-stubs/_typing.pyi
cmp0xff: `cfloat` and `complex_` works with `astype` but not with `dtype=...`. I have kept them in `PandasAstypeComplexDtypeArg`. What would you say? Personally, I prefer to keep numpy 2.x users happy.

tests/indexes/test_index_float.py
cmp0xff: d810f8d1ddf2a712302934d2228e1c8533429b0e

tests/indexes/test_index_float.py
cmp0xff: pandas-dev/pandas-stubs#1501 e7db317c4c0a181e8eb97ec47e1bc13958556c8a

pandas-stubs/_typing.pyi
Dr-Irv: So this is the tradeoff:
- If we keep `"cfloat"` and `"complex_"`, and someone is using numpy 1.22.4 (which is what pandas requires) and uses those types, their code works.  On the other hand, if someone is using numpy 2.x, and uses those types, we don't detect the error in static typing.
- If we remove `"cfloat"` and `"complex_"`, and someone is using numpy 1.22.4  and uses those types, then we raise an error in static typing even though it is valid.  On the other hand, if someone is using numpy 2.x, and uses those types, we do detect the error in static typing.

So now that I've written that all out, I think the removal makes sense, as we want to push people to the latest releases of required software.

tests/__init__.py
Dr-Irv: I created an issue.

pandas-stubs/core/indexes/base.pyi
Dr-Irv: OK

PR #1497, opened 2025-11-21
ENH: #1410 `DataFrame.loc` with `Iterable`

### LINKED ISSUE #1410: `DataFrame.loc` does not like `Sequence[Hashable]` for the columns
## Describe the bug

`DataFrame.loc` does not like `Sequence[Hashable]` for the columns

### Script

```py
from collections.abc import Hashable, Sequence
from typing import cast

import pandas as pd

a = cast(Sequence[Hashable], ["a"])
pd.DataFrame({"a": [1]}).loc[:, a]
```

### Typechecker
`pyright`

### Error message

```text
  ttest.py:7:1 - error: No overloads for "__getitem__" match the provided arguments (reportCallIssue)
  ttest.py:7:1 - error: Argument of type "tuple[slice[None, None, None], Sequence[Hashable]]" cannot be assigned to parameter "idx" of type "tuple[Scalar, slice[Any, Any, Any]]" in function "__getitem__"
    "tuple[slice[None, None, None], Sequence[Hashable]]" is not assignable to "tuple[Scalar, slice[Any, Any, Any]]"
        Type "slice[None, None, None]" is not assignable to type "Scalar"
          "slice[None, None, None]" is not assignable to "str"
          "slice[None, None, None]" is not assignable to "bytes"
          "slice[None, None, None]" is not assignable to "date"
          "slice[None, None, None]" is not assignable to "datetime"
          "slice[None, None, None]" is not assignable to "timedelta" (reportArgumentType)     
2 errors, 0 warnings, 0 informations
```

## Informations

 - version of type checker: 1.1.406
 - version of installed `pandas-stubs`: `main`

cmp0xff:
Note `str` is also `Iterable[HashableT]` but `DataFrame.loc[..., str]` produces `Series` instead of `DataFrame`, so I need to shuffle the order of the overloads.

- [x] Closes #1410 
- [x] Tests added

### DIFF
--- a/tests/test_frame.py
+++ b/tests/test_frame.py
@@ -2,7 +2,9 @@ from __future__ import annotations
 
 from collections import (
     OrderedDict,
+    UserList,
     defaultdict,
+    deque,
 )
 from collections.abc import (
     Callable,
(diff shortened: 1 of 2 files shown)


tests/test_frame.py
```diff
@@ -3718,12 +3720,28 @@ def test_loc_int_set() -> None:
...
+        assert_type(df.loc[:, (1,)], pd.DataFrame)
+
+        assert_type(df.loc[:, deque([None])], pd.DataFrame)
+        assert_type(df.loc[:, deque([1])], pd.DataFrame)
+
+        assert_type(df.loc[:, UserList([None])], pd.DataFrame)
+        assert_type(df.loc[:, UserList([1])], pd.DataFrame)
+
+        assert_type(df.loc[:, (None for _ in [0])], pd.DataFrame)
+        assert_type(df.loc[:, (1 for _ in [0])], pd.DataFrame)
```
Dr-Irv, line 3744: I really would prefer to always use the `check(assert_type()` pattern and not use `parameterize`.  I don't see why you just don't have `check(assert_type(` here.

cmp0xff: This is an alternative design pattern in the existing tests, see pandas-dev/pandas-stubs#1493 where I migrated such a test and added more.

To me, the advantage is that we see more runtime errors when several tests fail. I can switch back to the standard design pattern.

Dr-Irv: OK, it's fine.  In reviewing your other PR, I see that we use that pattern.

Issue #1485, opened 2025-11-13
Typing Index


MarcoGorelli:
Here's the methods in Index which are missing some type annotation or return type

- [x] is_
- [x] `__array_wrap__`
- [x] ravel
- [ ] view
- [x] take
- [x] repeat
- [x] `__copy__`
- [x] `to_flat_index`
- [x] `to_frame`
- [x] `set_names`
- [x] rename
- [x] `drop_level`
- [x] `__reduce__`
- [x] `isna`
- [x] `notna`
- [x] `fillna`
- [x] `memory_usage`
- [x] `__setitem__`
- [ ] `putmask`
- [x] `identical`
- [ ] `asof`
- [ ] `asof_locs`
- [ ] `sort_values`
- [ ] `get_indexer_non_unique`
- [ ] `get_indexer_for`
- [ ] `map`
- [ ] `isin`
- [ ] `slice_indexer`
- [ ] `get_slice_bound`
- [ ] `slice_locs`

Issue #1464, opened 2025-11-01
Handling of np.int64/np.float64 in `Series.sum()`


loicdiridollou:
**Describe the bug**

At runtime, `Series.sum` returns a np.int64 or np.float64 when summing numbers. Yet the typing says we should get an int or a float.
The issue arises when we use method on that result that are only defined on np.int64 or np.float64 (for example the clip method).

How would we want to handle those cases, are we open to adjusting the stubs to better type the behavior?

**To Reproduce**
1. Provide a minimal runnable `pandas` example that is not properly checked by the stubs.
2. Indicate which type checker you are using (`mypy` or  `pyright`).
3. Show the error message received from that type checker while checking your example.

```python
import pandas as pd
import numpy as np


a = pd.Series([1.,1.])
b = pd.Series([2.,2.])

a.mul(b).sum().clip(max=3)
```

The error we get is `Cannot access attribute "clip" for class "float"`

**Please complete the following information:**
 - OS: MacOS
 - OS Version latest
 - python version 3.13
 - version of type checker 1.18
 - version of installed `pandas-stubs` main

**Additional context**

Dr-Irv:
> I think 2/3 of the backends use `numpy` types. For now maybe we can redirect the element types and the results of such functions to the `numpy` types.

We can try that, but I think it will open up a whole host of issues.  I don't necessarily think that users should expect numpy types from the results of pandas operations.

The docs are unclear here because the examples show a result of a `float`, not `numpy float`.  I think this has to do with how `numpy` changed the `__repr__()` or `__str__()` of its results at some point.

I created a docs issue for discussion in the pandas team:  https://github.com/pandas-dev/pandas/issues/62966

loicdiridollou:
Let's close this one for now, I think it may remain a tricky issue to handle.

Dr-Irv:
Note that `pandas` introduced a `future.python_scalars` option in 3.0  where python scalars would always be returned.  So I think closing this was the right call.

PR #1505, opened 2025-11-25
type Index methods: putmask, asof, asof_locs, sort_values, get_indexer_non_unique, get_indexer_for, map, get_slice_bound

MarcoGorelli:
towards #1485 

### DIFF
--- a/tests/indexes/test_indexes.py
+++ b/tests/indexes/test_indexes.py
@@ -1625,9 +1625,9 @@ def test_index_sort_values() -> None:
 
 def test_index_get_indexer_non_unique() -> None:
     idx = pd.Index([1, 3])
-    res = idx.get_indexer_non_unique(pd.Index([3]))
-    check(assert_type(res[0], np_1darray_intp), np_1darray_intp)
-    check(assert_type(res[1], np_1darray_intp), np_1darray_intp)
+    indexer, missing = idx.get_indexer_non_unique(pd.Index([3]))
+    check(assert_type(indexer, np_1darray_intp), np_1darray_intp)
+    check(assert_type(missing, np_1darray_intp), np_1darray_intp)
 
 
 def test_index_get_indexer_for() -> None:
@@ -1655,6 +1655,6 @@ def test_index_get_slice_bound() -> None:
 
 def test_index_slice_locs() -> None:
     idx = pd.Index([1, 3])
-    res = idx.slice_locs(0, 1)
-    check(assert_type(res[0], np.intp | int), np.integer)
-    check(assert_type(res[1], np.intp | int), int)
+    start, end = idx.slice_locs(0, 1)
+    check(assert_type(start, np.intp | int), np.integer)
+    check(assert_type(end, np.intp | int), int)


pandas-stubs/core/indexes/base.pyi
```diff
@@ -532,39 +515,59 @@ class Index(IndexOpsMixin[S1], ElementOpsMixin[S1]):
     ) -> Index[C2]: ...
     @overload
     def append(self, other: Index | Sequence[Index]) -> Index: ...
-    def putmask(self, mask, value): ...
+    def putmask(
+        self,
+        mask: Sequence[bool] | np_ndarray_bool | BooleanArray | IndexOpsMixin[bool],
```
MarcoGorelli, line 519: i've kept this the same as 

https://github.com/pandas-dev/pandas-stubs/blob/54ccef83b44e9dbeffd4c0bd77a174ffbe17d978/pandas-stubs/core/indexes/base.pyi#L501

the original docstring seems to narrow, it says "np.ndarray[bool]" but then the example shows `list[bool]`

pandas-stubs/core/indexes/base.pyi
```diff
@@ -532,39 +515,59 @@ class Index(IndexOpsMixin[S1], ElementOpsMixin[S1]):
...
     def sort(self, *args: Any, **kwargs: Any) -> None: ...
     def argsort(self, *args: Any, **kwargs: Any) -> np_1darray_intp: ...
-    def get_indexer_non_unique(self, target): ...
+    def get_indexer_non_unique(
+        self, target: Index
+    ) -> tuple[np_1darray_intp, np_1darray_intp]: ...
     @final
-    def get_indexer_for(self, target, **kwargs: Any): ...
-    def map(self, mapper, na_action=...) -> Index: ...
+    def get_indexer_for(self, target: Index) -> np_1darray_intp: ...
```
MarcoGorelli, line 555: `get_indexer_for` doesn't accept kwargs

https://github.com/pandas-dev/pandas/blob/0e0bd864e6cce133ff94dffc3748d144a8839333/pandas/core/indexes/base.py#L6139

pandas-stubs/core/indexes/base.pyi
```diff
@@ -532,39 +515,59 @@ class Index(IndexOpsMixin[S1], ElementOpsMixin[S1]):
...
+        mask: Sequence[bool] | np_ndarray_bool | BooleanArray | IndexOpsMixin[bool],
+        value: Scalar,
+    ) -> Index: ...
     def equals(self, other: Any) -> bool: ...
     @final
     def identical(self, other: Any) -> bool: ...
     @final
-    def asof(self, label): ...
-    def asof_locs(self, where, mask): ...
+    def asof(self, label: Scalar) -> Scalar: ...
```
MarcoGorelli, line 526: the return type doesn't have to be `S1`, it can also be nan

```py
In [37]: i = pd.Index([1,2,3])

In [38]: i.asof(1.3)
Out[38]: np.int64(1)

In [39]: i.asof(-1)
Out[39]: nan
```

pandas-stubs/core/indexes/base.pyi
```diff
@@ -532,39 +515,59 @@ class Index(IndexOpsMixin[S1], ElementOpsMixin[S1]):
...
+    def asof_locs(
+        self, where: DatetimeIndex, mask: npt.NDArray[np.bool_]
```
cmp0xff:
```suggestion
        self, where: DatetimeIndex, mask: np_ndarray_bool
```

tests/indexes/test_indexes.py
```diff
@@ -1582,3 +1584,71 @@ def test_index_setitem() -> None:
     idx = pd.Index([1, 2])
     if TYPE_CHECKING_INVALID_USAGE:
         idx[0] = 999  # type: ignore[index]  # pyright: ignore[reportIndexIssue]
+
+
+def test_index_putmask() -> None:
+    idx = pd.Index([1, 2])
+    check(assert_type(idx.putmask([True, False], 11.4), "pd.Index"), pd.Index)
```
cmp0xff, line 1591: Could you test `Sequence[bool | np.bool_]` (note `np.bool_`), `np_ndarray_bool`, `BooleanArray` and `IndexOpsMixin[bool]`?

pandas-stubs/core/indexes/base.pyi
cmp0xff: Isn't it `S1 | float` (`float("nan")` to be precise, but we can't put it in a literal)?

The `S1` part has the same problem as in pandas-dev/pandas-stubs#1464, that for example, a user may expect native `int` but actually gets `np.integer`. Now that #1464 has not been resolved, how do we proceed, @Dr-Irv ?

tests/indexes/test_indexes.py
```diff
@@ -1582,3 +1584,71 @@ def test_index_setitem() -> None:
...
+        np_1darray_intp,
+    )
+
+
+def test_index_sort_values() -> None:
+    idx = pd.DatetimeIndex(["2020-01-01", "2020-01-02", "2020-01-03"])
+    check(assert_type(idx.sort_values(), pd.DatetimeIndex), pd.DatetimeIndex)
+    sorted_ = idx.sort_values(return_indexer=True)
+    check(assert_type(sorted_[0], pd.DatetimeIndex), pd.DatetimeIndex)
+    check(assert_type(sorted_[1], np_1darray_intp), np_1darray_intp)
```
cmp0xff: I think the following design pattern also implicitly tests the returned value is a 2-tuple
```suggestion
    sorted_index, indexer = idx.sort_values(return_indexer=True)
    check(assert_type(sorted_index, pd.DatetimeIndex), pd.DatetimeIndex)
    check(assert_type(indexer, np_1darray_intp), np_1darray_intp)
```

pandas-stubs/core/indexes/base.pyi
```diff
@@ -532,39 +515,59 @@ class Index(IndexOpsMixin[S1], ElementOpsMixin[S1]):
...
         end: Label | None = None,
         step: int | None = None,
-    ): ...
-    def get_slice_bound(self, label, side): ...
+    ) -> slice: ...
+    def get_slice_bound(self, label: Scalar, side: Literal["left", "right"]) -> int: ...
     def slice_locs(
         self, start: SliceType = None, end: SliceType = None, step: int | None = None
-    ): ...
+    ) -> tuple[int | np.integer, int | np.integer]: ...
```
cmp0xff: Maybe we can do `int | np.intp` here, to be consistent with `np_1darray_intp`. And also in the test.

cmp0xff (changes requested):

tests/indexes/test_indexes.py
MarcoGorelli: that would indeed raise at type-checking, but it's the same as it is in `where` / `case_when`

in practice i don't think this is common, and that this may require a project-wide change (int -> int | np.integer, bool -> bool | np.bool_)?

cmp0xff: Changing `Sequence[bool]` can be deferred pandas-dev/pandas-stubs#1506. Could you do `np_ndarray_bool`, `BooleanArray` and `IndexOpsMixin[bool]`?

cmp0xff (changes requested):

pandas-stubs/core/indexes/base.pyi
cmp0xff: pandas-dev/pandas#63209

cmp0xff (commented):
Let us wait a bit for https://github.com/pandas-dev/pandas-stubs/pull/1505/files#r2560788416.

tests/indexes/test_indexes.py
```diff
@@ -1582,3 +1584,77 @@ def test_index_setitem() -> None:
...
+    sorted_index, indexer = idx.sort_values(return_indexer=True)
+    check(assert_type(sorted_index, pd.DatetimeIndex), pd.DatetimeIndex)
+    check(assert_type(indexer, np_1darray_intp), np_1darray_intp)
+
+
+def test_index_get_indexer_non_unique() -> None:
+    idx = pd.Index([1, 3])
+    res = idx.get_indexer_non_unique(pd.Index([3]))
+    check(assert_type(res[0], np_1darray_intp), np_1darray_intp)
+    check(assert_type(res[1], np_1darray_intp), np_1darray_intp)
```
cmp0xff: The same reasoning as earlier
```suggestion
    indexer, missing = idx.get_indexer_non_unique(pd.Index([3]))
    check(assert_type(indexer, np_1darray_intp), np_1darray_intp)
    check(assert_type(missing, np_1darray_intp), np_1darray_intp)
```

tests/indexes/test_indexes.py
```diff
@@ -1582,3 +1584,77 @@ def test_index_setitem() -> None:
...
+def test_index_get_slice_bound() -> None:
+    idx = pd.Index([1, 3])
+    check(assert_type(idx.get_slice_bound(1, side="left"), int), int)
+
+
+def test_index_slice_locs() -> None:
+    idx = pd.Index([1, 3])
+    res = idx.slice_locs(0, 1)
+    check(assert_type(res[0], np.intp | int), np.integer)
+    check(assert_type(res[1], np.intp | int), int)
```
cmp0xff: The same reasoning as earlier
```suggestion
    start, end = idx.slice_locs(0, 1)
    check(assert_type(start, np.intp | int), np.integer)
    check(assert_type(end, np.intp | int), int)
```

pandas-stubs/core/indexes/base.pyi
Dr-Irv: Since the result can be `np.nan`, and `isinstance(np.nan, float)` is `True`, using `Scalar` here is appropriate.

Or you could use `S1 | float`

I don't think the issue is the same as in #1464 .  If you have `Index[S1]`,  you will get `S1` back OR `np.nan` .

cmp0xff: @Dr-Irv what I meant is the following:
```py
import pandas as pd

pd.Index([1, 2, 3]).asof(1)  # unexpectedly np.int64, not native int
pd.Index(["a", "b", "c"]).asof("c")  # native str, not np.str_
```
So it breaks the pattern of `S1 | float`, but instead something like `S1 | np.dtype[S1] | float` (pseudo-code, `np.dtype[S1]` won't work). That is the part that is related to #1464.

@MarcoGorelli could you add the example of `pd.Index(["a", "b", "c"]).asof("c")` in the test? I think it ensures the generality of `asof`, that it works not only for numeric values.

Dr-Irv: > So it breaks the pattern of `S1 | float`, but instead something like `S1 | np.dtype[S1] | float` (pseudo-code, `np.dtype[S1]` won't work). That is the part that is related to #1464.

That's why we have to use `Scalar`.  numpy types can also be returned like this:
```python
pd.Series([1,2,3]).iloc[1]
np.int64(2)
```

IMHO, I think in static typing, we should just say we return `int` or `float`, and NOT the numpy types.  There's some work going on in pandas to make this always the case in the future, via an option, which might eventually be the default.

So my philosophy here for `pandas-stubs` is:
- Our  methods should return the built-in types, and not do things like `-> int | np.integer`, because then users may have to cast.
- For parameters to our methods, we can accept both built-in and numpy types

cmp0xff: Thanks, let's do it like this for now.

PR #1509, opened 2025-11-30
GH1484 Upgrade ty

### LINKED ISSUE #1484: Update `ty` once UnionType is fixed
`ty` is currently blocked by some changes from 0.0.1a25 to 0.0.1a26. Once the following issue is resolved we can remove the constraint on `ty` version.

Blocked by https://github.com/astral-sh/ty/issues/1534#issuecomment-3524248647

loicdiridollou:
- [x] Closes #1484

A little bit of work on this one, needed to add some ignores but I think this should be good.

### DIFF
--- a/pandas-stubs/core/series.pyi
+++ b/pandas-stubs/core/series.pyi
@@ -337,9 +337,6 @@ class _LocIndexerSeries(_LocIndexer, Generic[S1]):
     ) -> None: ...
 
 _DataLike: TypeAlias = ArrayLike | dict[str, np_ndarray] | SequenceNotStr[S1]
-_DataLikeS1: TypeAlias = (
-    ArrayLike | dict[_str, np_ndarray] | Sequence[S1] | IndexOpsMixin[S1]
-)
 
 class Series(IndexOpsMixin[S1], ElementOpsMixin[S1], NDFrame):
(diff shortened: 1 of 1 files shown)


pandas-stubs/core/series.pyi
```diff
@@ -517,7 +517,11 @@ class Series(IndexOpsMixin[S1], ElementOpsMixin[S1], NDFrame):
     def __new__(
         cls,
         data: (
-            S1 | _DataLikeS1[S1] | dict[HashableT1, S1] | KeysView[S1] | ValuesView[S1]
+            S1
+            | _DataLikeS1[S1]  # ty: ignore[invalid-type-arguments]
```
loicdiridollou: Unsure here why we need to subscribe with `S1` since the definition mostly include `S1`, maybe we can refine:
```python
_DataLikeS1: TypeAlias = (
    ArrayLike | dict[_str, np_ndarray] | Sequence[S1] | IndexOpsMixin[S1]
)
```

cmp0xff: Could you do it or maybe just remove `_DataLikeS1`? It is used only here.

loicdiridollou: I have actually replaced the TypeAlias with all its elements and it is working without issues so that should be enough, just not sure why exactly.

PR #1510, opened 2025-11-30
Fix `Series.map` default for `na_action`

hamdanal:

`na_action` default to `None`


tests/series/test_series.py
cmp0xff, line 3721: I propose the following more comprehensive tests (L3587-L3602):

```suggestion
    check(assert_type(s.map(mapping), "pd.Series[str]"), pd.Series, str)

    def callable(x: int) -> str:
        return str(x)

    check(
        assert_type(s.map(callable, na_action="ignore"), "pd.Series[str]"),
        pd.Series,
        str,
    )
    check(assert_type(s.map(callable), "pd.Series[str]"), pd.Series, str)

    series = pd.Series(["a", "b", "c"])
    check(
        assert_type(s.map(series, na_action="ignore"), "pd.Series[str]"), pd.Series, str
    )
    check(assert_type(s.map(series), "pd.Series[str]"), pd.Series, str)

    unknown_series = pd.Series([1, 0, None])
    check(
        assert_type(
            unknown_series.map({1: True, 0: False, None: None}, na_action="ignore"),
            pd.Series,
        ),
        pd.Series,
    )
```

Your changes mark `check(assert_type(s.map(callable), "pd.Series[str]"), pd.Series, str)` as invalid, which actually works at runtime.

pandas-stubs/core/series.pyi
cmp0xff: If my comment on the tests is correct, we can do the following to make all tests work (L1107-L1117):

```suggestion
    def map(
        self,
        arg: Callable[[S1], S2 | NAType] | Mapping[S1, S2] | Series[S2],
        na_action: Literal["ignore"] | None = None,
    ) -> Series[S2]: ...
```

cmp0xff (changes requested):
I tried your idea myself and found one more unaddressed case.

tests/series/test_series.py
```diff
@@ -3709,16 +3710,35 @@ def callable(x: int) -> str:
...
     check(
         assert_type(s.map(series, na_action="ignore"), "pd.Series[str]"), pd.Series, str
     )
+    check(assert_type(s.map(series), "pd.Series[str]"), pd.Series, str)
 
     unknown_series = pd.Series([1, 0, None])
     check(
-        assert_type(unknown_series.map({1: True, 0: False, None: None}), pd.Series),
+        assert_type(
+            unknown_series.map({1: True, 0: False, None: None}), "pd.Series[bool]"
```
cmp0xff, line 3724: This is not a good behaviour, because `unknown_series.map({1: True, 0: False, None: None})` gives `dtype=object` at runtime. We would like to have `Series[Any]` in such case.

tests/series/test_series.py
```diff
@@ -3709,16 +3710,35 @@ def callable(x: int) -> str:
...
+        ),
+        pd.Series,
+        bool,
+    )
+    check(
+        assert_type(
+            unknown_series.map({1: True, 0: False, None: None}, na_action="ignore"),
+            "pd.Series[bool]",
```
cmp0xff, line 3732: `unknown_series.map({1: True, 0: False, None: None}, na_action="ignore")` also gives `dtype=object` at runtime. We would like to have `Series[Any]` in such case.

pandas-stubs/core/series.pyi
```diff
@@ -1111,20 +1111,18 @@ class Series(IndexOpsMixin[S1], ElementOpsMixin[S1], NDFrame):
...
+            | Series[S2]
+        ),
+        na_action: Literal["ignore"] | None = None,
     ) -> Series[S2]: ...
     @overload
     def map(
         self,
-        arg: Callable[[Any], Any] | Mapping[Any, Any] | Series,
-        na_action: Literal["ignore"] | None = ...,
+        arg: Callable[[Any], object] | Mapping[Any, object] | Series[Any],
```
cmp0xff, line 1124:
```suggestion
        arg: Callable[[Any], Any] | Mapping[Any, Any] | Series,
```

1. Currently we prefer `Any` over `object` to denote "arbitrary" types
2. `Series` defaults to `Series[Any]`, because `S1` defaults to `Any`.

cmp0xff (changes requested):

PR #1522, opened 2025-12-03
TYP: enable `reportUnknownVariableType`

cmp0xff:
Towards #1171

- In the stub files, the additions only include `pyright: ignore[reportUnknownVariableType]` for third party Python modules
- In the tests, added `pyright: ignore[reportUnknownVariableType]`'s are from
  - Third party Python modules
  - Dummy variables in invalid tests (e.g. `_1 = "a" / 2  # pyright: ignore[reportUnknownVariableType]`)


pandas-stubs/core/algorithms.pyi
```diff
@@ -88,14 +87,6 @@ def factorize(
...
     size_hint: int | None = ...,
 ) -> tuple[np_1darray_int64, Categorical]: ...
-def value_counts(
-    values: AnyArrayLike | list | tuple,
-    sort: bool = True,
-    ascending: bool = False,
-    normalize: bool = False,
-    bins: int | None = None,
-    dropna: bool = True,
-) -> Series: ...
```
cmp0xff: Not documented

pandas-stubs/io/excel/_base.pyi
```diff
@@ -35,6 +34,10 @@ from pandas._typing import (
     WriteExcelBuffer,
 )
 
+from odf.opendocument import (  # pyright: ignore[reportMissingTypeStubs] # isort: skip
+    OpenDocument,  # pyright: ignore[reportUnknownVariableType]
+)
+
```
cmp0xff, line 49: `read_excel` cannot be fully typed because of this

tests/extension/decimal/array.py
```diff
@@ -172,19 +172,19 @@ def __array_ufunc__(
...
+            return arraylike.dispatch_ufunc_with_out(  # type: ignore[attr-defined] # pyright: ignore[reportAttributeAccessIssue,reportUnknownVariableType]
                 self, ufunc, method, *inputs, **kwargs
             )
 
         inputs = tuple(x._data if isinstance(x, DecimalArray) else x for x in inputs)
         result = getattr(ufunc, method)(*inputs, **kwargs)
 
         if method == "reduce":
-            result = arraylike.dispatch_reduction_ufunc(  # type: ignore[attr-defined] # pyright: ignore[reportAttributeAccessIssue]
+            result = arraylike.dispatch_reduction_ufunc(  # type: ignore[attr-defined] # pyright: ignore[reportAttributeAccessIssue,reportUnknownVariableType]
```
Dr-Irv: Is it possible to `cast` the result of `dispatch_reduction_ufunc` so that you can then reduce some of the `reportUnknownVariableType` here?

tests/extension/decimal/array.py
Dr-Irv, line 1: Just a note on this file.  It was initially copied from the `pandas` tests.  I'm fine if we put `cast` in here instead of the `reportUnknownVariableType` stuff.

tests/frame/test_frame.py
```diff
@@ -76,7 +76,7 @@
     _PandasNamedTuple: TypeAlias = tuple
 
 if not PD_LTE_23:
-    from pandas.errors import Pandas4Warning  # type: ignore[attr-defined] # pyright: ignore[reportAttributeAccessIssue,reportRedeclaration] # isort: skip
+    from pandas.errors import Pandas4Warning  # type: ignore[attr-defined] # pyright: ignore[reportAttributeAccessIssue,reportRedeclaration,reportUnknownVariableType] # isort: skip
```
Dr-Irv: We are going to have to add this type when 3.0 comes out, which will be in a week or two, so why not add it now?  Then you can avoid `reportUnknownVariableType` (and maybe the other ignores as well)

You'd have to make the change in the other files that reference `Pandas4Warning`

tests/__init__.py
```diff
@@ -151,7 +156,7 @@ def check(
 
     if not isinstance(value, dtype):
         raise RuntimeError(f"Expected type '{dtype}' but got '{type(value)}'")
-    return actual
+    return actual  # pyright: ignore[reportUnknownVariableType]
```
Dr-Irv, line 163: I'm surprised that `reportUnknownVariableType` is needed here, but not in line 136

cmp0xff: microsoft/pyright#11190

tests/extension/decimal/array.py
cmp0xff: 0f6bd21faf4ac3da3e38da07344874bd0452ebed

tests/frame/test_frame.py
cmp0xff: 0e7433b66946daab5cf51b79a860d5f66a081e13

tests/__init__.py
```diff
@@ -137,25 +139,21 @@ def check(
...
     if isinstance(actual, pd.Series):
-        value = actual.iloc[  # pyright: ignore[reportUnknownVariableType]
-            index_to_check_for_type
-        ]
+        value = cast(pd.Series, actual).iloc[index_to_check_for_type]
     elif isinstance(actual, pd.Index):
-        value = actual[  # pyright: ignore[reportUnknownVariableType]
-            index_to_check_for_type
-        ]
+        value = cast(pd.Index, actual)[index_to_check_for_type]
```
Dr-Irv, line 146: can you put a comment in here that indicates why the `cast` is needed (probably related to https://github.com/microsoft/pyright/issues/11190)

tests/extension/decimal/array.py
Dr-Irv: I don't see the change that addresses this file in the commit you linked

Dr-Irv (commented):
still some open issues

tests/extension/decimal/array.py
cmp0xff: Sorry, now address in db97f62d3830ddc38a10de99f738e0fec93e05fd

tests/__init__.py
cmp0xff: a6655367a00e2deea06013e00a203a9dfa0a8cab also reported microsoft/pyright#11191

Dr-Irv: looks like he rejected 11191 and 11190 so delete the 11191 comments

tests/__init__.py
```diff
@@ -139,21 +139,27 @@ def check(
...
         value = actual.obj  # type: ignore[attr-defined] # pyright: ignore[reportAttributeAccessIssue,reportUnknownVariableType]
     elif isinstance(actual, Iterable):
+        # T_co in Iterable[T_co] does not have a default value and `actual` is Iterable[Unknown] by pyright
         value = next(iter(cast("Iterable[Any]", actual)))
     else:
         assert hasattr(actual, attr)
         value = getattr(actual, attr)
 
     if not isinstance(value, dtype):
+        # TODO: pyright ignore may lift by microsoft/pyright#11191
```
Dr-Irv: remove comment

tests/extension/decimal/array.py
```diff
@@ -165,23 +166,22 @@ def to_numpy(
...
-        #
+    ) -> Any:
         if not all(
             isinstance(t, self._HANDLED_TYPES + (DecimalArray,)) for t in inputs
         ):
             return NotImplemented
 
         if "out" in kwargs:
-            return arraylike.dispatch_ufunc_with_out(  # type: ignore[attr-defined] # pyright: ignore[reportAttributeAccessIssue]
+            return cast("Callable[Concatenate[Any, np.ufunc, str, ...], Any]", arraylike.dispatch_ufunc_with_out)(  # type: ignore[attr-defined] # pyright: ignore[reportAttributeAccessIssue]
```
Dr-Irv: create a `TypeAlias` for `Callable[Concatenate[Any, np.ufunc, str, ...], Any]` here and use that here and in line 184

tests/extension/decimal/array.py
```diff
@@ -200,19 +200,15 @@ def reconstruct(
...
 
     def __getitem__(self, item: ScalarIndexer | SequenceIndexer) -> Any:
         if isinstance(item, numbers.Integral):
             return self._data[item]
         # array, slice.
-        item = check_array_indexer(
-            self,
-            item,  # type: ignore[arg-type] # pyright: ignore[reportArgumentType,reportCallIssue]
+        item = check_array_indexer(  # pyright: ignore[reportCallIssue,reportUnknownVariableType]
+            self, item  # type: ignore[arg-type] # pyright: ignore[reportArgumentType]
```
Dr-Irv: Is it possible to fix the declaration of `check_array_indexer()` so that we don't need all these ignores?

tests/extension/decimal/array.py
```diff
@@ -351,10 +347,12 @@ def convert_values(param: Any) -> ExtensionArray | list[Any]:
 
     def value_counts(self, dropna: bool = True) -> Series:
         from pandas.core.algorithms import (  # type: ignore[attr-defined] # isort: skip
-            value_counts,  # pyright: ignore[reportAttributeAccessIssue]
+            value_counts,  # pyright: ignore[reportAttributeAccessIssue,reportAttributeAccessIssue,reportUnknownVariableType]
```
Dr-Irv: why list `reportAttributeAccessIssue` twice?

Can we just add `value_counts` to `algorithms`?  I know that violates our documentation rule, but in this case, since we see an example where its needed, I think we should add it

tests/extension/decimal/array.py
```diff
@@ -351,10 +347,12 @@ def convert_values(param: Any) -> ExtensionArray | list[Any]:
...
         )
 
-        return value_counts(self.to_numpy(), dropna=dropna)
+        return cast(
+            "Callable[Concatenate[np_ndarray, ...], Series[int]]", value_counts
+        )(self.to_numpy(), dropna=dropna)
```
Dr-Irv: if we add `value_counts` to algorithms, we can make this `cast` go away

Dr-Irv (changes requested):

tests/__init__.py
cmp0xff: 8de99ccdd4635190462fa5611b13261bfab3e6f8 I edited the comment which now says the ignore is by design

tests/__init__.py
cmp0xff: 8de99ccdd4635190462fa5611b13261bfab3e6f8 I edited the comment which now says the ignore is by design

tests/extension/decimal/array.py
cmp0xff: Following `value_counts`, I added the function to `arraylike` as well as a comment, saying the unpublished function is added for typing purposes fa0a5369736620c7bcb4d24addaedab7f5c55126

tests/extension/decimal/array.py
cmp0xff: 332554bf615d8d1d1e34d87751813e39264c2cc1

tests/extension/decimal/array.py
cmp0xff: fa0a5369736620c7bcb4d24addaedab7f5c55126

tests/extension/decimal/array.py
cmp0xff: fa0a5369736620c7bcb4d24addaedab7f5c55126

tests/extension/decimal/array.py
```diff
@@ -165,25 +169,20 @@ def to_numpy(
 
     def __array_ufunc__(
         self, ufunc: np.ufunc, method: str, *inputs: Any, **kwargs: Any
-    ) -> arraylike.dispatch_ufunc_with_out:  # type: ignore[name-defined] # pyright: ignore[reportAttributeAccessIssue]
-        #
+    ) -> Any:
```
Dr-Irv, line 172: Looks like the typing here was wrong in the first place.  It's wrong in the pandas source.

Dr-Irv (approved):
thanks @cmp0xff

PR #1526, opened 2025-12-03
TYP: `collections.abc.Callable`

cmp0xff:

`collections.abc.Callable` has no default value for the arguments. Replacing all bare `Callable` with `Callable[..., Any]`, so that `pyright_strict` is happier.

### DIFF
--- a/pandas-stubs/_typing.pyi
+++ b/pandas-stubs/_typing.pyi
@@ -612,7 +612,7 @@ AggFuncTypeDict: TypeAlias = (
     AggFuncTypeDictSeries[HashableT, P] | AggFuncTypeDictFrame[HashableT, P]
 )
 AggFuncType: TypeAlias = (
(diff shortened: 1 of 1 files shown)


loicdiridollou:
Maybe a little bit of a question but should we obsess with pyright_strict and would it become a requirement in the future? It is known to be quite difficult to mix pandas and pyright strict so wondering if that would cause a heavy burden for future PRs? Open to different views @Dr-Irv @MarcoGorelli

cmp0xff:
Indeed, `pyright_strict` errors will reduce from 3264 [in `main`](https://github.com/pandas-dev/pandas-stubs/actions/runs/19915891527/job/57094396751) to 1701 in [this pipeline](https://github.com/pandas-dev/pandas-stubs/actions/runs/19930857635/job/57142262552) (for now) if #1522 is finished

loicdiridollou:
Okay just wanted to make sure this was a route we wanted to take since there is quite a bit of work to make it happen, thanks @cmp0xff !

PR #1527, opened 2025-12-04
CLN: remove ignoring `bad-param-name-override` of `pyrefly` from `pyproject.toml`

cmp0xff:
- There are a few places where we need to add `# pyrefly: ignore[bad-param-name-override]`
- Other places can be fixed by correcting non-functional param names
- In most places we have one space between two ignores on the same line. I have fixed those with two spaces.

### DIFF
--- a/pandas-stubs/core/arrays/datetimelike.pyi
+++ b/pandas-stubs/core/arrays/datetimelike.pyi
@@ -72,7 +72,7 @@ class DatetimeLikeArrayMixin(ExtensionOpsMixin, ExtensionArray):
     def size(self) -> int: ...
     @overload
     def __getitem__(  # pyrefly: ignore[bad-override,bad-param-name-override]
-        self, item: ScalarIndexer
+        self, key: ScalarIndexer
     ) -> DTScalarOrNaT: ...
     @overload
     def __getitem__(  # ty: ignore[invalid-method-override]


pandas-stubs/core/arrays/datetimelike.pyi
```diff
@@ -71,11 +71,10 @@ class DatetimeLikeArrayMixin(ExtensionOpsMixin, ExtensionArray):
     @property
     def size(self) -> int: ...
     @overload
-    def __getitem__(self, key: ScalarIndexer) -> DTScalarOrNaT: ...
+    def __getitem__(self, item: ScalarIndexer) -> DTScalarOrNaT: ...
```
loicdiridollou: Why the change here? I think pandas code has key and not item. Same in a few other places, not sure item is used that much in the pandas code.

cmp0xff: Good point! The motivation was to keep `pyrefly` happy, which requires the param name identical to the one in the _base class_. But I believe we should really keep them identical to the one _in `pandas` in the concrete implementation_, which do not necessarily follow the `pyrefly` rule.

Incidently, in `pandas` there are _overloads_ of the concrete implementation which have different param names than the ones in the concrete implementation. I've chosen to follow the concrete implementation.

d71f9ab9fc557174b91fcc9858ee5987d0d636dd

cmp0xff: Basically I have the following example:

```py
from typing import overload


@overload
def foo(x: int) -> int: ...

@overload
def foo(x: str) -> str: ...

def foo(z: int | str):
    return z

foo(x=1)  # passes type check at this line, but fails at runtime
```

This is the implementation in `DatetimeLikeArrayMixin.__getitem__` in 2.3.3. in `main` of `pandas` they have been made identical.

Dr-Irv: > Good point! The motivation was to keep `pyrefly` happy, which requires the param name identical to the one in the _base class_. But I believe we should really keep them identical to the one _in `pandas` in the concrete implementation_, which do not necessarily follow the `pyrefly` rule.
> 
> Incidently, in `pandas` there are _overloads_ of the concrete implementation which have different param names than the ones in the concrete implementation. I've chosen to follow the concrete implementation.

I think the rule of thumb here that is equivalent is that the names we use in the stubs should also correspond to what is in the documentation, which I believe are also the names in the concrete implementation.

loicdiridollou: I see the argument name as `key` in the code:
https://github.com/pandas-dev/pandas/blob/fa1360d41a49173ddfc676c8194ca424871338c8/pandas/core/arrays/datetimelike.py#L376-L385

pandas-stubs/core/arrays/datetimelike.pyi
```diff
@@ -71,11 +71,12 @@ class DatetimeLikeArrayMixin(ExtensionOpsMixin, ExtensionArray):
...
     @overload
-    def __getitem__(self, key: ScalarIndexer) -> DTScalarOrNaT: ...
+    def __getitem__(  # pyrefly: ignore[bad-override,bad-param-name-override]
+        self, key: ScalarIndexer
+    ) -> DTScalarOrNaT: ...
     @overload
     def __getitem__(  # ty: ignore[invalid-method-override]
-        self,
-        item: SequenceIndexer | PositionalIndexerTuple,
+        self, key: SequenceIndexer | PositionalIndexerTuple
```
cmp0xff, line 79: @loicdiridollou regarding https://github.com/pandas-dev/pandas-stubs/pull/1527#discussion_r2591146608, I think I have already changed the argument names to `key`?

loicdiridollou: Arghh sorry I prob was looking at the wrong place, looks good.

loicdiridollou (approved):

loicdiridollou:
Thanks @cmp0xff

PR #1528, opened 2025-12-04
TYP: `tuple`

cmp0xff:

`tuple` has no default value for the arguments. Replacing all bare `tuple`'s with `tuple[`something`]`, so that `pyright_strict` is happier.


loicdiridollou (approved):
Looks good thanks @cmp0xff

PR #1532, opened 2025-12-08
TYP: `list`, `Sequence`, etc.

cmp0xff:

`list`, `Sequence` etc. has no default value for the arguments. Replacing all such typing annotations with `list[`something`]` etc., so that `pyright_strict` is happier.

### DIFF
--- a/pandas-stubs/io/formats/style.pyi
+++ b/pandas-stubs/io/formats/style.pyi
@@ -1,5 +1,6 @@
 from collections.abc import (
     Callable,
+    Hashable,
     MutableMapping,
     Sequence,
 )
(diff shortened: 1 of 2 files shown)


pandas-stubs/core/indexes/multi.pyi
```diff
@@ -91,16 +90,6 @@ class MultiIndex(Index):
...
-    def format(
-        self,
-        name: bool | None = ...,
-        formatter: Callable[..., Any] | None = ...,
-        na_rep: str | None = ...,
-        names: bool = ...,
-        space: int = ...,
-        sparsify: bool | None = ...,
-        adjoin: bool = ...,
-    ) -> list: ...
```
cmp0xff, line 103: Undocumented

PR #1537, opened 2025-12-09
TYP: `dict`, `Mapping`, etc. 

cmp0xff:

`dict`, `Mapping` etc. have no default value for the arguments. Replacing all such typing annotations with `dict[Hashable,`something`]` etc., so that `pyright_strict` is happier.


pandas-stubs/core/reshape/pivot.pyi
```diff
@@ -67,11 +67,17 @@ _Values: TypeAlias = SequenceNotStr[Any] | _ExtendedAnyArrayLike
 @overload
 def pivot_table(
     data: DataFrame,
-    values: _PivotTableValuesTypes = None,
-    index: _PivotTableIndexTypes = None,
-    columns: _PivotTableColumnsTypes = None,
+    values: _PivotTableValuesTypes[
```
loicdiridollou, line 70: Did we open an issue for that one on ty ?

loicdiridollou (approved):
Looks good, one minor thing about the `ty` issue, please check that but will go ahead and merge.

pandas-stubs/core/reshape/pivot.pyi
cmp0xff: Good idea, astral-sh/ty#1846

PR #1539, opened 2025-12-10
Type multi.pyi

MarcoGorelli:
the following just get inherited from Index:
- shape
- `__contains__`
- values
- duplicated
- get_indexer
- get_indexer_non_unique
- reindex
- equals
- insert
- delete

the following are undocumented:
- `__reduce__`

the rest i've typed

### DIFF
--- a/tests/indexes/test_multi_indexes.py
+++ b/tests/indexes/test_multi_indexes.py
@@ -1,5 +1,6 @@
 from __future__ import annotations
 
+import numpy as np
 import pandas as pd
 from typing_extensions import (
     assert_type,
@@ -45,7 +46,7 @@ def test_multiindex_set_codes() -> None:
 def test_multiindex_view() -> None:
     mi = pd.MultiIndex.from_arrays([[1, 2, 3], [4, 5, 6]])
     check(assert_type(mi.view(), pd.MultiIndex), pd.MultiIndex)
-    check(assert_type(mi.view(pd.Index), pd.MultiIndex), pd.MultiIndex)
+    check(assert_type(mi.view(np.ndarray), pd.MultiIndex), pd.MultiIndex)
 
 
 def test_multiindex_remove_unused_levels() -> None:


pandas-stubs/core/indexes/multi.pyi
```diff
@@ -70,19 +72,46 @@ class MultiIndex(Index):
...
+        *,
+        level: Level,
+        verify_integrity: bool = True,
+    ) -> MultiIndex: ...
     def copy(  # type: ignore[override] # pyright: ignore[reportIncompatibleMethodOverride] # pyrefly: ignore
         self, names: SequenceNotStr[Hashable] = ..., deep: bool = False
     ) -> Self: ...
-    def view(self, cls=...): ...
-    def __contains__(self, key) -> bool: ...
+    def view(self, cls: Any = None) -> MultiIndex: ...  # type: ignore[override] # pyrefly: ignore[bad-override] # pyright: ignore[reportIncompatibleMethodOverride]
```
MarcoGorelli: this is an odd one, for multiindex.view, cls is ignored 🤷

pandas-stubs/core/indexes/multi.pyi
```diff
@@ -125,36 +159,33 @@ class MultiIndex(Index):
...
+    def get_locs(self, seq: Level | Sequence[Level]) -> np_1darray_intp: ...
     def truncate(
         self, before: IndexLabel | None = None, after: IndexLabel | None = None
-    ): ...
-    def equals(self, other) -> bool: ...
-    def equal_levels(self, other): ...
-    def insert(self, loc, item): ...
-    def delete(self, loc): ...
+    ) -> MultiIndex: ...
+    def equal_levels(self, other: MultiIndex) -> bool: ...
```
loicdiridollou: Is this actually documented? I can't find it in the docs

pandas-stubs/core/indexes/multi.pyi
loicdiridollou: I think we may be able to refine it, docs suggest: `data-type or ndarray sub-class or None`

pandas-stubs/core/indexes/multi.pyi
loicdiridollou: Could not find it at all, prob worth deleting unless you have seen use cases in production.

loicdiridollou (changes requested):
One minor question but otherwise looking good!

pandas-stubs/core/indexes/multi.pyi
MarcoGorelli: sure, i've made it the same as index.view but without the overloads

pandas-stubs/core/indexes/multi.pyi
MarcoGorelli: well i'm always happy to delete code 🔥

Issue #1538, opened 2025-12-09
`__setitem__` should accept sequences containing None?


anentropic:
**Describe the bug**
I have code a bit like:

```python
In [1]: import pandas as pd
   ...:
   ...: df = pd.DataFrame({"flag": [True, False], "count": [1, 0]})
   ...:
   ...: df["result"] = [
   ...:     1 if flag and count > 0 else None
   ...:     for flag, count in zip(df["flag"], df["count"], strict=True)
   ...: ]

In [2]: df['result']
Out[2]:
0    1.0
1    NaN
Name: result, dtype: float64
```

The code works as intended.

Pylance complains:

> No overloads for `"__setitem__"` match the provided arguments Pylance [reportCallIssue](https://github.com/microsoft/pylance-release/blob/main/docs/diagnostics/reportCallIssue.md)
> frame.pyi(849, 9): Overload 2 is the closest match
> Argument of type "list[Any | None]" cannot be assigned to parameter "value" of type "Scalar | ArrayLike | NAType | NaTType | IndexOpsMixin[Any, Any] | Sequence[Scalar] | Sequence[Sequence[Scalar]] | DataFrame | Mapping[Hashable, Scalar | NAType | NaTType] | None" in function `"__setitem__"`
>   Type "list[Any | None]" is not assignable to type "Scalar | ArrayLike | NAType | NaTType | IndexOpsMixin[Any, Any] | Sequence[Scalar] | Sequence[Sequence[Scalar]] | DataFrame | Mapping[Hashable, Scalar | NAType | NaTType] | None"
>     "list[Any | None]" is not assignable to "str"
>     "list[Any | None]" is not assignable to "bytes"
>     "list[Any | None]" is not assignable to "date"
>     "list[Any | None]" is not assignable to "datetime"
>     "list[Any | None]" is not assignable to "timedelta"
>     "list[Any | None]" is not assignable to "datetime64[date | int | None]"
>     "list[Any | None]" is not assignable to "timedelta64[timedelta | int | None]"

Looking at the types in `frame.pyi` we have something like:
```python
    @overload
    def __setitem__(
        self,
        idx: (
            int
            | IndexType
            | tuple[int, int]
            | tuple[IndexType, int]
            | tuple[IndexType, IndexType]
            | tuple[int, IndexType]
        ),
        value: (
            Scalar
            | IndexOpsMixin
            | Sequence[Scalar]
            | DataFrame
            | np_ndarray
            | NAType
[...]

PR #1540, opened 2025-12-10
CLN: refactor `test_frame`

cmp0xff:
Towards #1538.

I would like to add tests, but `test_frame` is gigantic, and there are so many tests related to getitem and setitem.

This PR separates two groups of tests from `test_frame`: `groupby` and `indexing` (setitem + getitem).

- We should have exactly the same tests before and after this refactor.
- `test_groupby` should contain only `groupby`-related tests
- `test_indexing` should contain only tests related to getitem and setitem.


loicdiridollou (approved):
Looks good, will trust you that no tests fell through in the moving process, thanks @cmp0xff !

PR #1546, opened 2025-12-13
Add variance checks to aggregation tests and type assertions

### LINKED ISSUE #1545: Scalar type to float error with pyright
**Describe the bug**

Type checker error when converting the result of `Series.var()` (and other statistical methods) to `float`. The type checker reports that `Scalar` cannot be assigned to parameter `x` of type `ConvertibleToFloat`, even though this conversion works correctly at runtime.

**To Reproduce**

1. Minimal runnable example:

```python
import pandas as pd

variance = float(s.var())  # Type error here
```

2. Type checker: `pyright`

3. Error message:
```
Argument of type "Scalar" cannot be assigned to parameter "x" of type "ConvertibleToFloat" in function "__new__"
```

The same issue occurs with other statistical methods: `.mean()`, `.std()`, `.median()`, etc.

 - OS: AlmaLinux 9
 - python version: 3.13.1
 - version of type checker: pyright 1.1.407
 - version of installed `pandas-stubs`: 2.3.3.251201

**Additional context**

The code works correctly at runtime - `Series.var()` does return a numeric value that can be converted to `float`. The issue is that the return type is annotated as `Scalar` in the type stubs, which the type checker doesn't recognize as convertible to `float`. Current workaround is to use `# type: ignore` comments.

roman-y-wu:
- Updated aggregation tests in `test_agg.py` to include checks for the `var()` method across different data types (float, bool, int, and complex).
- Added a warning check for complex series when calculating variance to handle potential loss of imaginary parts.
- Enhanced type assertions in `test_series.py` for the `var()` method to ensure it returns a float type.

- [x] Closes #1545  (Replace xxxx with the Github issue number)

### DIFF
--- a/tests/series/test_agg.py
+++ b/tests/series/test_agg.py
@@ -90,6 +90,7 @@ def test_agg_ts() -> None:
     check(assert_type(series.std(), pd.Timedelta), pd.Timedelta)
 
     if TYPE_CHECKING_INVALID_USAGE:
+
         def _0() -> None:  # pyright: ignore[reportUnusedFunction]
             assert_type(series.var(), Never)
 


pandas-stubs/core/series.pyi
```diff
@@ -4610,14 +4610,33 @@ class Series(IndexOpsMixin[S1], ElementOpsMixin[S1], NDFrame):
...
+    ) -> float: ...
+    @overload
+    def var(
+        self: Series[complex],
+        axis: AxisIndex | None = 0,
+        skipna: _bool | None = True,
+        ddof: int = 1,
+        numeric_only: _bool = False,
+        **kwargs: Any,
+    ) -> np.float64: ...
```
cmp0xff:
```suggestion
    ) -> float: ...
```

tests/series/test_agg.py
```diff
@@ -83,3 +93,4 @@ def test_agg_td() -> None:
     check(assert_type(series.mean(), pd.Timedelta), pd.Timedelta)
     check(assert_type(series.median(), pd.Timedelta), pd.Timedelta)
     check(assert_type(series.std(), pd.Timedelta), pd.Timedelta)
+    # Note: var() is not supported for Timedelta series at runtime
```
cmp0xff: Please use `if TYPE_CHECKING_INVALID_USAGE` and make sure `series.var()` here gives either errors by type checkers (like in `test_agg_str`) or `Never` (`assert_type(series.var(), Never)`).

Error is preferable over `Never`.

tests/series/test_agg.py
cmp0xff, line 90: Please use `if TYPE_CHECKING_INVALID_USAGE` and make sure `series.var()` here gives either errors by type checkers (like in `test_agg_str`) or `Never` (`assert_type(series.var(), Never)`).

Error is preferable over `Never`.

tests/series/test_agg.py
```diff
@@ -57,6 +61,11 @@ def test_agg_complex() -> None:
         ),
     ):
         check(assert_type(series.std(), np.float64), np.float64)
+    with pytest_warns_bounded(
+        np.exceptions.ComplexWarning,
+        r"Casting complex values to real discards the imaginary part",
```
cmp0xff, line 69: please use `upper=2.3.99`

tests/series/test_agg.py
```diff
@@ -83,3 +94,5 @@ def test_agg_td() -> None:
     check(assert_type(series.mean(), pd.Timedelta), pd.Timedelta)
     check(assert_type(series.median(), pd.Timedelta), pd.Timedelta)
     check(assert_type(series.std(), pd.Timedelta), pd.Timedelta)
+    if TYPE_CHECKING_INVALID_USAGE:
+        series.var()  # type: ignore[misc]
```
cmp0xff: `Never` is considered as "valid" by type checkers. However if a function call returns `Never`, type checkers won't check the code after that call.

You can do something like this:

```suggestion
        def _0() -> None:  # pyright: ignore[reportUnusedFunction]
            assert_type(series.var(), Never)
```

tests/series/test_agg.py
cmp0xff: Please also add a check here for `series.var()`

tests/series/test_agg.py
```diff
@@ -75,6 +89,9 @@ def test_agg_ts() -> None:
     check(assert_type(series.median(), pd.Timestamp), pd.Timestamp)
     check(assert_type(series.std(), pd.Timedelta), pd.Timedelta)
 
+    if TYPE_CHECKING_INVALID_USAGE:
+        assert_type(series.var(), Never)
```
cmp0xff:
```suggestion
        def _0() -> None:  # pyright: ignore[reportUnusedFunction]
            assert_type(series.var(), Never)
```
So that in the future when we add further tests below, they will be checked by type checkers. Function calls returning `Never` will stop type checkers from checking the code below.

cmp0xff (changes requested):

tests/series/test_agg.py
loicdiridollou: Yes agreed, we may need to build a wrapper like `check_never` to handle those cases in a more "elegant" way, @Dr-Irv what are your thoughts on that?

Dr-Irv: > Yes agreed, we may need to build a wrapper like `check_never` to handle those cases in a more "elegant" way, @Dr-Irv what are your thoughts on that?

That's a good idea, if you can get it to work.  It would clean up a bunch of code in the arithmetic tests.

PR #1554, opened 2025-12-15
type indexing.pyi

MarcoGorelli:

### DIFF
--- a/tests/frame/test_frame.py
+++ b/tests/frame/test_frame.py
@@ -3959,26 +3959,3 @@ def test_frame_index_setter() -> None:
     check(assert_type(df.index, pd.Index), pd.Index)
     df.index = [2, 3]
     check(assert_type(df.index, pd.Index), pd.Index)
-
-
-def test_frame_at() -> None:
-    df = pd.DataFrame(data={"col1": [1.6, 2], "col2": [3, 4]})
-
-    check(assert_type(df.at[0, "col1"], Scalar), float)
-    df.at[0, "col1"] = 999
-    df.at[0, "col1"] = float("nan")
-
-    mi = pd.MultiIndex.from_arrays([[2, 3], [4, 5]])
-    df = pd.DataFrame(data={"col1": [1.6, 2], "col2": [3, 4]}, index=mi)
-
-    check(assert_type(df.at[(2, 4), "col1"], Scalar), float)
-    df.at[(2, 4), "col1"] = 999
-    df.at[(2, 4), "col1"] = float("nan")
-
-
-def test_frame_iat() -> None:
-    df = pd.DataFrame(data={"col1": [1, 2], "col2": [3, 4]})
-
-    check(assert_type(df.iat[0, 0], Scalar), np.integer)
-    df.iat[0, 0] = 999
-    df.iat[0, 0] = float("nan")
--- a/tests/series/test_indexing.py
+++ b/tests/series/test_indexing.py
@@ -8,6 +8,7 @@ from typing_extensions import assert_type
 
 from tests import (
     PD_LTE_23,
+    TYPE_CHECKING_INVALID_USAGE,
     check,
     pytest_warns_bounded,
 )
@@ -35,6 +36,9 @@ def test_types_iloc_iat() -> None:
     s2.loc[0]
     s2.iat[0]
     s2.iat[0] = None
+
+    if TYPE_CHECKING_INVALID_USAGE:
+        s.iat[0, 0]  # type: ignore[index]  # pyright: ignore[reportArgumentType]
 
 
 def test_types_loc_at() -> None:
--- a/tests/frame/test_indexing.py
+++ b/tests/frame/test_indexing.py
@@ -26,7 +26,10 @@ from typing_extensions import assert_type
 
 from pandas._typing import Scalar
 
-from tests import check
+from tests import (
+    TYPE_CHECKING_INVALID_USAGE,
+    check,
+)
 
 
 def test_types_getitem() -> None:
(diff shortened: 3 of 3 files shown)


pandas-stubs/core/indexing.pyi
```diff
@@ -38,20 +46,20 @@ class IndexingMixin:
...
-
-class _LocationIndexer(_NDFrameIndexer):
-    def __getitem__(self, key): ...
+    axis: AxisInt | None = None
+    def __call__(self, axis: Axis | None = None) -> Self: ...
 
-class _LocIndexer(_LocationIndexer): ...
-class _iLocIndexer(_LocationIndexer): ...
+class _LocIndexer(_NDFrameIndexer): ...
+class _iLocIndexer(_NDFrameIndexer): ...
```
MarcoGorelli, line 43: `_NDFrameIndexer` only subclassed by `_LocationIndexer` which is only subclassed by `_LocIndexer` and `_iLocIndexer`, both of which are subclassed by `_LocIndexerFrame`/ `_LocIndexerSeries` / ..., which define `__getitem__` / `__setitem__`

so, i think it's not necessary to have `_LocationIndexer` at all?

MarcoGorelli: ~correction - it is necessary for IndexingMixin~ `IndexingMixin` can be removed

pandas-stubs/core/frame.pyi
```diff
@@ -323,42 +323,21 @@ class _LocIndexerFrame(_LocIndexer, Generic[_T]):
...
-    def __getitem__(
-        self,
-        key: tuple[
-            int
-            | StrLike
-            | Timestamp
-            | tuple[Scalar, ...]
-            | Callable[[DataFrame], ScalarT],
-            int | StrLike | tuple[Scalar, ...],
-        ],
```
MarcoGorelli, line 343: this was way too wide to begin with, `.at` is for accessing single elements, all that you can pass are keys

pandas-stubs/core/frame.pyi
```diff
@@ -323,42 +323,21 @@ class _LocIndexerFrame(_LocIndexer, Generic[_T]):
...
-            | NAType
-            | NaTType
-            | ArrayLike
-            | IndexOpsMixin
-            | DataFrame
-            | Sequence[Scalar]
-            | Sequence[Sequence[Scalar]]
-            | Mapping[Hashable, Scalar | NAType | NaTType]
-            | None
-        ),
```
MarcoGorelli, line 361: same - all that can be set is a single value

pandas-stubs/core/indexing.pyi
```diff
@@ -27,31 +35,21 @@ class _IndexSlice:
...
 
-class IndexingMixin:
-    @property
-    def iloc(self) -> _iLocIndexer: ...
-    @property
-    def loc(self) -> _LocIndexer: ...
-    @property
-    def at(self) -> _AtIndexer: ...
-    @property
-    def iat(self) -> _iAtIndexer: ...
```
MarcoGorelli, line 38: both `Series` and `DataFrame` explicitly define `loc` `iloc` `at` and `iat`, so this is unneeded and just adds complexity

cmp0xff: I'm not too sure about changing inheritance hierarchies. If we leave an empty `class IndexingMixin` and let `NDFrame` be a subclass of it, will there be any differences? Let me ask @Dr-Irv @loicdiridollou .

tests/frame/test_frame.py
```diff
@@ -3959,3 +3959,26 @@ def test_frame_index_setter() -> None:
     check(assert_type(df.index, pd.Index), pd.Index)
     df.index = [2, 3]
     check(assert_type(df.index, pd.Index), pd.Index)
+
+
+def test_frame_at() -> None:
```
cmp0xff: Could you move the new tests to `frame/test_indexing.py`?

Could you add a few negative tests in `if TYPE_CHECKING_INVALID_USAGE` when users try to access several values instead of one value with `at` and `.iat`?

tests/series/test_indexing.py
```diff
@@ -31,17 +31,21 @@ def test_types_iloc_iat() -> None:
     s2 = pd.Series(data=[1, 2])
     s.loc["row1"]
     s.iat[0]
+    s.iat[0] = 999
```
cmp0xff, line 35: Could you add a few negative tests in `if TYPE_CHECKING_INVALID_USAGE` when users try to access several values instead of one value with `at` and `.iat`?

pandas-stubs/core/indexing.pyi
MarcoGorelli: i don't that having an empty mixin would have any effect

tests/series/test_indexing.py
MarcoGorelli: for `iat`, yes

for `at`, I don't think so, no, as `tuple[Hashable, ...]` is still `Hashable`

personally I think it's a design issue in pandas
```py
In [17]: s = pd.Series([1, 2, 3], index=pd.MultiIndex.from_arrays([[3,2,1], ['a', 'b', 'c']]))

In [18]: s
Out[18]:
3  a    1
2  b    2
1  c    3
dtype: int64

In [19]: s.at[(3, 'a')]
Out[19]: np.int64(1)

In [20]: s.at[3, 'a']
Out[20]: np.int64(1)
```

cmp0xff: In Python, `a[(3, 'a')]` is the same as `a[3, 'a']`.

I also learned that the idea of `Hashable` is flawed, see https://github.com/astral-sh/ty/issues/1846#issuecomment-3638888830.

So maybe the negative part is not testable after all.

pandas-stubs/core/indexing.pyi
Dr-Irv: The way that `.loc`, `.iloc`, etc. are handled in the stubs is by having specific declarations in `Series` and `DataFrame`.  So the "implementation" of those methods in the stubs was pushed down beyond `NDFrame` because we want to be more specific on the results.

In any case, `IndexingMixin` is not documented, so removing it is fine.

cmp0xff (approved):
All good. Thank you @MarcoGorelli !

PR #1544, opened 2025-12-11
TYP: enable `reportMissingTypeArgument`

cmp0xff:
- [x] Towards #1558, #1522 and #1171

PR #1556, opened 2025-12-15
TYP: `GroupBy` and native types

cmp0xff:
Towards #1544 

`GroupBy` and native types do not have default values in their type variables. Adding `Any`.

### DIFF
--- a/tests/test_resampler.py
+++ b/tests/test_resampler.py
@@ -1,8 +1,8 @@
+# pyright: reportMissingTypeArgument=false
 from collections.abc import (
     Hashable,
     Iterator,
 )
-from typing import Any  # noqa: F401
 from typing import TypeAlias
 
 import numpy as np
@@ -142,7 +142,7 @@ def test_asfreq() -> None:
 
 
 def test_getattr() -> None:
-    check(assert_type(DF.resample("ME").col1, "SeriesGroupBy[Any, Any]"), SeriesGroupBy)
+    check(assert_type(DF.resample("ME").col1, SeriesGroupBy), SeriesGroupBy)
 
 
 def test_interpolate() -> None:
@@ -467,10 +467,8 @@ def test_aggregate_frame_combinations() -> None:
 
 
 def test_getitem() -> None:
+    check(assert_type(DF.resample("ME")["col1"], SeriesGroupBy), SeriesGroupBy)
     check(
-        assert_type(DF.resample("ME")["col1"], "SeriesGroupBy[Any, Any]"), SeriesGroupBy
-    )
-    check(
-        assert_type(DF.resample("ME")[["col1", "col2"]], "DataFrameGroupBy[Any, Any]"),
+        assert_type(DF.resample("ME")[["col1", "col2"]], DataFrameGroupBy),
         DataFrameGroupBy,
     )


tests/test_api_typing.py
```diff
@@ -26,27 +30,32 @@
...
+        | TimedeltaIndexResamplerGroupby
+    )
 
 
 def test_dataframegroupby() -> None:
     df = pd.DataFrame({"a": [1, 2, 3]})
     group = df.groupby("a")
 
-    def f1(gb: DataFrameGroupBy) -> None:
+    def f1(gb: "DataFrameGroupBy[Any, Any]") -> None:
```
Dr-Irv: We need to be careful here. The generic forms are not documented, so we really should be testing just `DataFrameGroupBy` here in the tests because that would correspond to user code.

Maybe you can turn off that type checking for the `test` files in the CI.  

This would apply to other changes in this file as well

Dr-Irv (changes requested):

tests/test_api_typing.py
cmp0xff: > because that would correspond to user code.

I've slept over the idea. If users also turn on `reportMissingTypeArgument`, they will be forced to add these `Any`s as well, or they need to `pyright: ignore` manually.

Shall we try to use type variables with default values for `GroupBy`s? I thought we used type variables without default values with a good reason.

> Maybe you can turn off that type checking for the `test` files in the CI.

I checked `pyright` docs but did not find the option to have file-specific ignores.

Dr-Irv: You can do it in the file itself at the top:
```python
# pyright: reportMissingTypeArgument=false
```

I'm fine with doing things in the stubs, but I also think that our test code should test what is documented.  Now we don't always follow that rule (e.g., `Series[int]`), but things like `DataFrameGroupBy` are barely documented, and the goal of that documentation for that and related types is to allow people to declare those types in their code in function arguments.  

So for any types that come from `pandas.api.typing`, we should not test the `Generic` forms that are used internally.

tests/test_resampler.py
```diff
@@ -141,7 +142,7 @@ def test_asfreq() -> None:
 
 
 def test_getattr() -> None:
-    check(assert_type(DF.resample("ME").col1, SeriesGroupBy), SeriesGroupBy)
+    check(assert_type(DF.resample("ME").col1, "SeriesGroupBy[Any, Any]"), SeriesGroupBy)
```
Dr-Irv: remove `[Any, Any]` here

tests/test_resampler.py
```diff
@@ -466,8 +467,10 @@ def df2scalar(val: DataFrame) -> float:
 
 
 def test_getitem() -> None:
-    check(assert_type(DF.resample("ME")["col1"], SeriesGroupBy), SeriesGroupBy)
     check(
-        assert_type(DF.resample("ME")[["col1", "col2"]], DataFrameGroupBy),
+        assert_type(DF.resample("ME")["col1"], "SeriesGroupBy[Any, Any]"), SeriesGroupBy
+    )
+    check(
+        assert_type(DF.resample("ME")[["col1", "col2"]], "DataFrameGroupBy[Any, Any]"),
```
Dr-Irv: remove `[Any, Any]` here

Dr-Irv (changes requested):

tests/test_resampler.py
cmp0xff: 30f1758418f8c000207c4a9c13939aedc8fb6cdd

tests/test_resampler.py
cmp0xff: 30f1758418f8c000207c4a9c13939aedc8fb6cdd

PR #1561, opened 2025-12-17
TYP: enhance `ExcelWriter` with generic type support and overloads

9en9i:
- Closes https://github.com/pandas-dev/pandas-stubs/discussions/1559
- Tests for ExcelWriter fixed

### DIFF
--- a/pandas-stubs/io/excel/_base.pyi
+++ b/pandas-stubs/io/excel/_base.pyi
@@ -223,53 +223,53 @@ _WorkbookT = TypeVar("_WorkbookT", default=ExcelWriteWorkbook, bound=ExcelWriteW
 
 class ExcelWriter(Generic[_WorkbookT]):
     @overload
-    def __init__(
-        self,
-        path: FilePath | BinaryIO,
-        engine: Literal["auto"] | None = ...,
-        date_format: str | None = ...,
-        datetime_format: str | None = ...,
-        mode: Literal["w", "a"] = ...,
-        storage_options: StorageOptions = ...,
-        if_sheet_exists: ExcelWriterIfSheetExists | None = ...,
-        engine_kwargs: dict[str, Any] | None = ...,
-    ) -> None: ...
-    @overload
-    def __init__(
-        self: ExcelWriter[Workbook],
+    def __new__(
+        cls,
         path: FilePath | BinaryIO,
         engine: Literal["openpyxl"],
-        date_format: str | None = ...,
-        datetime_format: str | None = ...,
-        mode: Literal["w", "a"] = ...,
-        storage_options: StorageOptions = ...,
-        if_sheet_exists: ExcelWriterIfSheetExists | None = ...,
-        engine_kwargs: dict[str, Any] | None = ...,
-    ) -> None: ...
+        date_format: str | None = None,
+        datetime_format: str | None = None,
+        mode: Literal["w", "a"] = "w",
+        storage_options: StorageOptions = None,
+        if_sheet_exists: ExcelWriterIfSheetExists | None = None,
+        engine_kwargs: Mapping[str, Any] | None = None,
+    ) -> ExcelWriter[Workbook]: ...
     @overload
-    def __init__(
-        self: ExcelWriter[OpenDocument],
+    def __new__(
+        cls,
         path: FilePath | BinaryIO,
         engine: Literal["odf"],
-        date_format: str | None = ...,
-        datetime_format: str | None = ...,
-        mode: Literal["w", "a"] = ...,
-        storage_options: StorageOptions = ...,
-        if_sheet_exists: ExcelWriterIfSheetExists | None = ...,
-        engine_kwargs: dict[str, Any] | None = ...,
-    ) -> None: ...
+        date_format: str | None = None,
+        datetime_format: str | None = None,
+        mode: Literal["w", "a"] = "w",
+        storage_options: StorageOptions = None,
+        if_sheet_exists: ExcelWriterIfSheetExists | None = None,
+        engine_kwargs: Mapping[str, Any] | None = None,
+    ) -> ExcelWriter[OpenDocument]: ...
     @overload
-    def __init__(
-        self: ExcelWriter[XlsxWorkbook],
+    def __new__(
+        cls,
         path: FilePath | BinaryIO,
         engine: Literal["xlsxwriter"],
-        date_format: str | None = ...,
-        datetime_format: str | None = ...,
-        mode: Literal["w", "a"] = ...,
-        storage_options: StorageOptions = ...,
-        if_sheet_exists: ExcelWriterIfSheetExists | None = ...,
-        engine_kwargs: dict[str, Any] | None = ...,
-    ) -> None: ...
+        date_format: str | None = None,
+        datetime_format: str | None = None,
+        mode: Literal["w", "a"] = "w",
+        storage_options: StorageOptions = None,
+        if_sheet_exists: ExcelWriterIfSheetExists | None = None,
+        engine_kwargs: Mapping[str, Any] | None = None,
+    ) -> ExcelWriter[XlsxWorkbook]: ...
+    @overload
+    def __new__(
+        cls,
+        path: FilePath | BinaryIO,
+        engine: Literal["auto"] | None = None,
+        date_format: str | None = None,
+        datetime_format: str | None = None,
+        mode: Literal["w", "a"] = "w",
+        storage_options: StorageOptions = None,
+        if_sheet_exists: ExcelWriterIfSheetExists | None = None,
+        engine_kwargs: Mapping[str, Any] | None = None,
+    ) -> ExcelWriter[ExcelWriteWorkbook]: ...
     @property
     def supported_extensions(self) -> tuple[str, ...]: ...
     @property


pandas-stubs/io/excel/_base.pyi
```diff
@@ -209,11 +217,52 @@ def read_excel(
...
-class ExcelWriter:
+ExcelWriteWorkbook: TypeAlias = Workbook | OpenDocument | XlsxWorkbook
+
+_WorkbookT = TypeVar("_WorkbookT", default=ExcelWriteWorkbook, bound=ExcelWriteWorkbook)
+
+class ExcelWriter(Generic[_WorkbookT]):
+    @overload
     def __init__(
-        self,
+        self: ExcelWriter,
```
cmp0xff: I think
```suggestion
        self,
```
will do

pandas-stubs/io/excel/_base.pyi
```diff
@@ -209,11 +217,52 @@ def read_excel(
...
         path: FilePath | WriteExcelBuffer | ExcelWriter,
```
cmp0xff: From the [docs](https://pandas.pydata.org/docs/dev/reference/api/pandas.ExcelWriter.html) only `FilePath` and `BinaryIO` are supported
```suggestion
        path: FilePath | BinaryIO,
```
Also to other overloads

pandas-stubs/io/excel/_base.pyi
```diff
@@ -209,11 +217,52 @@ def read_excel(
...
-        engine: ExcelWriteEngine | Literal["auto"] | None = ...,
+        engine: Literal["auto"] | None = ...,
+        date_format: str | None = ...,
+        datetime_format: str | None = ...,
+        mode: Literal["w", "a"] = ...,
+        storage_options: StorageOptions = ...,
+        if_sheet_exists: ExcelWriterIfSheetExists | None = ...,
+        engine_kwargs: dict[str, Any] | None = ...,
```
cmp0xff: Please add default values
```suggestion
        engine: Literal["auto"] | None = None,
        date_format: str | None = None,
        datetime_format: str | None = None,
        mode: Literal["w", "a"] = "w",
        storage_options: StorageOptions = None,
        if_sheet_exists: ExcelWriterIfSheetExists = "error",
        engine_kwargs: Mapping[str, Any] | None = None,
```
Also to other overloads

pandas-stubs/io/excel/_base.pyi
```diff
@@ -209,11 +217,52 @@ def read_excel(
...
+    ) -> None: ...
+    @overload
+    def __init__(
+        self: ExcelWriter[Workbook],
+        path: FilePath | WriteExcelBuffer | ExcelWriter[Workbook],
+        engine: Literal["openpyxl"] = ...,
```
cmp0xff:
```suggestion
        engine: Literal["openpyxl"],
```
it has to be the fixed literal `openpyxl`, if I'm not mistaken

pandas-stubs/io/excel/_base.pyi
```diff
@@ -209,11 +217,52 @@ def read_excel(
...
+        mode: Literal["w", "a"] = ...,
+        storage_options: StorageOptions = ...,
+        if_sheet_exists: ExcelWriterIfSheetExists | None = ...,
+        engine_kwargs: dict[str, Any] | None = ...,
+    ) -> None: ...
+    @overload
+    def __init__(
+        self: ExcelWriter[OpenDocument],
+        path: FilePath | WriteExcelBuffer | ExcelWriter[OpenDocument],
+        engine: Literal["odf"] = ...,
```
cmp0xff:
```suggestion
        engine: Literal["odf"],
```

pandas-stubs/io/excel/_base.pyi
```diff
@@ -209,11 +217,52 @@ def read_excel(
...
+        mode: Literal["w", "a"] = ...,
+        storage_options: StorageOptions = ...,
+        if_sheet_exists: ExcelWriterIfSheetExists | None = ...,
+        engine_kwargs: dict[str, Any] | None = ...,
+    ) -> None: ...
+    @overload
+    def __init__(
+        self: ExcelWriter[XlsxWorkbook],
+        path: FilePath | WriteExcelBuffer | ExcelWriter[XlsxWorkbook],
+        engine: Literal["xlsxwriter"] = ...,
```
cmp0xff:
```suggestion
        engine: Literal["xlsxwriter"],
```

pandas-stubs/io/excel/_base.pyi
```diff
@@ -209,11 +217,52 @@ def read_excel(
...
 ) -> DataFrame: ...
 
-class ExcelWriter:
+ExcelWriteWorkbook: TypeAlias = Workbook | OpenDocument | XlsxWorkbook
+
+_WorkbookT = TypeVar("_WorkbookT", default=ExcelWriteWorkbook, bound=ExcelWriteWorkbook)
+
+class ExcelWriter(Generic[_WorkbookT]):
+    @overload
     def __init__(
```
cmp0xff: I am confused, maybe we should overload `__new__` instead of `__init__` here? cc @Dr-Irv @loicdiridollou

pandas-stubs/io/excel/_base.pyi
9en9i: I don't really understand where these default values came from. For example, here it's just None
https://github.com/pandas-dev/pandas/blob/main/pandas/io/excel/_base.py#L1159C9-L1159C65

And here it says dict
https://github.com/pandas-dev/pandas/blob/main/pandas/io/excel/_base.py#L1160

pandas-stubs/io/excel/_base.pyi
loicdiridollou: I think using `__new__` instead of `__init__` is when it will return a different type (like for Series we use that trick there), if it is only a matter of "these arguments can only be used in such combination" I believe `__init__` overloads should be sufficient.

pandas-stubs/io/excel/_base.pyi
loicdiridollou: For the second one, a `Mapping[str, Any]` is an abstract class to type elements like dict, OrderedDict, defaultdict and not just `dict`. `dict` is also the constructor (ie the concrete class) to define an object.
For `if_sheet_exists` argument, I think we should keep it without a default value since it would be a value passed by the user and not the default which is `None` as you said @9en9i , @cmp0xff what did you want to do with that `error` default value?

pandas-stubs/io/excel/_base.pyi
Dr-Irv: I think `__new__()` is needed here because the goal is to return a different type based on the value of `engine` .  So the overloads should be switched to `__new__()`, and where there is currently `self: ExcelWriter[something]`, the `__new__()` would return `ExcelWriter[something]`, where `something` relates to the value of `engine`

9en9i: So, if I understand correctly, the code will look like this? (the names of the overloads will change to `__new__` and the constructor will return)
```python
class ExcelWriter(Generic[_WorkbookT]):
    @overload
    def __new__(
        cls,
        path: FilePath | BinaryIO,
        engine: Literal["openpyxl"],
        date_format: str | None = None,
        datetime_format: str | None = None,
        mode: Literal["w", "a"] = "w",
        storage_options: StorageOptions = None,
        if_sheet_exists: ExcelWriterIfSheetExists = "error",
        engine_kwargs: Mapping[str, Any] | None = None,
    ) -> ExcelWriter[Workbook]: ...
    @overload
    def __new__(
        cls,
        path: FilePath | BinaryIO,
        engine: Literal["odf"],
        date_format: str | None = None,
        datetime_format: str | None = None,
        mode: Literal["w", "a"] = "w",
        storage_options: StorageOptions = None,
        if_sheet_exists: ExcelWriterIfSheetExists = "error",
        engine_kwargs: Mapping[str, Any] | None = None,
    ) -> ExcelWriter[OpenDocument]: ...
    @overload
    def __new__(
        cls,
        path: FilePath | BinaryIO,
        engine: Literal["xlsxwriter"],
        date_format: str | None = None,
        datetime_format: str | None = None,
        mode: Literal["w", "a"] = "w",
        storage_options: StorageOptions = None,
        if_sheet_exists: ExcelWriterIfSheetExists = "error",
        engine_kwargs: Mapping[str, Any] | None = None,
    ) -> ExcelWriter[XlsxWorkbook]: ...
    @overload
    def __new__(
        cls,
        path: FilePath | BinaryIO,
        engine: Literal["auto"] | None = ...,
        date_format: str | None = None,
        datetime_format: str | None = None,
        mode: Literal["w", "a"] = "w",
        storage_options: StorageOptions = None,
        if_sheet_exists: ExcelWriterIfSheetExists = "error",
        engine_kwargs: Mapping[str, Any] | None = None,
    ) -> ExcelWriter[ExcelWriteWorkbook]: ...
    def __init__(
        self,
        path: FilePath | BinaryIO,
        engine: ExcelWriteEngine | Literal["auto"] | None = None,
        date_format: str | None = None,
        datetime_format: str | None = None,
        mode: Literal["w", "a"] = "w",
        storage_options: StorageOptions = None,
        if_sheet_exists: ExcelWriterIfSheetExists = "error",
        engine_kwargs: Mapping[str, Any] | None = None,
    ) -> None: ...
```

pandas-stubs/io/excel/_base.pyi
cmp0xff: - `Mapping` is more generic than `dict`. From the [docs](https://pandas.pydata.org/docs/dev/reference/api/pandas.ExcelWriter.html), `engine_kwargs` are just sent to the engine with `**`, so `Mapping` should be fine. If you are unsure, you can just use `dict`.
- As for `if_sheet_exists`, in the docs it says both `None` and `"error"`. Let's follow the source code and use `None`.

pandas-stubs/io/excel/_base.pyi
cmp0xff: - I think the `__new__` version is correct
- Please remove `__init__`
- Please use `None` as the default value for `if_sheet_exists`

cmp0xff (changes requested):

pandas-stubs/io/excel/_base.pyi
9en9i: Done

pandas-stubs/io/excel/_base.pyi
9en9i: Done

pandas-stubs/io/formats/style.pyi
```diff
@@ -103,7 +107,9 @@ class Styler(StylerRenderer):
     ) -> Styler: ...
     def to_excel(
         self,
-        excel_writer: FilePath | WriteExcelBuffer | ExcelWriter,
+        excel_writer: (
+            FilePath | WriteExcelBuffer | ExcelWriter[OpenXlWorkbook | XlsxWorkbook]
```
cmp0xff, line 111: Why just `OpenXlWorkbook | XlsxWorkbook` and not `OpenDocument`?

9en9i: Because the argument below has the annotation (https://github.com/pandas-dev/pandas-stubs/blob/main/pandas-stubs/io/formats/style.pyi#L116)
```python
engine: Literal["openpyxl", "xlsxwriter"] | None = None,
```
I proceeded from the logic that if `odf` has no meaning, then it does not work with this class.

cmp0xff: Indeed, `odf` does not work at runtime, either.

PR #1562, opened 2025-12-17
Type `generic.pyi`

MarcoGorelli:


pandas-stubs/core/generic.pyi
```diff
@@ -312,115 +318,113 @@ class NDFrame:
         errors: OpenFileErrors = ...,
         storage_options: StorageOptions = ...,
     ) -> _str: ...
-    @final
-    def __delitem__(self, idx: Hashable) -> None: ...
     @overload
     def drop(
         self,
-        labels=...,
+        labels: Hashable | Sequence[Hashable] | Index | None = None,
```
loicdiridollou: I think we should be able to put `range` here too, docs says `single label or iterable of labels` so range should fit, maybe `ListLike` would be a slightly wider type.

MarcoGorelli: good call, thanks!

pandas-stubs/core/series.pyi
```diff
@@ -562,15 +563,9 @@ class Series(IndexOpsMixin[S1], ElementOpsMixin[S1], NDFrame):
...
     ) -> Any: ...
-    if sys.version_info >= (3, 11):
-        def __array__(
-            self, dtype: _str | np.dtype = ..., copy: bool | None = ...
-        ) -> np_1darray: ...
-    else:
-        def __array__(
-            self, dtype: _str | np.dtype[Any] = ..., copy: bool | None = ...
-        ) -> np_1darray: ...
-
```
cmp0xff, line 573: Please keep it, it is needed for the pyright strict transition.

MarcoGorelli: is it ok to handle this within `NpDType`?

cmp0xff: `NpDtype` is different from `str | np.dtype`. It will take a separate PR to handle, I think.

pandas-stubs/core/generic.pyi
```diff
@@ -312,115 +319,113 @@ class NDFrame:
         errors: OpenFileErrors = ...,
         storage_options: StorageOptions = ...,
     ) -> _str: ...
-    @final
-    def __delitem__(self, idx: Hashable) -> None: ...
     @overload
     def drop(
         self,
-        labels=...,
+        labels: Hashable | ListLike | None = None,
```
cmp0xff:
```suggestion
        labels: Hashable | ListLike = None,
```
`None` ia also `Hashable`. Also applies to other cases in the changes.

cmp0xff (commented):
I will leave the PR to @loicdiridollou since he has alread started reviewing. I just wanted to point out the `np.dtype` issue.

Issue #1548, opened 2025-12-14
DataFrame.index not reliably typed using Index[Any]


adamjstewart:
**Describe the bug**

`DataFrame.index` is currently typed as `Index[Any]`. As a result, certain statically typed valid code is flagged as invalid by mypy. I would argue that `Any` would be better until the index can be reliably typed.

**To Reproduce**

For the following code:
```python
import pandas as pd

data = {'col1': [1, 2], 'col2': [3, 4]}
index = pd.interval_range(start=0, end=2)
df = pd.DataFrame(data, index)
print(df.index.left)
print(df.index.right)
```
mypy complains with:
```
test.py:6: error: "Index[Any]" has no attribute "left"  [attr-defined]
test.py:7: error: "Index[Any]" has no attribute "right"  [attr-defined]
Found 2 errors in 1 file (checked 1 source file)
```
This is because I am using an `IntervalIndex`, not a generic `Index`.

**Please complete the following information:**
 - OS: macOS
 - OS Version: 26.1
 - python version: 3.13.8
 - version of type checker: 1.19.0
 - version of installed `pandas-stubs`: 2.3.2.250926

**Additional context**

The issue of DataFrame.index type hints was also raised in https://github.com/pandas-dev/pandas/issues/36708. This issue also impacts geopandas: https://github.com/python/typeshed/issues/15111.

adamjstewart:
Completely agree with making DataFrame a generic. I'm just wondering if index should be Any while we wait for it to become a generic.

cmp0xff:
In the example you provided, we can do as follows:

```py
from typing import cast, reveal_type

import pandas as pd

data = {'col1': [1, 2], 'col2': [3, 4]}
df = pd.DataFrame(data, pd.interval_range(start=0, end=2))

index = cast("pd.IntervalIndex[pd.Interval[int]]", df.index)
reveal_type(index.left)  # pyright gives Index[Any]
reveal_type(index.right)  # pyright gives Index[Any]
```

Neither `mypy` nor `pyright` complain any more. However, obtaining `Index[Any]` is not ideal, which is another topic.

adamjstewart:
We could also modify `Index[Any]` such that all unknown attributes (e.g., left, right) are `Any`.

Dr-Irv:
> We could also modify `Index[Any]` such that all unknown attributes (e.g., left, right) are `Any`.

[...]

PR #1566, opened 2025-12-18
ENH: Make `DataFrame` generic

cmp0xff:
Towards #1548

- I don't know how to make `setter` work so that the typings can be updated when one assigns `index` or `columns`.
- The two overlapping overloads of `__getitem__` in `_LocIndexerFrame` need to be fixed

Comments and suggestions are welcomed.

### DIFF
--- a/pandas-stubs/_typing.pyi
+++ b/pandas-stubs/_typing.pyi
@@ -977,6 +977,10 @@ ListLikeHashable: TypeAlias = (
     MutableSequence[HashableT] | np_1darray | tuple[HashableT, ...] | range
 )
 
+if TYPE_CHECKING:  # noqa: PYI002
+    IndexT0 = TypeVar("IndexT0", bound=Index, default=Index)
+    IndexStrT0 = TypeVar("IndexStrT0", bound=Index, default=Index[str])
+
 class SupportsDType(Protocol[GenericT_co]):
     @property
     def dtype(self) -> np.dtype[GenericT_co]: ...
--- a/pandas-stubs/core/frame.pyi
+++ b/pandas-stubs/core/frame.pyi
@@ -116,6 +116,8 @@ from pandas._typing import (
     IndexingInt,
     IndexKeyFunc,
     IndexLabel,
+    IndexStrT0,
+    IndexT0,
     IndexType,
     InterpolateOptions,
     IntervalClosedType,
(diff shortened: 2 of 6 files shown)


pandas-stubs/core/frame.pyi
```diff
@@ -373,10 +380,19 @@ _AstypeArgExt: TypeAlias = (
...
 
-class DataFrame(NDFrame, OpsMixin, _GetItemHack):
+class DataFrame(NDFrame, OpsMixin, _GetItemHack, Generic[IndexT0, IndexStrT0]):
 
     __hash__: ClassVar[None]  # type: ignore[assignment] # pyright: ignore[reportIncompatibleMethodOverride]
 
+    @overload
+    def __new__(
+        cls,
+        data: DataFrame[IndexT0, IndexStrT0],
```
loicdiridollou, line 354: Not totally sure if this is gonna break a lot of things, by default DataFrame will have `RangeIndex` as index and columns, isn't this gonna change that?

pandas-stubs/_typing.pyi
```diff
@@ -975,6 +972,10 @@ ListLikeHashable: TypeAlias = (
     MutableSequence[HashableT] | np_1darray | tuple[HashableT, ...] | range
 )
 
+if TYPE_CHECKING:  # noqa: PYI002
+    IndexT0 = TypeVar("IndexT0", bound=Index, default=Index)
+    IndexStrT0 = TypeVar("IndexStrT0", bound=Index, default=Index[str])
```
loicdiridollou, line 982: Shouldn't the default be RangeIndex here?

pandas-stubs/core/frame.pyi
cmp0xff: This is copying another DataFrame, I suppose, so nothing is changed. Will add tests later.

pandas-stubs/_typing.pyi
cmp0xff: For `df.columns` we somehow have a preference for it being `Index[str]`, see `def columns(self) -> Index[str]: ...` in the old code. That's the reason.

pandas-stubs/core/frame.pyi
cmp0xff: If you create a `DataFrame` from data without an index, `RangeIndex` will surely take over. That takes further overloads of `__new__`.

As for now I just want to show a prototype of what can happen.

Dr-Irv:
> I am not sure I see the whole idea about this PR

I'm in agreement.  If this were to work, what would be the benefit?

Note that `DatatFrame.columns` returns `Index[str]` since most people use strings to name their columns. Technically, they could be any `Hashable`, or even a `MultiIndex`  However, if we made `columns` be `Index[Any]` then for the majority of people, they'd have to do `cast` to get the actual string names.

cmp0xff:
## Benefits

Aside from paving the way to resolving #1548, we have the following issue:

```py
import pandas as pd
from typing_extensions import reveal_type

df = pd.DataFrame(
    {"a": [1, 2, 1, 2], "b": [1, 1, 2, 2], "c": [1, 2, 3, 4], "d": [5, 6, 7, 8]},
)

pivoted = df.pivot_table(["c", "d"], "a", "b")

reveal_type(pivoted["c"])  # pyright gives Series[Any], runtime is DataFrame
reveal_type(pivoted.columns)  # pyright gives Index[str], runtime is MultiIndex
```

## Default typing for `.columns`

I agree it should be `Index[str]`

## Default typing for `.index`

I think `RangeIndex` might be too radical. Keeping `Index[Any]` means keeping the current behaviour.

We can give `RangeIndex` in the following (but not exclusively) cases:
- when creating a `DataFrame` from unindex objects without specifying `index`
- when `ignore_index=True`
- when `.reset_index()`

PR #1567, opened 2025-12-18
type `*core/base.pyi`, remove SelectionMixin

MarcoGorelli:
`SelectionMixin` is undocumented, and its methods are:
- `obj`: undocumented
- `exclusions`: undocumented
- `ndim`: undocumented
- `__getitem__`: overwritten by subclasses `BaseGroupBy` and `BaseWindow`
- `aggregate`: overwritten by subclasses `DataFrameGroupBy`, `SeriesGroupBy`, and `BaseWindow`

### DIFF
--- a/tests/test_resampler.py
+++ b/tests/test_resampler.py
@@ -199,7 +199,7 @@ def test_pipe() -> None:
         kw: tuple[int],
     ) -> DataFrame:
         assert isinstance(res, DatetimeIndexResampler)
-        return res.obj  # type: ignore[return-value]  # pyright: ignore[reportReturnType]
+        return DataFrame({"a": [1, 2, 3]})
 
     check(
         assert_type(DF.resample("ME").pipe(j, 1, [1.0], arg2="hi", kw=(1,)), DataFrame),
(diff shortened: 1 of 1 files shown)


tests/test_resampler.py
```diff
@@ -203,7 +199,7 @@ def j(
         kw: tuple[int],
     ) -> DataFrame:
         assert isinstance(res, DatetimeIndexResampler)
-        return res.obj
+        return res.obj  # type: ignore[return-value]  # pyright: ignore[reportReturnType]
```
loicdiridollou: Since this is not documented is there a good reason to keep a function like this? The user should not in theory access the field anyway.

tests/test_resampler.py
```diff
@@ -263,7 +259,7 @@ def j(
 
     def k(x: int, t: "DatetimeIndexResampler[DataFrame]") -> DataFrame:
         assert isinstance(x, int)
-        return t.obj
+        return t.obj  # type: ignore[return-value] # pyright: ignore[reportReturnType]
```
loicdiridollou: Same here, since not documented is there a good reason to keep it? Or could we replace it with a function that use "allowed" fields to keep the same type coverage.

MarcoGorelli: sure, it's just used here to test `pipe` so might as well not use `obj` 👍

loicdiridollou: Okay so we can keep it for now

PR #1575, opened 2025-12-22
clean: remove unneded mixin classes

MarcoGorelli:
these classes aren't needed the methods they like `_add_arithmetic_ops` are just runtime tricks which serve no static typing purpose

### DIFF
--- a/tests/extension/decimal/array.py
+++ b/tests/extension/decimal/array.py
@@ -352,5 +352,12 @@ class DecimalArray(OpsMixin, ExtensionArray):
 
         return value_counts(self.to_numpy(), dropna=dropna)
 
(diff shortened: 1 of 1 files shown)


tests/extension/decimal/array.py
```diff
@@ -354,6 +351,3 @@ def value_counts(self, dropna: bool = True) -> Series:
         )
 
         return value_counts(self.to_numpy(), dropna=dropna)
-
-
-DecimalArray._add_arithmetic_ops()
```
cmp0xff, line 359: I think this can be functional, please keep

MarcoGorelli: suer thing

PR #1578, opened 2025-12-22
TYP: naming and default values

cmp0xff:
Towards #1544.

- Add default values to type variables
- Add type variables with a default value
- Rename to `_CrossTabValues`


pandas-stubs/io/json/_json.pyi
```diff
@@ -226,7 +227,7 @@ def read_json(
     engine: Literal["pyarrow"],
 ) -> DataFrame: ...
 
-class JsonReader(abc.Iterator, Generic[NDFrameT]):
+class JsonReader(abc.Iterator[Any], Generic[NDFrameT]):
```
loicdiridollou: Can you collapse `abc.Iterator` and `from collections.abc import Mapping`

pandas-stubs/io/stata.pyi
loicdiridollou: Collapse the two imports here if you can

pandas-stubs/io/parsers/readers.pyi
```diff
@@ -452,7 +452,7 @@ def read_fwf(
     **kwds: Any,
 ) -> DataFrame: ...
 
-class TextFileReader(abc.Iterator):
+class TextFileReader(abc.Iterator[Any]):
```
loicdiridollou: Same here if possible collapse `abc.Iterator` into `from collections.abc` since both are imported

loicdiridollou (changes requested):
Minor cosmetic requests but looks good otherwise!

PR #1581, opened 2025-12-23
MNT: use only `pyrefly: ignore` for `pyrefly`

cmp0xff:
`# type: ignore` used to also control `pyrefly` as well. In this PR I am making `pyrefly` respect `pyrefly: ignore` only.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -327,6 +327,18 @@ project-includes = ["pandas-stubs", "tests"]
 enabled-ignores = ["pyrefly"]
 python-version = "3.10"
 
+[tool.pyrefly.errors]
+implicit-abstract-class = true
+implicitly-defined-attribute = true
+open-unpacking = true
+unannotated-attribute = true
+# TODO: enable
+# implicit-any = true
+# missing-source = true
+# missing-override-decorator = true
+# unannotated-parameter = true
+# unannotated-return = true
+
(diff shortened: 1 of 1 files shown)


loicdiridollou:
On this one I am wondering how much `pyrefly` is used across projects our users are working on, I have discussed around and not heard a lot of `pyrefly` especially with `ty` getting a lot of traction (although it also has its caveats).
Prob a discussion for the future but I am wondering if having four type checkers is not creating too much maintenance burden (cc @MarcoGorelli @Dr-Irv), open to some feedback and views on the issue.

loicdiridollou (approved):
Great work on this one, seems quite painful with all the occurences.

Dr-Irv:
> On this one I am wondering how much `pyrefly` is used across projects our users are working on, I have discussed around and not heard a lot of `pyrefly` especially with `ty` getting a lot of traction (although it also has its caveats). Prob a discussion for the future but I am wondering if having four type checkers is not creating too much maintenance burden (cc @MarcoGorelli @Dr-Irv), open to some feedback and views on the issue.

The daily download count shown here: https://pypistats.org/packages/pyrefly does indicate a small, but growing user base. 

For now, I think we continue to support it and watch how things change over time.  If we remove the support now, we may have to put it back later...

PR #1591, opened 2025-12-25
TYP: Fix np.double dtype inference for numpy 2.4.0

### LINKED ISSUE #1590: numpy 2.4.0 compatibility: np.double dtype inference broken
- With numpy 2.4.0, `pd.Series([1.0], dtype=np.double)` incorrectly infers `Series[float64]` instead of `Series[float]`
- Caused by numpy 2.4.0 changing `double` from TypeAlias to concrete class reference

ritsuki1227:
- [x] Closes #1590 
- [x] Tests added (existing test at `tests/series/test_series_float.py:81`)

# PR: Fix type inference for `np.double` dtype with numpy 2.4.0

## Problem

With numpy 2.4.0, `pd.Series([1.0], dtype=np.double)` infers `Series[float64]` instead of `Series[float]`.

## Cause

numpy 2.4.0 changed `double` from abstract type alias to concrete class reference:
```python
# numpy 2.3.5
double: TypeAlias = floating[_NBitDouble]

# numpy 2.4.0
double = float64  # float64 is a concrete class
```

This causes the generic `dtype: type[S1]` overload to match before `dtype: FloatDtypeArg`.

## Solution

Reorder `Series.__new__` overloads: move `dtype: type[S1]` after `dtype: FloatDtypeArg`.

This matches the existing order in `Index.__new__` which doesn't have this issue.

## Testing

Passed on both numpy 2.3.5 and 2.4.0.

### DIFF
--- a/pandas-stubs/core/series.pyi
+++ b/pandas-stubs/core/series.pyi
@@ -454,16 +454,6 @@ class Series(IndexOpsMixin[S1], ElementOpsMixin[S1], NDFrame):
         copy: bool | None = None,
     ) -> Series[Interval[_OrderableT]]: ...
     @overload
-    def __new__(  # type: ignore[overload-overlap] # pyright: ignore[reportOverlappingOverload]
-        cls,
-        data: Scalar | _DataLike | dict[HashableT1, Any] | None,
-        index: AxesData | None = None,
-        *,
-        dtype: type[S1],
-        name: Hashable = None,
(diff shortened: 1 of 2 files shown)


ritsuki1227:
CI is passing with both numpy 2.2.6 (Python 3.10) and numpy 2.4.0 (Python 3.13/3.14).

PR #1592, opened 2025-12-25
MNT: ignore `untyped-import` in `pyrefly`

cmp0xff:
`pyrefly` has introduced a rule `untyped-import`. Since we are ignoring similar rules in `mypy` (`ignore_missing_imports = true`, `follow_imports_for_stubs = false`), it makes sense to also ignore it in `pyrefly`.


loicdiridollou (approved):
Nice work, thanks @cmp0xff

Issue #1580, opened 2025-12-23
2.1 compatibility tracker


cmp0xff:
This is the list of things that are in [pandas 2.1 release notes](https://pandas.pydata.org/docs/dev/whatsnew/v2.1.0.html) that need to be addressed in pandas-stubs. I have removed the sections `Performance improvements` and `Bug fixes`.

PR's welcome. If you do a PR, check off the item and put a link to the PR that closed it. One PR can address multiple issues.

Some of these may already have been taken care of, so if so, check them off and indicate with a comment such as "previously complete"

## Enhancements

- [ ] PyArrow will become a required dependency with pandas 3.0
- [ ] Avoid NumPy object dtype for strings by default
- [ ] DataFrame reductions preserve extension dtypes
- [x] Copy-on-Write improvements
- [ ] New `DataFrame.map` method and support for ExtensionArrays
- [x] New implementation of `DataFrame.stack`
- [ ] `Series.ffill` and `Series.bfill` are now supported for objects with `IntervalDtype` (pandas-dev/pandas#54247)
- [ ] Added ``filters`` parameter to :func:`read_parquet` to filter out data, compatible with both ``engines`` (pandas-dev/pandas#53212)
- [x] `.Categorical.map` and `CategoricalIndex.map` now have a ``na_action`` parameter.
  `.Categorical.map` implicitly had a default value of ``"ignore"`` for ``na_action``. This has formally been deprecated and will be changed to ``None`` in the future.
  Also notice that `Series.map` has default ``na_action=None`` and calls to series with categorical data will now use ``na_action=None`` unless explicitly set otherwise (pandas-dev/pandas#44279)
- [ ] `api.extensions.ExtensionArray` now has a `~api.extensions.ExtensionArray.map` method (pandas-dev/pandas#51809)
- [ ] `DataFrame.applymap` now uses the `~api.extensions.ExtensionArray.map` method of underlying `api.extensions.ExtensionArray` instances (pandas-dev/pandas#52219)
- [x] `MultiIndex.sort_values` now supports ``na_position`` (pandas-dev/pandas#51612)
- [x] `MultiIndex.sortlevel` and `Index.sortlevel` gained a new keyword ``na_position`` (pandas-dev/pandas#51612)
- [ ] `arrays.DatetimeArray.map`, `arrays.TimedeltaArray.map` and `arrays.PeriodArray.map` can now take a ``na_action`` argument (pandas-dev/pandas#51644)
[...]

PR #1594, opened 2025-12-25
TST: `xStringArray` and `NumpyExtensionArray`

cmp0xff:
Towards #1589, #1469, #1580

A `NumpyExtensionArray` is produced
- when no other Pandas array applies
- when a Python-native or a numpy dtype is imposed (provided that they are not datetime like)


tests/__init__.py
```diff
@@ -63,6 +183,10 @@
     # numpy float128
     np.longdouble: np.longdouble,
     "g": np.longdouble,
+    "longdouble": np.longdouble,
+    "f16": np.longdouble,  # NOTE: UNIX ONLY
+    # "float96": np.longdouble,  # NOTE: unsupported
+    "float128": np.longdouble,  # NOTE: UNIX ONLY
```
cmp0xff: I thought they would only work when `pd.Series([1]).astype(...)`, and I thought `pd.Series([1], dtype="longdouble")` wouldn't work. I was wrong. Adding them back.

tests/__init__.py
```diff
@@ -39,7 +39,127 @@
 PD_LTE_23 = Version(pd.__version__) < Version("2.3.999")
 NUMPY20 = np.lib.NumpyVersion(np.__version__) >= "2.0.0"
 
-NATIVE_FLOAT_ARGS = {float: np.floating, "float": np.floating}
+PYTHON_BOOL_ARGS = {bool: np.bool_, "bool": np.bool_}
```
cmp0xff: Python native types and Numpy types are relevant for constructing a `NumpyExtensionArray`.

tests/__init__.py
```diff
@@ -39,7 +39,127 @@
...
+PANDAS_BOOL_ARGS = {pd.BooleanDtype(): np.bool_, "boolean": np.bool_}
```
cmp0xff: Pandas types and PyArrow types are irrelevant to `NumpyExtensionArray`.

tests/series/test_series_float.py
```diff
@@ -46,7 +45,7 @@ def test_constructor() -> None:
     )
 
 
-@pytest.mark.parametrize(("dtype", "target_dtype"), TYPE_FLOAT_ARGS.items())
+@pytest.mark.parametrize(("dtype", "target_dtype"), ASTYPE_FLOAT_ARGS.items())
```
cmp0xff, line 45: I thought they would only work when `pd.Series([1]).astype(...)`, and I thought `pd.Series([1], dtype="longdouble")` wouldn't work. I was wrong. Adding them back.

tests/series/test_series.py
```diff
@@ -2950,23 +2759,23 @@ def test_all_astype_args_tested() -> None:
...
-        + ASTYPE_STRING_ARGS
-        + ASTYPE_CATEGORICAL_ARGS
-        + ASTYPE_OBJECT_ARGS
-        + ASTYPE_VOID_ARGS
+        | ASTYPE_INT_ARGS
+        | ASTYPE_UINT_ARGS
+        | ASTYPE_FLOAT_ARGS
+        | ASTYPE_COMPLEX_ARGS
+        | dict(ASTYPE_TIMEDELTA_ARGS)
+        | dict(ASTYPE_TIMESTAMP_ARGS)
```
cmp0xff: Timedelta, Timestamp and Categorical are irrelevant to `NumpyExtensionArray`.

tests/indexes/test_index_float.py
```diff
@@ -49,7 +48,9 @@ def test_constructor() -> None:
     )
 
 
-@pytest.mark.parametrize(("dtype", "target_dtype"), TYPE_FLOAT_NOT_NUMPY16_ARGS.items())
+@pytest.mark.parametrize(
+    ("dtype", "target_dtype"), ASTYPE_FLOAT_NOT_NUMPY16_ARGS.items()
+)
```
cmp0xff, line 52: I thought they would only work when `pd.Index([1]).astype(...)`, and I thought `pd.Index([1], dtype="longdouble")` wouldn't work. I was wrong. Adding them back.

pandas-stubs/_typing.pyi
```diff
@@ -538,8 +538,7 @@ CategoryDtypeArg: TypeAlias = CategoricalDtype | Literal["category"]
 BuiltinObjectDtypeArg: TypeAlias = type[object] | Literal["object"]
 # Numpy object type and its string alias
 # https://numpy.org/doc/stable/reference/arrays.scalars.html#numpy.object_
-# NOTE: "object_" not assigned
-NumpyObjectDtypeArg: TypeAlias = type[np.object_] | Literal["O"]
+NumpyObjectDtypeArg: TypeAlias = type[np.object_] | Literal["object_", "O"]
```
cmp0xff, line 538: `"object_"` works at least for `NumpyExtensionArray`. Adding back.

pandas-stubs/_typing.pyi
```diff
@@ -373,7 +373,7 @@ NumpyComplexDtypeArg: TypeAlias = (
     #  https://numpy.org/doc/stable/reference/arrays.scalars.html#numpy.clongdouble
     # NOTE: Alias np.complex256 only on Linux x86_64, use np.clongdouble instead
     | type[np.clongdouble]
-    | Literal["G", "clongdouble"]
+    | Literal["G", "c32", "complex256", "clongdouble"]
```
cmp0xff, line 363: `"c32"` and `"complex256"` work at least for `NumpyExtensionArray`. Adding back.

pandas-stubs/core/arrays/base.pyi
```diff
@@ -125,9 +124,6 @@ class ExtensionArray:
         skipna: bool = True,
         **kwargs: Any,
     ) -> Self: ...
-    def value_counts(
-        self, dropna: bool = True
-    ) -> Series[int]: ...  # probably to put in base class
```
cmp0xff: In `NDArrayBackedExtensionArray` instead of here.

pandas-stubs/core/arrays/base.pyi
```diff
@@ -55,7 +54,7 @@ class ExtensionArray:
     @property
     def dtype(self) -> ExtensionDtype: ...
     @property
-    def shape(self) -> tuple[int, ...]: ...
+    def shape(self) -> tuple[int]: ...
```
cmp0xff, line 57: Checked implementation, always 1d

pandas-stubs/core/arrays/categorical.pyi
```diff
@@ -32,7 +33,7 @@ from pandas.core.dtypes.dtypes import CategoricalDtype as CategoricalDtype
 
 def contains(cat, key, container): ...
 
-class Categorical(ExtensionArray):
+class Categorical(NDArrayBackedExtensionArray):
```
cmp0xff, line 41: Checked implementation

pandas-stubs/core/arrays/categorical.pyi
```diff
@@ -48,7 +49,6 @@ class Categorical(ExtensionArray):
     def ordered(self) -> Ordered: ...
     @property
     def dtype(self) -> CategoricalDtype: ...
-    def size(self) -> int: ...
```
cmp0xff: Removed methods that are already in the new base class

pandas-stubs/core/arrays/datetimelike.pyi
```diff
@@ -60,11 +61,9 @@ class TimelikeOps:
         nonexistent: TimeNonexistent = "raise",
     ) -> Self: ...
 
-class DatetimeLikeArrayMixin(ExtensionArray):
+class DatetimeLikeArrayMixin(OpsMixin, NDArrayBackedExtensionArray):
     @property
     def ndim(self) -> int: ...
-    @property
-    def shape(self) -> tuple[int]: ...
```
cmp0xff: Removed methods that are alread in the new base classes

pandas-stubs/core/arrays/base.pyi
loicdiridollou: Are we sure about this, seems like ExtensionArray does not inherit from any other class so wondering how the signature would percolate?

tests/arrays/test_numpy_extension_array.py
```diff
@@ -0,0 +1,197 @@
...
+    BuiltinDtypeArg,
+    NumpyNotTimeDtypeArg,
+)
+
+
+def test_constructor() -> None:
+    none_data = [pd.NA, None]
+    check(assert_type(pd.array(none_data), NumpyExtensionArray), NumpyExtensionArray)
+
+    mixed_data = [1, "🐼"]
```
loicdiridollou: Are we sure this character is supported across all platforms? Just want to make sure that won't cause issues.

cmp0xff:
Indeed it is too big, will create a few smaller PRs before the main one

tests/arrays/test_numpy_extension_array.py
cmp0xff: We don't ship tests to users. It also runs in CI (and on my local machine), so I think it's fine.

[🐼 has been added to Unicode in 2015.](https://emojipedia.org/panda) It's ten years old.

loicdiridollou: Good to know, have not been a big emoji user in the past!

pandas-stubs/_typing.pyi
```diff
@@ -887,15 +887,15 @@ SliceType: TypeAlias = Hashable | None
 ## All types below this point are only used in pandas-stubs
 ######
 
-BuiltinNotStrDtypeArg: TypeAlias = (
+BuiltinDtypeArg: TypeAlias = (
```
cmp0xff, line 890: Everything here goes to `NumpyExtensionArray`

pandas-stubs/core/arrays/base.pyi
cmp0xff: My stand is that either we keep the original inheritance, or we can omit some classes if there is no public API. `value_counts` is public, and lives in `NDArrayBackedExtensionArray`, so we should keep it there.

pandas-stubs/core/arrays/string_.pyi
```diff
@@ -53,6 +57,6 @@ class BaseStringArray(ExtensionArray, Generic[_StorageT]):
     def dtype(self) -> StringDtype[_StorageT]: ...
 
 class StringArray(BaseStringArray[Literal["python"]], NumpyExtensionArray):
-    def __init__(self, values: AnyArrayLike, copy: bool = False) -> None: ...
-    def __arrow_array__(self, type: DtypeArg | None = None) -> Any: ...
+    def __new__(cls, values: np_ndarray_object, copy: bool = False) -> Self: ...
```
Dr-Irv: why not just keep `__init__()` ?

pandas-stubs/core/arrays/string_arrow.pyi
```diff
@@ -2,5 +2,11 @@ from typing import Literal
...
 from pandas.core.arrays.arrow.array import ArrowExtensionArray
 from pandas.core.arrays.string_ import BaseStringArray
+import pyarrow as pa
+from typing_extensions import Self
 
-class ArrowStringArray(ArrowExtensionArray, BaseStringArray[Literal["pyarrow"]]): ...
+from pandas._typing import DtypeArg
+
+class ArrowStringArray(ArrowExtensionArray, BaseStringArray[Literal["pyarrow"]]):
+    def __new__(cls, values: pa.StringArray, copy: bool = False) -> Self: ...
```
Dr-Irv: why not just use `__init__()`

I think we only need to use `__new__()` when we want to use overloads that return different results.

pandas-stubs/core/arrays/string_.pyi
cmp0xff: My idea was that if we consistently use `__init__`, it will be less confusing for the contributors.

Here and in `ArrowStringArroy`, if we switch to `__init__`, `pyright` thinks that it is incompatible with `BaseStringArray`, because the annotation of `values` is to be changed to `np_ndarray_object`. I think keeping `__new__` is better than using `__init__` and adding a `pyright: ignore`.

Dr-Irv: Except I think we do that elsewhere when a subclass has different arguments than the parent (i.e., use `__init__()` with a `pyright: ignore`), so why not be consistent with our past practice?

Also, I tried changing `NumpyExtensionArray` and `StringArray` to use `__init__()` with the args as you have them and had no pyright message or mypy message.

So can you try putting everything back to `__init__()` and, at least for me to look at it, if you need an `ignore`, put it in there and I can then decide what to do about `__new__()` vs. `__init__()`.

cmp0xff: Now I'm using `__init__` and also added new tests for it

Dr-Irv (approved):
thanks @cmp0xff

PR #1599, opened 2025-12-26
API: pandas-dev/pandas#53694 rename `PandasArray` -> `NumpyExtensionArray`

cmp0xff:
Towards #1594, #1469, #1580

`NumpyExtensionArray` is added in pandas-dev/pandas#53694. It is produced
- when no other Pandas array applies
- when a Python-native or a numpy dtype is imposed (provided that they are not datetime like)

Tests will be added in pandas-dev/pandas-stubs#1594.

### DIFF
--- a/pandas-stubs/core/arrays/string_.pyi
+++ b/pandas-stubs/core/arrays/string_.pyi
@@ -20,17 +20,15 @@ from pandas._typing import (
 from pandas.core.dtypes.base import ExtensionDtype
 
 Storage: TypeAlias = Literal["python", "pyarrow"]
-_Unknown: TypeAlias = Literal["_unknown"]
-_Storage: TypeAlias = Storage | _Unknown
 StorageT = TypeVar("StorageT", bound=Storage)
-_StorageT = TypeVar("_StorageT", bound=_Storage, default=_Unknown)
+_StorageT = TypeVar("_StorageT", bound=Storage | None, default=None)
 
 # Trick to make mypy happy
 @type_check_only
 class _StringDtypeStorageDescriptor:
     @overload
     def __get__(
-        self, instance: StringDtype[_Unknown], owner: type[StringDtype[_Unknown]]
+        self, instance: StringDtype[None], owner: type[StringDtype[None]]
     ) -> Storage: ...
     @overload
     def __get__(


loicdiridollou:
This one looks good, I will do a second pass when all the ANN PRs are merged since they work on the same code so to make sure we have not broken anything.

pandas-stubs/core/construction.pyi
```diff
@@ -44,19 +50,9 @@ def array(
...
-        data: Sequence[object],
-        dtype: str | np.dtype[Any] | ExtensionDtype | None = None,
-        copy: bool = True,
-    ) -> ExtensionArray: ...
+@overload
+def array(
+    data: SequenceNotStr[object] | np_ndarray | NumpyExtensionArray | RangeIndex,
+    dtype: BuiltinDtypeArg | NumpyNotTimeDtypeArg | None = None,
+    copy: bool = True,
+) -> NumpyExtensionArray: ...
```
Dr-Irv, line 144: I think you need to add an overload ahead of this that returns `StringArray`.

```python
>>> pd.array(["abc", "def"])
<StringArray>
['abc', 'def']
Length: 2, dtype: string
```

Dr-Irv (changes requested):

pandas-stubs/core/construction.pyi
cmp0xff: 5145cdbb7a8c08080ed57dd9ab71a1cc470a1dcc

pandas-stubs/core/construction.pyi
```diff
@@ -51,8 +54,14 @@ def array(
     copy: bool = True,
 ) -> FloatingArray: ...
 @overload
+def array(
+    data: SequenceNotStr[str | np.str_ | NAType | None] | np_ndarray_str | StringArray,
```
Dr-Irv: Unfortunately, we need to support:
```python
pd.array(["abc", np.nan, "def"])
```
And that means we have to add `float` to the list of things that can be in the sequence.

So can you add that, and add a test for it?

pandas-stubs/core/construction.pyi
Dr-Irv: We have to support 
```python
pd.array([1, np.nan, 3], dtype=pd.Int64Dtype())
```

Similar for `BooleanArray`.   

Probably if the `dtype` is NOT `None`, we add `float` to the signature.  Might not be easy to deal with this because of the overload for `FloatingArray`

pandas-stubs/core/construction.pyi
cmp0xff: This is difficult.

In pandas 2.3, `pd.array([float("nan"), "1"])` gives `StringArray`, but in pandas 3.0, it gives `ArrowStringArray`.

In all versions, `pd.array([1.0, "1"])` gives `NumpyExtensionArray`.

091e502013f040d4d6144b98ff65472f248586b5

pandas-stubs/core/construction.pyi
cmp0xff: 091e502013f040d4d6144b98ff65472f248586b5

tests/arrays/test_boolean_array.py
```diff
@@ -0,0 +1,41 @@
...
+        "list[bool | np.bool[Literal[True]]]", [True, np.bool_(True)]
+    )
+
+    check(assert_type(pd.array(data), BooleanArray), BooleanArray)
+    check(assert_type(pd.array([*data, None]), BooleanArray), BooleanArray)
+    check(assert_type(pd.array([*data, pd.NA]), BooleanArray), BooleanArray)
+    check(assert_type(pd.array([*data, None, pd.NA]), BooleanArray), BooleanArray)
+
+    check(assert_type(pd.array(tuple(data)), BooleanArray), BooleanArray)
+    check(assert_type(pd.array(UserList(data)), BooleanArray), BooleanArray)
```
Dr-Irv: You should add a test where the data is inline, e.g.  `pd.array([True, np.bool_(True)])`

tests/arrays/test_floating_array.py
```diff
@@ -1,33 +1,94 @@
...
+    missing_values: tuple[Any, ...],
+    typ: Callable[[Sequence[Any]], Sequence[Any]],
+) -> None:
+    check(pd.array(typ([*nums, *missing_values])), FloatingArray)
+
+    if TYPE_CHECKING:
+        data = cast(  # pyright: ignore[reportUnnecessaryCast]
+            "list[float | np.float32]", [1.0, np.float32(1)]
+        )
+
```
Dr-Irv: same comment about adding an inline test

tests/arrays/test_integer_array.py
```diff
@@ -0,0 +1,38 @@
...
+from pandas.core.arrays.integer import IntegerArray
+from typing_extensions import assert_type
+
+from tests import check
+
+
+def test_construction_sequence() -> None:
+    data = cast(  # pyright: ignore[reportUnnecessaryCast]
+        "list[int | np.int8]", [1, np.int8(1)]
+    )
```
Dr-Irv: same comment about adding inline test

tests/arrays/test_string_array.py
```diff
@@ -0,0 +1,38 @@
...
+from pandas.core.arrays.string_ import StringArray
+from typing_extensions import assert_type
+
+from tests import check
+
+
+def test_construction_sequence() -> None:
+    data = cast(  # pyright: ignore[reportUnnecessaryCast]
+        "list[str | np.str_]", ["1", np.str_("1")]
+    )
```
Dr-Irv: same comment about adding inline test

pandas-stubs/core/construction.pyi
Dr-Irv: > In pandas 2.3, `pd.array([float("nan"), "1"])` gives `StringArray`, but in pandas 3.0, it gives `ArrowStringArray`.

Actually, the result in 3.0 depends on whether `pyarrow` is installed.

One option is that in 3.0, `StringArray` and `ArrowStringArray` derive from `BaseStringArray`, so we could use that when we worry about 3.0 compatibility.

> In all versions, pd.array([1.0, "1"]) gives NumpyExtensionArray.

Yes, this is where it is hard to do anything because there is not `Literal[np.nan]` .  It's also why we want to encourage the use of `pd.NA`

Dr-Irv (changes requested):

tests/arrays/test_boolean_array.py
cmp0xff: Now I am mostly using literals. Non-inline tests are there only when `mypy` is unhappy somehow.

4f9521e2443cd52bc566a5548719d20a53afa2d4

tests/arrays/test_base_string_array.py
```diff
@@ -0,0 +1,56 @@
...
+
+        assert_type(pd.array(["🐼", np.str_("🐼"), None]), BaseStringArray)
+        assert_type(pd.array(["🐼", np.str_("🐼"), pd.NA]), BaseStringArray)
+
+        assert_type(pd.array(["🐼", np.str_("🐼"), None, pd.NA]), BaseStringArray)
+
+        assert_type(pd.array(("🐼", np.str_("🐼"))), BaseStringArray)
+        assert_type(pd.array(("🐼", np.str_("🐼"), pd.NA)), BaseStringArray)
+
+        assert_type(pd.array(UserList(["🐼", np.str_("🐼")])), BaseStringArray)
```
Dr-Irv: why not just use a regular string like `"a"` here and elsewhere?

tests/arrays/test_boolean_array.py
```diff
@@ -1,40 +1,55 @@
...
+    typ: Callable[[Sequence[Any]], Sequence[Any]],
+) -> None:
+    check(pd.array(typ([*data, *missing_values])), BooleanArray)
 
-    check(assert_type(pd.array(data), BooleanArray), BooleanArray)
-    check(assert_type(pd.array([*data, None]), BooleanArray), BooleanArray)
-    check(assert_type(pd.array([*data, pd.NA]), BooleanArray), BooleanArray)
-    check(assert_type(pd.array([*data, None, pd.NA]), BooleanArray), BooleanArray)
+    if TYPE_CHECKING:
+        assert_type(pd.array([True, np.bool_(True)]), BooleanArray)
```
Dr-Irv, line 37: Should add a test for `pd.array([True, False])`

tests/arrays/test_base_string_array.py
```diff
@@ -0,0 +1,56 @@
...
+@pytest.mark.parametrize("missing_values", powerset([np.nan, None, pd.NA]))
+def test_construction_sequence(
+    data: tuple[str | np.str_, ...],
+    missing_values: tuple[Any, ...],
+    typ: Callable[[Sequence[Any]], Sequence[Any]],
+) -> None:
+    check(pd.array(typ([*data, *missing_values])), BaseStringArray)
+
+    if TYPE_CHECKING:
+        assert_type(pd.array(["🐼", np.str_("🐼")]), BaseStringArray)
```
Dr-Irv: Should add a test for `pd.array(["abc", "def"])`

tests/arrays/test_floating_array.py
```diff
@@ -17,48 +16,48 @@
...
     typ: Callable[[Sequence[Any]], Sequence[Any]],
 ) -> None:
-    check(pd.array(typ([*nums, *missing_values])), FloatingArray)
+    check(pd.array(typ([*data, *missing_values])), FloatingArray)
 
     if TYPE_CHECKING:
-        data = cast(  # pyright: ignore[reportUnnecessaryCast]
-            "list[float | np.float32]", [1.0, np.float32(1)]
-        )
+        assert_type(pd.array([1.0, np.float16(1)]), FloatingArray)
```
Dr-Irv: Should add a test for `pd.array([1.2, 3.4])`

tests/arrays/test_integer_array.py
```diff
@@ -1,37 +1,52 @@
...
+    typ: Callable[[Sequence[Any]], Sequence[Any]],
+) -> None:
+    check(pd.array(typ([*data, *missing_values])), IntegerArray)
 
-    check(assert_type(pd.array(data), IntegerArray), IntegerArray)
-    check(assert_type(pd.array([*data, None]), IntegerArray), IntegerArray)
-    check(assert_type(pd.array([*data, pd.NA]), IntegerArray), IntegerArray)
-    check(assert_type(pd.array([*data, None, pd.NA]), IntegerArray), IntegerArray)
+    if TYPE_CHECKING:
+        assert_type(pd.array([1, np.int8(1)]), IntegerArray)
```
Dr-Irv: Should add a test for `pd.array([1, 2])`

tests/arrays/test_numpy_extension_array.py
```diff
@@ -45,6 +40,7 @@ def test_construction_sequence(
         assert_type(pd.array([pd.NA]), NumpyExtensionArray)
         assert_type(pd.array([pd.NaT]), NumpyExtensionArray)
 
+        # mypy does not like the literal version pd.array([np.nan, pd.NaT])
         _10 = [np.nan, pd.NaT]
```
Dr-Irv: can we do the test with `pd.array([np.nan, pd.NaT])` and just do an appropriate `#type: ignore` for `mypy` and include your comment?

tests/arrays/test_string_array.py
```diff
@@ -1,37 +1,54 @@
...
 
-def test_construction_array_like() -> None:
-    data = cast(  # pyright: ignore[reportUnnecessaryCast]
-        "list[str | np.str_]", ["1", np.str_("1")]
-    )
-    np_arr = np.array(data, np.str_)
+        assert_type(pd.array(("🐼", np.str_("🐼"))), StringArray)
+        assert_type(pd.array(("🐼", np.str_("🐼"), pd.NA)), StringArray)
+
+        assert_type(pd.array(UserList(["🐼", np.str_("🐼")])), StringArray)
```
Dr-Irv: can we just use normal strings here and elsewhere in the tests, e.g., `"a"` ?

tests/arrays/test_string_array.py
```diff
@@ -1,37 +1,54 @@
...
+    typ: Callable[[Sequence[Any]], Sequence[Any]],
+) -> None:
+    check(pd.array(typ([*data, *missing_values])), StringArray)
 
-    check(assert_type(pd.array(data), StringArray), StringArray)
-    check(assert_type(pd.array([*data, None]), StringArray), StringArray)
-    check(assert_type(pd.array([*data, pd.NA]), StringArray), StringArray)
-    check(assert_type(pd.array([*data, None, pd.NA]), StringArray), StringArray)
+    if TYPE_CHECKING:
+        assert_type(pd.array(["🐼", np.str_("🐼")]), StringArray)
```
Dr-Irv: Need to add a test for `pd.array(["abc", "def"])`

Dr-Irv (changes requested):
Note: I've just been reviewing changes since last review, but when I don't see anything there to add, I will look again at the whole PR and see if anything else comes up.

tests/arrays/test_boolean_array.py
cmp0xff: d403b6ff2722fc989970995089c14934eff3d758

tests/arrays/test_base_string_array.py
cmp0xff: 583a510cf835917083fb77c7ae5768fbf02427cc

tests/arrays/test_base_string_array.py
cmp0xff: Resolved with #1608

tests/arrays/test_floating_array.py
cmp0xff: 583a510cf835917083fb77c7ae5768fbf02427cc

tests/arrays/test_integer_array.py
cmp0xff: a04f3fdfb3f0ad1f03e4ad852c12a950d6c34f60

tests/arrays/test_numpy_extension_array.py
cmp0xff: 0749327871d13b0ea97fac0252ee8abff2b80f29

tests/arrays/test_string_array.py
cmp0xff: Resolved with #1608 and also 08694177c5054e791d85692dd8997c7db5121a1b

tests/arrays/test_string_array.py
cmp0xff: Resolved with #1608

pandas-stubs/core/arrays/string_.pyi
```diff
@@ -14,18 +20,23 @@ from pandas._typing import (
 
 from pandas.core.dtypes.base import ExtensionDtype
 
-class StringDtype(ExtensionDtype):
+Storage: TypeAlias = Literal["python", "pyarrow"]
+StorageT = TypeVar("StorageT", bound=Storage, default=Storage)
```
Dr-Irv: If we do:
```python
Storage: TypeAlias = Literal["python", "pyarrow", "Unknown"]
StorageT = TypeVar("StorageT", bound=Storage, default=Literal["Unknown"])
```
and then in `StringDtype`, have `__new__()` look like this:
```python
    @overload
    def __new__(
        cls, storage: None = ..., na_value: NAType | float = ...
    ) -> StringDtype[Literal["Unknown"]]: ...
    @overload
    def __new__(
        cls, storage: Literal["python"], na_value: NAType | float = ...
    ) -> StringDtype[Literal["python"]]: ...
    @overload
    def __new__(
        cls, storage: Literal["pyarrow"], na_value: NAType | float = ...
    ) -> StringDtype[Literal["pyarrow"]]: ...
```
then the `mypy` ignore problems in `tests/arrays/test_base_string_array.py` go away, because the default becomes `"Unknown"`, which is really the case because the storage isn't known until runtime.

You have to make changes elsewhere to make the default `StringDtype[Literal["Unknown"]]` and `BaseStringArray[Literal["Unknown"]]` .

I tried it locally and it worked.

Now, it is true that at runtime, the storage can't be `Unknown`, but I think this is better than having to deal with the mypy problem.  It is just `Unknown` when looking at static typing.

tests/arrays/test_numpy_extension_array.py
```diff
@@ -40,30 +40,25 @@ def test_construction_sequence(
         assert_type(pd.array([pd.NA]), NumpyExtensionArray)
         assert_type(pd.array([pd.NaT]), NumpyExtensionArray)
 
-        # mypy does not like the literal version pd.array([np.nan, pd.NaT])
-        _10 = [np.nan, pd.NaT]
-        assert_type(pd.array(_10), NumpyExtensionArray)
+        # mypy does not like the literal versions
```
Dr-Irv: Modified comment:
```python
# Because np.nan is a float, mypy infers any list with np.nan and types like pd.NA and None
# to be list[object]
# It would be quite unusual for user code to be mixing np.nan with the other "NA"-like types.
```

pandas-stubs/core/arrays/string_.pyi
cmp0xff: I believe `"_unknown"` is now hidden from most users 5470c737cea1a8321b937da6377b6352cb989c3a

tests/arrays/test_numpy_extension_array.py
cmp0xff: 4a823091ac445fff6eff83f5f385dc2dd13f6674

pandas-stubs/core/arrays/string_.pyi
cmp0xff: Now I have even removed `Unknown` 5efd3f44040514f8e196f8e336e36878eed1510e

Dr-Irv: Nice.

PR #1600, opened 2025-12-26
TYP: dividing `ASTYPE_x_ARGS`

cmp0xff:
Towards #1594, #1469

- Following `ASTYPE_FLOAT_ARGS`, we need more detailed divisions for constants like `ASTYPE_BYTES_ARGS` as well, in order to add tests for `NumpyExtensionArray`. This division also helps with future tests for `ExtensionArray`, `Index` and `Series`.
- In addition, `c32`, `complex256` and `object_` can be used in constructing `Index` and `Series`, which are now added.

### DIFF
--- a/tests/_typing.py
+++ b/tests/_typing.py
@@ -1,8 +1,8 @@
 # This file serves as a stub file for static type checkers
 # (pyright does not like it if I call the file tests/_typing.pyi).
-# In tests/conftest.py, this file is copied to tests/_typing.pyi, and
-# pandas-stubs/_typing.pyi is copied to tests/_typing.py.
-# After tests are done, tests/_typing.pyi is copied back to
+# By tests/conftest.py, pytest copies this file to tests/_typing.pyi, and
+# pandas-stubs/_typing.pyi to tests/_typing.py.
+# After tests are done, pytest copies tests/_typing.pyi back to
 # pandas-stubs/_typing.pyi.
 # Do not commit this file if its content is identical to pandas-stubs/_typing.pyi.
 


tests/__init__.py
```diff
@@ -80,19 +208,200 @@
...
+    | NUMPY_INT_ARGS
+    | NUMPY_UINT_ARGS
+    | NUMPY_FLOAT16_ARGS
+    | NUMPY_FLOAT_NOT16_ARGS
+    | NUMPY_COMPLEX_ARGS
+    | NUMPY_STRING_ARGS
+    | NUMPY_BYTES_ARGS
+    | NUMPY_OBJECT_ARGS
+    | NUMPY_VOID_ARGS
+)
```
Dr-Irv: One concern here is whether we are testing all the types that are declared in `pandas-stubs/_typing.pyi` .  

I think we can test that by doing something like this:

- Create a mapping from the declarations in `_typing.pyi` to the declarations here
  - The map would go from the "typing type" to the "testing type", e.g., from a new type in `_typing.pyi` called `PandasDtypeUintArg` to `PANDAS_UINT_ARGS`
- Using `typing.get_args()` on each key of that map, it would make sure that key exists in the keys of the value part of that map.

This would probably require a big re-org of how the types are declared in `_typing.pyi`, but anything we can do to make sure we aren't missing anything would make the PR easier to review.

pandas-stubs/_typing.pyi
```diff
@@ -321,7 +313,6 @@ BuiltinFloatDtypeArg: TypeAlias = type[float] | Literal["float"]
 PandasFloatDtypeArg: TypeAlias = (
     pd.Float32Dtype | pd.Float64Dtype | Literal["Float32", "Float64"]
 )
-PandasAstypeFloatDtypeArg: TypeAlias = Literal["float_", "longfloat"]
```
cmp0xff, line 324: They work neither in `pd.Series([1], dtype="float_")` nor in `pd.Series([1]).astype("float_")`. Removing.

pandas-stubs/_typing.pyi
```diff
@@ -357,11 +348,6 @@ FloatDtypeArg: TypeAlias = (
 )
 # Builtin complex type and its string alias
 BuiltinComplexDtypeArg: TypeAlias = type[complex] | Literal["complex"]
-PandasAstypeComplexDtypeArg: TypeAlias = (
-    Literal["singlecomplex"]  # noqa: PYI030
-    | Literal["cfloat", "complex_"]
-    | Literal["c32", "complex256", "clongfloat", "longcomplex"]
-)
```
cmp0xff, line 364: They work neither in `pd.Series([1], dtype="complex_")` nor in `pd.Series([1]).astype("complex_")`. Removing.

pandas-stubs/_typing.pyi
```diff
@@ -373,7 +359,7 @@ NumpyComplexDtypeArg: TypeAlias = (
     #  https://numpy.org/doc/stable/reference/arrays.scalars.html#numpy.clongdouble
     # NOTE: Alias np.complex256 only on Linux x86_64, use np.clongdouble instead
     | type[np.clongdouble]
-    | Literal["G", "clongdouble"]
+    | Literal["G", "c32", "complex256", "clongdouble"]
```
cmp0xff, line 362: They work both in `pd.Series([1], dtype="c32")` and in `pd.Series([1]).astype("c32")`. Adding back.

pandas-stubs/_typing.pyi
```diff
@@ -538,8 +524,7 @@ CategoryDtypeArg: TypeAlias = CategoricalDtype | Literal["category"]
 BuiltinObjectDtypeArg: TypeAlias = type[object] | Literal["object"]
 # Numpy object type and its string alias
 # https://numpy.org/doc/stable/reference/arrays.scalars.html#numpy.object_
-# NOTE: "object_" not assigned
-NumpyObjectDtypeArg: TypeAlias = type[np.object_] | Literal["O"]
+NumpyObjectDtypeArg: TypeAlias = type[np.object_] | Literal["object_", "O"]
```
cmp0xff, line 527: `"object_"` also works, at least for `NumpyExtensionArray`.

tests/__init__.py
```diff
@@ -86,6 +90,9 @@
     "i": np.intc,
     "int32": np.int32,
     "i4": np.int32,
+    # numpy long
+    "l": np.long,
+    "long": np.long,
```
cmp0xff: A few missing cases in the previous implementation.

tests/__init__.py
```diff
@@ -141,6 +148,9 @@
     "I": np.uintc,
     "uint32": np.uint32,
     "u4": np.uint32,
+    # numpy ulong
+    "L": np.ulong,
+    "ulong": np.ulong,
```
cmp0xff: A few missing cases in the previous implementation.

tests/__init__.py
```diff
@@ -164,12 +174,21 @@
...
+    pd.Float64Dtype(): np.float64,
+    "Float64": np.float64,
+}
 NUMPY_FLOAT16_ARGS = {
     np.half: np.half,
     "half": np.half,
     "e": np.half,
     "float16": np.float16,
     "f2": np.float16,
+    "<f2": np.float16,
```
cmp0xff: A few missing cases in the previous implementation.

tests/__init__.py
```diff
@@ -193,25 +212,23 @@
     "float128": np.longdouble,  # NOTE: UNIX ONLY
 }
 PYARROW_FLOAT_ARGS = {
+    # pyarrow float16
+    "float16[pyarrow]": float,
```
cmp0xff: A few missing cases in the previous implementation.

tests/__init__.py
```diff
@@ -350,18 +387,19 @@
     "duration[us][pyarrow]": timedelta,
     "duration[ns][pyarrow]": timedelta,
 }
-ASTYPE_TIMEDELTA_ARGS = NUMPY_TIMEDELTA_ARGS | PYARROW_TIMEDELTA_ARGS
+TYPE_TIMEDELTA_ARGS = NUMPY_TIMEDELTA_ARGS | PYARROW_TIMEDELTA_ARGS
+ASTYPE_TIMEDELTA_ARGS = TYPE_TIMEDELTA_ARGS | PANDAS_ASTYPE_TIMEDELTA_ARGS
 
 PYTHON_STRING_ARGS = {str: str, "str": str}
-PANDAS_STRING_ARGS = {pd.StringDtype(): str}
+PANDAS_STRING_ARGS = {pd.StringDtype(): str, "string": str}
```
cmp0xff: A few missing cases in the previous implementation.

tests/__init__.py
```diff
@@ -554,3 +706,20 @@ def exception_on_platform(dtype: type | str | ExtensionDtype) -> type[Exception]
...
+    elif isinstance(dtype, type) and not str(dtype).startswith("type["):
+        if dtype is pd.DatetimeTZDtype:
+            yield dtype(tz=timezone.utc)
+        elif "pandas" in str(dtype):
+            yield dtype()
+        else:
+            yield dtype
+    else:
+        for arg in get_args(dtype):
+            yield from get_dtype(arg)
```
cmp0xff, line 214: `get_args` is not good enough. Introducing `get_dtype`.

tests/test_typing.py
```diff
@@ -0,0 +1,14 @@
...
+
+from tests import (
+    get_dtype,
+    get_dtype_arg_alias_maps,
+)
+
+
+@pytest.mark.parametrize(("dtype_arg", "alias_map"), get_dtype_arg_alias_maps().items())
+def test_dtype_arg_aliases(dtype_arg: Any, alias_map: Mapping[Any, Any]) -> None:
+    assert set(get_dtype(dtype_arg)) == set(alias_map)
```
cmp0xff, line 32: Ensuring `DtypeArgs` and the alias maps are the same.

tests/__init__.py
cmp0xff: 03f3f7db47c21b25df64daa60cd69d5dfc05e2aa

tests/_typing.py
Dr-Irv, line 7: do you need to update this comment??

tests/__init__.py
Dr-Irv: Nice work, thanks.

tests/__init__.py
```diff
@@ -39,13 +48,147 @@
 PD_LTE_23 = Version(pd.__version__) < Version("2.3.999")
 NUMPY20 = np.lib.NumpyVersion(np.__version__) >= "2.0.0"
 
-NATIVE_FLOAT_ARGS = {float: np.floating, "float": np.floating}
+PYTHON_BOOL_ARGS = {bool: np.bool_, "bool": np.bool_}
```
Dr-Irv: can you move all of these definitions to do with the dtype/astype checking to a separate file so that `__init__.py` remains short, and we then have anything to do with the type tests in one file that can be imported from any of the `test*` files.

cmp0xff: Moved to `tests/dtypes.py` 836c2b8e5179f05390c24dbc5b370ff433991489

tests/_typing.py
cmp0xff: The comment still looks valid to me, since the content is not identical to that of `pandas/_typing.pyi`.

Dr-Irv: Well, if we are making it part of the repo, then the last 5 lines of the comment are not correct.  And we shouldn't do the copy that we did before, otherwise, this gets clobbered by the copying.

Dr-Irv (changes requested):
only disagreement is on the comment in `tests/_typing.py`

tests/_typing.py
cmp0xff: So `tests/_typing.py` only does importing and does not duplicate `pandas-stubs/_typing.pyi`, as you can see. The copying and restoring is done automatically by `pytest`.

I've updated the comments a bit.

Dr-Irv (approved):
thanks @cmp0xff

PR #1603, opened 2025-12-26
TYP: `StringDtype`

cmp0xff:


loicdiridollou (approved):
Looks good, thanks @cmp0xff

PR #1604, opened 2025-12-26
BUG: 🌉 nightly

cmp0xff:
`value_counts` is probably Cython and can be imported only after compile (?)

### DIFF
--- a/tests/test_api_typing.py
+++ b/tests/test_api_typing.py
@@ -150,9 +150,9 @@ def test_json_reader(tmp_path: Path) -> None:
     def f1(gb: JsonReader) -> None:
         check(gb, JsonReader)
 
-    path = str(tmp_path / str(uuid.uuid4()))
-    check(assert_type(df.to_json(path), None), type(None))
-    json_reader = read_json(path, chunksize=1, lines=True)
+    path_str = str(tmp_path / str(uuid.uuid4()))
+    check(assert_type(df.to_json(path_str), None), type(None))
+    json_reader = read_json(path_str, chunksize=1, lines=True)
     f1(json_reader)
     json_reader.close()
 
@@ -206,13 +206,13 @@ def test_statereader(tmp_path: Path) -> None:
     df = pd.DataFrame([[1, 2], [3, 4]], columns=["col_1", "col_2"])
     time_stamp = pd.Timestamp(2000, 2, 29, 14, 21)
     variable_labels = {"col_1": "This is an example"}
-    path = str(tmp_path / str(uuid.uuid4()))
+    path_str = str(tmp_path / str(uuid.uuid4()))
     df.to_stata(
-        path, time_stamp=time_stamp, variable_labels=variable_labels, version=None
+        path_str, time_stamp=time_stamp, variable_labels=variable_labels, version=None
     )
 
     def f1(gb: StataReader) -> None:
         check(gb, StataReader)
 
-    with StataReader(path) as reader:
+    with StataReader(path_str) as reader:
         f1(reader)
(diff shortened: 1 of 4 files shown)


pandas-stubs/_testing/__init__.pyi
```diff
@@ -174,5 +174,3 @@ def assert_produces_warning(
     raise_on_extra_warnings: bool = True,
     match: str | None = None,
 ) -> Generator[list[warnings.WarningMessage], None, None]: ...
-@contextmanager
-def ensure_clean(filename: str | None = None) -> Generator[str, None, None]: ...
```
loicdiridollou, line 178: @Dr-Irv may have more context with this, but did this function have any other use than just creating a folder and deleting once the job was over?

cmp0xff: When I check https://github.com/pandas-dev/pandas/pull/63487, it seems that people are just doing what I described in https://github.com/pandas-dev/pandas-stubs/pull/1604#issuecomment-3694163433 and reducing indentation (which is annoying)

loicdiridollou: Okay thanks for checking, at work I have only used the pytest tmppath so seems to align.
+1 on the indentation thing 👍

tests/test_api_typing.py
```diff
@@ -203,17 +202,17 @@ def f1(gb: Window) -> None:
...
-def test_statereader() -> None:
+def test_statereader(tmp_path: Path) -> None:
     df = pd.DataFrame([[1, 2], [3, 4]], columns=["col_1", "col_2"])
     time_stamp = pd.Timestamp(2000, 2, 29, 14, 21)
     variable_labels = {"col_1": "This is an example"}
-    with ensure_clean() as path:
-        df.to_stata(
-            path, time_stamp=time_stamp, variable_labels=variable_labels, version=None
-        )
+    path = str(tmp_path / str(uuid.uuid4()))
```
loicdiridollou: maybe name it `path_str`

tests/test_styler.py
```diff
@@ -199,21 +199,21 @@ def test_text_gradient() -> None:
     check(assert_type(DF.style.text_gradient(), Styler), Styler)
 
 
-def test_to_excel() -> None:
-    with ensure_clean("test.xlsx") as path:
-        check(assert_type(DF.style.to_excel(path), None), type(None))
+def test_to_excel(tmp_path: Path) -> None:
+    path = str(tmp_path / f"{uuid.uuid4()}test.xlsx")
```
loicdiridollou: path_str where it is a str if not too much work

cmp0xff: That can be hard, `path` is everywhere. Making the name longer leads to more formatting changes.

pandas-stubs/_testing/__init__.pyi
Dr-Irv: > @Dr-Irv may have more context with this, but did this function have any other use than just creating a folder and deleting once the job was over?

My sense here is that the `tmp_path` fixture in `pytest` was there but not used, so `ensure_clean()` is what was used for a very long time, and then someone noticed that the `tmp_path` fixture could be used instead.

tests/extension/decimal/array.py
```diff
@@ -341,6 +340,8 @@ def convert_values(param: Any) -> ExtensionArray | list[Any]:
         return cast(np_1darray_bool, np.asarray(res, dtype=bool))
 
     def value_counts(self, dropna: bool = True) -> Series[int]:
+        from pandas.core.algorithms import value_counts
```
loicdiridollou, line 343: Will assume this one was also needed but in case it is fine to leave it there

cmp0xff: Yep, that one was the original error.

PR #1605, opened 2025-12-27
TYP: enable `report_x_ParameterType` and `reportUnknownLambdaType`

cmp0xff:
- `report_x_ParameterType`: all due to untyped external modules
- `reportUnknownLambdaType`: only in tests


loicdiridollou (approved):
Thanks @cmp0xff

PR #1607, opened 2025-12-27
TST: remove a few `pytest.skip*`

cmp0xff:
`pytables` is now compatible with `numpy` 2.x. We can enable the tests.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -71,7 +71,7 @@ html5lib = ">=1.1"
 python-calamine = ">=0.2.0"
 pyarrow-stubs = { version = ">=20.0.0.20250928", python = "<4" }
 fsspec = "^2025.10.0"
-tables = { version = ">=3.10.2", python = ">=3.11,<3.14" }
+tables = { version = ">=3.10.2", python = ">=3.11,<3.14" }  # TODO: add py314 when PyTables/PyTables#1261 is resolved
 
 [build-system]
 requires = ["poetry-core>=1.0.0"]
--- a/tests/test_io.py
+++ b/tests/test_io.py
@@ -336,35 +336,23 @@ def test_sas_xport() -> None:
         pass
 
 
-MESSAGE_PYTABLE_MISSING = (
-    r"Missing optional dependency 'pytables'.  Use pip or conda to install pytables."
+MESSAGE_PYTABLE_314 = (
+    "PyTables does not support Python 3.14+ yet, see PyTables/PyTables#1261"
 )
 
 
+@pytest.mark.skipif(sys.version_info < (3, 11), reason="PyTables requires Python 3.11+")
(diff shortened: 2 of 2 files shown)


pyproject.toml
```diff
@@ -71,6 +71,7 @@ html5lib = ">=1.1"
 python-calamine = ">=0.2.0"
 pyarrow-stubs = { version = ">=20.0.0.20250928", python = "<4" }
 fsspec = "^2025.10.0"
+tables = { version = ">=3.10.2", python = ">=3.11,<3.14" }
```
loicdiridollou: Is there a linked issue to enable pytables on 3.14 that we could reference here?

tests/test_api_typing.py
```diff
@@ -85,7 +84,6 @@ def f1(gb: ResamplerGroupBy) -> None:
     f1(gb_df.resample("1D"))
 
 
-@pytest.mark.skip("Resampling with a PeriodIndex is deprecated.")
```
loicdiridollou, line 88: So the deprecation is not effective anymore here?

tests/test_io.py
```diff
@@ -393,19 +419,33 @@ def test_read_hdf_iterator(tmp_path: Path) -> None:
...
+
+
+def _hdf_context_manager(path_str: str) -> None:
     check(assert_type(DF.to_hdf(path_str, key="df", format="table"), None), type(None))
     with HDFStore(path_str, mode="r") as store:
         check(assert_type(store.is_open, bool), bool)
         check(assert_type(store.get("df"), DataFrame | Series), DataFrame)
 
 
-@pytest.mark.skipif(NUMPY20, reason="numpy 2.0 not compatible with Pytables")
```
loicdiridollou, line 405: Wondering if we should just skip the test fro python314 for now instead of adding the version check (the same way we disabled the test for numpy above 2.0), what do you think? It will make things less verbose.

loicdiridollou (changes requested):
Minor questions about the best approach here

tests/test_io.py
cmp0xff: 540f1b91dfb320964a0840b60e2628b4bed61423

pyproject.toml
cmp0xff: PyTables/PyTables#1261

Added in 540f1b91dfb320964a0840b60e2628b4bed61423

tests/test_api_typing.py
cmp0xff: The deprecation is reverted pandas-dev/pandas#57033

loicdiridollou (approved):
Good to me, could have combined the two boolean tests for skipif in one line but this is bound to disappear so all good here. thanks @cmp0xff

PR #1608, opened 2025-12-29
TYP: add `BaseStringArray` and `ArrowStringArray`

cmp0xff:
Towards #1599

In order to support `pd.array([float("nan")], "string")`, we need changes that are complicated.

- Surprisingly, `{pd.DatetimeTZDtype(tz="UTC"), "datetime64[ns, UTC]"} == {"datetime64[ns, UTC]"}`. This is fixed.
- `pd.StringDtype` is made generic to accomodate the to-be-added `ArrowStringArray` (tested)
- `pd.array([np.nan], "string[python]")` gives a `StringArray` (tested)
- `pd.array([np.nan], "string")` gives a `BaseStringArray` (tested)

### DIFF
--- a/tests/arrays/test_arrow_string_array.py
+++ b/tests/arrays/test_arrow_string_array.py
@@ -1,21 +1,89 @@
-from typing import Literal
+from typing import (
+    TYPE_CHECKING,
+    Literal,
+)
 
 import numpy as np
 import pandas as pd
 from pandas.core.arrays.string_arrow import ArrowStringArray
+import pytest
 from typing_extensions import assert_type
 
+from pandas._libs.missing import NAType
+
 from tests import check
+from tests._typing import PyArrowStrDtypeArg
+from tests.dtypes import PYARROW_STRING_ARGS
+from tests.utils import powerset
 
 
-def test_construction_dtype_na() -> None:
-    check(
-        assert_type(pd.array([np.nan], "string[pyarrow]"), ArrowStringArray),
-        ArrowStringArray,
-    )
+@pytest.mark.parametrize("data", powerset(["pd", np.str_("pd")]))
+@pytest.mark.parametrize(("dtype", "target_dtype"), PYARROW_STRING_ARGS.items())
+def test_construction_dtype(
+    data: tuple[str | np.str_, ...], dtype: PyArrowStrDtypeArg, target_dtype: type
+) -> None:
+    dtype_notna = target_dtype if data else None
+    check(pd.array([*data], dtype=dtype), ArrowStringArray, dtype_notna)
+    check(pd.array([*data, *data], dtype=dtype), ArrowStringArray, dtype_notna)
+
+    dtype_na = target_dtype if data else NAType
+    check(pd.array([*data, np.nan], dtype=dtype), ArrowStringArray, dtype_na)
+    check(pd.array([*data, *data, np.nan], dtype=dtype), ArrowStringArray, dtype_na)
+
+    if TYPE_CHECKING:
+        assert_type(pd.array([], dtype=pd.StringDtype("pyarrow")), ArrowStringArray)
+        assert_type(pd.array([], dtype="string[pyarrow]"), ArrowStringArray)
+
+        assert_type(
+            pd.array([np.nan], dtype=pd.StringDtype("pyarrow")), ArrowStringArray
+        )
+        assert_type(pd.array([np.nan], dtype="string[pyarrow]"), ArrowStringArray)
+
+        assert_type(pd.array(["1"], dtype=pd.StringDtype("pyarrow")), ArrowStringArray)
+        assert_type(pd.array(["1"], dtype="string[pyarrow]"), ArrowStringArray)
+
+        assert_type(
+            pd.array(["1", "2"], dtype=pd.StringDtype("pyarrow")), ArrowStringArray
+        )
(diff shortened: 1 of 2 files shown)


tests/arrays/test_base_string_array.py
```diff
@@ -0,0 +1,56 @@
...
+
+        assert_type(pd.array(["🐼", np.str_("🐼"), None]), BaseStringArray)
+        assert_type(pd.array(["🐼", np.str_("🐼"), pd.NA]), BaseStringArray)
+
+        assert_type(pd.array(["🐼", np.str_("🐼"), None, pd.NA]), BaseStringArray)
+
+        assert_type(pd.array(("🐼", np.str_("🐼"))), BaseStringArray)
+        assert_type(pd.array(("🐼", np.str_("🐼"), pd.NA)), BaseStringArray)
+
+        assert_type(pd.array(UserList(["🐼", np.str_("🐼")])), BaseStringArray)
```
Dr-Irv: can we just use a normal string like `"abc"` here and elsewhere instead of the unusual one?

Also, need to add a test like `pd.array(["abc", "def"])`

Dr-Irv (changes requested):

tests/arrays/test_base_string_array.py
cmp0xff: - 🐼 removed in 1c3d3fb18fcf6bf5a2392586ef079883a109a42b
- new tests added in 5f49ed435b050f41c995fea7f43aed6a37afd31b

pandas-stubs/core/arrays/string_.pyi
```diff
@@ -45,11 +41,11 @@ class StringDtype(ExtensionDtype, Generic[StorageT]):
...
     def na_value(self) -> NAType | float: ...
 
-class BaseStringArray(ExtensionArray):
+class BaseStringArray(ExtensionArray, Generic[StorageT]):
     @property
-    def dtype(self) -> StringDtype: ...
+    def dtype(self) -> StringDtype[StorageT]: ...
 
-class StringArray(BaseStringArray, PandasArray):
+class StringArray(BaseStringArray[Literal["python"]], PandasArray):
```
Dr-Irv, line 40: This points out that we should add `core/arrays/string_arrow.pyi` and then `ArrowStringArray` inherits from `BaseStringArray[Literal["pyarrow"]]`

cmp0xff: Added in fe7a3e2701b5fee659f40f3c08151c150a6fcf4c

tests/arrays/test_arrow_string_array.py
```diff
@@ -0,0 +1,21 @@
...
+def test_construction_dtype_na() -> None:
+    check(
+        assert_type(pd.array([np.nan], "string[pyarrow]"), ArrowStringArray),
+        ArrowStringArray,
+    )
+
+
+def test_dtype() -> None:
+    arr = pd.array(["a"], "string[pyarrow]")
+    check(assert_type(arr.dtype, 'pd.StringDtype[Literal["pyarrow"]]'), pd.StringDtype)
```
Dr-Irv: don't think this is formatted with `black` ?  Usually the double quotes are on the outside, single quotes on the inside

tests/arrays/test_arrow_string_array.py
```diff
@@ -0,0 +1,21 @@
...
+
+def test_construction_dtype_na() -> None:
+    check(
+        assert_type(pd.array([np.nan], "string[pyarrow]"), ArrowStringArray),
+        ArrowStringArray,
+    )
+
+
+def test_dtype() -> None:
+    arr = pd.array(["a"], "string[pyarrow]")
```
Dr-Irv, line 87: can we add a test for constructing a `ArrowStringArray` from `["abc", "def"]`

cmp0xff: e8f6fbe4d803d6cb7514259ab177a91d2cba8c6c

tests/arrays/test_arrow_string_array.py
cmp0xff: I think neither is necessarily better. Imagine we want to search the string `Literal["`, so `Literal['` won't match. Nevertheless e8f6fbe4d803d6cb7514259ab177a91d2cba8c6c

Dr-Irv: Oh, I thought `black` would always reformat it

tests/arrays/test_arrow_string_array.py
```diff
@@ -0,0 +1,89 @@
...
+            ArrowStringArray,
+        )
+        assert_type(
+            pd.array([np.str_("1"), "2"], dtype="string[pyarrow]"), ArrowStringArray
+        )
+
+
+def test_dtype() -> None:
+    arr = pd.array(["a"], "string[pyarrow]")
+    check(assert_type(arr.dtype, "pd.StringDtype[Literal['pyarrow']]"), pd.StringDtype)
```
Dr-Irv, line 88: I'm not a fan of `pd.StringDtype[Literal["pyarrow"]]` as a type, but I don't see an easy way around it without introducing stub-specific classes.  It seems a Generic can't be based on a choice of literal string constants.

cmp0xff: It is indeed lengthy. I hope users won't need to cast or assert types themselves here.

PR #1610, opened 2025-12-30
ENH: make `np_ndarray`'s more generic

cmp0xff:
We can expose the shape type variable in `np_ndarray_xxx` and simplify the stubs.

### DIFF
--- a/tests/_typing.py
+++ b/tests/_typing.py
@@ -79,6 +79,7 @@ from pandas._typing import (
     np_ndarray_int64,
     np_ndarray_intp,
     np_ndarray_num,
+    np_ndarray_object,
     np_ndarray_str,
     np_ndarray_td,
 )
@@ -156,6 +157,7 @@ __all__ = [
     "np_ndarray_int64",
     "np_ndarray_intp",
     "np_ndarray_num",
+    "np_ndarray_object",
     "np_ndarray_str",
     "np_ndarray_td",
 ]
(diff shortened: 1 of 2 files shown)


pandas-stubs/_typing.pyi
```diff
@@ -942,10 +932,24 @@ GenericT_contra = TypeVar(
     "GenericT_contra", bound=np.generic, default=Any, contravariant=True
 )
 NpNumT = TypeVar("NpNumT", bound=np_num, default=np_num)
-ShapeT = TypeVar("ShapeT", bound=tuple[int, ...], default=tuple[Any, ...])
+if sys.version_info >= (3, 11):
+    ShapeT = TypeVar("ShapeT", bound=tuple[int, ...], default=tuple[Any, ...])
+else:
+    ShapeT = TypeVar("ShapeT", bound=tuple[int, ...], default=tuple[int, ...])
```
cmp0xff, line 938: Here might be a subtle change

loicdiridollou: What is requiring that change?

cmp0xff: I think this can explain:

```py
from typing import Any
import numpy as np

assert_type(np.array([1]), np.ndarray[tuple[int, ...], np.dtype[Any]])  # works in 3.10, fails in 3.11+
assert_type(np.array([1]), np.ndarray[tuple[Any, ...], np.dtype[Any]])  # fails in 3.10, works in 3.11+
```

loicdiridollou: gotta love those subtleties, it seemed to me like it was another one of those 3.10/3.11 issues, thanks for confirming!

PR #1611, opened 2025-12-31
CLN: move tests for `Series.astype`

cmp0xff:
Move tests for `Series.astype` to dedicated `test_series_[dtype].py` files, so that `test_series.py` becomes smaller, and further tests for constructors can be added more easily.


cmp0xff:
>     * `tests/series/astype/` then has `test_bool.py`, `test_int.py`, etc.
> 
>     * `tests/series/constructors/` then has files with the same name

This would work if we still have `tests/series/arithmetic`.

Unfortunately contributors have reported that `tests/series/arithmetic/int` has too much hierarchies, and we have moved the structure to `tests/series/int`.

If we now make `tests/series/astype/`, `astype/` will stand in parallel with `int/`, which I find strange..

Issue #1579, opened 2025-12-23
2.0 compatibility tracker


cmp0xff:
The original 2.0 compatibility tracker #624 is very long. Here I list only the unresolved items as of 2025-12-23.

This is the list of things that are in pandas 2.0 release notes that need to be addressed in `pandas-stubs`.  PR's welcome.  If you do a PR, check off the item and put a link to the PR that closed it.  One PR can address multiple issues.

Some of these may already have been taken care of, so if so, check them off and indicate with a comment such as "previously complete"

- [ ] Copy-on-Write can be enabled through one of
  - [ ] pd.set_option("mode.copy_on_write", True)
  - [ ] pd.options.mode.copy_on_write = True
  - [ ] with pd.option_context("mode.copy_on_write", True):
- [ ] Construction with datetime64 or timedelta64 dtype with unsupported resolution (check allowable resolutions for `pd.Series()`)
- [ ] Disallow astype conversion to non-supported datetime64/timedelta64 dtypes
- [ ] The pandas latex options below are no longer used and have been removed. The generic max rows and columns arguments remain but for this functionality should be replaced by the Styler equivalents. The alternative options giving similar functionality are indicated below:
  - [ ] display.latex.escape: replaced with styler.format.escape,
  - [ ] display.latex.longtable: replaced with styler.latex.environment,
  - [ ] display.latex.multicolumn, display.latex.multicolumn_format and display.latex.multirow: replaced with styler.sparse.rows, styler.sparse.columns, styler.latex.multirow_align and styler.latex.multicol_align,
  - [ ] display.latex.repr: replaced with styler.render.repr,
  - [ ] display.max_rows and display.max_columns: replace with styler.render.max_rows, styler.render.max_columns and styler.render.max_elements.
  - [ ] The freq, tz, nanosecond, and unit keywords in the [Timestamp](https://pandas.pydata.org/docs/reference/api/pandas.Timestamp.html#pandas.Timestamp) constructor are now keyword-only
[...]

PR #1612, opened 2025-12-31
BUG: propagate pandas-dev/pandas#41789

cmp0xff:
Towards #1579

Since pandas-dev/pandas#41789, in `DataFrame.xs` and `Series.xs`, `key` can no longer be a `list`.


Dr-Irv (approved):
thanks @cmp0xff

Issue #1587, opened 2025-12-24
2.2 compatibility tracker


cmp0xff:
This is the list of things that are in [pandas 2.2 release notes](https://pandas.pydata.org/docs/dev/whatsnew/v2.2.0.html) that need to be addressed in pandas-stubs. I have removed the sections `Performance improvements` and `Bug fixes`.

PR's welcome. If you do a PR, check off the item and put a link to the PR that closed it. One PR can address multiple issues.

Some of these may already have been taken care of, so if so, check them off and indicate with a comment such as "previously complete"

## Upcoming changes in pandas 3.0

- [ ] Dedicated string data type (backed by Arrow) by default

## Enhancements

- [ ] ADBC Driver support in to_sql and read_sql
- [ ] Create a pandas Series based on one or more conditions
- [ ] Series.struct accessor for PyArrow structured data
- [ ] Series.list accessor for PyArrow list data
- [ ] Calamine engine for `read_excel`
- [ ] `~DataFrame.to_sql` with method parameter set to ``multi`` works with Oracle on the backend
- [ ] :attr:`Series.attrs` / :attr:`DataFrame.attrs` now uses a deepcopy for propagating ``attrs`` (/pandas-dev/pandas#54134).
- [ ] `get_dummies` now returning  extension dtypes ``boolean`` or ``bool[pyarrow]`` that are compatible with the input dtype (/pandas-dev/pandas#56273)
- [ ] `read_csv` now supports ``on_bad_lines`` parameter with ``engine="pyarrow"`` (/pandas-dev/pandas#54480)
- [ ] `read_sas` returns ``datetime64`` dtypes with resolutions better matching those stored natively in SAS, and avoids returning object-dtype in cases that cannot be stored with ``datetime64[ns]`` dtype (/pandas-dev/pandas#56127)
- [ ] `read_spss` now returns a `DataFrame` that stores the metadata in :attr:`DataFrame.attrs` (/pandas-dev/pandas#54264)
- [ ] `tseries.api.guess_datetime_format` is now part of the public API (/pandas-dev/pandas#54727)
- [ ] `DataFrame.apply` now allows the usage of numba (via ``engine="numba"``) to JIT compile the passed function, allowing for potential speedups (/pandas-dev/pandas#54666)
- [ ] `ExtensionArray._explode` interface method added to allow extension type implementations of the ``explode`` method (/pandas-dev/pandas#54833)
[...]

PR #1613, opened 2025-12-31
DEPR: `ravel` in `Index` and `Series` pandas-dev/pandas#36900 pandas-dev/pandas#52511

cmp0xff:
Towards pandas-dev/pandas-stubs#1579 and pandas-dev/pandas-stubs#1587. Propagate pandas-dev/pandas#36900 and pandas-dev/pandas#52511.

- `Index.ravel` returns `Self`
- `Series.ravel` is deprecated


tests/indexes/test_indexes.py
```diff
@@ -1751,3 +1754,39 @@ def test_index_drop() -> None:
...
+        ([1.0, 2.0, 3.0], pd.Index, np.floating),
+        ([1j, 2 + 2j, 3j], pd.Index, np.complexfloating),
+        (pd.RangeIndex(0, 3), pd.RangeIndex, int),
+        ([pd.Timestamp("2022-01-01")], pd.DatetimeIndex, pd.Timestamp),
+        ([pd.Timedelta(days=1)], pd.TimedeltaIndex, pd.Timedelta),
+        ([pd.Period("2022-01", freq="M")], pd.PeriodIndex, pd.Period),
+        (["a", "b", "c"], pd.Index, str),
+    ],
+)
+def test_ravel(
```
loicdiridollou, line 1773: Maybe more a discussion than a request but when we have not too many cases (like here we have 9), wouldn't it be easier to just in line everything since there is already the type checking section below.
Open to suggestions @Dr-Irv @cmp0xff, not sure which is more readable for the maintainer.

cmp0xff: In some cases we'll have tests (check + assert_type) with multiple lines, like

```py
    check(
        assert_type(
            s0.dt.tz_localize(pytz.timezone("US/Eastern")), "pd.Series[pd.Timestamp]"
        ),
        pd.Series,
        pd.Timestamp,
    )
    check(
        assert_type(s0_local.dt.tz_convert("EST"), "pd.Series[pd.Timestamp]"),
        pd.Series,
        pd.Timestamp,
    )
    check(
        assert_type(s0_local.dt.tz_convert(None), "pd.Series[pd.Timestamp]"),
        pd.Series,
        pd.Timestamp,
    )
    check(
        assert_type(s0_local.dt.tz_convert(pytz.UTC), "pd.Series[pd.Timestamp]"),
        pd.Series,
        pd.Timestamp,
    )
    check(
        assert_type(s0_local.dt.tz_convert(1), "pd.Series[pd.Timestamp]"),
        pd.Series,
        pd.Timestamp,
    )
    check(
        assert_type(
            s0_local.dt.tz_convert(pytz.timezone("US/Eastern")),
            "pd.Series[pd.Timestamp]",
        ),
        pd.Series,
        pd.Timestamp,
    )
```

If we do it with `@pytest.mark.parametrize`, separating `check` and `assert_type`, I personally find it easier to read, because we'll have the code more compact with less short lines.

Dr-Irv: My concern is using `parameterize` is that it is easy to have something in the `parameterize` statements that you miss in the `assert_type` statement, or vice versa.

That's why I like the `check(assert_type` approach -  you are testing the type *and* the runtime in one place and you can't make a mistake like that.

PR #1615, opened 2026-01-02
GH1614 Add more agg func methods for pivot_table

### LINKED ISSUE #1614: nunique missing in _PivotAggFunc raises an assignment error for pd.DataFrame.pivot_table
**Describe the bug**
pd.DataFrame.pivot_table does not accept `'nunique'`

**To Reproduce**
```python
my_df.pivot_table(index="index_col", aggfunc="nunique", values="value_col")
```
I use pyright to check my code.
Message:
```
Argument of type "Literal['nunique']" cannot be assigned to parameter "aggfunc" of type "_PivotAggFunc[Unknown] | Sequence[_PivotAggFunc[Unknown]] | Mapping[Hashable, _PivotAggFunc[Unknown]]" in function "pivot_table"
  Type "Literal['nunique']" is not assignable to type "_PivotAggFunc[Unknown] | Sequence[_PivotAggFunc[Unknown]] | Mapping[Hashable, _PivotAggFunc[Unknown]]"
    Type "Literal['nunique']" is not assignable to type "_PivotAggCallable[Unknown]"
    "Literal['nunique']" is not assignable to "ufunc"
    "Literal['nunique']" is not assignable to "Sequence[_PivotAggFunc[Unknown]]"
      Type parameter "_T_co@Sequence" is covariant, but "str" is not a subtype of "_PivotAggFunc[Unknown]"
        Type "str" is not assignable to type "_PivotAggFunc[Unknown]"
          Type "str" is not assignable to type "_PivotAggCallable[Unknown]"
          "str" is not assignable to "ufunc"
```

 - OS: WSL-Ubuntu
 - OS Ubuntu 24.04.1 LTS
 - python version: 3.13.2
 - pyright 1.1.407
 - pandas-stubs 2.3.3.251219

**Additional context**
Proposed solution:
add "nunique" [here](https://github.com/pandas-dev/pandas-stubs/blob/main/pandas-stubs/core/reshape/pivot.pyi#L36C54-L36C60)
  
And thank you for providing the stubs!

loicdiridollou:
- [x] Closes #1614

### DIFF
--- a/pandas-stubs/core/frame.pyi
+++ b/pandas-stubs/core/frame.pyi
@@ -1374,13 +1374,12 @@ class DataFrame(NDFrame, OpsMixin, _GetItemHack):
         index: IndexLabel = ...,
         values: IndexLabel = ...,
     ) -> Self: ...
-    # TODO: pandas-dev/pandas-stubs#1614 improve list of available functions for aggfunc
     def pivot_table(
         self,
         values: _PivotTableValuesTypes = None,
         index: _PivotTableIndexTypes = None,
         columns: _PivotTableColumnsTypes = None,
-        aggfunc: _PivotAggFuncTypes[Scalar] | str = "mean",
+        aggfunc: _PivotAggFuncTypes[Scalar] = "mean",
         fill_value: Scalar | None = None,
         margins: _bool = False,
         dropna: _bool = True,
--- a/pandas-stubs/core/reshape/pivot.pyi
+++ b/pandas-stubs/core/reshape/pivot.pyi
@@ -30,24 +30,23 @@ from pandas._typing import (
 )
 
 _PivotAggCallable: TypeAlias = Callable[[Series], ScalarT]
-_PivotAggFunc: TypeAlias = (
-    _PivotAggCallable[ScalarT]
-    | np.ufunc
-    | Literal[
-        "mean",
-        "sum",
-        "count",
-        "min",
-        "max",
-        "median",
-        "std",
-        "var",
-        "nunique",
-        "ohlc",
-        "quantile",
-        "rank",
-    ]
-)
+_PivotAggFunc: TypeAlias = _PivotAggCallable[ScalarT] | np.ufunc | str
+
+# TODO: pandas-dev/pandas-stubs#1614 improve list of available functions for aggfunc
+# | Literal[
+#     "mean",
+#     "sum",
+#     "count",
+#     "min",
+#     "max",
+#     "median",
+#     "std",
+#     "var",
+#     "nunique",
+#     "ohlc",
+#     "quantile",
+#     "rank",
+# ]
 
 _PivotAggFuncTypes: TypeAlias = (
     _PivotAggFunc[ScalarT]
--- a/tests/frame/test_frame.py
+++ b/tests/frame/test_frame.py
@@ -1523,6 +1523,7 @@ def test_types_pivot() -> None:
 
 
 def test_types_pivot_table() -> None:
+    # TODO: pandas-dev/pandas-stubs#1614 improve tests to cover more aggfunc
     df = pd.DataFrame(
         data={
             "col1": ["first", "second", "third", "fourth"],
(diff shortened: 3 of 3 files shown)


pandas-stubs/core/reshape/pivot.pyi
```diff
@@ -33,7 +33,20 @@ _PivotAggCallable: TypeAlias = Callable[[Series], ScalarT]
...
+        "min",
+        "max",
+        "median",
+        "std",
+        "var",
+        "nunique",
+        "ohlc",
+        "quantile",
+        "rank",
+    ]
```
cmp0xff: - It seems that the allowed str functions is governed by that for `DataFrameGroupBy`. So we can change the two places together.
- The ground truth seems to be in [core/groupby/base.py](https://github.com/pandas-dev/pandas/blob/main/pandas/core/groupby/base.py)
- I've tried a few from there, like "cumprod" and "sem". All works.
- Curiously `pivot_table` with `cumcount` gives a `Series` instead of a `DataFrame`
- The original request was on `DataFrame.pivot_table`, not on `pd.pivot_table`. So the tests are to be added in `test_pandas.py`.
   - I suppose we need to add a new test `test_pivot_table_aggfunc_str`
   - The design pattern with `@pytest.mark.parametrize` could be helpful, since we are adding a lot more tests, and it really reduces line breaks

loicdiridollou: Okay will need a bit of work then, depending how urgent this one is, we can ship this one as is to unblock the issue and I can work on a follow up to clean it all up. Let me know, in any case this will need a bit more love.

c-w-feldmann: From my side it is not urgent, so feel free to take time if you prefer :)

cmp0xff: - Problem in #1614: the `Literal` in `_PivotAggFunc` is too restrictive
- My concern with the current PR: it is still not complete

What if we replace the `Literal` with `str` for now?

- #1614 will be resolved, and my concern is also not there any more
- New problem: it does not tell user what available string functions are there
- It might be fine, since I did not easily found the available string functions in the docs, either

I would say using `str` instead of `Literal` seems to be a good idea to me, if we think there are other higher priorities.

loicdiridollou: Okay I have relaxed the string names to `str`, so should be fine to merge now, please keep #1614 open I will take care of it separately.

cmp0xff: - The original request was on `DataFrame.pivot_table`, not on `pd.pivot_table`. So the tests are to be added in `test_pandas.py`.
   - I suppose we need to add a new test `test_pivot_table_aggfunc_str`

loicdiridollou: Okay I have reverted for pd.pivot_table and only allowed `str` for DataFrame.pivot_table.
I will group all the tests in the subsequent PR, not sure we need it as this stage.

cmp0xff: Sorry, somehow I mistyped twice 😅 Let me correct myself.

- We have to test that `df.pivot_table("value_col", "index_col" aggfunc="nunique")` works. What has been tested in the PR is that `pd.pivot_table(df, "value_col", "index_col" aggfunc="nunique")` works.
   - So we need to add tests in test_**frame**.py.
- Also I prefer the version
   ```suggestion
       | str
   ```
   You can keep the `Literal` commented out, if you want

pandas-stubs/core/frame.pyi
```diff
@@ -1374,12 +1374,13 @@ class DataFrame(NDFrame, OpsMixin, _GetItemHack):
...
         values: IndexLabel = ...,
     ) -> Self: ...
+    # TODO: pandas-dev/pandas-stubs#1614 improve list of available functions for aggfunc
     def pivot_table(
         self,
         values: _PivotTableValuesTypes = None,
         index: _PivotTableIndexTypes = None,
         columns: _PivotTableColumnsTypes = None,
-        aggfunc: _PivotAggFuncTypes[Scalar] = "mean",
+        aggfunc: _PivotAggFuncTypes[Scalar] | str = "mean",
```
cmp0xff:
```suggestion
        aggfunc: _PivotAggFuncTypes[Scalar] = "mean",
```

cmp0xff (changes requested):

pandas-stubs/core/frame.pyi
loicdiridollou: This is fixed, please let me know if this is what you had in mind!

PR #1616, opened 2026-01-02
BUG: rename to `NoDefaultDoNotUse`

cmp0xff:
`_NoDefaultDoNotUse` has been called `NoDefault` since pandas-dev/pandas#47045.

### DIFF
--- a/pandas-stubs/_libs/lib.pyi
+++ b/pandas-stubs/_libs/lib.pyi
@@ -12,7 +12,7 @@ class _NoDefault(Enum):
     no_default = "NO_DEFAULT"
 
 no_default: Final = _NoDefault.no_default
-NoDefault: TypeAlias = Literal[_NoDefault.no_default]
+NoDefaultDoNotUse: TypeAlias = Literal[_NoDefault.no_default]
 
 def infer_dtype(value: object, skipna: bool = True) -> str: ...
 def is_iterator(obj: object) -> bool: ...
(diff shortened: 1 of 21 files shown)


Dr-Irv:
> pending `pyright` strict rule `reportPrivateUsage`, and importing `_NoDefaultDoNotUse` is a part of the violation.

I should mention that I disagree that `pyright` applies that rule to stub files.  We use private things in the stubs to "hide" them from users.  Other stub packages do the same.  See the issue he dismissed at https://github.com/microsoft/pyright/issues/10992

cmp0xff:
Anyway, I have added a test to show that `NoDefaultDoNotUse` cannot be used by users. Importing it leads to `ImportError`. That was what I meant by we could even call it `NoDefault_`, because that is different from what exists in `pandas`, `NoDefault`.

```py
def test_no_default_alias() -> None:
    from pandas._libs.lib import no_default

    assert no_default

    msg = r"cannot import name 'NoDefaultDoNotUse' from 'pandas._libs.lib'"
    with pytest.raises(ImportError, match=msg):
        from pandas._libs.lib import (  # isort: skip
            NoDefaultDoNotUse as NoDefaultDoNotUse,  # noqa: PLC0414
        )
```
If you reopen the PR, you'll see it.

Dr-Irv (approved):
thanks @cmp0xff

Dr-Irv:
> Anyway, I have added a test to show that `NoDefaultDoNotUse` cannot be used by users. Importing it leads to `ImportError`. That was what I meant by we could even call it `NoDefault_`, because that is different from what exists in `pandas`, `NoDefault`.

Sure, but the `DoNotUse` makes it clear that you shouldn't use it!!

> 
> If you reopen the PR, you'll see it.

Done.  Just waiting for the tests to pass to merge.

PR #1617, opened 2026-01-03
ENH: enable `reportPrivateUsage`

cmp0xff:
I made several types of changes that make `reportPrivateUsage` pass.

### DIFF
--- a/pandas-stubs/_libs/tslibs/timestamps.pyi
+++ b/pandas-stubs/_libs/tslibs/timestamps.pyi
@@ -7,7 +7,7 @@ from datetime import (
     tzinfo as _tzinfo,
 )
 
-# The class is private in pandas implementation. We have to ignore the private usage in the stubs.
+# The class is private in python implementation. We have to ignore the private usage in the stubs.
 from datetime import _IsoCalendarDate  # pyright: ignore[reportPrivateUsage]
 import sys
 from time import struct_time


pandas-stubs/_libs/tslibs/period.pyi
```diff
@@ -27,7 +27,7 @@ from pandas._typing import (
 
 class IncompatibleFrequency(ValueError): ...
 
-_PeriodAddSub: TypeAlias = (
+PeriodAddSub: TypeAlias = (
```
cmp0xff: First type: just remove the leading `_`.

This one should be harmless, since it should be fine even if users use it for their typing.

Furthermore, it is in `_libs`, which is private, so users can't easily import it.

pandas-stubs/_libs/indexing.pyi
```diff
@@ -1,4 +1,4 @@
-class _NDFrameIndexerBase:
+class NDFrameIndexerBase:
```
cmp0xff, line 1: Second type: this one is already renamed in `pandas`

pandas-stubs/_libs/interval.pyi
```diff
@@ -19,17 +19,18 @@ from typing_extensions import Self
...
 VALID_CLOSED: frozenset[str]
 
-_OrderableScalarT = TypeVar("_OrderableScalarT", bound=int | float)
-_OrderableTimesT = TypeVar("_OrderableTimesT", bound=Timestamp | Timedelta)
-_OrderableT = TypeVar(
-    "_OrderableT", bound=int | float | Timestamp | Timedelta, default=Any
-)
+_OrderableScalarT = TypeVar("_OrderableScalarT", bound=OrderableScalars)
+_OrderableTimesT = TypeVar("_OrderableTimesT", bound=OrderableTimes)
+_OrderableT = TypeVar("_OrderableT", bound=Orderables, default=Any)
```
cmp0xff: Third type: now I have `OrderableScalars` etc. in `_typing`, and I am creating type variables in each `.pyi` file when needed.

Users should not import from `_typing`, since it's private. So if users do import `OrderableScalars` from `_typing`, they should be prepared for the runtime `ImportError`.

pandas-stubs/core/indexes/base.pyi
```diff
@@ -144,6 +142,9 @@ FloatNotNumpy16DtypeArg: TypeAlias = (
     | PyArrowFloatDtypeArg
 )
 
+_T_contra = TypeVar("_T_contra", contravariant=True)
```
cmp0xff: Type 3, just create `_T_contra` again, instead of importing.

pandas-stubs/core/indexes/base.pyi
```diff
@@ -1217,10 +1218,10 @@ class Index(IndexOpsMixin[S1], ElementOpsMixin[S1]):
     def infer_objects(self, copy: bool = True) -> Self: ...
 
 @type_check_only
-class _IndexSubclassBase(Index[S1], Generic[S1, GenericT_co]):
+class IndexSubclassBase(Index[S1], Generic[S1, GenericT_co]):
```
cmp0xff: Type 1: this is already `@type_check_only`, and does not exist in `pandas`. I am removing the leading `_`.

pandas-stubs/core/reshape/pivot.pyi
```diff
@@ -29,16 +29,16 @@ from pandas._typing import (
...
 )
-_PivotAggFuncTypes: TypeAlias = (
-    _PivotAggFunc[ScalarT]
-    | Sequence[_PivotAggFunc[ScalarT]]
-    | Mapping[Any, _PivotAggFunc[ScalarT]]
+PivotAggFuncTypes: TypeAlias = (
+    PivotAggFunc[ScalarT]
+    | Sequence[PivotAggFunc[ScalarT]]
+    | Mapping[Any, PivotAggFunc[ScalarT]]
 )
```
cmp0xff, line 58: Type 1

pandas-stubs/core/frame.pyi
```diff
@@ -2769,5 +2768,6 @@ class DataFrame(NDFrame, OpsMixin, _GetItemHack):
     @final
     def __bool__(self) -> NoReturn: ...
 
-class _PandasNamedTuple(tuple[Any, ...]):
+@type_check_only
+class PandasNamedTuple(tuple[Any, ...]):
     def __getattr__(self, field: str) -> Scalar: ...
```
cmp0xff, line 2775: Maybe this one is tricky. I remember there were some discussions on this. Is it fine to rename?

tests/extension/decimal/array.py
```diff
@@ -1,3 +1,4 @@
+# pyright: reportPrivateUsage=false
```
cmp0xff, line 1: For testing files, especially for the current `array.py`, I just ignore the rule.

pandas-stubs/core/frame.pyi
```diff
@@ -50,18 +51,16 @@ from pandas.core.indexes.interval import IntervalIndex
...
-    _iAtIndexer,
-    _iLocIndexer,
-    _IndexSliceTuple,
-    _LocIndexer,
-)
+from pandas.core.indexing import _AtIndexer  # pyright: ignore[reportPrivateUsage]
+from pandas.core.indexing import _IndexSliceTuple  # pyright: ignore[reportPrivateUsage]
+from pandas.core.indexing import _LocIndexer  # pyright: ignore[reportPrivateUsage]
+from pandas.core.indexing import _iAtIndexer  # pyright: ignore[reportPrivateUsage]
+from pandas.core.indexing import _iLocIndexer  # pyright: ignore[reportPrivateUsage]
```
cmp0xff, line 66: Fourth type: I cannot rename classes that exist in `pandas` with a leading `_`, so I just use `pyright: ignore` on them.

pandas-stubs/_libs/tslibs/period.pyi
Dr-Irv: Maybe we should do this for the stubs.  

Create a file (or maybe a sub-package)  `_private_for_stubs` (or something like that), where we put ALL the private symbols like this, removing the underscores, i.e., it contains any symbols that we created for use in the stubs package.

Because if `pyright` doesn't complain about importing `pandas._libs.tslib.period.PeriodAddSub`, it won't complain about importing `pandas._private_for_stubs.PeriodAddSub`

Then we know that any stub-private symbols are all in one place and using a different name means that people wouldn't import it in their code.

pandas-stubs/core/indexes/base.pyi
Dr-Irv: So we could move it to the "private" idea I suggested above.

pandas-stubs/core/indexes/interval.pyi
```diff
@@ -63,6 +65,10 @@ _EdgesTimedelta: TypeAlias = (
 _TimestampLike: TypeAlias = pd.Timestamp | np.datetime64 | dt.datetime
 _TimedeltaLike: TypeAlias = pd.Timedelta | np.timedelta64 | dt.timedelta
 
+_OrderableScalarT = TypeVar("_OrderableScalarT", bound=OrderableScalars)
+_OrderableTimesT = TypeVar("_OrderableTimesT", bound=OrderableTimes)
+_OrderableT = TypeVar("_OrderableT", bound=Orderables, default=Any)
```
Dr-Irv: Not a fan of repeating this across files.  But if we put it (without the leading underscore) in the "private" file, then it is defined in one place.

pandas-stubs/core/reshape/pivot.pyi
Dr-Irv: could move to the "private" file

pandas-stubs/core/reshape/pivot.pyi
```diff
@@ -67,22 +67,20 @@ _NonIterableHashable: TypeAlias = (
...
 
-_PivotTableIndexTypes: TypeAlias = Label | Sequence[Hashable] | Series | Grouper | None
-_PivotTableColumnsTypes: TypeAlias = (
-    Label | Sequence[Hashable] | Series | Grouper | None
-)
-_PivotTableValuesTypes: TypeAlias = Label | Sequence[Hashable] | None
+PivotTableIndexTypes: TypeAlias = Label | Sequence[Hashable] | Series | Grouper | None
+PivotTableColumnsTypes: TypeAlias = Label | Sequence[Hashable] | Series | Grouper | None
+PivotTableValuesTypes: TypeAlias = Label | Sequence[Hashable] | None
```
Dr-Irv: could move to the "private" file

pandas-stubs/core/base.pyi
```diff
@@ -52,6 +51,7 @@ from pandas._typing import (
     np_ndarray_td,
 )
 
+_T_contra = TypeVar("_T_contra", contravariant=True)
```
Dr-Irv: Or we create `T_contra` in the "private" place and import from there to avoid multiple repeats of the same thing

pandas-stubs/core/frame.pyi
Dr-Irv: agree

pandas-stubs/core/frame.pyi
Dr-Irv: Yes, and we can move it to the "private" file

pandas-stubs/core/series.pyi
```diff
@@ -250,6 +245,10 @@ from pandas.core.dtypes.dtypes import CategoricalDtype
 
 from pandas.plotting import PlotAccessor
 
+_T_co = TypeVar("_T_co", covariant=True)
+_T_contra = TypeVar("_T_contra", contravariant=True)
+_OrderableT = TypeVar("_OrderableT", bound=Orderables, default=Any)
```
Dr-Irv: rename without the leading underscore and put in "private" file.

pandas-stubs/plotting/_core.pyi
```diff
@@ -31,7 +32,8 @@ from pandas._typing import (
     np_ndarray_object,
 )
 
-class _BoxPlotT(NamedTuple):
+@type_check_only
+class BoxPlotT(NamedTuple):
```
Dr-Irv, line 36: move to private file

pandas-stubs/_typing.pyi
```diff
@@ -919,6 +919,9 @@ PyArrowNotStrDtypeArg: TypeAlias = (
 )
 
 StrLike: TypeAlias = str | np.str_
+OrderableScalars: TypeAlias = int | float
+OrderableTimes: TypeAlias = Timestamp | Timedelta
+Orderables: TypeAlias = OrderableScalars | OrderableTimes
```
Dr-Irv: can move to private file

tests/test_groupby.py
```diff
@@ -39,7 +39,9 @@
 )
 
 if TYPE_CHECKING:
-    from pandas.core.groupby.groupby import _ResamplerGroupBy  # noqa: F401
+    from pandas.core.groupby.groupby import (  # noqa: F401 # isort: skip
+        _ResamplerGroupBy,  # pyright: ignore[reportPrivateUsage]
```
Dr-Irv: Maybe we should put this one without underscore in our "private" file instead?  Then it's clear we are doing a test on a private type (and we'll still need the `ignore`)

Dr-Irv (changes requested):
I suggested in my first comment (and now that I looked at the whole PR), that we create a "private" stubs file (leading underscore), and put any of the types that we created just for typing in there.  They would have no leading underscores.  I'd suggest a trailing underscore as a reminder that they are there for typing purposes.

Then we don't keep repeating code everywhere just to deal with how pyright is handling things.

Fortunately, it doesn't complain when you import from a private package, just private symbols.

pandas-stubs/core/frame.pyi
cmp0xff: This one cannot live in the private file. We also use it in static type checking in tests. There, when I import anything that does not have an underlying `.py` file in `pandas`, I get something like `Import "pandas._stubs_only" could not be resolved from source ` from `pyright`.

pandas-stubs/core/series.pyi
cmp0xff: I think recreating type variables might be a good practice. We have created numbered type variables, e.g. S2, S3, to avoid "overlapping". However we can also just create distinct, new type variables in the scripts when we need to use them, so that they are not imported and have different "scope".

What would you say?

pandas-stubs/plotting/_core.pyi
cmp0xff: The same problem as `PandasNamedTuple`

tests/test_groupby.py
cmp0xff: The same problem as `PandasNamedTuple`

pandas-stubs/core/frame.pyi
Dr-Irv: Agreed that this one can't live in the private file.  It needs to be a retuned result from our stubs.

pandas-stubs/core/series.pyi
Dr-Irv: The overlapping was done because we had to refer to multiple `TypeVar` in the same method declaration.

I'd like to avoid this kind of repeated code, and it gives us one place to look at them.

pandas-stubs/plotting/_core.pyi
Dr-Irv: Agreed.  Now that I read the docs and look at the test.

tests/test_groupby.py
Dr-Irv: OK, but I think we should just remove the underscore and leave it where it is.

pandas-stubs/core/frame.pyi
```diff
@@ -50,19 +57,11 @@ from pandas.core.indexes.interval import IntervalIndex
...
-    _PivotAggFuncTypes,
-    _PivotTableColumnsTypes,
-    _PivotTableIndexTypes,
-    _PivotTableValuesTypes,
-)
+from pandas.core.indexing import _AtIndexer  # pyright: ignore[reportPrivateUsage]
+from pandas.core.indexing import _IndexSliceTuple  # pyright: ignore[reportPrivateUsage]
+from pandas.core.indexing import _LocIndexer  # pyright: ignore[reportPrivateUsage]
+from pandas.core.indexing import _iAtIndexer  # pyright: ignore[reportPrivateUsage]
+from pandas.core.indexing import _iLocIndexer  # pyright: ignore[reportPrivateUsage]
```
Dr-Irv, line 66: Can you add a comment here indicating why the ignores are needed (you said in the conversation here it's because they exist in pandas)

pandas-stubs/core/series.pyi
```diff
@@ -106,13 +110,11 @@ from pandas.core.indexes.interval import IntervalIndex
...
-    _iAtIndexer,
-    _iLocIndexer,
-    _IndexSliceTuple,
-    _LocIndexer,
-)
+from pandas.core.indexing import _AtIndexer  # pyright: ignore[reportPrivateUsage]
+from pandas.core.indexing import _IndexSliceTuple  # pyright: ignore[reportPrivateUsage]
+from pandas.core.indexing import _LocIndexer  # pyright: ignore[reportPrivateUsage]
+from pandas.core.indexing import _iAtIndexer  # pyright: ignore[reportPrivateUsage]
+from pandas.core.indexing import _iLocIndexer  # pyright: ignore[reportPrivateUsage]
```
Dr-Irv, line 119: add comment here too why ignores are needed

pandas-stubs/io/formats/style_render.pyi
```diff
@@ -16,7 +16,7 @@ from jinja2.environment import (
 )
 from jinja2.loaders import PackageLoader
 from pandas import Index
-from pandas.core.indexing import _IndexSlice
+from pandas.core.indexing import _IndexSlice  # pyright: ignore[reportPrivateUsage]
```
Dr-Irv, line 21: and add comment here too

pandas-stubs/core/frame.pyi
cmp0xff: 44d91fff0b691b96dc85247b6bdd6dc0d4fcd305

pandas-stubs/core/series.pyi
cmp0xff: 44d91fff0b691b96dc85247b6bdd6dc0d4fcd305

pandas-stubs/io/formats/style_render.pyi
cmp0xff: 44d91fff0b691b96dc85247b6bdd6dc0d4fcd305

Dr-Irv (approved):
thanks @cmp0xff

PR #1619, opened 2026-01-03
TYP: construction for `DatetimeArray`

cmp0xff:
Towards #1469.

- Fixed inheritance of several classes
- Added typing for `DatetimeArray`
- Added tests for `pd.array` that gives `DatetimeArray`

### DIFF
--- a/tests/frame/test_frame.py
+++ b/tests/frame/test_frame.py
@@ -1779,6 +1779,7 @@ def test_pivot_table_aggfunc_numpy_ufunc(sample_df: pd.DataFrame) -> None:
         FutureWarning,
         r"The provided callable.*",
         lower="2.3.0",
+        upper="2.99",
     ):
         check(
             assert_type(
@@ -1793,6 +1794,7 @@ def test_pivot_table_aggfunc_numpy_ufunc(sample_df: pd.DataFrame) -> None:
         FutureWarning,
         r"The provided callable.*",
         lower="2.3.0",
+        upper="2.99",
     ):
         check(
             assert_type(
@@ -1807,6 +1809,7 @@ def test_pivot_table_aggfunc_numpy_ufunc(sample_df: pd.DataFrame) -> None:
         FutureWarning,
         r"The provided callable.*",
         lower="2.3.0",
+        upper="2.99",
     ):
         check(
             assert_type(
@@ -1822,6 +1825,7 @@ def test_pivot_table_aggfunc_numpy_ufunc(sample_df: pd.DataFrame) -> None:
         FutureWarning,
         r"The provided callable.*",
         lower="2.3.0",
+        upper="2.99",
     ):
         check(
             assert_type(
@@ -1836,6 +1840,7 @@ def test_pivot_table_aggfunc_numpy_ufunc(sample_df: pd.DataFrame) -> None:
         FutureWarning,
         r"The provided callable.*",
         lower="2.3.0",
+        upper="2.99",
     ):
         check(
             assert_type(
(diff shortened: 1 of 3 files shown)


tests/arrays/test_datetime_array.py
```diff
@@ -0,0 +1,225 @@
...
+        )
+        assert_type(
+            pd.array([datetime(2052, 1, 5), pd.Timestamp(2, 1, 6)]), DatetimeArray
+        )
+
+        assert_type(pd.array([datetime(2061, 1, 5, 1), None]), DatetimeArray)
+        assert_type(pd.array([pd.Timestamp(1902, 1, 5, 3), None]), DatetimeArray)
+
+        assert_type(pd.array([datetime(1921, 1, 5, 1), pd.NaT]), DatetimeArray)  # type: ignore[assert-type]
+        assert_type(pd.array([pd.Timestamp(1872, 1, 5, 3), pd.NaT]), DatetimeArray)  # type: ignore[assert-type]
```
cmp0xff, line 86: Mypy did not recognise these two. I've already removed `SequenceNotStr` in `construction.pyi` in this PR and fixed some errors, but I don't know how to fix further.

pandas-stubs/core/arrays/datetimes.pyi
```diff
@@ -64,34 +63,55 @@ class DatetimeArray(DatetimeLikeArrayMixin, TimelikeOps, DatelikeOps):
...
-    is_leap_year = ...
+    def day_name(self, locale: str | None = None) -> np_1darray_object: ...
+    @property
+    def time(self) -> np_1darray_object: ...
+    @property
+    def timetz(self) -> np_1darray_object: ...
+    @property
+    def date(self) -> np_1darray_object: ...
+    @property
+    def year(self) -> np_1darray_int32: ...
```
loicdiridollou, line 76: I think they may be int64 here (see in #1621)

tests/arrays/test_datetime_array.py
loicdiridollou: Do we have the correct overloads in pandas.core.construction? I don't see any that goes to DatetimeArray, maybe we need one like:
```python
@overload
 def array(  # type: ignore[overload-overlap] # pyright: ignore[reportOverlappingOverload]
    data: Sequence[pd.Timestamp | NAType | NaTType | None],
    dtype: DtypeArg | None = None,
    copy: bool = True,
) -> DatetimeArray: ...
```

cmp0xff: At the moment we have
```py
@overload
def array(
    data: (
        tuple[_NaTDatetimeElement, ...]
        | MutableSequence[_NaTDatetimeElement]
        | np_ndarray
        | DatetimeArray
    ),
    dtype: (
        DatetimeTZDtype
        | PandasTimestampDtypeArg
        | np.dtype[np.datetime64]
        | NumpyTimestampDtypeArg
    ),
    copy: bool = True,
) -> DatetimeArray: ...
@overload
def array(  # type: ignore[overload-overlap]
    data: (  # TODO: merge the two Sequence's after 3.0 pandas-dev/pandas#57064
        Sequence[datetime | NaTType | None]
        | Sequence[np.datetime64 | NaTType | None]
        | np_ndarray_dt
        | DatetimeArray
    ),
    dtype: None = None,
    copy: bool = True,
) -> DatetimeArray: ...
```

pandas-stubs/core/arrays/datetimes.pyi
cmp0xff: Tests added 864007eeab2af9e2e764b6585c6f6a80d8bb6371

tests/arrays/test_datetime_array.py
loicdiridollou: I think once you have a pd.NA and a pd.NaT pandas at constrcution will move the pd.NA into pd.NaT so the bottom one we may need to have the pd.NA added to it

loicdiridollou: I think it is an edge case anyway so not sure that should be a good practice to mix NA and NaT together.

PR #1620, opened 2026-01-04
BUG: regression on #1594 for nightly

cmp0xff:
- `pd.array([], str)` gives `NumpyExtensionArray` before 3.0, but gives `BaseStringArray` after 3.0. Fixing.
- `pyarrow` is required in 3.0 (pandas-dev/pandas#54466). Adding notes about results giving `ArrowStringArray`.
- In 3.0, `pd.array([], str).storage.na_value` is `float("nan")`, instead of `pd.NA`. Adding notes and reporting in pandas-dev/pandas#63567.


Dr-Irv:
> * `pyarrow` is required in 3.0 ([FEEDBACK: PyArrow as a required dependency and PyArrow backed strings pandas#54466](https://github.com/pandas-dev/pandas/issues/54466)). Adding notes about results giving `ArrowStringArray`.

This is NOT true.  `pyarrow` is not required in 3.0.  If installed, it will get used.  If not, it won't get used.

I don't know if it will affect this PR, but please adjust accordingly. 

I don't think we need to worry about supporting type checking without arrow installed, as there is no way for us to test it in the stubs.  But we can test it in the tests if we want.

PR #1625, opened 2026-01-07
GH1570 Fix nightly

### LINKED ISSUE #1570: Failure in nightly for pandas 3.0
See https://github.com/pandas-dev/pandas-stubs/actions/runs/20354692300/job/58487440040

loicdiridollou:
- [x] Closes #1570

Fix the nightly because of FutureWarning not being raised with 3.0.0, NumpyArray vs DatetimeArray in some tests with 3.0.0 as well as an old #1570 since it was fixed on pandas side.

### DIFF
--- a/tests/_typing.py
+++ b/tests/_typing.py
@@ -1,6 +1,5 @@
 # This file serves as a stub file for static type checkers
 # (pyright does not like it if I call the file tests/_typing.pyi).
-# It can only import from pandas._typing.
 # By tests/conftest.py, pytest copies this file to tests/_typing.pyi, and
 # pandas-stubs/_typing.pyi to tests/_typing.py.
 # After tests are done, pytest copies tests/_typing.pyi back to


tests/arrays/test_numpy_extension_array.py
```diff
@@ -40,7 +39,7 @@ def test_construction_sequence(
 ) -> None:
     # `pd.NaT in [pd.NA, pd.NaT]` leads to an exception
     if missing_values[-1] is pd.NaT:
-        expected_type = DatetimeArray if PD_LTE_23 else NumpyExtensionArray
+        expected_type = DatetimeArray
```
cmp0xff, line 43: If the behaviour is like this, they do not belong to tests for `NumpyExtensionArray` at all. we need to remove the relevant tests and typing in `pd.array`.

tests/frame/test_indexing.py
```diff
@@ -430,18 +429,17 @@ def test_frame_setitem_na() -> None:
     df.at["a", "y"] = None
     df.iat[0, 0] = None
 
-    if PD_LTE_23:
-        # TODO: pandas-dev/pandas#63420, this is failing on latest build, should work
-        df.loc[:, "x"] = [None, pd.NA, pd.NaT]
-        df.iloc[:, 0] = [None, pd.NA, pd.NaT]
+    # TODO: pandas-dev/pandas#63420, this is failing on latest build, should work
```
cmp0xff:
```suggestion
```

cmp0xff (changes requested):

tests/arrays/test_numpy_extension_array.py
loicdiridollou: The test is parametrize so it will take a bit of work to clean it up, maybe in this version the PR is good, I will add a todo to clean it up and we can ship this. What do you think?

tests/arrays/test_numpy_extension_array.py
```diff
@@ -40,7 +39,8 @@ def test_construction_sequence(
 ) -> None:
     # `pd.NaT in [pd.NA, pd.NaT]` leads to an exception
     if missing_values[-1] is pd.NaT:
-        expected_type = DatetimeArray if PD_LTE_23 else NumpyExtensionArray
+        # TODO: pandas-dev/pandas-stubs#1589 clean up this and move to the right test file
```
cmp0xff, line 42: We can also remove the statement about "exception". I don't see the exception here.
```suggestion
    if missing_values[-1] is pd.NaT:
        # TODO: pandas-dev/pandas-stubs#1634 clean up this and move to the right test file
```

cmp0xff (changes requested):
Just one more place to clean up.

tests/arrays/test_numpy_extension_array.py
cmp0xff: This is fine, we can fix it in #1619.

tests/_typing.py
cmp0xff, line 1: Please revert

loicdiridollou: Yes not sure where it drifted but this is now fixed.

PR #1630, opened 2026-01-13
Bump ty

Dr-Irv:
Supports ty 0.0.11

Thought I was fixing the numpy issue, but that was already addressed!

`ty` and `mypy` have disagreements on various overload override messages, so now we separate `mypy` ignores and `ty` ignores.

Edits to `pyproject.toml` also did some reformatting - not exactly sure why...


cmp0xff (approved):
Looks good to me. Thank you @Dr-Irv !

PR #1636, opened 2026-01-18
MNT: regression on #1620 regarding `PyArrow`

cmp0xff:
`PyArrow` will not become a mandatory depdendency in 3.0. Several comments are to be removed.


Dr-Irv (approved):
thanks @cmp0xff

PR #1638, opened 2026-01-19
TYP: Accept broader iterables in MultiIndex.from_product

### LINKED ISSUE #1637: `MutliIndex.from_product` should accept more general types
```python
import pandas as pd

pd.MultiIndex.from_product([{1,2,3}, {"a", "b", "c"}])
```

with `pyrefly`:

```
ERROR Argument `list[set[int] | set[str]]` is not assignable to parameter `iterables` with type `Sequence[Index[Any] | SequenceNotStr[Hashable] | Series[Any] | range]` in function `pandas.core.indexes.multi.MultiIndex.from_product` [bad-argument-type]
```

The stubs seem out of sync:

https://github.com/pandas-dev/pandas-stubs/blob/96328d3294957eac96a3b7ecfa5c7474b63e7354/pandas-stubs/core/indexes/multi.pyi#L66-L72

On the other hand, [pandas itself defines](https://github.com/pandas-dev/pandas/blob/d51494eeae96b4118057153c70555c31731605f0/pandas/core/indexes/multi.py#L593-L613)

```python
    @classmethod
    def from_product(
        cls,
        iterables: Sequence[Iterable[Hashable]],
        sortorder: int | None = None,
        names: Sequence[Hashable] | Hashable | lib.NoDefault = lib.no_default,
    ) -> MultiIndex:
        """
        Make a MultiIndex from the cartesian product of multiple iterables.

        Parameters
        ----------
        iterables : list / sequence of iterables
            Each iterable has unique labels for each level of the index.
        sortorder : int or None
            Level of sortedness (must be lexicographically sorted by that
            level).
        names : list / sequence of str, optional
            Names for the levels in the index.
            If not explicitly provided, names will be inferred from the
            elements of iterables if an element has a name attribute.
        """
```

randolf-scholz:
Change iterables parameter from Sequence[SequenceNotStr[Hashable] | pd.Series | pd.Index | range] to Sequence[Iterable[Hashable]] to accept sets and other iterables, matching pandas implementation.

- [x] Closes #1637 (Replace xxxx with the Github issue number)
- [x] If I used AI to develop this pull request, I prompted it to follow `AGENTS.md`

### DIFF
--- a/pandas-stubs/core/indexes/multi.pyi
+++ b/pandas-stubs/core/indexes/multi.pyi
@@ -14,7 +14,10 @@ from typing import (
 import numpy as np
 import pandas as pd
 from pandas.core.indexes.base import Index
-from typing_extensions import Self
+from typing_extensions import (
+    Never,
+    Self,
+)
 
 from pandas._typing import (
     AnyAll,
@@ -63,6 +66,15 @@ class MultiIndex(Index):
         sortorder: int | None = None,
         names: SequenceNotStr[Hashable] | None = None,
     ) -> Self: ...
+    @overload
+    @classmethod
+    def from_product(
+        cls,
+        iterables: Sequence[str],
+        sortorder: int | None = None,
+        names: SequenceNotStr[Hashable] | None = None,
+    ) -> Never: ...
+    @overload
     @classmethod
     def from_product(
         cls,
--- a/tests/indexes/test_indexes.py
+++ b/tests/indexes/test_indexes.py
@@ -1417,6 +1417,25 @@ def test_multiindex_from_product_with_sets() -> None:
     check(assert_type(midx, pd.MultiIndex), pd.MultiIndex)
 
 
+def test_multiindex_from_product_with_iterables() -> None:
+    """Test using genuinely iterable inputs in `MultiIndex.from_product` GH1637."""
+    # Test with generator expressions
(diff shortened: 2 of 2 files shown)


pandas-stubs/core/indexes/multi.pyi
```diff
@@ -66,7 +66,7 @@ class MultiIndex(Index):
     @classmethod
     def from_product(
         cls,
-        iterables: Sequence[SequenceNotStr[Hashable] | pd.Series | pd.Index | range],
+        iterables: Sequence[Iterable[Hashable]],
```
cmp0xff, line 81: The current code allows `pd.MultiIndex.from_product(["12", "34"])`, which fails at runtime. We need to forbid such cases.

I propose to add an overload with `iterables: Sequence[str]`, returning `Never`, and add a test block in `if TYPE_CHECKING_INVALID_USAGE` with `assert_type(pd.MultiIndex.from_product(["12", "34"]), Never)`. You'll need to use an unused function to protect the test, see `def _0() -> None:  # pyright: ignore[reportUnusedFunction]` elsewhere.

cmp0xff: And please also add tests for genuinely iterable inputs. Currently only sequences and collections are tested.

PR #1639, opened 2026-01-20
TYP: enable `reportUnknownMemberType`

### LINKED ISSUE #1171: Strict report as of 3/13/2025

cmp0xff:
- [x] Closes #1171

We can also optionally remove `pyright_strict`.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -123,10 +123,6 @@ script = "scripts.test.run:pyrefly"
 help = "Run pyright on 'tests' (using the local stubs) and on the local stubs"
 script = "scripts.test.run:pyright_src"
 
-[tool.poe.tasks.pyright_strict]
-help = "Run pyright on 'tests' (using the local stubs) and on the local stubs in full strict mode"
-script = "scripts.test.run:pyright_src_strict"
-
 [tool.poe.tasks.pyright_dist]
 help = "Run pyright on 'tests' using the installed stubs"
 script = "scripts.test:run_tests(dist=True, type_checker='pyright')"
--- a/pyrightconfig-strict.json
+++ b/pyrightconfig-strict.json
@@ -1,8 +1,0 @@
-{
-  "typeCheckingMode": "strict",
-  "stubPath": ".",
-  "include": ["tests", "pandas-stubs"],
-  "enableTypeIgnoreComments": false,
-  "reportUnnecessaryTypeIgnoreComment": true,
-  "reportMissingModuleSource": true
-}
--- a/.github/workflows/comment_commands.yml
+++ b/.github/workflows/comment_commands.yml
@@ -8,9 +8,9 @@ permissions:
 
 env:
   # store mapping of commands to use with poetry
-  RUN_COMMAND: '{"/pandas_nightly": "pytest --nightly", "/pyright_strict": "pyright_strict", "/mypy_nightly": "mypy --mypy_nightly"}'
+  RUN_COMMAND: '{"/pandas_nightly": "pytest --nightly", "/mypy_nightly": "mypy --mypy_nightly"}'
   # store mapping of labels to display in the check runs
-  DISPLAY_COMMAND: '{"/pandas_nightly": "Pandas nightly tests", "/pyright_strict": "Pyright strict tests", "/mypy_nightly": "Mypy nightly tests"}'
+  DISPLAY_COMMAND: '{"/pandas_nightly": "Pandas nightly tests", "/mypy_nightly": "Mypy nightly tests"}'
 
 jobs:
   optional_tests:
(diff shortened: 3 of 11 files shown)


pyproject.toml
```diff
@@ -300,12 +300,10 @@ typeCheckingMode = "strict"
 stubPath = "."
 include = ["tests", "pandas-stubs"]
 enableTypeIgnoreComments = false    # use pyright-specific ignores
-# disable subset of strict
 reportUnnecessaryTypeIgnoreComment = true
-reportUnknownMemberType = false
 # enable optional checks
 reportMissingModuleSource = true
-useLibraryCodeForTypes = false
+useLibraryCodeForTypes = true
```
cmp0xff: Note I am enabling `useLibraryCodeForTypes`. This was disabled in pandas-dev/pandas-stubs#687 because of `matplotlib`. When I enable it, only two `# pyright: ignore[reportUnknownMemberType]` upon `plt.subplots` are added. I think it is fine.

tests/frame/test_frame.py
```diff
@@ -1,4 +1,4 @@
-# pyright: reportUnknownArgumentType=false,reportUnknownLambdaType=false
+# pyright: reportUnknownArgumentType=false,reportUnknownLambdaType=false,reportUnknownMemberType=false
```
Dr-Irv: Can we split this file into 2 parts, one where this is not needed, and the other where it is needed?  So then we know which tests require us to relax this setting.

tests/frame/test_indexing.py
```diff
@@ -1,4 +1,4 @@
-# pyright: reportUnknownLambdaType=false
+# pyright: reportUnknownLambdaType=false,reportUnknownMemberType=false
```
Dr-Irv: same comment as in `test_frame`

tests/series/test_series.py
```diff
@@ -1,4 +1,4 @@
-# pyright: reportUnknownArgumentType=false,reportUnknownLambdaType=false
+# pyright: reportUnknownArgumentType=false,reportUnknownLambdaType=false,reportUnknownMemberType=false
```
Dr-Irv: same comment as in `test_frame`

tests/test_plotting.py
```diff
@@ -383,7 +383,7 @@ def test_scatter_matrix(close_figures: None) -> None:
 
 def test_table(close_figures: None) -> None:
     df = pd.DataFrame(np.random.randn(1000, 4), columns=["A", "B", "C", "D"])
-    _, ax = plt.subplots(1, 1)
+    _, ax = plt.subplots(1, 1)  # pyright: ignore[reportUnknownMemberType]
```
Dr-Irv, line 387: can you add a comment here (and below) that this is due to `matplotlib` not having `subplots` defined?

I'd also be curious as to which member is unknown?  Is it something missing in the `matplotlib` code that we can report to them?

pyproject.toml
Dr-Irv: add a comment here that we actually don't need this in `pyproject.toml` because it is the default. (or you could delete the line)

pyrightconfig-strict.json
```diff
@@ -5,6 +5,6 @@
   "enableTypeIgnoreComments": false,
   "reportUnnecessaryTypeIgnoreComment": true,
   "reportMissingModuleSource": true,
-  "useLibraryCodeForTypes": false,
+  "useLibraryCodeForTypes": true,
```
Dr-Irv: can delete this - it's the default

cmp0xff: 240158e0f2880028d9efdfeca213ef43730034ea

pyproject.toml
cmp0xff: 240158e0f2880028d9efdfeca213ef43730034ea

tests/test_plotting.py
cmp0xff: 8fe3351296c6157308db6edae2a506951c498bdd

tests/series/test_series.py
cmp0xff: 7985c9b0b8ec0dedca3fae36a1e539a386bdee27

tests/frame/test_indexing.py
cmp0xff: 7985c9b0b8ec0dedca3fae36a1e539a386bdee27

tests/frame/test_frame.py
cmp0xff: 7985c9b0b8ec0dedca3fae36a1e539a386bdee27

tests/test_plotting.py
Dr-Irv: I created https://github.com/matplotlib/matplotlib/issues/31003

PR #1643, opened 2026-01-22
GH1641 Pandas 3.0 support

loicdiridollou:

### DIFF
--- a/tests/test_timefuncs.py
+++ b/tests/test_timefuncs.py
@@ -1867,8 +1867,12 @@ def test_timestamp_to_list_add() -> None:
     tslist = list(pd.to_datetime(["2022-01-01", "2022-01-02"]))
     check(assert_type(tslist, list[pd.Timestamp]), list, pd.Timestamp)
     sseries = pd.Series(tslist)
-    with pytest_warns_bounded(Pandas4Warning, "'d' is deprecated", lower="2.3.99"):
-        _0 = sseries + pd.Timedelta(1, "d")
+    with pytest_warns_bounded(Pandas4Warning, "'d' is deprecated", lower="2.99"):
+        check(
+            assert_type(sseries + pd.Timedelta(1, "d"), "pd.Series[pd.Timestamp]"),
+            pd.Series,
+            pd.Timestamp,
+        )
 
     check(
         assert_type(sseries + pd.Timedelta(1, "D"), "pd.Series[pd.Timestamp]"),
--- a/tests/__init__.py
+++ b/tests/__init__.py
@@ -38,7 +38,6 @@ TYPE_CHECKING_INVALID_USAGE: Final = TYPE_CHECKING
 LINUX = sys.platform == "linux"
 WINDOWS = sys.platform in {"win32", "cygwin"}
 MAC = sys.platform == "darwin"
-PD_LTE_23 = Version(pd.__version__) < Version("2.3.999")
 
 
 def check(


pandas-stubs/core/frame.pyi
```diff
@@ -889,20 +889,6 @@ class DataFrame(NDFrame, OpsMixin, _GetItemHack):
...
-        mapper: Renamer | None = ...,
-        *,
-        index: Renamer | None = ...,
-        columns: Renamer | None = ...,
-        axis: Axis | None = ...,
-        copy: bool = ...,
-        inplace: Literal[True],
-        level: Level | None = None,
-        errors: IgnoreRaise = ...,
-    ) -> None: ...
```
Dr-Irv, line 910: I don't think you can remove this

pandas-stubs/core/frame.pyi
```diff
@@ -911,46 +897,24 @@ class DataFrame(NDFrame, OpsMixin, _GetItemHack):
         columns: Renamer | None = ...,
         axis: Axis | None = ...,
         copy: bool = ...,
-        inplace: Literal[False] = False,
+        inplace: bool = False,
```
Dr-Irv: have to switch this back.  

You could remove the `copy` argument - it's deprecated in 3.0

pandas-stubs/core/frame.pyi
```diff
@@ -994,27 +958,14 @@ class DataFrame(NDFrame, OpsMixin, _GetItemHack):
...
-        self,
-        level: Level | Sequence[Level] = ...,
-        *,
-        drop: _bool = ...,
-        col_level: int | _str = ...,
-        col_fill: Hashable = ...,
-        inplace: Literal[True],
-        allow_duplicates: _bool = ...,
-        names: Hashable | Sequence[Hashable] = ...,
-    ) -> None: ...
```
Dr-Irv, line 1014: cannot remove this `inplace=True` returns `None`

pandas-stubs/core/frame.pyi
```diff
@@ -994,27 +958,14 @@ class DataFrame(NDFrame, OpsMixin, _GetItemHack):
...
-    @overload
     def reset_index(
         self,
         level: Level | Sequence[Level] = ...,
         *,
         col_level: int | _str = ...,
         col_fill: Hashable = ...,
         drop: _bool = ...,
-        inplace: Literal[False] = False,
+        inplace: bool = False,
```
cmp0xff: [inplace=True still returns None](https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.DataFrame.reset_index.html). Please fix and also add a test.

pandas-stubs/core/frame.pyi
```diff
@@ -1373,20 +1324,10 @@ class DataFrame(NDFrame, OpsMixin, _GetItemHack):
...
-    ) -> Self | Series: ...
-    @overload
     def stack(
         self,
         level: IndexLabel = ...,
-        dropna: _bool = ...,
-        sort: _bool = ...,
-        future_stack: Literal[False] = False,
+        future_stack: Literal[True] = True,
     ) -> Self | Series: ...
```
cmp0xff, line 1396: [Both implementations still coexist](https://pandas.pydata.org/docs/dev/reference/api/pandas.DataFrame.stack.html). I guess we should keep the old implementation.

tests/arrays/test_arrow_string_array.py
```diff
@@ -107,8 +106,7 @@ def test_dtype() -> None:
     check(assert_type(arr.dtype, "pd.StringDtype[Literal['pyarrow']]"), pd.StringDtype)
     assert assert_type(arr.dtype.storage, Literal["pyarrow"]) == "pyarrow"
 
-    if not PD_LTE_23:
-        # pandas-dev/pandas#63567
-        arr_arrow = cast("ArrowStringArray", pd.array([pd.NA], str))
-        check(assert_type(arr_arrow, ArrowStringArray), ArrowStringArray, float)
-        assert pd.isna(assert_type(arr_arrow.dtype.na_value, NAType | float))
+    # pandas-dev/pandas#63567
```
cmp0xff: pandas-dev/pandas#54466 will need to be propagated, maybe in a separate PR.
```suggestion
    # TODO: pandas-dev/pandas#54466 should give BaseStringArray after 3.0
```

tests/arrays/test_base_string_array.py
```diff
@@ -80,23 +77,19 @@ def test_construction_dtype(
...
         # pandas-dev/pandas#63567 Pandas 3.0 gives StringDtype(na_value=nan) if dtype is str or "str"
         else float if is_builtin_str else NAType
     )
-    check(pd.array([*data, np.nan], dtype), target_type, dtype_na)
-    check(pd.array([*data, *data, np.nan], dtype), target_type, dtype_na)
+    check(pd.array([*data, np.nan], dtype), BaseStringArray, dtype_na)
+    check(pd.array([*data, *data, np.nan], dtype), BaseStringArray, dtype_na)
 
     if TYPE_CHECKING:
         # TODO: pandas-dev/pandas#54466 should give BaseStringArray after 3.0
```
cmp0xff, line 95: pandas-dev/pandas#54466 will need to be propagated, maybe in a separate PR.

tests/arrays/test_datetime_array.py
```diff
@@ -239,14 +238,8 @@ def test_construction_dtype(
...
-    check(
-        assert_type(arr.month_name(), np_1darray_object),
-        np_1darray_object if PD_LTE_23 else BaseStringArray,
-    )
-    check(
-        assert_type(arr.day_name(), np_1darray_object),
-        np_1darray_object if PD_LTE_23 else BaseStringArray,
-    )
+    check(assert_type(arr.month_name(), np_1darray_object), BaseStringArray)
+    check(assert_type(arr.day_name(), np_1darray_object), BaseStringArray)
```
cmp0xff: The following needs to work
```suggestion
    check(assert_type(arr.month_name(), BaseStringArray), BaseStringArray)
    check(assert_type(arr.day_name(), BaseStringArray), BaseStringArray)
```

tests/frame/test_groupby.py
```diff
@@ -163,21 +162,6 @@ def test_types_groupby() -> None:
...
-        with pytest_warns_bounded(
-            FutureWarning,
-            "DataFrameGroupBy.apply operated on the grouping columns",
-            upper="2.3.99",
-        ):
-            if PD_LTE_23:
-                check(
-                    assert_type(df.groupby(by="col1").apply(sum), pd.DataFrame),
-                    pd.DataFrame,
-                )
```
cmp0xff, line 180: Could it be made a negative test (in `if TYPE_CHECKING_INVALID_USAGE`)?

tests/test_timefuncs.py
```diff
@@ -366,7 +360,7 @@ def test_series_dt_accessors() -> None:
     ):
         check(
             assert_type(s0.dt.to_pydatetime(), np_1darray_object),
```
cmp0xff:
```suggestion
            assert_type(s0.dt.to_pydatetime(), pd.Series),
```
I think `dtype` would be `object`, hence it's not `pd.Series[Timstamp]`. May be worthy reporting.

tests/test_pandas.py
```diff
@@ -955,30 +952,30 @@ def test_index_unqiue() -> None:
...
     check(
-        assert_type(pd.unique(tdi), np_1darray_td),
-        np_1darray if PD_LTE_23 else pd.TimedeltaIndex,
+        assert_type(pd.unique(tdi), pd.TimedeltaIndex),
+        pd.TimedeltaIndex,
     )
     check(
         assert_type(pd.unique(mi), np_ndarray),
-        np_ndarray if PD_LTE_23 else pd.MultiIndex,
+        pd.MultiIndex,
```
cmp0xff:
```suggestion
        assert_type(pd.unique(mi), pd.MultiIndex), pd.MultiIndex
```

tests/test_pandas.py
```diff
@@ -955,30 +952,30 @@ def test_index_unqiue() -> None:
...
+        pd.Index,
     )
-    check(assert_type(pd.unique(ii_pd), np_1darray | pd.Index), pd.Index)
+    check(assert_type(pd.unique(ii_pd), pd.Index), pd.Index)
     check(assert_type(pd.unique(pi), pd.PeriodIndex), pd.PeriodIndex)
     check(
-        assert_type(pd.unique(ri), np_1darray_int64),
-        np_1darray_int64 if PD_LTE_23 else pd.Index,
+        assert_type(pd.unique(ri), "pd.Index[int]"),
+        pd.Index,
```
cmp0xff:
```suggestion
        assert_type(pd.unique(ri), "pd.Index[int]"), pd.Index, np.integer
```

tests/test_groupby.py
```diff
@@ -467,8 +448,6 @@ def test_frame_groupby_rolling() -> None:
     # props
     check(assert_type(GB_DF.rolling(1).on, str | Index | None), type(None))
     check(assert_type(GB_DF.rolling(1).method, Literal["single", "table"]), str)
-    if PD_LTE_23:
-        check(assert_type(GB_DF.rolling(1).axis, int), int)
```
cmp0xff, line 471: Could you find a document for these properties?

tests/series/test_indexing.py
```diff
@@ -10,23 +10,13 @@
...
 def test_types_select() -> None:
     s = pd.Series(data={"row1": 1, "row2": 2})
-    if PD_LTE_23:
-        # Not valid in 3.0
-        with pytest_warns_bounded(
-            FutureWarning,
-            "Series.__getitem__ treating keys as positions is deprecated",
-            lower="2.0.99",
-        ):
-            s[0]
```
cmp0xff, line 29: We need to disable `s[0]`. I think we can create a ticket and put a `TODO` here, adding `if TYPE_CHECKING_INVALID_USAGE` later.

tests/_typing.py
cmp0xff, line 1: Revert

tests/indexes/timedeltaindex/test_floordiv.py
```diff
@@ -64,11 +63,7 @@ def test_floordiv_py_sequence(left: pd.TimedeltaIndex) -> None:
         _03 = left // c  # type: ignore[operator] # pyright: ignore[reportOperatorIssue,reportUnknownVariableType]
         _04 = left // s  # type: ignore[operator] # pyright: ignore[reportOperatorIssue,reportUnknownVariableType]
     # TODO: pandas-dev/pandas#62552 switch to np.integer after Pandas 3.0
```
cmp0xff:
```suggestion
```

tests/series/timedelta/test_floordiv.py
```diff
@@ -92,7 +91,7 @@ def test_floordiv_py_sequence(left: "pd.Series[pd.Timedelta]") -> None:
     check(
         assert_type(left // d, "pd.Series[int]"),
         pd.Series,
-        int if PD_LTE_23 else np.integer,
+        np.integer,
     )
```
cmp0xff:
```suggestion
check(assert_type(left // d, "pd.Series[int]"), pd.Series, np.integer)
```

tests/frame/test_frame.py
```diff
@@ -3369,37 +3202,17 @@ def test_frame_stack() -> None:
         [[1.0, 2.0], [3.0, 4.0]], index=["cat", "dog"], columns=multicol2
     )
 
-    with pytest_warns_bounded(
-        FutureWarning,
-        "The previous implementation of stack is deprecated",
```
cmp0xff, line 3378: Probably need to disable the previous implementation and add negative tests in `if TYPE_CHECKING_INVALID_USAGE`

tests/frame/test_frame.py
```diff
@@ -3007,49 +2901,6 @@ class ReadCsvKwargs(TypedDict):
...
-        ):
-            check(
-                assert_type(
-                    pd.read_csv(path_str, parse_dates=parse_dates_2), pd.DataFrame
-                ),
-                pd.DataFrame,
-            )
-        parse_dates_3 = {"combined_date": [1, 2, 3]}
-        with pytest_warns_bounded(
-            FutureWarning, "Support for nested sequences", lower="2.1.99"
```
cmp0xff, line 3027: Probably need to disable the previous implementation and add negative tests in `if TYPE_CHECKING_INVALID_USAGE`

tests/frame/test_frame.py
```diff
@@ -3007,49 +2901,6 @@ class ReadCsvKwargs(TypedDict):
...
-            check(
-                assert_type(
-                    pd.read_csv(path_str, parse_dates=parse_dates_3), pd.DataFrame
-                ),
-                pd.DataFrame,
-            )
-        # MyPy calls this Dict[str, object] by default which necessitates the explicit annotation (Pyright does not)
-        parse_dates_4: dict[str, list[str | int]] = {"combined_date": [1, "Month", 3]}
-        with pytest_warns_bounded(
-            FutureWarning, "Support for nested sequences", lower="2.1.99"
```
cmp0xff, line 3038: Probably need to disable the previous implementation and add negative tests in `if TYPE_CHECKING_INVALID_USAGE`

tests/test_groupby.py
```diff
@@ -247,16 +246,6 @@ def df2scalar(val: DataFrame) -> float:
             DatetimeIndexResamplerGroupby,
         )
 
-        # interpolate
-        if PD_LTE_23:
-            check(assert_type(GB_DF.resample("ME").interpolate(), DataFrame), DataFrame)
```
cmp0xff, line 252: Do we need negative tests here?

tests/test_groupby.py
```diff
@@ -865,64 +840,6 @@ def test_frame_groupby_ewm() -> None:
     check(assert_type(GB_DF.ewm(1).std(), DataFrame), DataFrame)
     check(assert_type(GB_DF.ewm(1).var(), DataFrame), DataFrame)
 
-    # aggregate
-    if PD_LTE_23:
-        with pytest_warns_bounded(
-            FutureWarning,
-            r"The provided callable <function (sum|mean) .*> is currently using ",
```
cmp0xff, line 872: Probably need to disable the previous implementation and add negative tests in `if TYPE_CHECKING_INVALID_USAGE`

tests/test_groupby.py
```diff
@@ -967,31 +884,6 @@ def test_series_groupby_ewm() -> None:
     check(assert_type(GB_S.ewm(1).std(), "Series[float]"), Series, float)
     check(assert_type(GB_S.ewm(1).var(), "Series[float]"), Series, float)
 
-    # aggregate
-    with pytest_warns_bounded(
-        FutureWarning,
-        r"The provided callable <function (sum|mean) .*> is currently using ",
```
cmp0xff, line 973: Probably need to disable the previous implementation and add negative tests in `if TYPE_CHECKING_INVALID_USAGE`

tests/test_windowing.py
```diff
@@ -350,24 +349,6 @@ def test_ewm_basic_math() -> None:
 
 
 def test_ewm_aggregate() -> None:
-    if PD_LTE_23:
-        with pytest_warns_bounded(
-            FutureWarning,
-            r"The provided callable <function (sum|mean) .*> is currently using ",
```
cmp0xff, line 356: Probably need to disable the previous implementation and add negative tests in `if TYPE_CHECKING_INVALID_USAGE`

tests/test_windowing.py
```diff
@@ -381,24 +362,6 @@ def test_ewm_basic_math_series() -> None:
 
 
 def test_ewm_aggregate_series() -> None:
-    if PD_LTE_23:
-        with pytest_warns_bounded(
-            FutureWarning,
-            r"The provided callable <function (sum|mean) .*> is currently using ",
```
cmp0xff, line 387: Probably need to disable the previous implementation and add negative tests in `if TYPE_CHECKING_INVALID_USAGE`

pandas-stubs/core/frame.pyi
loicdiridollou: Reverted and added

tests/arrays/test_base_string_array.py
loicdiridollou: Yes I plan to do those in a separate one, otherwise it is just a massive bunch of changes and so easy to miss something.

tests/arrays/test_datetime_array.py
loicdiridollou: Fixed

tests/test_timefuncs.py
loicdiridollou: The behavior is deprecated (see the warning) so I think we should remove it, will add a TODO there so we don't forget and we can discuss.

tests/series/test_series.py
```diff
@@ -746,32 +708,25 @@ def test_types_clip() -> None:
...
+    check(
+        assert_type(s.clip(lower=lower, axis="index", inplace=True), "pd.Series[int]"),
+        pd.Series,
+        np.integer,
+    )
 
     if TYPE_CHECKING_INVALID_USAGE:
-        s.clip(lower=lower, axis=1)  # type: ignore[call-overload] # pyright: ignore[reportArgumentType]
-        s.clip(lower=lower, axis="column")  # type: ignore[call-overload] # pyright: ignore[reportArgumentType]
+        s.clip(lower=lower, axis=1)  # type: ignore[call-overload] # pyright: ignore[reportCallIssue, reportArgumentType]
```
cmp0xff:
```suggestion
        s.clip(lower=lower, axis=1)  # type: ignore[call-overload] # pyright: ignore[reportCallIssue,reportArgumentType]
```

tests/series/test_series.py
```diff
@@ -2145,44 +2071,6 @@ def test_logical_operators() -> None:
 
     check(assert_type(True ^ (df["a"] >= 2), "pd.Series[bool]"), pd.Series, np.bool_)
 
-    if PD_LTE_23:
-        with pytest_warns_bounded(
-            FutureWarning,
-            r"Logical ops \(and, or, xor\) between Pandas objects and dtype-less sequences "
-            r"\(e.g. list, tuple\) are deprecated",
-            lower="2.0.99",
```
cmp0xff, line 2153: Maybe negative tests

tests/series/test_series.py
```diff
@@ -2097,22 +2039,6 @@ def test_bitwise_operators() -> None:
     check(assert_type(s ^ s2, "pd.Series[int]"), pd.Series, np.integer)
     check(assert_type(s2 ^ s, "pd.Series[int]"), pd.Series, np.integer)
 
-    if PD_LTE_23:
-        with pytest_warns_bounded(
-            FutureWarning,
-            r"Logical ops \(and, or, xor\) between Pandas objects and dtype-less sequences "
-            r"\(e.g. list, tuple\) are deprecated",
-            lower="2.0.99",
```
cmp0xff, line 2105: Maybe negative tests

pandas-stubs/core/frame.pyi
loicdiridollou: According to the old warning message, the future_stack=False is a deprecated implementation so I thought we should not let the user use them, unless they willfully ignore the warning. What do you think?

tests/frame/test_groupby.py
loicdiridollou: This is not an issue with the types but with the content of the dataframe so I am not sure we can do anything with the typing here.

tests/series/test_indexing.py
loicdiridollou: If 0 is an index in the series `s[0]` is accepted, in 3.0 pandas treat `s[...]` as getting the value for the index (similar to how it works for dataframe retrieving the column).
There is nothing we can do here.

tests/test_groupby.py
loicdiridollou: The attribute is not exposed anymore (and RollingGroupBy is not documented either).

tests/test_windowing.py
loicdiridollou: I think something is broken there:

```python
import numpy as np
from pandas import DataFrame

df = DataFrame({"A": ["a"] * 4, "B": range(4)})

df.groupby("A").ewm(1).aggregate(sum)  # fails
df.groupby("A").ewm(1).aggregate("sum") # works
df.groupby("A").ewm(1).aggregate(np.sum) # fails
```
Raised issue in https://github.com/pandas-dev/pandas/issues/63855

tests/test_windowing.py
loicdiridollou: Same issue as https://github.com/pandas-dev/pandas/issues/63855

tests/test_windowing.py
loicdiridollou: Seems like they fixed it on main but it will remain broken on 3.0.0, we can decide to ignore the tests until 3.0.1 is released.

pandas-stubs/core/frame.pyi
Dr-Irv: I vote for removing the `dropna`, `sort` and `future_stack` parameters.  See https://github.com/pandas-dev/pandas/issues/63818

tests/arrays/test_arrow_string_array.py
```diff
@@ -107,8 +106,7 @@ def test_dtype() -> None:
...
     assert assert_type(arr.dtype.storage, Literal["pyarrow"]) == "pyarrow"
 
-    if not PD_LTE_23:
-        # pandas-dev/pandas#63567
-        arr_arrow = cast("ArrowStringArray", pd.array([pd.NA], str))
-        check(assert_type(arr_arrow, ArrowStringArray), ArrowStringArray, float)
-        assert pd.isna(assert_type(arr_arrow.dtype.na_value, NAType | float))
+    # TODO: pandas-dev/pandas#54466 should give BaseStringArray after 3.0
+    arr_arrow = cast("ArrowStringArray", pd.array([pd.NA], str))
+    check(assert_type(arr_arrow, ArrowStringArray), ArrowStringArray, float)
```
Dr-Irv, line 112: So the type from the stubs here should be `BaseStringArray`.  But the runtime type will be `ArrowStringArray` because we test with pyarrow installed.

tests/frame/test_groupby.py
Dr-Irv: agree

tests/series/test_indexing.py
Dr-Irv: I agree.  The following works:
```python
s = pd.Series(data={4: 1, 5: 2})
s[4]
```

tests/test_groupby.py
Dr-Irv: Yes, I think we should not allow `ResamplerGroupby.interpolate()`.  This would mean splitting `Resampler()` into 2 classes, a `BaseResampler` class that has everything but `interpolate()`, and `Resampler` inherits from `BaseResampler` and includes `interpolate()`, and then `DatetimeIndexResampler` would need to inherit from `BaseResampler` instead.

Probably best at this point to create a new issue where we need to make this happen and add a test.

tests/test_groupby.py
```diff
@@ -403,21 +392,13 @@ def f(val: Series) -> float:
     # asfreq
     check(assert_type(GB_S.resample("ME").asfreq(-1.0), "Series[float]"), Series, float)
 
-    # interpolate
-    if PD_LTE_23:
-        check(
-            assert_type(GB_S.resample("ME").interpolate(), "Series[float]"),
```
Dr-Irv, line 409: would need to disallow as mentioned above

tests/test_groupby.py
```diff
@@ -247,16 +246,6 @@ def df2scalar(val: DataFrame) -> float:
             DatetimeIndexResamplerGroupby,
         )
 
-        # interpolate
-        if PD_LTE_23:
-            check(assert_type(GB_DF.resample("ME").interpolate(), DataFrame), DataFrame)
-            check(
-                assert_type(
-                    GB_DF.resample("ME").interpolate(method="linear"), DataFrame
```
Dr-Irv, line 255: for this and the above test, could also add `GB_DF.apply(lambda x: x.resample("ME").interpolate(method="linear")` as a replacement test

tests/test_groupby.py
Dr-Irv: `axis` was removed from all groupby operatoins, as well as rolling, resample

tests/test_groupby.py
Dr-Irv: agree

tests/test_groupby.py
Dr-Irv: agree

tests/test_pandas.py
```diff
@@ -954,32 +950,14 @@ def test_index_unqiue() -> None:
...
-        assert_type(pd.unique(tdi), np_1darray_td),
-        np_1darray if PD_LTE_23 else pd.TimedeltaIndex,
-    )
-    check(
-        assert_type(pd.unique(mi), np_ndarray),
-        np_ndarray if PD_LTE_23 else pd.MultiIndex,
-    )
+    check(assert_type(pd.unique(ri), "pd.Index[int]"), pd.Index)
+    check(assert_type(pd.unique(tdi), pd.TimedeltaIndex), pd.TimedeltaIndex)
+    check(assert_type(pd.unique(mi), np_ndarray), pd.MultiIndex)
```
Dr-Irv: this test seems incorrect.  If the runtime result is `MultiIndex`, then the typed result should be the same

tests/test_timefuncs.py
Dr-Irv: For pandas 3.0, it returns a `Series`, so fix the stubs and make the test check for `Series`

tests/test_timefuncs.py
```diff
@@ -543,7 +538,7 @@ def test_series_dt_accessors() -> None:
             upper="2.99",
         ),
```
Dr-Irv: should remove this

tests/test_timefuncs.py
```diff
@@ -543,7 +538,7 @@ def test_series_dt_accessors() -> None:
...
         pytest_warns_bounded(
-            Pandas4Warning,  # should be Pandas4Warning but only exposed starting pandas 3.0.0
+            Pandas4Warning,
             "The behavior of TimedeltaProperties.to_pytimedelta is deprecated",
             lower="2.99",
             upper="3.0.99",
```
Dr-Irv:
```suggestion
upper="3.0.99",
```
since we'll test with 3.0, no need for lower bound any more

pyproject.toml
```diff
@@ -34,10 +34,11 @@ exclude = ["pandas-stubs/__init__.py"]
 python = ">=3.11"
 types-pytz = ">=2022.1.1"
 numpy = ">=1.23.5"
+pytz = "^2025.2"
```
Dr-Irv: Not sure why you added this.  With pandas 3.0, `tzdata` is required on Windows, but not on Linux or mac.

tests/frame/test_frame.py
Dr-Irv: agree

tests/frame/test_frame.py
Dr-Irv: agree

tests/frame/test_frame.py
```diff
@@ -3007,49 +2901,6 @@ class ReadCsvKwargs(TypedDict):
...
-            )
-
-        parse_dates_6 = [[1, 2, 3]]
-        with pytest_warns_bounded(
-            FutureWarning, "Support for nested sequences", lower="2.1.99"
-        ):
-            check(
-                assert_type(
-                    pd.read_csv(path_str, parse_dates=parse_dates_6), pd.DataFrame
-                ),
```
Dr-Irv, line 3054: change stubs here too and have negative tests

tests/frame/test_frame.py
```diff
@@ -3369,37 +3202,17 @@ def test_frame_stack() -> None:
...
-                assert_type(
-                    df_multi_level_cols2.stack(0, future_stack=False),
-                    pd.DataFrame | pd.Series,
-                ),
-                pd.DataFrame,
-            )
-            check(
-                assert_type(
-                    df_multi_level_cols2.stack(0, dropna=True, sort=True),
-                    pd.DataFrame | pd.Series,
```
Dr-Irv, line 3403: these would become negative tests

pandas-stubs/core/frame.pyi
Dr-Irv: And I think we need a test for `df.rename(..., inplace=True)`, because we don't have one right now as best as I can tell.

tests/series/test_series.py
Dr-Irv: yes, and the stubs have to be modified to make that happen

tests/series/test_series.py
Dr-Irv: and modify the stubs

pandas-stubs/core/frame.pyi
loicdiridollou: There was no test thus I made the mistake and did not catch it.
Test is added and this is reverted, thanks for pointing it out.

pandas-stubs/core/frame.pyi
loicdiridollou: Done and test added.

tests/arrays/test_arrow_string_array.py
loicdiridollou: Improved the TODO so we can make a follow up PR with those.

tests/arrays/test_base_string_array.py
loicdiridollou: Improved the TODO so we can do a follow up

tests/series/test_series.py
loicdiridollou: I assume `MaskType` in the stubs is having a dtype defined in those so just needed to drop `list[int]` from the stubs.

tests/series/test_series.py
loicdiridollou: Addressed, turns out MaskType has list[bool] in it so had to create a new typealias

tests/test_groupby.py
loicdiridollou: Created a new issue for it

tests/test_groupby.py
loicdiridollou: Will do in a follow up

tests/test_groupby.py
loicdiridollou: There is a conversation in pandas-dev about it, the agg seems to only allow the "standard" methods lime sum mean var cov corr std so they should be passed as str and not callable. Will add a TODO there

tests/test_pandas.py
loicdiridollou: Fixed, good catch!

tests/test_windowing.py
loicdiridollou: Actually here it should only allow "str" and the apply method is not defined on ewm so callable would not work.

tests/test_windowing.py
loicdiridollou: Made it a TODO

pyproject.toml
loicdiridollou: I am getting an error when running the tests:
```
tests/test_timefuncs.py:22: in <module>
    import pytz
E   ModuleNotFoundError: No module named 'pytz'
```

Dr-Irv: Now that I look at it, `pytz` is only used in our tests, so it (and `types-pytz`) should be moved down to the dev dependencies section.

tests/test_groupby.py
```diff
@@ -890,6 +890,8 @@ def test_series_groupby_ewm() -> None:
     check(assert_type(next(iterator), "Series[float]"), Series, float)
     check(assert_type(list(GB_S.ewm(1)), "list[Series[float]]"), list, Series)
 
+    # TODO: pandas-dev/pandas-stubes#1641 in pandas 3.0 agg only supports str function and not callable
```
Dr-Irv, line 830: So I think https://github.com/pandas-dev/pandas/issues/63855 will get resolved, and we may need to bring back the old tests.  So can you leave those tests in there and just comment them out with a link to that issue?

tests/_typing.py
Dr-Irv: This is still showing the wrong version.  It should not be changed.

tests/test_groupby.py
Dr-Irv: #1644

tests/test_groupby.py
Dr-Irv: I think you should add this test

tests/test_groupby.py
Dr-Irv: Keep these tests in there as comments until https://github.com/pandas-dev/pandas/issues/63855 is resolved, and add a pointer to it in the comments

tests/test_pandas.py
```diff
@@ -954,32 +950,14 @@ def test_index_unqiue() -> None:
...
-    )
-    check(
-        assert_type(pd.unique(tdi), np_1darray_td),
-        np_1darray if PD_LTE_23 else pd.TimedeltaIndex,
-    )
-    check(
-        assert_type(pd.unique(mi), np_ndarray),
-        np_ndarray if PD_LTE_23 else pd.MultiIndex,
-    )
+    check(assert_type(pd.unique(ri), "pd.Index[int]"), pd.Index)
```
Dr-Irv: From @cmp0xff comment here:  https://github.com/pandas-dev/pandas-stubs/pull/1643#discussion_r2721090565

Need to add `np.integer` in the `check` call

tests/test_windowing.py
Dr-Irv: Keep the tests in comments until https://github.com/pandas-dev/pandas/issues/63855 is resolved

tests/test_windowing.py
Dr-Irv: Keep the tests in comments until pandas-dev/pandas#63855 is resolved

tests/test_groupby.py
Dr-Irv: See this comment:  
https://github.com/pandas-dev/pandas-stubs/pull/1643#pullrequestreview-3694937337

These tests have to be changed

tests/test_groupby.py
loicdiridollou: To be fair I am not sure this issue will get solved any time soon, it is hard to apply an ewm with a "random" function. It is just that in the past it would convert sum, np.sum to "sum". The apply method was never implemented.
So we need to remove Callable from the allowed types in the aggregate for ewm. Not sure if it is clear

tests/test_groupby.py
loicdiridollou: I was going to do a follow up PR for all those tests but will address in this one.

tests/test_groupby.py
loicdiridollou: So that behavior is not the correct behavior, np.sum is just relying on `ewm(1).sum()` so I think this is a deprecated behavior and we should remove aggregate(Callable).
What do you think?

tests/test_windowing.py
loicdiridollou: Will keep the tests but feel like this will stay in there forever

tests/test_groupby.py
loicdiridollou: Fixed, the warning is just not there anymore

tests/test_groupby.py
Dr-Irv: I'm fine to remove the Callable as an allowed type, but let's keep the tests in there, and use `TYPE_CHECKING_INVALID_USAGE` .  Include a comment pointing to the pandas issue, and if that gets resolved to allow the callables, then we'll change the tests and the stubs.  If not, then we have tests that invalidate using a callable.

tests/test_groupby.py
Dr-Irv: So you'll have to check all of those `pytest_warns_bounded` throughout the tests to either remove that part, or alter the test/stubs to conform.

tests/test_groupby.py
Dr-Irv: Yes, see my other comment.  Remove the `Callable`, and then use `TYPE_CHECKING_INVALID_USAGE` around the current tests so that we are testing that we reject the callable versions.

tests/test_windowing.py
Dr-Irv: Or remove `Callable`, and create the tests as `TYPE_CHECKING_INVALID_USAGE`

tests/_typing.py
loicdiridollou: This is fixed

tests/test_groupby.py
loicdiridollou: That test already exists below so did not need to add it.

tests/test_groupby.py
loicdiridollou: Brought those back in `TYPE_CHECKING_INVALID_USAGE`

tests/test_windowing.py
loicdiridollou: I have added a TODO and put the tests into a `TYPE_CHECKING_INVALID_USAGE`.

tests/test_groupby.py
loicdiridollou: This is added.

tests/test_pandas.py
loicdiridollou: Fixed

tests/test_groupby.py
loicdiridollou: Cleaned all of those.

tests/test_resampler.py
```diff
@@ -143,21 +133,6 @@ def test_interpolate() -> None:
...
-        Pandas4Warning,
-        r"The 'inplace' keyword in DatetimeIndexResampler.interpolate is deprecated and will be removed in a future version. resample\(...\).interpolate is never inplace.",
-        lower="2.99",
-    ):
-        check(
-            assert_type(DF.resample("ME").interpolate(inplace=False), DataFrame),
-            DataFrame,
-        )
-    if TYPE_CHECKING_INVALID_USAGE:
-        DF.resample("ME").interpolate(inplace=True)  # type: ignore[arg-type] # pyright: ignore[reportArgumentType]
```
Dr-Irv, line 163: Leave both in as a negative test with `TYPE_CHECKING_INVALID_USAGE`

tests/test_pandas.py
```diff
@@ -2144,65 +2112,6 @@ def g(x: pd.Series) -> int:
...
-        )
-        check(
-            assert_type(
-                pd.pivot_table(
-                    df, index=Grouper(freq="YE"), columns=Grouper(key="dt", freq="M")
-                ),
-                pd.DataFrame,
-            ),
-            pd.DataFrame,
-        )
```
Dr-Irv, line 2227: This test actually works if we use `freq="ME"`, so can you put it back.

And import `Grouper` from `pandas`:
```python
from pandas import Grouper
```

tests/test_groupby.py
Dr-Irv: I still think you have to add the test as `GB_DF.apply(lambda x: x.resample("ME").interpolate(method="linear")` .  I don't see anything equivalent to using the `lambda` on the `DataFrameGroupBy` object

tests/test_groupby.py
Dr-Irv: I am still seeing some of them (e.g., in `test_pandas.py`.  Just do a search for `pytest_warns` and keep ones you need to keep (if any)

Dr-Irv (changes requested):
There are still some unresolved comments.  See them in here.  Still have some `pytest_warns_bounded` to address

tests/test_groupby.py
loicdiridollou: Added back with the lambda inside.

tests/test_groupby.py
loicdiridollou: Indeed not sure how I missed those

tests/test_resampler.py
loicdiridollou: So for this one it is recognized as part of the kwargs so not sure how we could write a negative test there (unless we keep the `inplace` argument in the definition and set it to False.
Let me know

tests/test_pandas.py
loicdiridollou: The Grouper throws the following error:
`TypeError: Only valid with DatetimeIndex, TimedeltaIndex or PeriodIndex, but got an instance of 'RangeIndex'`
I will add it as a TODO

tests/test_pandas.py
```diff
@@ -2346,6 +2232,18 @@ def test_pivot_table_aggfunc_string_reduction(sample_df: pd.DataFrame) -> None:
         ),
         pd.DataFrame,
     )
+    # TODO: pandas-dev/pandas-stubs#1641, pandas 3.0 support
+    # check(
+    #     assert_type(
+    #         pd.pivot_table(
+    #             sample_df,
+    #             index=Grouper(freq="YE"),
+    #             columns=Grouper(freq="ME"),
```
Dr-Irv:
```suggestion
    #             columns=Grouper(key="dt", freq="ME"),
```
If you put that in there, then the test should work.  the `key="dt"` was there in the previous test.

tests/test_resampler.py
Dr-Irv: You could add an overload that says if `inplace` is specified, we return `Never` and test for `Never`

loicdiridollou:
When it comes to pytest_warns_bounded, the bounds are excluded so we can not do with 3.0.0, we are forced to do with 2.99. I maxed those bounds to 2.99 for now but happy to make a PR to change the behavior to include to bounds (please open an issue, I have for now improved the docs to clarify).

tests/test_timefuncs.py
Dr-Irv, line 1: there is one more lower bound to fix here in this file in line 1870

Dr-Irv (changes requested):
missed one lower bound to change, otherwise good to go

tests/test_timefuncs.py
loicdiridollou: fixed good catch, I have also added the check assert type in that test as it was missing

PR #1649, opened 2026-01-29
BLD: remove invalid `pyproject.toml` property

jorenham:
Only properties specified in https://packaging.python.org/en/latest/specifications/pyproject-toml/#pyproject-toml-spec are allowed in `pyproject.toml`, and `packages`  isn't one of those.


jorenham:
Hmm, CI doesn't seem happy. Is this a Poetry thing or something? 
If you're open to it, I wouldn't mind converting this into a uv-based project, instead.

Issue #1110, opened 2025-02-08
Series.diff returns Series['float'] even for nullable integers


MarcoGorelli:
**Describe the bug**

**To Reproduce**
1. Provide a minimal runnable `pandas` example that is not properly checked by the stubs.
2. Indicate which type checker you are using (`mypy` or  `pyright`).
3. Show the error message received from that type checker while checking your example.

```python
import pandas as pd


reveal_type(pd.Series([1,2,3], dtype='Int64').diff())
#  note: Revealed type is "pandas.core.series.Series[builtins.float]"
```

**Please complete the following information:**
 - OS: [e.g. Windows, Linux, MacOS] linux
 - OS Version [e.g. 22] 22.04
 - python version 3.12.5
 - version of type checker mypy 1.15
 - version of installed `pandas-stubs` 2.2.3.241126

**Additional context**
noticed looking into https://github.com/pandas-dev/pandas-stubs/pull/1108

cmp0xff:
I am replying here so that the idea can have a broader audience.

It seems to me that we can add a generic variable to `Index` and `Series` (and maybe even `DataFrame`) to __indicate the type of `value`__.
- By default, say `Index([1])`, `Series([1])`, `.value` gives `np.ndarray`. They are not nullable.
  - `.array` gives `NumpyExtensionArray`.
- We can also have `Index([1], dtype="Int64")`, `Series([1], dtype="Int64")`. `.value` gives `IntegerArray`. They are nullable.
  - `.array` gives `IntegerArray`.
- We can also have `Index([1], dtype="int64[pyarrow]")`, `Series([1], dtype="int64[pyarrow]")`. `.value` gives `ArrowExtensionArray`. They are nullable.
  - `.array` gives `ArrowExtensionArray`.
- We can also have `Index([1], dtype="category")`, `Series([1], dtype="category")`. `.value` gives `Categorical`. They are kind of nullable.
  - `.array` gives `Categorical`.

What do you think?

Dr-Irv:
> It also preserved the current typing overloads, like arithmetic, because the storage backend is an _additional_ piece of information, and we are not doing something breaking like `Series[NullableInt]`, `Series[NonNullableInt]`.

So are you suggesting we'd have something like:
- `Series[int, "numpy"]
- `Series[int, "nullable"]
- `Series[int, "pyarrow"]

[...]

PR #1650, opened 2026-01-29
TYP: fix `Series[Any].diff()` return type

### LINKED ISSUE #1646: `Series[Any].diff()` and `Series[Timestamp].diff()`  are incompatible
The problem is that `Series[Any].diff() -> Series[float]` and `Series[Timestamp].diff() -> Series[Timedelta]` are incompatible, i.e. `Series[Timedelta]` is not assignable to `Series[float]`, even though `Series[Timestamp]` is assignable to `Series[Any]`. 

```pyi
from typing import Any, reveal_type

import pandas as pd

d1: pd.Series[pd.Timestamp]
reveal_type(d1.diff())  # Series[Timedelta]

d2: pd.Series[Any] = d1
reveal_type(d2.diff())  # Series[float]
```

This seems to be caused by incompatible overlapping overloads. The solution is to change the first overload with `self: Series[Never]` (to which `Series[Any]` is assignable) to return `Series[Any]` or `Series[float | Timedelta | ...]`. It might be cleanest to split it off as a separate overload, so that this won't affect the other `self: Series[int] | ...` cases.

jorenham:
- [x] Closes #1646
- [x] Tests added

### DIFF
--- a/tests/series/test_series.py
+++ b/tests/series/test_series.py
@@ -2775,7 +2775,10 @@ def test_diff() -> None:
     df = pd.DataFrame({"a": range(4), "b": pd.date_range("2026-01-30", "2026-02-02")})
     check(assert_type(df["a"].diff(), "pd.Series[Any]"), pd.Series, float)
     check(
-        assert_type(df["b"].diff(), "pd.Series[Any]"), pd.Series, pd.Timedelta, -1)
+        assert_type(df["b"].diff(), "pd.Series[Any]"),
+        pd.Series,
+        pd.Timedelta,
+        index_to_check_for_type=-1,
     )
     # complex -> complex
     check(


pandas-stubs/core/series.pyi
```diff
@@ -948,12 +948,10 @@ class Series(IndexOpsMixin[S1], ElementOpsMixin[S1], NDFrame):
         self, other: Series[S1], min_periods: int | None = None, ddof: int = 1
     ) -> float: ...
     @overload
-    def diff(  # type: ignore[overload-overlap]
```
loicdiridollou, line 951: This one is actually used when we do things like `df['A'].diff()` and `df['A']` is seen as Series[Any] so we are using that overload to match the common usage, I think if you to remove that it will break the usage of the stubs. @Dr-Irv was this the good reasoning for that overload?

jorenham: I use the same trick a lot as well in numpy and scipy-stubs. But because we can now use `Series[bool_] -> Series[Any]` as first overload, it's no longer needed (as the passing tests show) :)

loicdiridollou: Can you add a test like, I tried to reproduce your changes and the below is not passing
```python
df = pd.DataFrame({"a": range(4)})
check(assert_type(df["a"].diff(), "pd.Series[float]"), pd.Series)
```
I think with your changes it would give `Series[Any]` and not `Series[float]`.
Thanks

jorenham: That test will (and should) fail because `df["a"]` is `Series[Any]`, so `df["A"].diff()` should be `Series[Any]`, not `Series[float]`.

jorenham: Also note that this is backwards-compatible change because of the [gradual guarantee](https://typing.python.org/en/latest/spec/concepts.html#the-gradual-guarantee).

loicdiridollou: On paper yes it should fail but in the "common" usage, this is often a Series[float] which is why there is this overload. i had tried on our codebase at work and it would create hundreds of errors. Let's wait for @Dr-Irv to conform this is the reason we have this slightly unorthodox overload.

jorenham: The problem is that with `df = pd.DataFrame({"a": pd.date_range(1, 4)})`, saying that `df["a"].diff()` is a `Series[float]` is incorrect. `Any` was invented exactly for cases like these, i.e. where it's impossible to statically know a type, even though it is known at runtime.

Dr-Irv: This is a tough call.  Consider this `DataFrame`:
```python
df = pd.DataFrame({"a": range(4), "b": pd.date_range(1, 4)})
```
Our stubs will infer that `df["a"]` and `df["b"]` are both `Series[Any]`.

But at runtime, `df["a"].diff()` is really `Series[float]` and `df["b"].diff()`  is really `Series[Timedelta]`.  

From a pure static typing standpoint, both should be `Series[Any]` because `.diff()` is being applied to a `Series[Any]`.  That is what @jorenham is suggesting.

But if we do that, then this might break code where the stubs require a `Series[float]` to be used.  And since most people just have numerical data in their `DataFrame` columns, we return `Series[float]` so that people don't have to keep doing `cast`.  At least that is the philosophy that @loicdiridollou has indicated above.

Having said that, I think that `Series.diff()` and `Series.__sub__()` should be consistent, so right now, if we do:
```python
import pandas as pd
df = pd.DataFrame({"a": [1,2,3,4], "b": [5,4,3,2]})

dd = df["a"] - df["b"]

reveal_type(dd)
```
The result is `Type of "dd" is "Series[Any]"`

So that suggests that what @jorenham is proposing is the better choice.

pandas-stubs/core/series.pyi
```diff
@@ -948,12 +948,10 @@ class Series(IndexOpsMixin[S1], ElementOpsMixin[S1], NDFrame):
...
-        self: Series[Never] | Series[int], periods: int = ...
-    ) -> Series[float]: ...
-    @overload
     def diff(self: Series[_bool], periods: int = ...) -> Series: ...
     @overload
+    def diff(self: Series[int], periods: int = ...) -> Series[float]: ...
```
Dr-Irv, line 953: This relates to https://github.com/pandas-dev/pandas-stubs/issues/1110 

The issue here is that if the `Series` has dtype `Int64`, the result should be `Series[int]`.  However, since most people aren't using the nullable types yet, it will need to be `Series[float]`

tests/series/test_series.py
```diff
@@ -2768,10 +2768,10 @@ def test_diff() -> None:
         np.bool_,
         index_to_check_for_type=-1,
     )
-    # Any -> float
+    # Any -> Any
     s_o = s.astype(object)
     assert_type(s_o, pd.Series)
-    check(assert_type(s_o.diff(), "pd.Series[float]"), pd.Series, float)
+    assert_type(s_o.diff(), "pd.Series[Any]")
```
Dr-Irv:
```suggestion
    check(assert_type(s_o.diff(), "pd.Series[Any]", pd.Series, float)
```

Dr-Irv (changes requested):
can you also add a test using the example
```python
df = pd.DataFrame({"a": range(4), "b": pd.date_range(1, 4)})
```
and then tests the types and results of `df["a"].diff()` and `df["b"].diff()` ?

pandas-stubs/core/series.pyi
jorenham: > But if we do that, then this might break code where the stubs require a Series[float] to be used.

It won't, because `Series[Any]` is assignable to `Series[float]` (that's what I meant with the gradual guarantee); so there's no need to cast. These changes should be fully backwards compatible.

pandas-stubs/core/series.pyi
jorenham: I suppose it helps that `Series[int]` is assignable to `Series[float]` then. So technically speaking, even if this returns an integer series at runtime, this return annotation isn't wrong. In the worst case it's just a bit too broad.

But just to be clear; is this something you want me to address here? I suppose it would be kinda out of scope though.

pandas-stubs/core/series.pyi
Dr-Irv: OK - but in the back of my mind there is some case that could cause an issue, but I can't think of it right now.  Nevertheless, I'm fine with the change.

pandas-stubs/core/series.pyi
Dr-Irv: No, you don't have to address it.

tests/series/test_series.py
```diff
@@ -2768,10 +2768,15 @@ def test_diff() -> None:
...
-    # Any -> float
+    # Any -> Any
     s_o = s.astype(object)
     assert_type(s_o, pd.Series)
-    check(assert_type(s_o.diff(), "pd.Series[float]"), pd.Series, float)
+    check(assert_type(s_o.diff(), "pd.Series[Any]"), pd.Series, float)
+    df = pd.DataFrame({"a": range(4), "b": pd.date_range("2026-01-30", "2026-02-02")})
+    check(assert_type(df["a"].diff(), "pd.Series[Any]"), pd.Series, float)
+    check(
+        assert_type(df["b"].diff(), "pd.Series[Any]"), pd.Series, pd.api.typing.NaTType
```
Dr-Irv:
```suggestion
        assert_type(df["b"].diff(), "pd.Series[Any]"), pd.Series, pd.Timedelta, -1)
```
That will make it check the last element of the series.

Dr-Irv (changes requested):

tests/series/test_series.py
jorenham: hmm that broke a bunch of stuff

Dr-Irv: oops - I forgot about the `attr` argument, but you figured it out

PR #1655, opened 2026-02-05
GH1654 I/O changes for 3.0

loicdiridollou:
Addresses the following changes from the 3.0 release notes:
- [x] [read_parquet()](https://pandas.pydata.org/docs/dev/reference/api/pandas.read_parquet.html#pandas.read_parquet) accepts to_pandas_kwargs which are forwarded to [pyarrow.Table.to_pandas()](https://arrow.apache.org/docs/python/generated/pyarrow.Table.html#pyarrow.Table.to_pandas) which enables passing additional keywords to customize the conversion to pandas, such as maps_as_pydicts to read the Parquet map data type as python dictionaries (https://github.com/pandas-dev/pandas/issues/56842)
- [x] [read_spss()](https://pandas.pydata.org/docs/dev/reference/api/pandas.read_spss.html#pandas.read_spss) now supports kwargs to be passed to pyreadstat (https://github.com/pandas-dev/pandas/issues/56356)
- [x] Added support to read and write from and to Apache Iceberg tables with the new [read_iceberg()](https://pandas.pydata.org/docs/dev/reference/api/pandas.read_iceberg.html#pandas.read_iceberg) and [DataFrame.to_iceberg()](https://pandas.pydata.org/docs/dev/reference/api/pandas.DataFrame.to_iceberg.html#pandas.DataFrame.to_iceberg) functions (https://github.com/pandas-dev/pandas/issues/61383)

### DIFF
--- a/tests/_typing.py
+++ b/tests/_typing.py
@@ -1,1261 +1,170 @@
-# ruff: noqa: PYI002
-from builtins import type as type_t
-from collections.abc import (
-    Callable,
-    Hashable,
-    Iterator,
-    KeysView,
-    Mapping,
-    MutableSequence,
-    Sequence,
-)
-import datetime
-from datetime import tzinfo
-from os import PathLike
-from re import Pattern
-import sys
-from typing import (
-    TYPE_CHECKING,
-    Any,
-    Generic,
-    Literal,
-    Protocol,
-    SupportsIndex,
-    TypeAlias,
-    TypedDict,
-    overload,
-)
+# This file serves as a stub file for static type checkers
+# (pyright does not like it if I call the file tests/_typing.pyi).
+# It can only import from pandas._typing.
+# By tests/conftest.py, pytest copies this file to tests/_typing.pyi, and
+# pandas-stubs/_typing.pyi to tests/_typing.py.
+# After tests are done, pytest copies tests/_typing.pyi back to
+# pandas-stubs/_typing.pyi.
+# Do not commit this file if its content is identical to pandas-stubs/_typing.pyi.
 
-import numpy as np
-from numpy import typing as npt
-import pandas as pd
-from pandas.core.arrays import (
-    ExtensionArray,
-    IntegerArray,
-)
-from pandas.core.frame import DataFrame
-from pandas.core.generic import NDFrame
-from pandas.core.groupby.grouper import Grouper
-from pandas.core.indexes.base import Index
-from pandas.core.series import Series
-from pandas.core.tools.datetimes import FulldatetimeDict
-from typing_extensions import (
-    ParamSpec,
-    TypeVar,
-    override,
-)
-
-from pandas._libs.interval import Interval
-from pandas._libs.missing import NAType
-from pandas._libs.tslibs import (
-    BaseOffset,
-    Period,
-    Timedelta,
-    Timestamp,
-)
-from pandas._libs.tslibs.nattype import NaTType
-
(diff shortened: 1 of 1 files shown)


tests/test_io.py
```diff
@@ -1683,3 +1703,11 @@ def test_converters_partial(tmp_path: Path) -> None:
 
     result = pd.read_excel(path_str, converters={"field_1": partial_func})
     check(assert_type(result, pd.DataFrame), pd.DataFrame)
+
+
+def test_iceberg(tmp_path: Path) -> None:
+    path_str = str(tmp_path / str(uuid.uuid4()))
+
+    if TYPE_CHECKING:
```
loicdiridollou: There is quite a bit of setup to run iceberg tests, I chose to only make them type check and not runtime, let me know what you think.

pandas-stubs/core/generic.pyi
```diff
@@ -473,3 +473,13 @@ class NDFrame:
         level: IndexLabel | None = None,
         drop_level: _bool = True,
     ) -> Self | Series: ...
+    def to_iceberg(
```
Dr-Irv: this should be in `frame.pyi`, since there is no `Series.to_iceberg()`

tests/test_io.py
Dr-Irv: We can install `pyiceberg` as part of the testing env and then do the tests.

loicdiridollou: Took a bit of moving stuff around but we have a test that runs (the test is a bit clunky because there are so many warnings getting raised and we need to clean up after the test, let me know if you'd like more comments around the test)

pandas-stubs/core/generic.pyi
loicdiridollou: Moved

tests/_typing.py
Dr-Irv, line 1: don't commit this!

pyproject.toml
```diff
@@ -72,6 +72,7 @@ fsspec = "^2025.10.0"
 tables = { version = ">=3.10.2", python = "<3.14" }  # TODO: add py314 when PyTables/PyTables#1261 is resolved
 pytz = "^2025.2"
 types-pytz = ">=2022.1.1"
+pyiceberg = "^0.10.0"
```
Dr-Irv, line 75: this should go to the `tool.poetry.group.dev.dependencies` section.

tests/_typing.py
loicdiridollou: Arghhh every time the tests fail this appears will revert

pyproject.toml
loicdiridollou: It is already in that group, or maybe I missed something

tests/test_io.py
```diff
@@ -1683,3 +1704,130 @@ def test_converters_partial(tmp_path: Path) -> None:
 
     result = pd.read_excel(path_str, converters={"field_1": partial_func})
     check(assert_type(result, pd.DataFrame), pd.DataFrame)
+
+
+@pytest.mark.filterwarnings("ignore::ResourceWarning")
```
loicdiridollou, line 1709: Weirdly it does not always raise the warning with the gc call at the end so I do a larger ignore if that is okay @Dr-Irv

Dr-Irv: yes

pyproject.toml
Dr-Irv: ooops - my mistake

PR #1661, opened 2026-02-06
GH1654 Styler module

loicdiridollou:

### DIFF
--- a/tests/test_styler.py
+++ b/tests/test_styler.py
@@ -267,7 +267,9 @@ def test_styler_map() -> None:
 
 def test_to_typst(tmp_path: Path) -> None:
     """Test Styler.to_typst."""
-    path_str = str(tmp_path / f"{uuid.uuid4()}test.typ")
+    path = tmp_path / f"{uuid.uuid4()}test.typ"
+    path_str = str(path)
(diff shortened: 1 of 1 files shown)


tests/test_styler.py
```diff
@@ -253,3 +263,33 @@ def color_negative(v: Scalar, /, color: str) -> str | None:
     df = DataFrame(np.random.randn(5, 2), columns=["A", "B"])
 
     check(assert_type(df.style.map(color_negative, color="red"), Styler), Styler)
+
+
+def test_to_typst(tmp_path: Path) -> None:
+    """Test Styler.to_typst."""
+    path_str = str(tmp_path / f"{uuid.uuid4()}test.typ")
```
cmp0xff: Can we also check a `Path` object?
```suggestion
    path = tmp_path / f"{uuid.uuid4()}test.typ"
    path_str = str(path)
```
and try `path` in addition to `path_str`.

cmp0xff (changes requested):
Just one comment.

tests/test_styler.py
loicdiridollou: Added the test, good idea

PR #1668, opened 2026-02-09
use pandas.api.typing.aliases

Dr-Irv:
With 3.0, the goal of adding `pandas.api.typing.aliases` was to get people to stop importing from `pandas._typing`.  This PR adds the aliases, and modifies the tests to import from that module.

There are a few aliases that need to be added back into pandas, which I will handle via a PR there

- `IntervalClosedType`
- `IntervalLeftRight`
- `ListLike`
- `ScalarIndexer`
- `SequenceIndexer`

Once a new version of pandas is released with those changes, we can modify the tests that need to use `TYPE_CHECKING` to no longer need them.

### DIFF
--- a/tests/_typing.py
+++ b/tests/_typing.py
@@ -1,1263 +1,171 @@
-# ruff: noqa: PYI002
-from builtins import type as type_t
-from collections.abc import (
-    Callable,
-    Hashable,
-    Iterator,
-    KeysView,
-    Mapping,
-    MutableSequence,
-    Sequence,
-)
-import datetime
-from datetime import tzinfo
-from os import PathLike
-from re import Pattern
-import sys
-from typing import (
-    TYPE_CHECKING,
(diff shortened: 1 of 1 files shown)


Dr-Irv:
PR for pandas is here:  https://github.com/pandas-dev/pandas/pull/64098

tests/_typing.py
```diff
@@ -1,171 +1,1263 @@
-# ruff: noqa: PLC0414
```
loicdiridollou, line 1: Do we want to make the changes to this file? Always confused if it needs to be committed or not

Dr-Irv: No - that was my mistake.

loicdiridollou (commented):
Great, thanks @Dr-Irv

loicdiridollou (approved):

PR #1672, opened 2026-02-13
GH1671 Improve testing for Transformation/Reduction methods on Groupby

### LINKED ISSUE #1671: Complete tests for `ReductionKernelType`
pandas-dev/pandas-stubs#1242 and pandas-dev/pandas-stubs#1665 introduced `ReductionKernelType` and `test_groupby_kurt`. However other `Literal`s in `ReductionKernelType` are not tested, which is to be completed.

loicdiridollou:
- [x] Closes #1671

### DIFF
--- a/tests/test_groupby.py
+++ b/tests/test_groupby.py
@@ -31,9 +31,12 @@ from pandas.core.window import (
     RollingGroupby,
 )
 
+from pandas.errors import Pandas4Warning
+
 from tests import (
     TYPE_CHECKING_INVALID_USAGE,
     check,
+    pytest_warns_bounded,
 )
 
 if TYPE_CHECKING:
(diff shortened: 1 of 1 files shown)


tests/test_groupby.py
```diff
@@ -925,3 +928,171 @@ def test_frame_groupby_aggregate() -> None:
...
     check(assert_type(df.groupby("b").agg(a=("a", "mean")), DataFrame), DataFrame)
     check(assert_type(df.groupby("b").agg(**dico), DataFrame), DataFrame)
+
+
+def test_frame_groupby_transform_reduction_kernels() -> None:
+    """Test DataFrameGroupBy.transform with ReductionKernelType literals."""
+    check(assert_type(GB_DF.transform("all"), DataFrame), DataFrame)
+    check(assert_type(GB_DF.transform("any"), DataFrame), DataFrame)
+    with pytest_warns_bounded(Pandas4Warning, "corrwith is deprecated", lower="2.99"):
+        check(assert_type(GB_DF.transform("corrwith", other=DF), DataFrame), DataFrame)
```
cmp0xff, line 938: We don't support 2.x anymore, do we?

tests/test_groupby.py
```diff
@@ -925,3 +928,171 @@ def test_frame_groupby_aggregate() -> None:
 
     check(assert_type(df.groupby("b").agg(a=("a", "mean")), DataFrame), DataFrame)
     check(assert_type(df.groupby("b").agg(**dico), DataFrame), DataFrame)
+
+
+def test_frame_groupby_transform_reduction_kernels() -> None:
+    """Test DataFrameGroupBy.transform with ReductionKernelType literals."""
+    check(assert_type(GB_DF.transform("all"), DataFrame), DataFrame)
```
cmp0xff, line 935: I think for `DataFrame.transform`, all `str`'s are accepted by the stubs at the moment. Shall we switch to a `Literal` and also add one or two negative tests in `TYPE_CHECKING_INVALID_USAGE`?

cmp0xff (changes requested):

tests/test_groupby.py
loicdiridollou: in `pytest_warns_bounded` the check for version is strict so if you put `3.0` it will not trigger for the `3.0` version, that way I am forced to stick to 2.99. (we had the issue in a different PR).

tests/test_groupby.py
loicdiridollou: Happy to make the switch, I was just worried there are some functions not in the list and we would break for the user. What do you think? Should we take the risk?

cmp0xff: Postponed to #1676

PR #1673, opened 2026-02-14
GH1644 Prevent interpolate on datetime-like objects in groupby.resample

### LINKED ISSUE #1644: Improve typing of ResamplerGroupBy.interpolate
> Yes, I think we should not allow `ResamplerGroupby.interpolate()`.  This would mean splitting `Resampler()` into 2 classes, a `BaseResampler` class that has everything but `interpolate()`, and `Resampler` inherits from `BaseResampler` and includes `interpolate()`, and then `DatetimeIndexResampler` would need to inherit from `BaseResampler` instead.
> 
> Probably best at this point to create a new issue where we need to make this happen and add a test. 

 _Originally posted by @Dr-Irv in [#1643](https://github.com/pandas-dev/pandas-stubs/pull/1643/changes#r2733494573)_

loicdiridollou:
- [x] Closes #1644

### DIFF
--- a/pandas-stubs/core/resample.pyi
+++ b/pandas-stubs/core/resample.pyi
@@ -55,18 +55,18 @@ _SeriesGroupByFuncArgs: TypeAlias = (
     _SeriesGroupByFuncTypes | Mapping[Hashable, _SeriesGroupByFunc | str]
 )
 
-class BaseResampler(BaseGroupBy[NDFrameT]):
+class Resampler(BaseGroupBy[NDFrameT]):
     def __getattr__(self, attr: str) -> SeriesGroupBy[Any, Any]: ...
     @overload
     def aggregate(
-        self: BaseResampler[DataFrame],
+        self: Resampler[DataFrame],
         func: _FrameGroupByFuncArgs | None = ...,
         *args: Any,
         **kwargs: Any,
     ) -> DataFrame: ...
     @overload
     def aggregate(
-        self: BaseResampler[Series],
+        self: Resampler[Series],
         func: _SeriesGroupByFuncArgs | None = ...,
         *args: Any,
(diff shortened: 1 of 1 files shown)


pandas-stubs/core/resample.pyi
```diff
@@ -55,18 +55,18 @@ _SeriesGroupByFuncArgs: TypeAlias = (
     _SeriesGroupByFuncTypes | Mapping[Hashable, _SeriesGroupByFunc | str]
 )
 
-class Resampler(BaseGroupBy[NDFrameT]):
+class BaseResampler(BaseGroupBy[NDFrameT]):
```
cmp0xff: I think @Dr-Irv is at a better position to review this PR. I would just like to mention that `BaseResampler` would better be decorated by `type_check_only`.

pandas-stubs/core/resample.pyi
```diff
@@ -182,6 +184,16 @@ class DatetimeIndexResamplerGroupby(
...
+    def interpolate(
+        self,
+        method: InterpolateOptions = ...,
+        *,
+        axis: Axis = ...,
+        limit: int | None = ...,
+        limit_direction: Literal["forward", "backward", "both"] = ...,
+        limit_area: Literal["inside", "outside"] | None = ...,
+        **kwargs: Any,
+    ) -> Never: ...
```
Dr-Irv, line 190: Create an `InterpolateMixin` class that defines `interpolate`, then just change the 3 classes  that need `interpolate()` to include that mixin.

loicdiridollou: Let me know if this is what you had in mind

Dr-Irv: yes

PR #1677, opened 2026-02-18
GH1654 set_options with dict

loicdiridollou:
- [x] Add tests for [DataFrame.iloc()](https://pandas.pydata.org/docs/dev/reference/api/pandas.DataFrame.iloc.html#pandas.DataFrame.iloc) and [Series.iloc()](https://pandas.pydata.org/docs/dev/reference/api/pandas.Series.iloc.html#pandas.Series.iloc) now support boolean masks in getitem for more consistent indexing behavior (https://github.com/pandas-dev/pandas/issues/60994)
- [x] [set_option()](https://pandas.pydata.org/docs/dev/reference/api/pandas.set_option.html#pandas.set_option) now accepts a dictionary of options, simplifying configuration of multiple settings at once (https://github.com/pandas-dev/pandas/issues/61093)
- [x] Changes in the error ignore for pyrefly with version 0.53.0 (change from 0.52 to 0.53)


pandas-stubs/_config/config.pyi
```diff
@@ -11,7 +11,10 @@ from typing import (
 )
 
 def get_option(pat: str) -> Any: ...
+@overload
 def set_option(pat: str, val: object) -> None: ...
```
cmp0xff, line 15: From the source code, it can be `set_option(*args)`. In particular, it can be several pat-val pairs. But I don't know how we can type this 😕

tests/frame/test_frame.py
```diff
@@ -3952,6 +3926,15 @@ def _constructor(self) -> type[MyClass]:
     check(assert_type(df[["a", "b"]], MyClass), MyClass)
 
 
+def test_frame_iloc_series_bool() -> None:
```
cmp0xff, line 3929: Is there an issue for this new test?

tests/series/test_series.py
```diff
@@ -2074,6 +2073,16 @@ def func() -> MySeries[float]:
     assert result.iloc[1] == 2
 
 
+def test_series_iloc_series_bool() -> None:
```
cmp0xff, line 2076: Same question as above

pyproject.toml
```diff
@@ -41,7 +41,7 @@ pyarrow = ">=10.0.1"
 pytest = ">=8.4.2"
 pyright = ">=1.1.408"
 ty = ">=0.0.12"
-pyrefly = ">=0.48.0"
+pyrefly = "0.53.0"
```
cmp0xff: Is there an issue behind fixing the version?

pandas-stubs/_config/config.pyi
Dr-Irv: If you have an overload of 
```python
def set_option(*args: tuple[str, object]) -> None: ...
```
that will do it.  This is saying you can have the variable length arguments, and that they come in pairs

tests/frame/test_frame.py
loicdiridollou: These were just new things in pandas 3.0 and I could not see a test for them so decided to create ones to make sure the behavior is tested.

tests/series/test_series.py
loicdiridollou: Same

pyproject.toml
loicdiridollou: Sorry should stay with >=

tests/frame/test_frame.py
cmp0xff: OK for now, but please separate distinct changes next time if you can, thanks 🙂

tests/series/test_series.py
cmp0xff: OK for now, but please separate distinct changes next time if you can, thanks 🙂

pandas-stubs/_config/config.pyi
cmp0xff: Thank you @Dr-Irv. @loicdiridollou could you try the idea and add a few new tests?

pyproject.toml
cmp0xff: Have you fixed this one?

tests/frame/test_frame.py
loicdiridollou: Originally the PR was made for those changes but turned out we already had them so eventually added set_option on top, will split in the future.

pyproject.toml
loicdiridollou: Sorry thought I had added it but did not, it is now fixed.

pandas-stubs/_config/config.pyi
loicdiridollou: `tuple` here is the wrong typing since what I am trying to do is with `dict`, but it does with it so I have put it in the code.

cmp0xff: #1680

PR #1713, opened 2026-03-30
GH1396 Add concat compatibility for subclasses of DataFrame

### LINKED ISSUE #1396: `pd.concat` with `GeoDataFrame`s
```python
import geopandas as gpd
import pandas as pd
from shapely import Point

gdf1 = gpd.GeoDataFrame(geometry=[Point(0, 0)])
gdf2 = gpd.GeoDataFrame(geometry=[Point(1, 1)])

gdf: gpd.GeoDataFrame = pd.concat([gdf1, gdf2], ignore_index=True)
assert isinstance(gdf, gpd.GeoDataFrame)
print(gdf)
```

```
      geometry
0  POINT (0 0)
1  POINT (1 1)
```

```
t.py:8: error: Incompatible types in assignment (expression has type "DataFrame", variable has type "GeoDataFrame")  [assignment]
```

* pandas-stubs 2.3.2.250926
* mypy 1.18.2
* geopandas 1.1.1

loicdiridollou:
- [x] Closes #1396

### DIFF
--- a/pandas-stubs/core/reshape/concat.pyi
+++ b/pandas-stubs/core/reshape/concat.pyi
@@ -6,6 +6,7 @@ from collections.abc import (
 from typing import (
     Literal,
     Never,
+    TypeVar,
     overload,
 )
 
@@ -22,6 +23,8 @@ from pandas._typing import (
     HashableT3,
     HashableT4,
 )
+
+DataFrameCompatType = TypeVar("DataFrameCompatType", bound=DataFrame)
 
 @overload
 def concat(
@@ -62,6 +65,19 @@ def concat(  # type: ignore[overload-overlap]
     verify_integrity: bool = False,
     sort: bool = False,
 ) -> Series: ...
+@overload
+def concat(
+    objs: Iterable[DataFrameCompatType | None],
+    *,
+    axis: Axis = 0,
+    join: Literal["inner", "outer"] = "outer",
+    ignore_index: bool = False,
+    keys: Iterable[HashableT2] | None = None,
+    levels: Sequence[list[HashableT3] | tuple[HashableT3, ...]] | None = None,
+    names: list[HashableT4] | None = None,
+    verify_integrity: bool = False,
+    sort: bool = False,
+) -> DataFrameCompatType: ...
 @overload
 def concat(
     objs: Iterable[NDFrame | None] | Mapping[HashableT1, NDFrame | None],
--- a/pandas-stubs/core/reshape/concat.pyi
+++ b/pandas-stubs/core/reshape/concat.pyi
@@ -24,7 +24,7 @@ from pandas._typing import (
     HashableT4,
 )
 
-DataFrameCompatType = TypeVar("DataFrameCompatType", bound=DataFrame)
+DataFrameT0 = TypeVar("DataFrameT0", bound=DataFrame)
 
 @overload
 def concat(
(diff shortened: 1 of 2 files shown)


pandas-stubs/core/reshape/concat.pyi
```diff
@@ -23,6 +24,8 @@ from pandas._typing import (
     HashableT4,
 )
 
+DataFrameCompatType = TypeVar("DataFrameCompatType", bound=DataFrame)
```
cmp0xff: We usually do something like `DataFrameT0`, don't we

pandas-stubs/core/reshape/concat.pyi
cmp0xff: Would it make sense to merge these two overloads?

tests/test_pandas.py
```diff
@@ -2544,3 +2544,18 @@ def test_argmin_and_argmax_return() -> None:
     i1 = df.a.abs().argmax()
     check(assert_type(i, np.int64), np.int64)
     check(assert_type(i1, np.int64), np.int64)
+
+
+def test_frame_subclass_concat() -> None:
```
cmp0xff: Can we move it to somewhere near L112-L404 where other tests for concat live?

cmp0xff (changes requested):

tests/test_pandas.py
loicdiridollou: Fixed

pandas-stubs/core/reshape/concat.pyi
loicdiridollou: Not super clear to me how we could combine both since everything that is NDFrame should go to DataFrame and anything DataFrame compatible should maintain the type. Did you have an idea about how to implement that?

pandas-stubs/core/reshape/concat.pyi
loicdiridollou: Fixed

pandas-stubs/core/reshape/concat.pyi
cmp0xff: It seems to me that DataFrame is the only subclass of NDFrame. What if you use `NDFrameT` (already in `_typing`) instead of `DataFrameT0`?

loicdiridollou: Series is also a subclass of NDFrame, the issue arises when you have a list of `[None, Series, DataFrame]` it will give the return type as `DataFrame | Series` but we want `DataFrame` (the `NDFrameT` matches to the type of the union and not the "bigger" type). Not sure if that was clear

loicdiridollou: https://github.com/pandas-dev/pandas-stubs/actions/runs/23871184309/job/69603313845?pr=1713
See that run with all the lints

cmp0xff: Oops, my bad, somehow forgot `Series` is also a subclass of `NDFrame`.

If mixing `Series` and `DataFrame` gives `DataFrame`, what if people mix `Series` and _subclasses_ of `DataFrame` and complain that `pandas-stubs` doesn't serve them right?

loicdiridollou: Not sure there is a way to do so, I tried with `Iterable[DataFrameT0 | NDFrame | None]` and it is raising a lot of issues with `Unknown`. I would say in general that this is not what is intended like a Geopandas and a Series should not be allowed.

cmp0xff (approved):
Let's see if the users of downstream projects will be happy with the changes. Thank you @loicdiridollou !

PR #1721, opened 2026-04-11
GH1654 Deprecations

loicdiridollou:

- [x] [IncompatibleFrequency](https://pandas.pydata.org/docs/dev/reference/api/pandas.errors.IncompatibleFrequency.html#pandas.errors.IncompatibleFrequency) now subclasses TypeError instead of ValueError. As a result, joins with mismatched frequencies now cast to object like other non-comparable joins, and arithmetic with indexes with mismatched frequencies align (https://github.com/pandas-dev/pandas/issues/55782)
- [x] Deprecated [Timestamp.utcfromtimestamp()](https://pandas.pydata.org/docs/dev/reference/api/pandas.Timestamp.utcfromtimestamp.html#pandas.Timestamp.utcfromtimestamp), use Timestamp.fromtimestamp(ts, "UTC") instead (https://github.com/pandas-dev/pandas/issues/56680)
- [x] Deprecated [Timestamp.utcnow()](https://pandas.pydata.org/docs/dev/reference/api/pandas.Timestamp.utcnow.html#pandas.Timestamp.utcnow), use Timestamp.now("UTC") instead (https://github.com/pandas-dev/pandas/issues/56680)
- [x] Deprecated option future.no_silent_downcasting, as it is no longer used. In a future version accessing this option will raise (https://github.com/pandas-dev/pandas/issues/59502)
- [x] Deprecated using epoch date format in [DataFrame.to_json()](https://pandas.pydata.org/docs/dev/reference/api/pandas.DataFrame.to_json.html#pandas.DataFrame.to_json) and [Series.to_json()](https://pandas.pydata.org/docs/dev/reference/api/pandas.Series.to_json.html#pandas.Series.to_json), use iso instead. (https://github.com/pandas-dev/pandas/issues/57063)


pandas-stubs/core/series.pyi
```diff
@@ -4118,6 +4150,15 @@ class Series(IndexOpsMixin[S1], ElementOpsMixin[S1], NDFrame):
         *args: Any,
         **kwargs: Any,
     ) -> Series[S1]: ...
+    @overload
+    def divmod(
```
cmp0xff: What if you don't add these `Never` overloads? I think at least some of the results can be `undefined`

pandas-stubs/core/arrays/integer.pyi
```diff
@@ -12,8 +12,7 @@ class IntegerDtype(NumericDtype):
     base: None
     @property
     def itemsize(self) -> int: ...
-    @classmethod
-    def construct_array_type(cls) -> type[IntegerArray]: ...
+    def construct_array_type(self) -> type[IntegerArray]: ...
```
cmp0xff, line 15: I think we need more tests for these methods

tests/test_io.py
```diff
@@ -524,6 +524,19 @@ def test_json_chunk(tmp_path: Path) -> None:
...
 
 
+def test_to_json_date_format(tmp_path: Path) -> None:
+    """Test date_format deprecated option `epoch`."""
+    path_str = str(tmp_path / str(uuid.uuid4()))
+    sr = DF.iloc[:, 0]
+
+    if TYPE_CHECKING_INVALID_USAGE:
+        _0 = DF.to_json(path_str, date_format="epoch")  # type: ignore[call-overload]  # pyright: ignore[reportCallIssue,reportArgumentType,reportUnknownVariableType]
+        _1 = DF.to_json(None, date_format="epoch")  # type: ignore[call-overload]  # pyright: ignore[reportCallIssue,reportArgumentType,reportUnknownVariableType]
```
cmp0xff: Might have already been tested in `test_to_json_mode` in `test_frame.py`

tests/test_io.py
```diff
@@ -524,6 +524,19 @@ def test_json_chunk(tmp_path: Path) -> None:
...
+
+        _2 = sr.to_json(path_str, date_format="epoch")  # type: ignore[call-overload]  # pyright: ignore[reportCallIssue,reportArgumentType,reportUnknownVariableType]
+        _3 = sr.to_json(None, date_format="epoch")  # type: ignore[call-overload]  # pyright: ignore[reportCallIssue,reportArgumentType,reportUnknownVariableType]
```
cmp0xff: Might have already been tested in `test_to_json_mode` in `test_series.py`

cmp0xff (changes requested):

tests/test_io.py
loicdiridollou: Correct the tests are the same, will drop those

pandas-stubs/core/series.pyi
loicdiridollou: I was wondering if we wanted to block the user from doing so since it will blow up at runtime or put a Never, open to both, I think the Never will block the user early on so easier to identify the issue, let me know

cmp0xff: I mean, if you remove these new overloads, you will still get `reportCallIssue` in the corresponding tests from `pyright`, and `type-var` or `arg-type` from `mypy`.

Users will be blocked even with the new overloads removed.

The new tests are good, we just see different error types.

cmp0xff (changes requested):

tests/arrays/test_integer.py
```diff
@@ -62,6 +62,14 @@ def test_construction_sequence(
         assert_type(pd.array(UserList([610, np.int64(987)])), IntegerArray)
 
 
+def test_construct_array_type() -> None:
+    """Test the constructor_array_type method."""
+    check(
+        assert_type(pd.array([-1]).dtype.construct_array_type(), type[IntegerArray]),
+        type,
+    )
```
cmp0xff: I think `check(..., type)` is not strong enough. `check(int, type)` will also pass. Maybe the following?
```suggestion
    assert assert_type(pd.array([-1]).dtype.construct_array_type(), type[IntegerArray]) is IntegralArray
```
and also for other new tests.

cmp0xff (changes requested):
Just one more thing!

tests/arrays/test_integer.py
loicdiridollou: Agreed it was a little too wide

cmp0xff (approved):
Thank you @loicdiridollou ! ~~It would be good if you say "Deprecations `pandas` 3.0" in the MR title, so that we know it is about pandas 3.0 deprecations, not 2.3 or 2.2~~

OK I realised `GH1654` it is, but I didn't see the link so my brain omitted it. Maybe #1654 so it links automatically, but it's just personal 🙂

loicdiridollou:
Last time I had put the issue link and when merging it closed the issue which was annoying since we are far from done on the 3.0 migration. Will add a better title next time.

cmp0xff:
I think only if you write `closes #xxxx` in the description will the merging close that Issue; and if you edit the PR title / description before merging, the effect will be removed. We can give it a try next time.

PR #1725, opened 2026-04-15
BUG: allow kwargs in Styler apply_index and map_index

### LINKED ISSUE #1723: Styler `apply_*` and `map` functions do not accept `**kwargs`
**Describe the bug**
This is an extension of https://github.com/pandas-dev/pandas-stubs/issues/919 which was fixed by https://github.com/pandas-dev/pandas-stubs/pull/921. 

There's a few `Styler` functions that accept `**kwargs` being passed to `func`, but the stubs only support `**kwargs` in the `func` signature for `Styler.apply()`. I'm working with `apply_index` right now, but `map` and `map_index` both should be able to accept `**kwargs` according to the docs.  

**To Reproduce**
```
import numpy as np
import pandas as pd


def highlight_odd(index: pd.Series, /, color: str) -> list[str]:
    style = [f"color: {color}" if x % 2 else "" for x in index]
    return style


df = pd.DataFrame(np.random.randint(0, high=100, size=(10, 3)))
styler = df.style.apply_index(highlight_odd, color='red')
styler.to_html("output.html")
```

Using `pyright`:
```test_stub.py:11:31 - error: Argument of type "(index: Series[Any], /, color: str) -> list[str]" cannot be assigned to parameter "func" of type "(Series[Any]) -> (NDArray[str_] | list[str] | Series[str])" in function "apply_index"
    Type "(index: Series[Any], /, color: str) -> list[str]" is not assignable to type "(Series[Any]) -> (NDArray[str_] | list[str] | Series[str])"
      Extra parameter "color" (reportArgumentType)
1 error, 0 warnings, 0 informations
```

 - OS Version: 26.3.1
 - python version: 3.12.12
 - version of type checker: `pyright 1.1.408`
 - version of installed `pandas-stubs`: ﻿ 3.0.0.260204

eosti:
Adds ability to pass a function with `**kwargs` to Styler `apply_index` and `map_index`. The `apply` and `map` functions already supported this. 

- [x] Closes #1723

### DIFF
--- a/pandas-stubs/io/formats/style.pyi
+++ b/pandas-stubs/io/formats/style.pyi
@@ -57,6 +57,11 @@ class _SeriesFunc(Protocol):
         self, series: Series, /, *args: Any, **kwargs: Any
     ) -> list[Any] | Series: ...
 
+class _SeriesStrFunc(Protocol):
+    def __call__(
+        self, series: Series[str], /, *args: Any, **kwargs: Any
+    ) -> list[str] | Series[str]: ...
+
 class _DataFrameFunc(Protocol):
     def __call__(
         self, series: DataFrame, /, *args: Any, **kwargs: Any
@@ -277,14 +282,17 @@ class Styler(StylerRenderer):
     ) -> Styler: ...
     def apply_index(
         self,
-        func: Callable[[Series], list[str] | np_ndarray_str | Series[str]],
+        func: (
+            _SeriesStrFunc
+            | Callable[[Series], list[str] | np_ndarray_str | Series[str]]
+        ),
         axis: Axis = ...,
         level: Level | list[Level] | None = ...,
         **kwargs: Any,
     ) -> Styler: ...
     def map_index(
         self,
-        func: Callable[[Scalar], str | None],
+        func: _MapCallable | Callable[[Scalar], str | None],
         axis: Axis = ...,
         level: Level | list[Level] | None = ...,
         **kwargs: Any,
--- a/tests/test_styler.py
+++ b/tests/test_styler.py
@@ -76,12 +76,25 @@ def test_apply_index() -> None:
 
     check(assert_type(DF.style.apply_index(f1), Styler), Styler)
 
+    # GH 1723
+    def highlight_odd(index: pd.Series, /, color: str) -> list[str]:
+        return [f"color: {color}" if x % 2 else "" for x in index]
+
+    check(
+        assert_type(DF.style.apply_index(highlight_odd, color="purple"), Styler), Styler
+    )
+
 
 def test_map_index() -> None:
     def f(s: Scalar) -> str | None:
         return "background-color: yellow;" if s == "B" else None
 
     check(assert_type(DF.style.map_index(f), Styler), Styler)
+
+    def f1(s: Scalar, /, color: str) -> str | None:
+        return f"background-color: {color};" if s == "b" else None
+
+    check(assert_type(DF.style.map_index(f1, color="pink"), Styler), Styler)
 
 
 def test_background_gradient() -> None:
--- a/tests/test_styler.py
+++ b/tests/test_styler.py
@@ -77,11 +77,14 @@ def test_apply_index() -> None:
     check(assert_type(DF.style.apply_index(f1), Styler), Styler)
 
     # GH 1723
-    def highlight_odd(index: pd.Series, /, color: str) -> list[str]:
+    def highlight_odd(index: pd.Series, color: str) -> list[str]:
         return [f"color: {color}" if x % 2 else "" for x in index]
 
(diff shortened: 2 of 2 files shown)


pandas-stubs/io/formats/style.pyi
```diff
@@ -57,6 +57,11 @@ class _SeriesFunc(Protocol):
         self, series: Series, /, *args: Any, **kwargs: Any
     ) -> list[Any] | Series: ...
 
+class _SeriesStrFunc(Protocol):
```
loicdiridollou, line 60: I tried on my end and this protocol thing works but is a bit of overkill for what you are trying to do here. Turns out since `args` and `kwargs` here are just `Any`, you can drop it in exchange for:
`Callable[Concatenate[Series, ...]], list[str] | np_ndarray_str | Series[str]]`
This will make it a lot cleaner and simpler to understand.

tests/test_styler.py
```diff
@@ -76,13 +76,26 @@ def f1(s: Series) -> Series[str]:
 
     check(assert_type(DF.style.apply_index(f1), Styler), Styler)
 
+    # GH 1723
+    def highlight_odd(index: pd.Series, /, color: str) -> list[str]:
```
loicdiridollou, line 80: No need for the `/`, it should work without it since it is hardly a usecase you see in production code.

loicdiridollou (commented):
There is an opportunity to simplify here, the tests though are looking good!

loicdiridollou (changes requested):
See comments

loicdiridollou:
Hey @eosti,
Do you have time to respond to the feedback soon?
Thanks!

eosti:
Thanks for the ping, this completely slipped my mind! I'll take a look now.

pandas-stubs/io/formats/style.pyi
eosti: I tried swapping it out for the `Callable` that you mentioned and it fails on `mypy`:

```
tests/test_styler.py:72: error: Argument 1 to "apply_index" of "Styler" has incompatible type "Callable[[Series[Any]], ndarray[tuple[Any, ...], dtype[str_]]]"; expected "Callable[[[Series[Any], VarArg(Any), KwArg(Any)]], list[str] | ndarray[tuple[Any, ...], dtype[str_]] | Series[str]]"  [arg-type]
tests/test_styler.py:77: error: Argument 1 to "apply_index" of "Styler" has incompatible type "Callable[[Series[Any]], Series[str]]"; expected "Callable[[[Series[Any], VarArg(Any), KwArg(Any)]], list[str] | ndarray[tuple[Any, ...], dtype[str_]] | Series[str]]"  [arg-type]
tests/test_styler.py:84: error: Argument 1 to "apply_index" of "Styler" has incompatible type "Callable[[Series[Any], str], list[str]]"; expected "Callable[[[Series[Any], VarArg(Any), KwArg(Any)]], list[str] | ndarray[tuple[Any, ...], dtype[str_]] | Series[str]]"  [arg-type]
```

This is a bit of a complex way of doing it, but it aligns with what is used in the other `Styler` functions (see https://github.com/pandas-dev/pandas-stubs/pull/921) so I'm inclined to match that style.

tests/test_styler.py
eosti: The `/`s were added in response to https://github.com/pandas-dev/pandas-stubs/issues/919, but I tried it out without the `/`s and it worked fine... So I guess something changed in the last two years; I'll remove these.

pandas-stubs/io/formats/style.pyi
loicdiridollou: Sorry made a typo above, you should pass:
`Callable[Concatenate[Series, ...], list[str] | np_ndarray_str | Series[str]]`
(there was a bracket misplaced).
Let me know if that works. The use of Protocol is needed if you have a `/` in the function definition but since here it works without we should try to go with simplicity and readability if you agree

loicdiridollou: @Dr-Irv any guidance on which you see as the most appropriate way forward?

Dr-Irv: I ran the tests with the current version of this PR, so I think this should be acceptable.  @loicdiridollou do final review.  

It's possible that @eosti wasn't using mypy 2.0.0 or higher.  The tests in CI were run with mypy 2.1.

loicdiridollou: Yes this works, I was just looking for a simpler solution (suggested above), compared to the Protocol class.
Eventually let's not block this one, this works well and should fix users errors.

loicdiridollou (approved):
Nice one @eosti, thanks for your contribution!

PR #1728, opened 2026-04-17
MNT: Add more defaults to the stubs

loicdiridollou:

About 500 instances of ellipsis removed and replaced with the proper default value so help with IDE hovering.
It remains around 1900 instances of ellipsis, we won't be able to clean them all since some use `lib.no_default` but we should be able to do some good progress there.


pandas-stubs/_libs/tslibs/timedeltas.pyi
```diff
@@ -100,7 +100,7 @@ class Timedelta(timedelta):
     def __new__(
         cls,
         value: str | float | Timedelta | Tick | timedelta | np.timedelta64 = ...,
-        unit: TimeDeltaUnitChoices = ...,
+        unit: TimeDeltaUnitChoices = "ns",
```
cmp0xff: In `Timedelta.__new__`, the `unit` parameter actually defaults to `None` at runtime, although the docstring mentions `"ns"`. Using `None` allows the constructor to handle input resolution correctly.

**Ground Truth:** [timedeltas.pyx#L2147](https://github.com/pandas-dev/pandas/blob/main/pandas/_libs/tslibs/timedeltas.pyx#L2147)

Gemini 3 created, I verified

pandas-stubs/io/sql.pyi
```diff
@@ -42,11 +42,11 @@ _SQLStatement: TypeAlias = (
...
-    parse_dates: list[str] | dict[str, str] | dict[str, dict[str, Any]] | None = ...,
-    columns: list[str] | None = ...,
+    schema: str | None = None,
+    index_col: str | list[str] | None = None,
+    coerce_float: bool = True,
+    parse_dates: list[str] | dict[str, str] | dict[str, dict[str, Any]] | None = None,
+    columns: list[str] | None = None,
     *,
     chunksize: int,
     dtype_backend: DtypeBackend | NoDefault = ...,
```
cmp0xff, line 52: The `read_sql_table` function does not enforce keyword-only arguments for `chunksize` in the `pandas` 3.0 source. The `*` should be removed to maintain positional compatibility.

**Ground Truth:** [io/sql.py#L257](https://github.com/pandas-dev/pandas/blob/main/pandas/io/sql.py#L257)

Gemini 3 created, I verified

pandas-stubs/plotting/_core.pyi
```diff
@@ -137,37 +137,37 @@ class PlotAccessor:
         kind: Literal["pie"],
         ax: Axes | None = None,
         subplots: Literal[False] | None = False,
-        sharex: bool = ...,
-        sharey: bool = ...,
-        layout: tuple[int, int] = ...,
-        figsize: tuple[float, float] = ...,
+        sharex: bool = False,
```
cmp0xff: For `sharex`, `sharey`, and `legend`, the runtime often uses `None` as a default to trigger context-aware logic (e.g., `sharex` depends on whether an `ax` is passed). Setting these to `False` in the stubs might be misleading. I suggest using `None` where the runtime signature does.

**Ground Truth:** [plotting/_core.py#L1112](https://github.com/pandas-dev/pandas/blob/main/pandas/plotting/_core.py#L1112)

Signed, Gemini 3

pandas-stubs/core/arrays/timedeltas.pyi
```diff
@@ -34,37 +35,27 @@ class TimedeltaArray(TimelikeOps):
...
+        keepdims: bool = False,
+        initial: object | None = None,
+        skipna: bool = True,
+        min_count: int = 0,
+    ) -> Timedelta: ...
+    def std(
+        self,
+        *,
+        axis: AxisInt | None = None,
+        dtype: DtypeArg | None = None,
```
cmp0xff, line 79: These reduction methods (`sum`, `std`, etc.) can return `NaT` (represented as `NaTType`) when `skipna=False` or on empty inputs. The return type should be expanded to `Timedelta | NaTType`.

**Ground Truth:** [core/arrays/timedeltas.py#L367](https://github.com/pandas-dev/pandas/blob/main/pandas/core/arrays/timedeltas.py#L367) (via `_box_func`)

Gemini 3 created, I verified

pandas-stubs/io/json/_json.pyi
```diff
@@ -32,199 +32,199 @@ from pandas._typing import (
...
     chunksize: int,
-    compression: CompressionOptions = ...,
-    nrows: int | None = ...,
-    storage_options: StorageOptions = ...,
+    compression: CompressionOptions = "infer",
+    nrows: int | None = None,
+    storage_options: StorageOptions = None,
     dtype_backend: DtypeBackend | NoDefault = ...,
-    engine: Literal["ujson"] = ...,
+    engine: Literal["ujson"] = "ujson",
```
cmp0xff, line 54: The `engine` parameter supports more than just `"ujson"`. It should use the `JSONEngine` type alias to include `"pyarrow"`.

**Ground Truth:** [io/json/_json.py#L535](https://github.com/pandas-dev/pandas/blob/main/pandas/io/json/_json.py#L535)

Gemini 3 created, I verified

pandas-stubs/plotting/_core.pyi
cmp0xff: Edit: maybe github.com/pandas-dev/pandas/blob/main/pandas/plotting/_core.py#L1050, where it says `None` for `sharex`, but `False` for `sharey`. This is confusing.

cmp0xff (changes requested):

pandas-stubs/core/arrays/timedeltas.pyi
Dr-Irv: I don't think we should return `Timedelta | NaTType` in this case because it will force users to do a `cast` when they know the result is valid.  I'd be OK with an overload with `skipna=False` that returned `Timedelta | NaTType`

pandas-stubs/io/sql.pyi
loicdiridollou: This is here so we can pass the chunksize without the position (chunksize does not have a default here).

pandas-stubs/plotting/_core.pyi
loicdiridollou: Quite confusing indeed, I have realigned them

pandas-stubs/io/json/_json.pyi
loicdiridollou: The `pyarrow` engine is used in another overload so I believe we are good here

tests/__init__.py
```diff
@@ -39,6 +39,7 @@
 LINUX = sys.platform == "linux"
 WINDOWS = sys.platform in {"win32", "cygwin"}
 MAC = sys.platform == "darwin"
+PD_LTE_31 = Version(pd.__version__) < Version("3.0.99")
```
cmp0xff, line 42: Maybe we'll need to always keep such a thing for the next major version

cmp0xff (approved):
Thank you @loicdiridollou !

PR #1730, opened 2026-04-21
MNT: `pyrefly`, `ruff`, GitHub action, type checker versions

cmp0xff:
- Update `pyrefly` ignores
- Update `ruff`
- Update GitHub action
- Bump the versions of `mypy`, `pyrefly` and `ty`
- ~~[ ] Closes #xxxx (Replace xxxx with the Github issue number)~~
- ~~[ ] Tests added (Please use `assert_type()` to assert the type of any return value)~~
- ~~[ ] If I used AI to develop this pull request, I prompted it to follow `AGENTS.md`.~~


cmp0xff:
Hi @Dr-Irv , I remember you had Windows. Could you help me to check? The branch passes `poe test_all` on my local Windows machine with python 3.13. However the CI pipelines fail, also the ones on Windows. Thank you!

loicdiridollou:
Are you sure about removing all the ignore for `pyrefly` with the bad-specialization? I have pyrefly 0.0.62 on my side and it does not say that those should be removed.

cmp0xff:
> Are you sure about removing all the ignore for `pyrefly` with the bad-specialization? I have pyrefly 0.0.62 on my side and it does not say that those should be removed.

That's the weird thing, I ran `poetry update` then `poetry run pyrefly check --remove-unused-ignores` and here are the results, from my local machine, which do not agree with those from the CI pipeline.

Dr-Irv:
> Hi @Dr-Irv , I remember you had Windows. Could you help me to check? The branch passes `poe test_all` on my local Windows machine with python 3.13. However the CI pipelines fail, also the ones on Windows. Thank you!

What I am seeing is that the CI ran with pyrefly 0.61.0.  On my machine, that also fails.  But if you upgrade to 0.62.0, then it passes pyrefly.  So update `pyproject.toml` in your branch to constraint pyrefly to `>= 0.62` and see if that fixes things.

cmp0xff:
> What I am seeing is that the CI ran with pyrefly 0.61.0. On my machine, that also fails. But if you upgrade to 0.62.0, then it passes pyrefly. So update `pyproject.toml` in your branch to constraint pyrefly to `>= 0.62` and see if that fixes things.

Thank you, I thought CI would use the latest type checker versions, because we do not upload `poetry.lock`, and every time CI would lock from scratch, getting the newest version. Apparently that is not the case.

loicdiridollou (approved):
Looks good, thanks @cmp0xff !

PR #1732, opened 2026-04-22
BUG: Fix Index.droplevel() stub — use Sequence[Never] to allow only droplevel([]) (GH#1678)

### LINKED ISSUE #1678: `Index.droplevel()` should not be permitted
**Describe the bug**
`Index.droplevel()` is not supported.

So if a user knows that they have a `MultiIndex` (e.g., `df.index.droplevel()`, they should be forced to do a cast.

**To Reproduce**
```python
>>> import pandas as pd
>>> df = pd.DataFrame({"x":[1,2]})
>>> df
   x
0  1
1  2
>>> df.columns.droplevel([0])
  File "<python-input-3>", line 1, in <module>
    df.columns.droplevel([0])
    ~~~~~~~~~~~~~~~~~~~~^^^^^
  File "C:\Condadirs\envs\base313\Lib\site-packages\pandas\core\indexes\base.py", line 2301, in droplevel
    return self._drop_level_numbers(levnums)
           ~~~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^
  File "C:\Condadirs\envs\base313\Lib\site-packages\pandas\core\indexes\base.py", line 2312, in _drop_level_numbers
    raise ValueError(
    ...<2 lines>...
    )
ValueError: Cannot remove 1 levels from an index with 1 levels: at least one level must be left.
```

So it doesn't work at runtime

 - OS: Windows 11
 - OS Version 2H25
 - python version   3.13
 - version of type checker   pyright 1.1.408 - it allows the above
 - version of installed `pandas-stubs`  - dev version

**Additional context**

We could not fix this because it forces a `cast` if you do something like this.  But I think this is helpful to users to make them acknowledge that the corresponding index *is* a `MultiIndex`

tinezivic:
- [x] Closes #1678

## Problem

`Index.droplevel()` was defined in the `Index` base stub, but calling it on a plain `Index` with any non-empty argument **always** raises `ValueError` at runtime. A plain `Index` has exactly one level — you cannot remove it.

Type checkers accepted `idx.droplevel(0)` without complaint, misleading users into thinking this call is safe.

## Fix

Use `Sequence[Never]` as the parameter type for `Index.droplevel()`:

```python
# pandas-stubs/core/indexes/base.pyi
def droplevel(self, level: Sequence[Never]) -> Self: ...
```

`Sequence[Never]` means "a sequence with no possible elements" — in practice, only the empty list `[]` satisfies this type. Both mypy and pyright (verified with actual tests) correctly:

- **allow** `idx.droplevel([])` — which is a documented no-op at runtime
- **reject** `idx.droplevel()`, `idx.droplevel(0)`, `idx.droplevel([0])`, `idx.droplevel("name")` — all of which raise `ValueError`

This is more precise than deleting the method entirely, because `droplevel([])` is documented behavior.

## Changes

- **`base.pyi`**: add `droplevel(level: Sequence[Never]) -> Self`
- **`multi.pyi`**: `droplevel` unchanged; `# type: ignore[override]` stays because the return type `MultiIndex | Index` is wider than `Self` (needed for correctness, unrelated to our change)
- **`tests/indexes/test_indexes.py`**: add `idx.droplevel([])` as a valid usage; expand `TYPE_CHECKING_INVALID_USAGE` block with four invalid cases (`droplevel()`, `droplevel(0)`, `droplevel([0])`, `droplevel("name")`)

## Why Sequence[Never] works

Both mypy and pyright treat `[]` as assignable to `Sequence[Never]` (an empty sequence has no elements, so the element type constraint is vacuously satisfied), but reject any non-empty sequence. This is the most accurate annotation for the actual runtime contract on a plain `Index`.


tests/indexes/test_indexes.py
```diff
@@ -1608,15 +1608,16 @@ def test_index_set_names() -> None:
 
 
 def test_index_droplevel() -> None:
-    idx = pd.Index([1, 2])
-    check(assert_type(idx.droplevel([]), "pd.Index[int]"), pd.Index, np.integer)
```
loicdiridollou, line 1612: This behavior is allowed and documented in https://pandas.pydata.org/docs/dev/reference/api/pandas.Index.droplevel.html#pandas.Index.droplevel

Can you keep the tests and the signature?

loicdiridollou (changes requested):
`pd.Index().droplevel([])` does not raise an error and I believe should be kept, @Dr-Irv what do you think?

Dr-Irv:
> `pd.Index().droplevel([])` does not raise an error and I believe should be kept, @Dr-Irv what do you think?

Well, I'm the one that created the issue, and that means it happened in code my team was working on, so I'm fine with having the type checker say that `pd.Index().droplevel([])` is invalid.

loicdiridollou:
Sorry not clear, my argument was that `pd.Index().droplevel([])` is a valid code but `pd.Index().droplevel([0])` is not so my point was if we should type `Sequence[Never]` as valid and leave the rest as is in the PR

Dr-Irv:
> Sorry not clear, my argument was that `pd.Index().droplevel([])` is a valid code but `pd.Index().droplevel([0])` is not so my point was if we should type `Sequence[Never]` as valid and leave the rest as is in the PR

From what I've read, type checkers are not able to distinguish between empty sequences and non-empty sequences.  But if you can get it to work, go ahead.

tinezivic:
Hey @loicdiridollou — you were right, I tested it and `Sequence[Never]` actually works here. Both mypy and pyright allow `droplevel([])` and reject `droplevel(0)`, `droplevel([0])`, `droplevel("name")` — exactly the behavior we want.

I updated the PR: `Index.droplevel` is back in `base.pyi` with `Sequence[Never]` as the parameter type, and the tests now cover both the valid no-op case (`droplevel([])`) and the three invalid cases in `TYPE_CHECKING_INVALID_USAGE`.

The `# type: ignore[override]` on `MultiIndex.droplevel` is also back — turns out it was never about argument types, it's needed because the return type `MultiIndex | Index` is wider than `Self`. Sorry for dropping it earlier without understanding why it was there.

pandas-stubs/core/indexes/base.pyi
```diff
@@ -452,6 +451,11 @@ class Index(IndexOpsMixin[S1], ElementOpsMixin[S1]):
     def dropna(self, how: AnyAll = "any") -> Self: ...
     def unique(self, level: Hashable | None = None) -> Self: ...
     def drop_duplicates(self, *, keep: DropKeep = "first") -> Self: ...
+    # droplevel() always raises ValueError on a plain (single-level) Index
+    # unless you pass an empty sequence — then it's a no-op.
+    # Use Sequence[Never] so type checkers allow droplevel([]) but reject
+    # anything else (droplevel(0), droplevel([0]), droplevel("name"), ...).
```
cmp0xff: Please drop these comments

tests/indexes/test_indexes.py
```diff
@@ -1608,15 +1608,22 @@ def test_index_set_names() -> None:
...
     mi = pd.MultiIndex.from_arrays([[1, 2, 3], [4, 5, 6]], names=["elk", "owl"])
     check(assert_type(mi.droplevel([]), pd.MultiIndex | pd.Index), pd.MultiIndex)
     check(assert_type(mi.droplevel([0]), pd.MultiIndex | pd.Index), pd.Index)
     check(assert_type(mi.droplevel((0,)), pd.MultiIndex | pd.Index), pd.Index)
     check(assert_type(mi.droplevel(["elk"]), pd.MultiIndex | pd.Index), pd.Index)
     check(assert_type(mi.droplevel(("elk",)), pd.MultiIndex | pd.Index), pd.Index)
     check(assert_type(mi.droplevel(0), pd.MultiIndex | pd.Index), pd.Index)
+    # droplevel([]) on a plain Index is a documented no-op — returns the same Index
+    idx = pd.Index([1, 2])
+    check(assert_type(idx.droplevel([]), "pd.Index[int]"), pd.Index)
```
cmp0xff: Can you drop the comment and move the unchanged lines back to their original place?

tinezivic:
Done, removed all comments.

pandas-stubs/core/indexes/base.pyi
tinezivic: Dropped the comments as suggested.

tests/indexes/test_indexes.py
tinezivic: Done, dropped the comment and moved the lines back to their original position.

tests/indexes/test_indexes.py
```diff
@@ -1609,14 +1609,18 @@ def test_index_set_names() -> None:
 
 def test_index_droplevel() -> None:
     idx = pd.Index([1, 2])
-    check(assert_type(idx.droplevel([]), "pd.Index[int]"), pd.Index, np.integer)
+    check(assert_type(idx.droplevel([]), "pd.Index[int]"), pd.Index)
```
loicdiridollou: Please keep the subtype of the Index (`np.integer` here)

pandas-stubs/core/indexes/multi.pyi
```diff
@@ -140,7 +140,7 @@ class MultiIndex(Index):
     @property
     def is_monotonic_decreasing(self) -> bool: ...
     def dropna(self, how: AnyAll = "any") -> Self: ...
-    def droplevel(self, level: Level | Sequence[Level] = 0) -> MultiIndex | Index: ...  # type: ignore[override] # pyrefly: ignore[bad-override]
+    def droplevel(self, level: Level | Sequence[Level] = 0) -> MultiIndex | Index: ...  # type: ignore[override]  # return type is wider than Self
```
loicdiridollou: Seems like the `pyrefly` ignore got dropped.
Also if you want to test locally feel free to run`poetry run poe test_all` to replicate the CI process.

loicdiridollou (changes requested):
On minor thing, rest looks good, check CI I kicked it to make sure it passes or not

pandas-stubs/core/indexes/base.pyi
```diff
@@ -452,6 +451,7 @@ class Index(IndexOpsMixin[S1], ElementOpsMixin[S1]):
     def dropna(self, how: AnyAll = "any") -> Self: ...
     def unique(self, level: Hashable | None = None) -> Self: ...
     def drop_duplicates(self, *, keep: DropKeep = "first") -> Self: ...
+    def droplevel(self, level: Sequence[Never] = ...) -> Self: ...
```
cmp0xff:
```suggestion
    def droplevel(self, level: Sequence[Never]) -> Self: ...
```

tests/indexes/test_indexes.py
```diff
@@ -1617,6 +1617,10 @@ def test_index_droplevel() -> None:
     check(assert_type(mi.droplevel(["elk"]), pd.MultiIndex | pd.Index), pd.Index)
     check(assert_type(mi.droplevel(("elk",)), pd.MultiIndex | pd.Index), pd.Index)
     check(assert_type(mi.droplevel(0), pd.MultiIndex | pd.Index), pd.Index)
+    if TYPE_CHECKING_INVALID_USAGE:
+        idx.droplevel(0)  # type: ignore[arg-type] # pyright: ignore[reportArgumentType]
```
cmp0xff, line 1622:
```suggestion
        idx.droplevel()  # type: ignore[arg-type] # pyright: ignore[reportArgumentType]
        idx.droplevel(0)  # type: ignore[arg-type] # pyright: ignore[reportArgumentType]
```

cmp0xff (changes requested):

tests/indexes/test_indexes.py
tinezivic: Index subtype (np.integer) is kept.

tests/indexes/test_indexes.py
tinezivic: Both kept! `Sequence[Never]` is exactly what models this — `[]` passes fine (nothing to violate the constraint), while `droplevel()`, `droplevel(0)`, `droplevel([0])`, and `droplevel("name")` all get flagged. Verified locally with `poetry run poe test`, all green.

pandas-stubs/core/indexes/multi.pyi
tinezivic: Restored in the latest commit. And thanks for the tip — ran poetry run poe test locally, all green (mypy: 0 errors, pyright: 0 errors, pytest: 3198 passed).

cmp0xff (approved):
Looks good to me. I would not move `droplevel` from L434 to L454, but it's not a bit deal.

I'll hand over to @loicdiridollou to merge.

loicdiridollou (approved):
All good, thanks for your contribution @tinezivic !

Issue #1415, opened 2025-10-09
How should we type `Series` with a categorical dtype?


cmp0xff:
After pandas-dev/pandas-stubs#1383 and pandas-dev/pandas-stubs#1391, `pd.Series([1], dtype="category")` gives `Series[CategoricalDtype]`.

Is it the best idea to follow? Typically we use the type of elements as the `Generic` type of a `Series` or `Index`, for example `Series[int]`, instead of `Series[Int64Dtype]`. The current approach seems inconsistent to me.

This is also related to pandas-dev/pandas-stubs#1395, where
- `Series([1], dtype=int)`: `dtype` is `np.dtype("int64")`, element type is `np.int64`. Currently this is `Series[int]` in `pandas-stubs`.
- `Series([1], dtype="Int64")`: `dtype` is `pd.Int64Dtype`, element type is (built-in) `int`. Currently this is `Series[Any]` in `pandas-stubs`.

We will discuss the last part in pandas-dev/pandas-stubs#1395, here it is just an illustrative example.

PR #1733, opened 2026-04-25
TYP: add CategoricalSeries type alias for Series[CategoricalDtype] (#1415)

MukundaKatta:
## Summary
Refs #1415. Adds the additive piece of "how should we type Series with categorical dtype?" — a `CategoricalSeries: TypeAlias = Series[CategoricalDtype]` declared under `TYPE_CHECKING` in `pandas-stubs/_typing.pyi`.

## Design choice
The broader question — re-parameterizing `Series` by element type, or introducing a `CategoricalItem[C1]` generic mirroring `Interval` — was discussed in the issue thread but is genuinely open. Re-parameterizing would break every existing user that wrote `Series[CategoricalDtype]`. This PR ships only the additive subset: a short, discoverable spelling. Maintainers can layer the generic design on top later without breaking anything here.

## Files changed
- `pandas-stubs/_typing.pyi` — alias under `TYPE_CHECKING` (Series is not subscriptable at runtime; mirrors `MaskType`/`IndexType` pattern).
- `tests/series/categorical/test_dtypes.py` — 3 new tests: assignment compatibility, cat accessor typing, copy-preserves-parameterization.

## Test plan
- ruff check, black --check, isort --check-only — all clean
- pyright — 0 errors on changed files
- pytest — 5/5 categorical tests pass

### DIFF
--- a/pandas-stubs/_typing.pyi
+++ b/pandas-stubs/_typing.pyi
@@ -531,6 +531,14 @@ BytesDtypeArg: TypeAlias = (
     BuiltinBytesDtypeArg | NumpyBytesDtypeArg | PyArrowBytesDtypeArg
 )
 CategoryDtypeArg: TypeAlias = CategoricalDtype | Literal["category"]
+# Alias for a Series with a categorical dtype. The element-type of a categorical
+# Series is not statically known (it can hold any hashable categories), so we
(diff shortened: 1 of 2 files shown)


Dr-Irv:
We don't add things in the stubs that are syntactic sugar, which is what this PR provides.

PR #1736, opened 2026-05-08
GH1727 Narrow Series.to_numpy return types based on the dtype argument

### LINKED ISSUE #1727: Missing specialization for to_numpy for Series[float] and Series[int]
For Series[float] and Series[int], the overloaded to_numpy functions are not expecting a np.dtype for the dtype arg. Maybe overloads missing ?

**To Reproduce**
Pyright will report an error for this line

`pd.Series(np.empty(0, np.dtype("f8"))).to_numpy(dtype=np.dtype("f8"))`

loicdiridollou:
- [x] Closes #1727

Added the following deprecation:
- [x] Enforced deprecation of axis=None acting the same as axis=0 in the DataFrame reductions sum, prod, std, var, and sem, passing axis=None will now reduce over both axes; this is particularly the case when doing e.g. numpy.sum(df) (https://github.com/pandas-dev/pandas/issues/21597)


pandas-stubs/core/series.pyi
```diff
@@ -4592,6 +4607,70 @@ class Series(IndexOpsMixin[S1], ElementOpsMixin[S1], NDFrame):
...
+    @overload
+    def to_numpy(
+        self,
+        dtype: NumpyObjectDtypeArg | BuiltinObjectDtypeArg,
+        copy: bool = False,
+        na_value: Scalar = ...,
+        **kwargs: Any,
+    ) -> np_1darray_object: ...
+    @overload
+    def to_numpy(
```
loicdiridollou, line 4704: Not totally sure if this one is really in the spirit of the stubs since we have all the other `np_1darray_xxx` but thought it was worth offering since once we do the to_numpy it goes into the numpy ground.

tests/series/test_series.py
```diff
@@ -1986,6 +1992,218 @@ def test_to_numpy() -> None:
...
+    )
+    check(
+        assert_type(s_date.to_numpy(dtype=np.dtype("i8")), np_1darray[np.int64]),
+        np_1darray,
+        np.integer,
+    )
+    check(
+        assert_type(s_period.to_numpy(dtype=np.dtype("O")), np_1darray[np.object_]),
+        np_1darray,
+        pd.Period,
```
cmp0xff:
```suggestion
        assert_type(s_period.to_numpy(dtype=np.dtype("O")), np_1darray_object),
        np_1darray_object,
        pd.Period,
```

tests/series/test_series.py
```diff
@@ -1917,9 +1918,14 @@ def test_types_to_numpy() -> None:
 
     check(assert_type(pd.Series().to_numpy(), np_1darray), np_1darray)
 
+    check(
+        assert_type(
+            pd.Series(np.empty(0, np.dtype("f8"))).to_numpy(dtype=np.dtype("f8")),
+            np_1darray[np.float64],
+        ),
+        np_1darray[np.float64],  # parameterized form checks dtype without iterating
```
cmp0xff: In the comment. do you suggest the following won't work?
```suggestion
        np_1darray[np.float64],  # parameterized form checks dtype without iterating
        np.floating,
```

pandas-stubs/core/series.pyi
```diff
@@ -4592,6 +4607,70 @@ class Series(IndexOpsMixin[S1], ElementOpsMixin[S1], NDFrame):
         **kwargs: Any,
     ) -> np_1darray_bytes: ...
     @overload
+    def to_numpy(
+        self,
+        dtype: NumpyFloat16DtypeArg | NumpyFloatNot16DtypeArg | BuiltinFloatDtypeArg,
+        copy: bool = False,
+        na_value: Scalar = ...,
+        **kwargs: Any,
+    ) -> np_1darray_float: ...
```
cmp0xff, line 4654: Would it make more sense, if this one is located directly below the overload of `to_numpy` that also leads to `np_1darray_float` but without a `dtype`?

tests/series/test_series.py
```diff
@@ -1986,6 +1992,218 @@ def test_to_numpy() -> None:
...
+    )
+
+    # dtype type objects -> np_1darray[T] via type[GenericT] overload
+    check(
+        assert_type(s_timedelta.to_numpy(dtype=np.float64), np_1darray[np.float64]),
+        np_1darray,
+        np.floating,
+    )
+    check(
+        assert_type(s_date.to_numpy(dtype=np.int64), np_1darray[np.int64]),
```
cmp0xff, line 2081: Maybe we can be more systematic here, for example I don't see `np.int8` or `np.int16` or `np.uint64` and related symbols are located far away. New issue is welcome.

cmp0xff (changes requested):

tests/series/test_series.py
loicdiridollou: This is just to say that we don't need to operate on the iteration of the results to check the type of what is inside.
If you add the floating you get a StopIteration error so will clarify the comment.

tests/series/test_series.py
loicdiridollou: Added more

pandas-stubs/core/series.pyi
loicdiridollou: I was unable to move it as it created an issue with the `sr_bool.to_numpy(dtype=bool)` would match to a float

tests/series/test_series.py
```diff
@@ -1986,6 +1994,248 @@ def test_to_numpy() -> None:
         np.timedelta64,
     )
 
+    # np.dtype[T] instance -> np_1darray[T] via generic overload
+    check(
+        assert_type(s_timedelta.to_numpy(dtype=np.dtype("f8")), np_1darray[np.float64]),
```
cmp0xff: I still believe we are missing systematicality here. when I look for `"f4"`, there is no match. And `dtype=np.int64` and `dtype=np.int32` are far away from each other (below).

Feel free to postpone the idea in a separate issue or challenge it.

cmp0xff (changes requested):

cmp0xff (changes requested):
In my example in https://github.com/pandas-dev/pandas-stubs/issues/1727#issuecomment-4261104630, `np.dtypes.Float64DType()` and `np.dtypes.Float64DType` etc should also work. Currently they are missing from the tests.

loicdiridollou:
Added a lot more tests to fit, let me know what you think!

tests/series/test_series.py
```diff
@@ -1986,6 +1994,575 @@ def test_to_numpy() -> None:
...
+    )
+
+    # --- Object ---
+    check(
+        assert_type(s_td_small.to_numpy(dtype=object), np_1darray_object),
+        np_1darray_object,
+    )
+    check(
+        assert_type(s_td_small.to_numpy(dtype=np.object_), np_1darray[np.object_]),
+        np_1darray,
```
cmp0xff: This one is different from other ones. I'd prefer

```suggestion
        assert_type(s_td_small.to_numpy(dtype=np.object_), np_1darray_object),
        np_1darray_object,
```

tests/series/test_series.py
```diff
@@ -1986,6 +1994,575 @@ def test_to_numpy() -> None:
...
+        np_1darray_object,
+    )
+    check(
+        assert_type(s_td_small.to_numpy(dtype="object_"), np_1darray_object),
+        np_1darray_object,
+    )
+    check(
+        assert_type(s_interval.to_numpy(dtype="O"), np_1darray_object),
+        np_1darray,
+        pd.Interval,
```
cmp0xff, line 2378:
```suggestion
        assert_type(s_interval.to_numpy(dtype="O"), np_1darray_object),
        np_1darray_object,
        pd.Interval,
```

tests/series/test_series.py
```diff
@@ -1986,6 +1994,575 @@ def test_to_numpy() -> None:
...
+    )
+    check(
+        assert_type(s_period.to_numpy(dtype=np.dtype("O")), np_1darray[np.object_]),
```
cmp0xff:
```suggestion
        assert_type(s_period.to_numpy(dtype=np.dtype("O")), np_1darray_object),
```

cmp0xff (changes requested):
Just a few small changes!

tests/series/test_series.py
loicdiridollou: All fixed, should be good now!

cmp0xff (approved):
Good, the PR should make `to_numpy` typed much better. Thank you @loicdiridollou !

PR #1745, opened 2026-05-17
Read-only (covariant) list parameter annotations

### LINKED ISSUE #153: BUG: set_index rejects `List[str]`

hamdanal:
- [x] Closes #153 (already closed but similar issues still exist)
- [x] No AI used

Some pandas functions expect a list as argument but not any sequence. Because `list` is invariant, parameters annotated with `list[float]` will not accept `list[int]`, for examlpe, even though we know that the list is read-only and will not be modified. To fix this, I propose the following trick using the `CovariantList` protocol. This protocol can replace `list` annotations in parameter positions (never use as a return type, instead keep using `list` there).

Note that this problem has manifested before and one trick you use today is to parametrize the list with a type variable that has an appropriate upper bound (for example you use `list[HashableT]` where you want a list of hashable objects). This is cumbersome because for every type of list element, you'll need to define a corresponding type variable. Also, there are some places where a list of a very wide type like a union is still used (for example `list[AggFuncTypeBase]` in `DataFrame.aggregate`). All of these can be replaced with the `CovariantList` protocol.

In this PR I only introduce the protocol and change one method to use it to demonstrate its value. I think it would be better to incrementally use this protocol in follow up PR(s) for easier work and review. For example a PR could replace all `list[HashableT]` and `list[Hashable]` in parameter positions by `CovariantList[Hashable]` then other uses could be audited and changed in separate follow-up PRs.

### DIFF
--- a/pandas-stubs/_typing.pyi
+++ b/pandas-stubs/_typing.pyi
@@ -17,6 +17,7 @@ import sys
 from typing import (
     TYPE_CHECKING,
     Any,
+    ClassVar,
     Generic,
     Literal,
     Protocol,
@@ -1231,12 +1232,13 @@ class Just(Protocol, Generic[T]):
     def __class__(self, t: type[T], /) -> None: ...
 
 class CovariantList(Protocol[_T_co]):
+    __hash__: ClassVar[None]  # type: ignore[assignment] # pyright: ignore[reportIncompatibleMethodOverride]
     @property  # type: ignore[override]
     def __class__(self) -> type[list[Any]]: ...  # pyrefly: ignore[bad-override]
     @__class__.setter
-    def __class__(
+    def __class__(  # pyright: ignore[reportIncompatibleMethodOverride]
         self, value: type[list[Any]], /
-    ) -> None: ...  # pyright: ignore[reportIncompatibleMethodOverride]
+    ) -> None: ...
     def __iter__(self) -> Iterator[_T_co]: ...
     # copy() is only TEMPORARILY needed because `__class__` is a property
     # and ty doesn't support property protocol members. Remove when
--- a/pandas-stubs/_typing.pyi
+++ b/pandas-stubs/_typing.pyi
@@ -1230,6 +1230,19 @@ class Just(Protocol, Generic[T]):
     @override
     def __class__(self, t: type[T], /) -> None: ...
 
+class CovariantList(Protocol[_T_co]):
+    @property  # type: ignore[override]
+    def __class__(self) -> type[list[Any]]: ...  # pyrefly: ignore[bad-override]
+    @__class__.setter
+    def __class__(
+        self, value: type[list[Any]], /
+    ) -> None: ...  # pyright: ignore[reportIncompatibleMethodOverride]
+    def __iter__(self) -> Iterator[_T_co]: ...
+    # copy() is only TEMPORARILY needed because `__class__` is a property
+    # and ty doesn't support property protocol members. Remove when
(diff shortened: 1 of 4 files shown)


tests/test_typing.py
```diff
@@ -36,3 +40,21 @@ def test_dtype_arg_aliases(dtype_arg: Any, alias_map: Mapping[Any, Any]) -> None
...
+    bad2: list[str] = ["a", "b", "c"]  # Error, list[str] !< list[float]
+    bad3: list[object] = [1, "a", 3.0]  # Error, list[object] !< list[float]
+    bad4: float = 1.0  # Error, float !< list[float]
+
+    f(good1)
+    f(good2)
+    f(bad1)  # type: ignore[arg-type] # pyright: ignore[reportArgumentType] # pyrefly: ignore[bad-argument-type] # ty: ignore[invalid-argument-type]
+    f(bad2)  # type: ignore[arg-type] # pyright: ignore[reportArgumentType] # pyrefly: ignore[bad-argument-type] # ty: ignore[invalid-argument-type]
+    f(bad3)  # type: ignore[arg-type] # pyright: ignore[reportArgumentType] # pyrefly: ignore[bad-argument-type] # ty: ignore[invalid-argument-type]
+    f(bad4)  # type: ignore[arg-type] # pyright: ignore[reportArgumentType] # pyrefly: ignore[bad-argument-type] # ty: ignore[invalid-argument-type]
```
Dr-Irv: Can you surround these 4 lines with `if TYPE_CHECKING_INVALID_USAGE`

pandas-stubs/_typing.pyi
```diff
@@ -1230,6 +1231,20 @@ class Just(Protocol, Generic[T]):
     @override
     def __class__(self, t: type[T], /) -> None: ...
 
+class CovariantList(Protocol[_T_co]):
```
Dr-Irv, line 1235: If you picked this up from somewhere, can you add a comment that refers to that repo?

Otherwise add a comment referring to this PR (`# GH 1745`)

Dr-Irv (changes requested):
thanks.  This looks like an interesting idea!

pandas-stubs/_typing.pyi
hamdanal: This is my idea, added a comment referring to this PR

tests/test_typing.py
hamdanal: Done

Dr-Irv (approved):
thanks @hamdanal

PR #1746, opened 2026-05-17
Modernize project settings in pyproject.toml

hamdanal:
- Use standards compliant `project` table supported since [poetry 2.0](https://github.com/python-poetry/poetry/releases/tag/2.0.0) released in Jan 2025 and update the poetry version in the documentation
- Remove redundant python version specifiers for tools that support the `project.requires-python` field, namely [black](https://black.readthedocs.io/en/stable/usage_and_configuration/the_basics.html#command-line-options), [ruff](https://docs.astral.sh/ruff/settings/#target-version), [ty](https://docs.astral.sh/ty/python-version/#python-version)
- Remove the [deprecated](https://packaging.python.org/en/latest/specifications/pyproject-toml/#classifiers) license classifier

I verified that `poetry build` produces the correct wheel and source distributions

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -77,7 +77,6 @@ build-backend = "poetry.core.masonry.api"
 
 [tool.poetry]
 packages = [{ include = "pandas-stubs" }]
-exclude = ["pandas-stubs/__init__.py"]
 
 [tool.poe.tasks.test_all]
 help = "Run all tests"
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,17 +1,15 @@
-[tool.poetry]
+[project]
 name = "pandas-stubs"
 version = "3.0.0.260204"
 description = "Type annotations for pandas"
-authors = ["The Pandas Development Team <pandas-dev@python.org>"]
+authors = [{name = "The Pandas Development Team", email = "pandas-dev@python.org"}]
 license = "BSD-3-Clause"
 readme = "README.md"
-homepage = "https://pandas.pydata.org"
-repository = "https://github.com/pandas-dev/pandas-stubs"
+requires-python = '>=3.11'
 classifiers = [
   "Development Status :: 5 - Production/Stable",
   "Environment :: Console",
   "Intended Audience :: Science/Research",
-  "License :: OSI Approved :: BSD License",
   "Operating System :: OS Independent",
   "Programming Language :: Python",
   "Programming Language :: Python :: 3",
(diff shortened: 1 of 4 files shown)


pyproject.toml
```diff
@@ -23,61 +21,64 @@ classifiers = [
...
+  "types-python-dateutil>=2.8.19",
+  "beautifulsoup4>=4.14.2",
+  "html5lib>=1.1",
+  "python-calamine>=0.2.0",
+  "pyarrow-stubs>=20.0.0.20250928;python_version<'4'",
+  "fsspec>=2025.10.0",
+  "tables>=3.10.2",
+  "pytz>=2025.2",
+  "types-pytz>=2022.1.1",
+  "pyiceberg>=0.10.0",
```
loicdiridollou, line 71: This is the reason your CI is failing (on main it is `^0.10.0` meaning it caps to 0.10.99) and the `tests/test_io.py` is not raising a warning.

loicdiridollou (changes requested):
One minor thing but it is a good improvement.
Please try `poetry run poe test_all` to run the CI on your local machine before pushing.

pyproject.toml
hamdanal: Was it deliberately set to `^0.10.0` though? I thought it was an oversight because the `^` specifier is(was?) the default behavior of `poetry add`. If it was deliberate, it would be nice to have a comment explaining why this package version needs to be strictly capped.

loicdiridollou: pyiceberg does not need to be strongly restricted for the version so you can keep this one and remove both `pytest_ward_bounded`

pyproject.toml
```diff
@@ -1,17 +1,15 @@
...
 license = "BSD-3-Clause"
 readme = "README.md"
-homepage = "https://pandas.pydata.org"
-repository = "https://github.com/pandas-dev/pandas-stubs"
+requires-python = '>=3.11'
 classifiers = [
   "Development Status :: 5 - Production/Stable",
   "Environment :: Console",
   "Intended Audience :: Science/Research",
-  "License :: OSI Approved :: BSD License",
```
Dr-Irv, line 14: I think this gets reflected in `pypi`, so not sure it is OK to remove.

pyproject.toml
```diff
@@ -23,61 +21,64 @@ classifiers = [
...
+]
 
 [build-system]
-requires = ["poetry-core>=1.0.0"]
+requires = ["poetry-core>=2.0.0"]
 build-backend = "poetry.core.masonry.api"
 
+[tool.poetry]
+packages = [{ include = "pandas-stubs" }]
+exclude = ["pandas-stubs/__init__.py"]
```
Dr-Irv: This was needed due to some earlier poetry issues in building a release.  There is no `pandas-stubs/__init__.py` any more, so we may be able to delete this.  I'm going to have to try a build to see.

pyproject.toml
hamdanal: This was deprecated in [PEP 639](https://peps.python.org/pep-0639/#deprecate-license-classifiers) in favor of the `project.license` field. See the link in the PR description.

pyproject.toml
hamdanal: Indeed, I just tested. Removed

pyproject.toml
Dr-Irv: ok

loicdiridollou (approved):
Looks good, thanks for this one @hamdanal !

Issue #1743, opened 2026-05-13
Breaking nightly tests due to pandas-dev/pandas#65466


cmp0xff:
After pandas-dev/pandas#65466, we see the following additional error in the nightly tests:
[83](https://github.com/pandas-dev/pandas-stubs/actions/runs/25737318860/job/75577704586#step:4:584)
```
FAILED tests/frame/test_frame.py::test_types_rename_inplace - pandas.errors.Pandas4Warning: The inplace keyword in DataFrame.rename is deprecated and will be removed in a future version. See `PDEP-8 for more details:https://pandas.pydata.org/pdeps/0008-inplace-methods-in-pandas.html
```

Dr-Irv:
there will be more of these `inplace` related deprecations coming - someone is working on them one by one

cmp0xff:
Should be closed by #1747

Issue #1742, opened 2026-05-13
Breaking nightly tests due to pandas-dev/pandas#65429


cmp0xff:
After pandas-dev/pandas#65429, we see the following additional errors in nightly tests
[81](https://github.com/pandas-dev/pandas-stubs/actions/runs/25737318860/job/75577704586#step:4:582)
```
FAILED tests/arrays/test_timedeltas.py::test_timedelta_array_freq - AttributeError: 'TimedeltaArray' object has no attribute 'freq'
FAILED tests/arrays/test_timedeltas.py::test_timedelta_array_freqstr - AttributeError: 'TimedeltaArray' object has no attribute 'freqstr'
```

cmp0xff:
Should be closed by #1747

PR #1747, opened 2026-05-18
GH1742 fix nightly CI

### LINKED ISSUE #1734: Breaking nightly test due to pandas-dev/pandas#64798
After pandas-dev/pandas#64798, `pd.Index([<unhashables>])` is no more allowed.

We probably need to tweak `AxesData` in `_typing.pyi` and also adapt many tests.

loicdiridollou:
- [x] Closes #1734, #1742 and #1743 

### DIFF
--- a/tests/frame/test_frame.py
+++ b/tests/frame/test_frame.py
@@ -2643,7 +2643,7 @@ def test_types_rename_inplace() -> None:
     df = pd.DataFrame(columns=["a"])
 
     with pytest_warns_bounded(
-        Pandas4Warning, "The inplace keyboard in DataFrame", lower="3.0.99"
+        Pandas4Warning, "The inplace keyword in DataFrame", lower="3.0.99"
     ):
         check(
             assert_type(df.rename(columns={"a": "b"}, inplace=True), None), type(None)
--- a/tests/frame/test_frame.py
+++ b/tests/frame/test_frame.py
@@ -48,6 +48,7 @@ import pytest
 import xarray as xr
 
 from tests import (
+    PD_LTE_31,
     TYPE_CHECKING_INVALID_USAGE,
     check,
 )
@@ -2638,7 +2639,10 @@ def test_types_rename() -> None:
 def test_types_rename_inplace() -> None:
     """Test DataFrame.rename with inplace argument."""
     df = pd.DataFrame(columns=["a"])
-    check(assert_type(df.rename(columns={"a": "b"}, inplace=True), None), type(None))
+
+    if PD_LTE_31:
+        check(assert_type(df.rename(columns={"a": "b"}, inplace=True), None), type(None))
+
     check(assert_type(df.rename(columns={"a": "b"}), pd.DataFrame), pd.DataFrame)
 
     if TYPE_CHECKING_INVALID_USAGE:
--- a/tests/frame/test_frame.py
+++ b/tests/frame/test_frame.py
@@ -2641,7 +2641,9 @@ def test_types_rename_inplace() -> None:
     df = pd.DataFrame(columns=["a"])
 
     if PD_LTE_31:
-        check(assert_type(df.rename(columns={"a": "b"}, inplace=True), None), type(None))
+        check(
+            assert_type(df.rename(columns={"a": "b"}, inplace=True), None), type(None)
+        )
 
     check(assert_type(df.rename(columns={"a": "b"}), pd.DataFrame), pd.DataFrame)
 
--- a/tests/frame/test_frame.py
+++ b/tests/frame/test_frame.py
@@ -47,10 +47,12 @@ from pandas.core.window.rolling import Rolling
 import pytest
 import xarray as xr
 
+from pandas.errors import Pandas4Warning
+
 from tests import (
-    PD_LTE_31,
     TYPE_CHECKING_INVALID_USAGE,
     check,
+    pytest_warns_bounded,
 )
 from tests._typing import (
     np_1darray,
@@ -2640,7 +2642,9 @@ def test_types_rename_inplace() -> None:
     """Test DataFrame.rename with inplace argument."""
     df = pd.DataFrame(columns=["a"])
 
-    if PD_LTE_31:
+    with pytest_warns_bounded(
+        Pandas4Warning, "The inplace keyboard in DataFrame", lower="3.0.99"
+    ):
         check(
             assert_type(df.rename(columns={"a": "b"}, inplace=True), None), type(None)
         )
--- a/tests/indexes/test_indexes.py
+++ b/tests/indexes/test_indexes.py
@@ -211,6 +211,7 @@ def test_str_split() -> None:
     ind = pd.Index(["a-b", "c-d"])
 
     if PD_LTE_31:
+        # TODO: pandas-dev/pandas#64798 may end up working in 3.1.0, issue in discussion
         check(assert_type(ind.str.split("-"), "pd.Index[list[str]]"), pd.Index, list)
 
     check(assert_type(ind.str.split("-", expand=True), pd.MultiIndex), pd.MultiIndex)
@@ -1569,7 +1570,7 @@ def test_index_where() -> None:
         val_idx = idx.where(mask, idx)
         check(assert_type(val_idx, "pd.Index[int]"), pd.Index, int)
 
-    val_sr = idx.where(mask, (idx).to_series())
+    val_sr = idx.where(mask, idx.to_series())
 
     if PD_LTE_31:
         check(assert_type(val_sr, "pd.Index[int]"), pd.Index, int)
--- a/tests/indexes/test_indexes.py
+++ b/tests/indexes/test_indexes.py
@@ -209,25 +209,35 @@ def test_difference_none() -> None:
 def test_str_split() -> None:
     # GH 194
     ind = pd.Index(["a-b", "c-d"])
-    check(assert_type(ind.str.split("-"), "pd.Index[list[str]]"), pd.Index, list)
+
+    if PD_LTE_31:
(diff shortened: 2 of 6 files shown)


tests/indexes/test_indexes.py
```diff
@@ -1554,11 +1564,17 @@ def test_index_where() -> None:
     """Test Index.where with multiple types of other GH1419."""
     idx = pd.Index(range(48))
     mask = np.ones(48, dtype=bool)
-    val_idx = idx.where(mask, idx)
-    check(assert_type(val_idx, "pd.Index[int]"), pd.Index, int)
+
+    if PD_LTE_31:
```
loicdiridollou: See pandas issue https://github.com/pandas-dev/pandas/issues/65685

tests/indexes/test_indexes.py
```diff
@@ -209,25 +209,35 @@ def test_difference_none() -> None:
 def test_str_split() -> None:
     # GH 194
     ind = pd.Index(["a-b", "c-d"])
-    check(assert_type(ind.str.split("-"), "pd.Index[list[str]]"), pd.Index, list)
+
+    if PD_LTE_31:
+        check(assert_type(ind.str.split("-"), "pd.Index[list[str]]"), pd.Index, list)
```
Dr-Irv: There is still debate in the pandas dev team about what to do in this case.  So if this becomes allowable code, I'm not sure how we can know that it is permitted in the future if we put the `PD_LTE_31` test here.  

I guess for now can you add a comment that says something like `# May end up working when pandas 3.1 is released`  And do this for the other tests where we are checking if the result is `Index[list[str]]`

tests/indexes/test_indexes.py
```diff
@@ -1554,11 +1564,17 @@ def test_index_where() -> None:
...
+        val_idx = idx.where(mask, idx)
+        check(assert_type(val_idx, "pd.Index[int]"), pd.Index, int)
 
     val_sr = idx.where(mask, (idx).to_series())
-    check(assert_type(val_sr, "pd.Index[int]"), pd.Index, int)
+
+    if PD_LTE_31:
+        check(assert_type(val_sr, "pd.Index[int]"), pd.Index, int)
+    else:
+        check(assert_type(val_sr, "pd.Index[int]"), pd.Index, np.int64)
```
Dr-Irv: surprised at this one as well

tests/test_string_accessors.py
```diff
@@ -276,8 +283,12 @@ def test_string_accessors_string_index() -> None:
             idx_bytes.str.decode("utf-8", dtype=pd.StringDtype()), "pd.Index[str]"
         )
     )
-    idx_list = pd.Index([["apple", "banana"], ["cherry", "date"], ["one", "eggplant"]])
-    _check(assert_type(idx_list.str.join("-"), "pd.Index[str]"))
+
+    if PD_LTE_31:
+        idx_list = pd.Index(
+            [["apple", "banana"], ["cherry", "date"], ["one", "eggplant"]]
+        )
```
Dr-Irv: needs similar comment as stated elsewhere.

Dr-Irv (changes requested):

tests/indexes/test_indexes.py
loicdiridollou: Let me open a separate issue to track that and put the link in the code

tests/indexes/test_indexes.py
loicdiridollou: Yes I am worried it is related to the same changes

tests/test_string_accessors.py
loicdiridollou: Added

tests/indexes/test_indexes.py
loicdiridollou: This is now fixed.

tests/frame/test_frame.py
```diff
@@ -2638,7 +2639,12 @@ def test_types_rename() -> None:
 def test_types_rename_inplace() -> None:
     """Test DataFrame.rename with inplace argument."""
     df = pd.DataFrame(columns=["a"])
-    check(assert_type(df.rename(columns={"a": "b"}, inplace=True), None), type(None))
+
+    if PD_LTE_31:
+        check(
+            assert_type(df.rename(columns={"a": "b"}, inplace=True), None), type(None)
```
Dr-Irv, line 2649: this should be done with `pytest_warns_bounded`

tests/indexes/test_indexes.py
```diff
@@ -209,25 +209,36 @@ def test_difference_none() -> None:
 def test_str_split() -> None:
     # GH 194
     ind = pd.Index(["a-b", "c-d"])
-    check(assert_type(ind.str.split("-"), "pd.Index[list[str]]"), pd.Index, list)
+
+    if PD_LTE_31:
+        # TODO: pandas-dev/pandas#64798 may end up working in 3.1.0, issue in discussion
+        check(assert_type(ind.str.split("-"), "pd.Index[list[str]]"), pd.Index, list)
```
Dr-Irv: now this is valid code, as the change was reverted, so remove the `if PD_LTE_31`

tests/indexes/test_indexes.py
```diff
@@ -209,25 +209,36 @@ def test_difference_none() -> None:
...
+        )
 
 
 def test_str_rsplit() -> None:
     # GH 1074
     ind = pd.Index(["a-b", "c-d"])
-    check(assert_type(ind.str.rsplit("-"), "pd.Index[list[str]]"), pd.Index, list)
+
+    if PD_LTE_31:
+        check(assert_type(ind.str.rsplit("-"), "pd.Index[list[str]]"), pd.Index, list)
```
Dr-Irv: same comment

tests/indexes/test_indexes.py
```diff
@@ -209,25 +209,36 @@ def test_difference_none() -> None:
...
     check(assert_type(ind.str.rsplit("-", expand=True), pd.MultiIndex), pd.MultiIndex)
-    check(
-        assert_type(ind.str.rsplit("-", expand=False), "pd.Index[list[str]]"),
-        pd.Index,
-        list,
-    )
+
+    if PD_LTE_31:
+        check(
+            assert_type(ind.str.rsplit("-", expand=False), "pd.Index[list[str]]"),
```
Dr-Irv: same comment

tests/test_string_accessors.py
```diff
@@ -234,10 +235,16 @@ def test_string_accessors_string_index() -> None:
     check(assert_type(idx.str.cat(sep="X"), str), str)
     _check(assert_type(idx.str.center(10), "pd.Index[str]"))
     _check(assert_type(idx.str.get(2), "pd.Index[str]"))
-    idx_dict = pd.Index(
-        [{"name": "Hello", "value": "World"}, {"name": "Goodbye", "value": "Planet"}]
-    )
-    _check(assert_type(idx_dict.str.get("name"), "pd.Index[str]"))
+
+    if PD_LTE_31:
+        idx_dict = pd.Index(
```
Dr-Irv: still valid so remove the `if`

tests/test_string_accessors.py
```diff
@@ -276,8 +283,13 @@ def test_string_accessors_string_index() -> None:
             idx_bytes.str.decode("utf-8", dtype=pd.StringDtype()), "pd.Index[str]"
         )
     )
-    idx_list = pd.Index([["apple", "banana"], ["cherry", "date"], ["one", "eggplant"]])
-    _check(assert_type(idx_list.str.join("-"), "pd.Index[str]"))
+
+    if PD_LTE_31:
```
Dr-Irv: still valid so remove the `if`

tests/test_string_accessors.py
```diff
@@ -317,14 +329,16 @@ def test_string_accessors_list_series() -> None:
...
+        _check(assert_type(idx.str.findall("pp"), "pd.Index[list[str]]"))
+        _check(assert_type(idx.str.findall(re.compile(r"pp")), "pd.Index[list[str]]"))
+        _check(assert_type(idx.str.split("a"), "pd.Index[list[str]]"))
+        _check(assert_type(idx.str.split(re.compile(r"a")), "pd.Index[list[str]]"))
+        # GH 194
+        _check(assert_type(idx.str.split("a", expand=False), "pd.Index[list[str]]"))
+        _check(assert_type(idx.str.rsplit("a"), "pd.Index[list[str]]"))
+        _check(assert_type(idx.str.rsplit("a", expand=False), "pd.Index[list[str]]"))
```
Dr-Irv: all still valid so remove the `if`

Dr-Irv (changes requested):
Now the code has been reverted that made non-hashable list items be rejected.

It might come back.  There are 2 options:

1. Just remove the `if PD_LTE_31` tests that you inserted for those
2. Create a new constraint `PD_INDEX_HASHABLE` which we set to `True` for now, and use that instead of `if PD_LTE_31`, and if we do end up putting back where we don't allow non-hashable index items, we just change the definition of `PD_INDEX_HASHABLE` to be `PD_LTE_31`, so we only have to manage it in one place.

tests/frame/test_frame.py
loicdiridollou: Fixed!

tests/indexes/test_indexes.py
loicdiridollou: Yes will remove the PD_LTE_31, let's not overcomplicate the code for now with something like option 2 for the moment and see where pandas goes with this issue.

loicdiridollou:
All cleaned up, still needs the PD_LTE_31 for some types in some instances where the return type is int before and np.int64 after.

Dr-Irv (approved):
thanks @loicdiridollou

cmp0xff:
Somehow only the first issue after `closes` will be closed automatically.

PR #1748, opened 2026-05-19
GH1415 Enhance typing of Series[Categorical]

### LINKED ISSUE #1415: How should we type `Series` with a categorical dtype?

loicdiridollou:
- [x] Closes #1415


pandas-stubs/_stubs_only/__init__.pyi
```diff
@@ -84,6 +86,11 @@ OrderableScalarT = TypeVar("OrderableScalarT", bound=OrderableScalars)
 OrderableTimesT = TypeVar("OrderableTimesT", bound=OrderableTimes)
 OrderableT = TypeVar("OrderableT", bound=Orderables, default=Any)
 
+CategoricalT = TypeVar("CategoricalT", str, int, float, object)
```
Dr-Irv: Can we rename this as `CategoricalValueT` to make it clear we are referring to the type of values used?

pandas-stubs/core/arrays/categorical.pyi
```diff
@@ -146,20 +151,24 @@ class CategoricalAccessor(PandasDelegate, NoNewAttributesMixin):
     def ordered(self) -> bool | None: ...
     def rename_categories(
         self, new_categories: ListLike | dict[Any, Any] | Callable[[Any], Any]
-    ) -> Series: ...
+    ) -> Series[C1[Any]]: ...
```
Dr-Irv: I think  you can restrict the types here.  Because we know that the only possible values are `CategoricalValueT` so then the result will be `Series[C1[CateogricalValueT]]`

pandas-stubs/core/series.pyi
```diff
@@ -360,6 +370,36 @@ class Series(IndexOpsMixin[S1], ElementOpsMixin[S1], NDFrame):
...
+    @overload
+    def __new__(
+        cls,
+        data: Sequence[float | np.floating],
+        index: AxesData | None = None,
+        *,
+        dtype: CategoryDtypeArg,
+        name: Hashable = None,
+        copy: bool | None = None,
+    ) -> Series[C1[float]]: ...
```
Dr-Irv: If we are only having `str`, `int` and `float` here, maybe delete `object` from `CategoricalValueT` ??

pandas-stubs/core/series.pyi
```diff
@@ -403,7 +443,7 @@ class Series(IndexOpsMixin[S1], ElementOpsMixin[S1], NDFrame):
         dtype: CategoryDtypeArg,
         name: Hashable = None,
         copy: bool | None = None,
-    ) -> Series[CategoricalDtype]: ...
+    ) -> Series[C1[Any]]: ...
```
Dr-Irv: can we avoid this?  I think we should only accept data that we know are appropriate labels in a `Categorical`

tests/series/categorical/test_dtypes.py
```diff
@@ -12,6 +12,13 @@
 from tests._typing import CategoryDtypeArg
 from tests.dtypes import ASTYPE_CATEGORICAL_ARGS
 
+if TYPE_CHECKING:
+    from typing import Any  # noqa: F401
+
+    from pandas._stubs_only import (  # pyright: ignore[reportMissingModuleSource]  # isort:skip
+        C1,  # noqa: F401
```
Dr-Irv: the whole point of having `_stubs_only` is so that the tests do *not* import from there.  So I think this would need to change.

I think the solution is to make `CategoricalDType` generic in the stubs and use that instead of `C1` in the tests.

tests/series/test_series.py
```diff
@@ -85,6 +85,11 @@
 )
 from tests.extension.decimal.array import DecimalDtype
 
+if TYPE_CHECKING:
+    from pandas._stubs_only import (  # pyright: ignore[reportMissingModuleSource]
+        C1,  # noqa: F401
+    )
```
Dr-Irv: same comment

pandas-stubs/core/arrays/categorical.pyi
```diff
@@ -146,20 +151,24 @@ class CategoricalAccessor(PandasDelegate, NoNewAttributesMixin):
...
+    def add_categories(
+        self, new_categories: Scalar | ListLike
+    ) -> Series[C1[CategoricalT]]: ...
+    def remove_categories(
+        self, removals: Scalar | ListLike
+    ) -> Series[C1[CategoricalT]]: ...
+    def remove_unused_categories(self) -> Series[C1[CategoricalT]]: ...
     def set_categories(
         self,
         new_categories: ListLike,
```
Dr-Irv: this should be restricted to `CategoricalT` in the lists

tests/series/test_series.py
```diff
@@ -1672,30 +1677,54 @@ def test_cat_accessor() -> None:
         pd.Categorical(["a", "b", "a"], categories=["a", "b"])
     )
     check(assert_type(s.cat.codes, "pd.Series[int]"), pd.Series, np.int8)
-    # GH 139
+
     ser = pd.Series([1, 2, 3], name="A").astype("category")
```
Dr-Irv, line 1721: So if you set things up where `Series[int].astype("category")` gives you `Series[CategoryDtype[int]]` then you don't need to use `C1[Any]` in the tests.

tests/series/test_series.py
```diff
@@ -1672,30 +1677,54 @@ def test_cat_accessor() -> None:
...
+        assert_type(ser.cat.as_unordered(), "pd.Series[C1[Any]]"), pd.Series, np.integer
+    )
+    check(ser.cat.set_categories([1, 2, 3]), pd.Series, np.integer)
+    check(ser.cat.reorder_categories([2, 3, 1], ordered=True), pd.Series, np.integer)
+    check(ser.cat.rename_categories([1, 2, 3]), pd.Series, np.integer)
+    check(ser.cat.remove_unused_categories(), pd.Series, np.integer)
+    check(ser.cat.remove_categories([2]), pd.Series, np.integer)
+    check(ser.cat.add_categories([4]), pd.Series, np.integer)
+    check(ser.cat.as_ordered(), pd.Series, np.integer)
+    check(ser.cat.as_unordered(), pd.Series, np.integer)
```
Dr-Irv: Why aren't there `assert_type()` statements here?

Dr-Irv (changes requested):
I didn't comment on everything in the tests.  The main issue is the introduction of `C1`.  As suggested in one of my comments, I think `CategoricalDtype` should be `Generic[CategoricalValueT]` .  This is similar to how we are doing `IntervalIndex`, where we do things like `IntervalIndex[Interval[int]]` 

Then you don't need `C1`.  You can also be more precise with some of the methods because you only want to allow `CategoricalValueT` as being arguments for the categories.

pandas-stubs/core/series.pyi
loicdiridollou: I was having an issue with the default value for the TypeVar

tests/series/categorical/test_dtypes.py
loicdiridollou: I changed it all, it should be much better included with what we have.

tests/series/test_series.py
loicdiridollou: Dropped.

tests/series/test_series.py
loicdiridollou: Yes it is now done.

tests/series/test_series.py
loicdiridollou: Fixed

loicdiridollou:
@Dr-Irv thanks for the review, I redesigned the PR, this one should be much closer to what we want.
I was forced to set a default to the TypeVar to object, I kept str | float | int as allowd values, will check if we can relax it.

pandas-stubs/core/arrays/categorical.pyi
```diff
@@ -45,23 +49,23 @@ class Categorical(NDArrayBackedExtensionArray):
...
+        dtype: CategoricalDtype[Any] | None = None,
         copy: bool = True,
     ) -> None: ...
     @property
     def categories(self) -> Index: ...
     @property
     def ordered(self) -> Ordered: ...
     @property
-    def dtype(self) -> CategoricalDtype: ...
+    def dtype(self) -> CategoricalDtype[object]: ...
```
Dr-Irv: I think this should be `Any`.  That is similar to what we do with `DataFrame["some_column"]` being `Series[Any]`, and if we make `Categorical` a `Generic` type, then we can return that specific type.

pandas-stubs/core/arrays/categorical.pyi
```diff
@@ -45,23 +49,23 @@ class Categorical(NDArrayBackedExtensionArray):
         values: SequenceNotStr[Hashable] | AnyArrayLike,
         categories: SequenceNotStr[Hashable] | AnyArrayLike | None = None,
```
Dr-Irv, line 115: Can we add overloads here?  If we are getting a `SequenceNotStr[some_type]` for `values`, and possibly `categories`, where `some_type` is `float`, `str`, `int`, then we could make `Categorical` a `Generic` type as well, dependent on that, which means we could then return `dtype` as the corresponding value.

Would require a number of overloads to handle this.

pandas-stubs/core/arrays/categorical.pyi
```diff
@@ -45,23 +49,23 @@ class Categorical(NDArrayBackedExtensionArray):
...
+    def dtype(self) -> CategoricalDtype[object]: ...
     def tolist(self) -> list[Scalar]: ...
     @classmethod
     def from_codes(
         cls,
         codes: AnyArrayLike | Sequence[int],
         categories: Index | None = ...,
         ordered: bool | None = ...,
-        dtype: CategoricalDtype | None = ...,
+        dtype: CategoricalDtype[Any] | None = ...,
```
Dr-Irv: could do overloads here as well, because if `categories` is `Index[int]`, then we can say this is `Categorical[int]`

pandas-stubs/core/indexes/accessors.pyi
```diff
@@ -461,7 +461,7 @@ class ArrayDescriptor:
     ) -> ExtensionArray: ...
     @overload
     def __get__(
-        self, instance: IndexOpsMixin[CategoricalDtype], owner: type[IndexOpsMixin]
+        self, instance: IndexOpsMixin[CategoricalDtype[Any]], owner: type[IndexOpsMixin]
     ) -> Categorical: ...
```
Dr-Irv: If `Categorical` also becomes `Generic`, then  you could do

```suggestion
        self, instance: IndexOpsMixin[CategoricalDtype[CategoricalValueT]], owner: type[IndexOpsMixin]
) -> Categorical[CategoricalValureT]: ...
```

pandas-stubs/core/series.pyi
```diff
@@ -360,6 +373,36 @@ class Series(IndexOpsMixin[S1], ElementOpsMixin[S1], NDFrame):
         copy: bool | None = None,
     ) -> Series[list[_str]]: ...
     @overload
+    def __new__(  # type: ignore[overload-overlap] # pyright: ignore[reportOverlappingOverload]
+        cls,
+        data: Sequence[_str],
```
Dr-Irv: For this and the other overloads, you have to accept `Series[_str]` and `Index[_str]`  (and similar for `float` and `int`) as `data`.  E.g., the following works:
```python
>>> pd.Series(pd.Series([1,2,3,1,3]), dtype="category")
0    1
1    2
2    3
3    1
4    3
dtype: category
Categories (3, int64): [1, 2, 3]
>>> pd.Series(pd.Index([1,2,3,1,3]), dtype="category")
0    1
1    2
2    3
3    1
4    3
dtype: category
Categories (3, int64): [1, 2, 3]
```
You could possibly do an overload like this:
```python
   def __new__(cls,
        data: Sequence[CategoricalValueT] | Index[CategoricalValueT] | Series[CategoricalValueT]
        index: AxesData | None = None,
        *,
        dtype: CategoryDtypeArg,
        name: Hashable = None,
        copy: bool | None = None,
    ) -> Series[CategoricalDtype[CategoricalValueT]]: ...
```
instead of having separate overloads

pandas-stubs/core/series.pyi
```diff
@@ -1507,10 +1550,10 @@ class Series(IndexOpsMixin[S1], ElementOpsMixin[S1], NDFrame):
     ) -> Series[Timestamp]: ...
     @overload
     def astype(
-        self,
+        self: Series[CategoricalValueT1],
         dtype: CategoryDtypeArg,
         errors: IgnoreRaise = ...,
-    ) -> Series[CategoricalDtype]: ...
+    ) -> Series[CategoricalDtype[CategoricalValueT1]]: ...
```
Dr-Irv, line 1575: I don't believe this is correct, because you are only saying that categorical series can be changed to categorical series.  Because the idea of `astype` is to change the type of the `Series`, so for example, the following works:
```python
pd.Series([1,2,3,1,3]).astype("category")
0    1
1    2
2    3
3    1
4    3
dtype: category
Categories (3, int64): [1, 2, 3]
```
So I think you need another overload for when the `Series` is `Series[Any]`

tests/series/test_properties.py
```diff
@@ -81,9 +81,8 @@ def test_property_dt() -> None:
 
 def test_property_array() -> None:
     """Test that Series.array returns ExtensionArray and its subclasses"""
-    check(
-        assert_type(Series([1], dtype="category").array, Categorical), Categorical, int
-    )
+    sr = Series([1], dtype="category")
+    check(assert_type(sr.array, Categorical), Categorical, int)
```
Dr-Irv: I would hope this would work:

```suggestion
    check(assert_type(sr.array, "Categorical[int]"), Categorical, int)
```

Dr-Irv (changes requested):
I think we can make `Categorical` a `Generic` type as well

pandas-stubs/core/series.pyi
loicdiridollou: This is fixed, test is added.

tests/series/test_properties.py
loicdiridollou: This works in the latest version.

pandas-stubs/core/series.pyi
loicdiridollou: Yes this work, added in the last version.

pandas-stubs/core/arrays/categorical.pyi
loicdiridollou: Yes that works.

pandas-stubs/core/arrays/categorical.pyi
loicdiridollou: Do you want to take care of it in this one or defer it to another issue? Feels like there are already a ton of changes here

Dr-Irv: If we're going to do `Categorical`, let's get it all done in one PR

pandas-stubs/core/arrays/categorical.pyi
```diff
@@ -36,32 +37,35 @@ from pandas._typing import (
...
-class Categorical(NDArrayBackedExtensionArray):
+class Categorical(NDArrayBackedExtensionArray, Generic[CategoricalValueT]):
     __array_priority__: int = ...
     def __init__(
         self,
         values: SequenceNotStr[Hashable] | AnyArrayLike,
         categories: SequenceNotStr[Hashable] | AnyArrayLike | None = None,
         ordered: bool | None = None,
-        dtype: CategoricalDtype | None = None,
+        dtype: CategoricalDtype[Any] | None = None,
```
cmp0xff: `CategoricalDtype` is generic only in stubs and not in runtime, which is the sames as `Series` in this regards. I think the convention is to use `CategoricalDtype` and omit `[Any]`, or the new code should allow us and allow users to do so.

cmp0xff (commented):
This is an exciting PR! I won't modify @Dr-Irv 's review process, just adding small ideas.

tests/test_api_types.py
```diff
@@ -368,7 +368,12 @@ def test_infer_dtype() -> None:
 
 def test_union_categoricals() -> None:
     to_union = [pd.Categorical([1, 2, 3]), pd.Categorical([3, 4, 5])]
-    check(assert_type(api.union_categoricals(to_union), pd.Categorical), pd.Categorical)
+    check(
+        assert_type(  # pyrefly: ignore[assert-type]
```
loicdiridollou, line 372: Not super clear why pyrefly was complaining, I may open an open on their end to see if they have a clue.

pandas-stubs/core/construction.pyi
```diff
@@ -107,15 +108,20 @@ def array(  # type: ignore[overload-overlap] # pyright: ignore[reportOverlapping
...
 def array(
     data: Sequence[Any] | np_ndarray | ExtensionArray | Index | Series,
     dtype: CategoryDtypeArg,
     copy: bool = True,
 ) -> Categorical: ...
 @overload
 def array(
-    # TODO: Categorical Series pandas-dev/pandas-stubs#1415
-    data: Categorical | CategoricalIndex,
+    data: CategoricalIndex,
```
loicdiridollou, line 124: Right now CategoricalIndex does not have the same generic as Categorical, could be left as a TODO

pandas-stubs/core/arrays/categorical.pyi
```diff
@@ -34,60 +36,138 @@ from pandas._typing import (
...
+        cls,
+        values: (
+            list[int]
+            | np_ndarray_anyint
+            | SequenceNotStr[int]
+            | Series[int]
+            | Index[int]
+        ),
+        categories: (
+            SequenceNotStr[int] | Series[int] | Index[int] | np_ndarray_anyint | None
```
loicdiridollou: Since we have the np_ndarray variations here it was hard to compress all the overloads into one.

cmp0xff:
Let me raise the old design issue here again: We don't do `Series[Int64Dtype]`. Instead we do `Series[int]`, where `int` is roughly the python type of Series element. The same for `Series[Interval[int]]`, etc.

How to justify `Series[CategoricalDtype[...]]` then? I know `Series[CategoricalDtype]` has been introduced a while ago, but now I think it may not be a good idea.

Dr-Irv:
> Let me raise the old design issue here again: We don't do `Series[Int64Dtype]`. Instead we do `Series[int]`, where `int` is roughly the python type of Series element. The same for `Series[Interval[int]]`, etc.
> 
> How to justify `Series[CategoricalDtype[...]]` then? I know `Series[CategoricalDtype]` has been introduced a while ago, but now I think it may not be a good idea.

It's similar to `Series[Interval[...]]`

pandas-stubs/core/arrays/categorical.pyi
```diff
@@ -34,60 +36,138 @@ from pandas._typing import (
...
-        self,
+    @overload
+    def __new__(  # type: ignore[overload-overlap]
+        cls,
+        values: Sequence[Never],
+        categories: SequenceNotStr[Hashable] | AnyArrayLike | None = None,
+        ordered: bool | None = None,
+        dtype: CategoricalDtype | None = None,
+        copy: bool = True,
+    ) -> Categorical: ...
```
Dr-Irv, line 60: I would think this overload would go last, i.e., we return the non-generic version if nothing else matched

pandas-stubs/core/arrays/categorical.pyi
```diff
@@ -34,60 +36,138 @@ from pandas._typing import (
...
+        dtype: CategoricalDtype | None = None,
+        copy: bool = True,
+    ) -> Categorical[str]: ...
+    @overload
+    def __new__(  # pyright: ignore[reportOverlappingOverload]
+        cls,
+        values: (
+            list[int]
+            | np_ndarray_anyint
+            | SequenceNotStr[int]
```
Dr-Irv: you can use `Sequence[int]` here. We only need `SequenceNotStr[str]` to differentiate between a pure `str` where `Sequence[str]` would match.

There are other places below where you can use `Sequence` instead of `SequenceNotStr`

pandas-stubs/core/arrays/categorical.pyi
```diff
@@ -34,60 +36,138 @@ from pandas._typing import (
...
+        ordered: bool | None = None,
+        dtype: CategoricalDtype | None = None,
+        copy: bool = True,
+    ) -> Categorical[float]: ...
+    @overload
+    def __new__(
+        cls,
+        values: Categorical[CategoricalValueT],
+        categories: (
+            SequenceNotStr[str] | Series[str] | Index[str] | np_1darray_str | None
```
Dr-Irv: I think these need to be `CategoricalValueT`, not `str` because you want the values and categories to have the same type.

pandas-stubs/core/arrays/categorical.pyi
```diff
@@ -34,60 +36,138 @@ from pandas._typing import (
...
     def ordered(self) -> Ordered: ...
     @property
-    def dtype(self) -> CategoricalDtype: ...
-    def tolist(self) -> list[Scalar]: ...
+    def dtype(self) -> CategoricalDtype[CategoricalValueT]: ...
+    def tolist(self) -> list[CategoricalValueT]: ...
     @classmethod
     def from_codes(
         cls,
         codes: AnyArrayLike | Sequence[int],
```
Dr-Irv: `AnyArrayLike` is too wide here.  `codes` has to be "array-like of int" so you can use `Series[int]`, `Index[int]`, `np_ndarray_anyint` and `Sequence[int]`

pandas-stubs/core/arrays/categorical.pyi
```diff
@@ -34,60 +36,138 @@ from pandas._typing import (
...
         categories: Index | None = ...,
         ordered: bool | None = ...,
-        dtype: CategoricalDtype | None = ...,
+        dtype: CategoricalDtype[Any] | None = ...,
         validate: bool = True,
     ) -> Categorical: ...
```
Dr-Irv:
```suggestion
        dtype: CategoricalDtype[CategoricalValueT] | None = ...,
validate: bool = True,
) -> Categorical[CategoricalValueT]: ...
```

pandas-stubs/core/arrays/categorical.pyi
```diff
@@ -34,60 +36,138 @@ from pandas._typing import (
...
     @property
-    def dtype(self) -> CategoricalDtype: ...
-    def tolist(self) -> list[Scalar]: ...
+    def dtype(self) -> CategoricalDtype[CategoricalValueT]: ...
+    def tolist(self) -> list[CategoricalValueT]: ...
     @classmethod
     def from_codes(
         cls,
         codes: AnyArrayLike | Sequence[int],
         categories: Index | None = ...,
```
Dr-Irv: This should be same as used in `__new__()`.  MIght need an overload, because docs say you either do `categories`, and if not, you do the `dtype`

pandas-stubs/core/dtypes/concat.pyi
```diff
@@ -8,6 +11,27 @@ from pandas import (
...
+    ignore_order: bool = False,
+) -> Categorical[int]: ...
+@overload
+def union_categoricals(
+    to_union: (
+        list[Categorical[float]] | list[Series[float]] | list[CategoricalIndex[float]]
+    ),
+    sort_categories: bool = False,
+    ignore_order: bool = False,
+) -> Categorical[float]: ...
```
Dr-Irv: Could these be combined like this into one overload?  (Not sure)
```python
def union_categoricals(
    to_union: list[Categorical[CategoricalValueT1]] | list[Series[CategoricalValueT1]] | list[CategoricalIndex[CategoricalValueT1]],
    sort_categories: bool = False,
    ignore_order: bool = False,
) -> Categorical[CategoricalValueT1]: ...

File: pandas-stubs/core/series.pyi [hunk 28]
```diff
@@ -360,6 +373,20 @@ class Series(IndexOpsMixin[S1], ElementOpsMixin[S1], NDFrame):
         copy: bool | None = None,
     ) -> Series[list[_str]]: ...
     @overload
+    def __new__(  # type: ignore[overload-overlap]
+        cls,
+        data: (
+            Sequence[CategoricalValueT1]
```
Dr-Irv: Use `SequenceNotStr[CategoricalValueT1]` here because we don't want  a `str` to match

pandas-stubs/core/series.pyi
```diff
@@ -1507,10 +1534,16 @@ class Series(IndexOpsMixin[S1], ElementOpsMixin[S1], NDFrame):
...
         dtype: CategoryDtypeArg,
         errors: IgnoreRaise = ...,
-    ) -> Series[CategoricalDtype]: ...
+    ) -> Series[CategoricalDtype[CategoricalValueT1]]: ...
+    @overload
+    def astype(
+        self: Series[Any],
+        dtype: CategoryDtypeArg,
+        errors: IgnoreRaise = ...,
+    ) -> Series[CategoricalDtype[CategoricalValueT1]]: ...
```
```suggestion
    ) -> Series[CategoricalDtype[Any]]: ...
```
this overload doesn't have `CategoricalValueT1` in it, and if it's a generic `Series`, we don't know the type of the categories.

tests/arrays/test_categorical.py
```diff
@@ -49,74 +50,75 @@ def test_construction_dtype() -> None:
...
-    check(assert_type(cat, Categorical), Categorical)
+    dd = ["a", "b", "c", "a"]
+    cat = Categorical(dd)
+    check(assert_type(cat, "Categorical[str]"), Categorical, str)
 
     values = np.array(["a", "b", "c", "a"])
-    cat = Categorical(values)
-    check(assert_type(cat, Categorical), Categorical)
+    cat_np = Categorical(values)
+    check(cat_np, Categorical)
```
Dr-Irv: why no `assert_type` ?

loicdiridollou:
I think we need to look at what we are worried by in this PR. Is it a maintenance issue? Is it the complexity?
Right now it more or less matches our implementation of the `Interval` generic type.
Let me know what you think

Dr-Irv (changes requested):
definitely getting closer!

pandas-stubs/core/arrays/categorical.pyi
loicdiridollou: This one is for empty list, if I move it to the bottom `pd.Categorical([])` gets identified as `Categorical[str]` and not `Categorical`

Dr-Irv:
@loicdiridollou I resolved the comments of everything you did address from my previous reviews, but found this one that is still open:  https://github.com/pandas-dev/pandas-stubs/pull/1748#discussion_r3274551527

cmp0xff:
> It's similar to `Series[Interval[...]]`

I don't think so
- `pd.Series([1]).iloc[0]` gives `np.int64(1)` (pandas 2.x, well `np.int64` is similar to `int`). `pd.Series([1])` is typed as `pd.Series[int]`, where `int` is kind of the python type of the element.
- `pd.Series([pd.Interval(1, 2)]).iloc[0]` gives `Interval(1, 2, closed='right')`. `pd.Series([pd.Interval(1, 2)])` is typed as `pd.Series[pd.Interval[int]]`, where `Interval[int]` is the python type of the element.
- `pd.Series(pd.Categorical([1])).iloc[0]` gives `np.int64(1)` (pandas 2.x). `pd.Series(pd.Categorical([1]))` will be typed as `pd.Series[pd.CategoricalDtype[int]]`, but `pd.CategoricalDtype[int]` is NOT the python type of the element.

Dr-Irv:
> * `pd.Series(pd.Categorical([1])).iloc[0]` gives `np.int64(1)` (pandas 2.x). `pd.Series(pd.Categorical([1]))` will be typed as `pd.Series[pd.CategoricalDtype[int]]`, but `pd.CategoricalDtype[int]` is NOT the python type of the element.

Yes, but we certainly want to have at least `pd.Series[pd.CategoricalDtype]`, and `pd.CategoricalDtype` is not the type of the element either.

pandas-stubs/core/arrays/categorical.pyi
```diff
@@ -34,60 +36,142 @@ from pandas._typing import (
...
+        ordered: bool | None = None,
+        dtype: CategoricalDtype | None = None,
+        copy: bool = True,
+    ) -> Categorical[str]: ...
+    @overload
+    def __new__(
+        cls,
+        values: Sequence[int] | np_ndarray_anyint | Series[int] | Index[int],
+        categories: (
+            SequenceNotStr[int] | Series[int] | Index[int] | np_ndarray_anyint | None
```
Dr-Irv: Don't use `SequenceNotStr[int]` - use `Sequence[int]`

pandas-stubs/core/arrays/categorical.pyi
```diff
@@ -34,60 +36,142 @@ from pandas._typing import (
...
+        ordered: bool | None = None,
+        dtype: CategoricalDtype | None = None,
+        copy: bool = True,
+    ) -> Categorical[int]: ...
+    @overload
+    def __new__(
+        cls,
+        values: Sequence[float] | np_ndarray_float | Series[float] | Index[float],
+        categories: (
+            SequenceNotStr[float]
```
Dr-Irv: Don't use `SequenceNotStr[float]` - use `Sequence[float]`

pandas-stubs/core/arrays/categorical.pyi
```diff
@@ -34,60 +36,142 @@ from pandas._typing import (
...
         cls,
-        codes: AnyArrayLike | Sequence[int],
+        codes: Series[int] | Index[int] | np_ndarray_anyint | Sequence[int],
         categories: Index | None = ...,
         ordered: bool | None = ...,
-        dtype: CategoricalDtype | None = ...,
+        dtype: CategoricalDtype[int] | None = ...,
         validate: bool = True,
-    ) -> Categorical: ...
+    ) -> Categorical[int]: ...
```
Dr-Irv: the idea here should be if `categories` is NOT specified, then the type of the result is dependent on the type of `dtype`.  So if what is passed is `CategoricalDtype`, then result is `Categorical[Any]`, but if `CategoricalDtype[CategoryValueT]` is passed, result is `Categorical[CategoricalValueT]`

So you have to play a bit with the `*` in there to handle this correctly.

pandas-stubs/core/arrays/categorical.pyi
```diff
@@ -240,7 +244,7 @@ class CategoricalAccessor(
     ) -> Series[CategoricalDtype[CategoricalValueT]]: ...
     def reorder_categories(
         self,
-        new_categories: ListLike,
+        new_categories: Sequence[CategoricalValueT] | AnyArrayLike,
```
Dr-Irv: create 2 overloads.  If the parameter is `Sequence[CategoricalValueT]`, result is `Series[CategoricalDtype[CategoricalValueT]]`.  If the parameter is `AnyArrayLike`, then it  is `Series[Any]`

pandas-stubs/core/series.pyi
```diff
@@ -373,10 +373,45 @@ class Series(IndexOpsMixin[S1], ElementOpsMixin[S1], NDFrame):
...
+        index: AxesData | None = None,
+        *,
+        dtype: CategoryDtypeArg,
+        name: Hashable = None,
+        copy: bool | None = None,
+    ) -> Series[CategoricalDtype[int]]: ...
+    @overload
     def __new__(  # type: ignore[overload-overlap]
+        cls,
+        data: Sequence[_str] | np_1darray_str | Index[_str] | Series[_str],
```
Dr-Irv: Use `SequenceNotStr[_str]` here

pandas-stubs/core/arrays/categorical.pyi
```diff
@@ -34,60 +36,142 @@ from pandas._typing import (
...
-    def as_ordered(self) -> Categorical: ...
-    def as_unordered(self) -> Categorical: ...
+    def as_ordered(self) -> Self: ...
+    def as_unordered(self) -> Self: ...
     def set_categories(
         self,
         new_categories: AnyArrayLike | SequenceNotStr[Hashable],
         ordered: bool | None = False,
         rename: bool = False,
     ) -> Self: ...
```
Dr-Irv, line 168: this has to be fixed still., probably with an overload.  Because if you pass in `SequenceNotStr[CategoricalValueT1]`, `Index[CategoricalValueT1]` or `Series[CategoricalValueT1]` , the result is `Categorical[CategoricalValueT1]`.  But if an  untyped `ArrayLike` is passed, then result is `Categorical[Any]`

Also, `SequenceNotStr[Hashable]` is too wide.  We only want to accept `str`, `int`, `float`

pandas-stubs/core/arrays/categorical.pyi
Dr-Irv: I"m resolving this because I did a new comment about it

pandas-stubs/core/arrays/categorical.pyi
Dr-Irv: ok

Dr-Irv (changes requested):
almost there

pandas-stubs/core/arrays/categorical.pyi
loicdiridollou: Do we want series[Any] or Series[CategoricalDtype[Any]] (without a specific subtype)

pandas-stubs/core/arrays/categorical.pyi
loicdiridollou: Fixed

pandas-stubs/core/arrays/categorical.pyi
Dr-Irv: `Series[CategoricalDtype[Any]]` because we know it is a categorical series

loicdiridollou: Perfect, in that case we are good, I put `Series[CategoricalDtype]` in the last commit.

Dr-Irv (approved):
thanks @loicdiridollou  .  Great work!

PR #1749, opened 2026-05-23
MNT: Fix missing check/assert_type harness where missing

loicdiridollou:

Went through the tests and found places where we don't have the check/assert_type on statements that deserve to be checked.

### DIFF
--- a/tests/arrays/test_base_string_.py
+++ b/tests/arrays/test_base_string_.py
@@ -38,8 +38,8 @@ def test_construction_sequence(
     missing_values: tuple[Any, ...],
     typ: Callable[[Sequence[Any]], Sequence[Any]],
 ) -> None:
-    check(assert_type(pd.array(typ([*data, *missing_values])), BaseStringArray), BaseStringArray)  # type: ignore[assert-type] # pyright: ignore[reportAssertTypeFailure]
-    check(assert_type(pd.array(typ([*data, *data, *missing_values])), BaseStringArray), BaseStringArray)  # type: ignore[assert-type] # pyright: ignore[reportAssertTypeFailure]
+    check(pd.array(typ([*data, *missing_values])), BaseStringArray)
+    check(pd.array(typ([*data, *data, *missing_values])), BaseStringArray)
 
     if TYPE_CHECKING:
         assert_type(pd.array(["pd", np.str_("pd")]), BaseStringArray)
--- a/tests/arrays/test_base_string_.py
+++ b/tests/arrays/test_base_string_.py
@@ -38,8 +38,8 @@ def test_construction_sequence(
     missing_values: tuple[Any, ...],
     typ: Callable[[Sequence[Any]], Sequence[Any]],
 ) -> None:
-    check(pd.array(typ([*data, *missing_values])), BaseStringArray)
-    check(pd.array(typ([*data, *data, *missing_values])), BaseStringArray)
+    check(assert_type(pd.array(typ([*data, *missing_values])), BaseStringArray), BaseStringArray)  # type: ignore[assert-type] # pyright: ignore[reportAssertTypeFailure]
+    check(assert_type(pd.array(typ([*data, *data, *missing_values])), BaseStringArray), BaseStringArray)  # type: ignore[assert-type] # pyright: ignore[reportAssertTypeFailure]
 
     if TYPE_CHECKING:
         assert_type(pd.array(["pd", np.str_("pd")]), BaseStringArray)
@@ -81,8 +81,16 @@ def test_construction_dtype(
     is_builtin_str = dtype in PYTHON_STRING_ARGS
 
     dtype_notna = target_dtype if data else None
-    check(pd.array([*data], dtype), BaseStringArray, dtype_notna)
-    check(pd.array([*data, *data], dtype), BaseStringArray, dtype_notna)
+    check(
+        assert_type(pd.array([*data], dtype), BaseStringArray),
+        BaseStringArray,
+        dtype_notna,
+    )
+    check(
+        assert_type(pd.array([*data, *data], dtype), BaseStringArray),
+        BaseStringArray,
+        dtype_notna,
+    )
 
     dtype_na = (
         target_dtype
@@ -90,8 +98,16 @@ def test_construction_dtype(
         # pandas-dev/pandas#63567 Pandas 3.0 gives StringDtype(na_value=nan) if dtype is str or "str"
         else float if is_builtin_str else NAType
     )
-    check(pd.array([*data, np.nan], dtype), BaseStringArray, dtype_na)
-    check(pd.array([*data, *data, np.nan], dtype), BaseStringArray, dtype_na)
+    check(
+        assert_type(pd.array([*data, np.nan], dtype), BaseStringArray),
+        BaseStringArray,
+        dtype_na,
+    )
+    check(
+        assert_type(pd.array([*data, *data, np.nan], dtype), BaseStringArray),
+        BaseStringArray,
+        dtype_na,
+    )
 
     if TYPE_CHECKING:
         assert_type(pd.array([], str), BaseStringArray)
--- a/tests/arrays/test_datetime_array.py
+++ b/tests/arrays/test_datetime_array.py
@@ -65,10 +65,10 @@ def test_construction_sequence_pandas(
     missing_values: tuple[Any, ...],
     typ: Callable[[Sequence[Any]], Sequence[Any]],
 ) -> None:
-    check(pd.array(typ([*data, *missing_values])), DatetimeArray)
-    check(pd.array(typ([datetime(2077, 1, 1), *data, *missing_values])), DatetimeArray)
+    check(assert_type(pd.array(typ([*data, *missing_values])), DatetimeArray), DatetimeArray)  # type: ignore[assert-type] # pyright: ignore[reportAssertTypeFailure]
+    check(assert_type(pd.array(typ([datetime(2077, 1, 1), *data, *missing_values])), DatetimeArray), DatetimeArray)  # type: ignore[assert-type] # pyright: ignore[reportAssertTypeFailure]
     check(
-        pd.array(typ([pd.Timestamp(1988, 1, 1), *data, *missing_values])), DatetimeArray
+        assert_type(pd.array(typ([pd.Timestamp(1988, 1, 1), *data, *missing_values])), DatetimeArray), DatetimeArray  # type: ignore[assert-type] # pyright: ignore[reportAssertTypeFailure]
     )
 
     if TYPE_CHECKING:
@@ -146,17 +146,27 @@ def test_construction_dtype(
     target_dtype: type,
 ) -> None:
     dtype_notna = target_dtype if data else None
-    check(pd.array([*data], dtype), DatetimeArray, dtype_notna)
     check(
-        pd.array([np.datetime64("1748-12-24"), *data], dtype),
+        assert_type(pd.array([*data], dtype), DatetimeArray), DatetimeArray, dtype_notna
+    )
+    check(
+        assert_type(
+            pd.array([np.datetime64("1748-12-24"), *data], dtype), DatetimeArray
+        ),
         DatetimeArray,
         dtype_notna,
     )
 
     dtype_na = target_dtype if data else NaTType
-    check(pd.array([*data, np.nan], dtype), DatetimeArray, dtype_na)
     check(
-        pd.array([np.datetime64("2048-12-24"), *data, np.nan], dtype),
+        assert_type(pd.array([*data, np.nan], dtype), DatetimeArray),
+        DatetimeArray,
+        dtype_na,
+    )
+    check(
+        assert_type(
+            pd.array([np.datetime64("2048-12-24"), *data, np.nan], dtype), DatetimeArray
+        ),
         DatetimeArray,
         target_dtype,
     )
--- a/tests/arrays/test_datetime_array.py
+++ b/tests/arrays/test_datetime_array.py
@@ -65,10 +65,10 @@ def test_construction_sequence_pandas(
     missing_values: tuple[Any, ...],
     typ: Callable[[Sequence[Any]], Sequence[Any]],
 ) -> None:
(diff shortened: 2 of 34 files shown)


tests/arrays/test_base_string_.py
```diff
@@ -38,8 +38,8 @@ def test_construction_sequence(
     missing_values: tuple[Any, ...],
     typ: Callable[[Sequence[Any]], Sequence[Any]],
 ) -> None:
-    check(pd.array(typ([*data, *missing_values])), BaseStringArray)
-    check(pd.array(typ([*data, *data, *missing_values])), BaseStringArray)
+    check(assert_type(pd.array(typ([*data, *missing_values])), BaseStringArray), BaseStringArray)  # type: ignore[assert-type] # pyright: ignore[reportAssertTypeFailure]
```
loicdiridollou: Debating whether to have this here or not, the issue is that we check the `assert_type` below in the `TYPE_CHECKING` control statement.
It is the same for other places.
Open to debate on this one.

loicdiridollou: Actually reverted that change, did not like it and it made the code messier than needed.

tests/arrays/test_datetime_array.py
```diff
@@ -65,10 +65,10 @@ def test_construction_sequence_pandas(
     missing_values: tuple[Any, ...],
     typ: Callable[[Sequence[Any]], Sequence[Any]],
 ) -> None:
-    check(pd.array(typ([*data, *missing_values])), DatetimeArray)
-    check(pd.array(typ([datetime(2077, 1, 1), *data, *missing_values])), DatetimeArray)
+    check(assert_type(pd.array(typ([*data, *missing_values])), DatetimeArray), DatetimeArray)  # type: ignore[assert-type] # pyright: ignore[reportAssertTypeFailure]
+    check(assert_type(pd.array(typ([datetime(2077, 1, 1), *data, *missing_values])), DatetimeArray), DatetimeArray)  # type: ignore[assert-type] # pyright: ignore[reportAssertTypeFailure]
     check(
-        pd.array(typ([pd.Timestamp(1988, 1, 1), *data, *missing_values])), DatetimeArray
+        assert_type(pd.array(typ([pd.Timestamp(1988, 1, 1), *data, *missing_values])), DatetimeArray), DatetimeArray  # type: ignore[assert-type] # pyright: ignore[reportAssertTypeFailure]
```
Dr-Irv: I'm not a fan of having `#type: ignore` on lines that the type checker should see as correct.

I think this can be fixed by changing `typ` to be `Callable[[Sequence[datetime | pd.Timestamp]], Sequence[datetime | pd.Timestamp]]`  If not, then put in a comment indicating why we don't have `assert_type` here.

tests/arrays/test_floating.py
```diff
@@ -66,7 +66,7 @@ def test_construction_sequence(
 def test_construction_sequence_nan(
     data: tuple[Any, ...], typ: Callable[[Sequence[Any]], Sequence[Any]]
 ) -> None:
-    check(pd.array(typ(data)), FloatingArray)
+    check(assert_type(pd.array(typ(data)), FloatingArray), FloatingArray)  # type: ignore[assert-type]
```
Dr-Irv, line 70: Need a comment here as to why `mypy` can't figure it out, but `pyright` can

tests/test_groupby.py
```diff
@@ -390,14 +396,43 @@ def s2series(val: Series) -> Series:
...
-    check(GB_S.resample("ME").aggregate([np.mean]), DataFrame)
-    check(GB_S.resample("ME").aggregate(["sum", np.mean]), DataFrame)
-    check(GB_S.resample("ME").aggregate({"sum": np.sum}), DataFrame)
-    check(GB_S.resample("ME").aggregate({"sum": np.sum, "mean": np.mean}), DataFrame)
-    check(GB_S.resample("ME").aggregate("sum"), Series)
-    check(GB_S.resample("ME").aggregate(s2series), Series)
-    check(GB_S.resample("ME").aggregate(s2scalar), Series)
+    check(
+        assert_type(GB_S.resample("ME").aggregate(np.sum), "Series | DataFrame"), Series
+    )
```
Dr-Irv: I think this is a bug in the types.  Because `GB_S.resample("ME")` has type `DatetimeIndexResamplerGroupby[Series[float]] | PeriodIndexResamplerGroupby[Series[float]] | TimedeltaIndexResamplerGroupby[Series[float]]`, so if `aggregate` with a single argument is passed to any of these, the result should be `Series`.  I think the type here is incorrect and should return `Series`
https://github.com/pandas-dev/pandas-stubs/blob/ee82af586ddcaf86bb9bfc64cb4f07a7793ffa16/pandas-stubs/core/resample.pyi#L68-L73

tests/test_groupby.py
```diff
@@ -390,14 +396,43 @@ def s2series(val: Series) -> Series:
...
+        assert_type(
+            GB_S.resample("ME").aggregate({"sum": np.sum}), "Series | DataFrame"
+        ),
+        DataFrame,
+    )
+    check(
+        assert_type(
+            GB_S.resample("ME").aggregate({"sum": np.sum, "mean": np.mean}),
+            "Series | DataFrame",
+        ),
```
Dr-Irv: types wrong here, because if the groupby operation is a list or dict, then the result should be `DataFrame`

tests/test_groupby.py
```diff
@@ -390,14 +396,43 @@ def s2series(val: Series) -> Series:
...
+        assert_type(GB_S.resample("ME").aggregate("sum"), "Series | DataFrame"), Series
+    )
+    check(
+        assert_type(GB_S.resample("ME").aggregate(s2series), "Series | DataFrame"),
+        Series,
+    )
+    check(
+        assert_type(GB_S.resample("ME").aggregate(s2scalar), "Series | DataFrame"),
+        Series,
+    )
```
Dr-Irv, line 426: All of these should be `Series`

tests/test_resampler.py
```diff
@@ -370,14 +370,33 @@ def s2series(val: Series) -> Series:
...
 
-    check(S.resample("ME").aggregate(np.sum), Series)
-    check(S.resample("ME").aggregate([np.mean]), DataFrame)
-    check(S.resample("ME").aggregate(["sum", np.mean]), DataFrame)
-    check(S.resample("ME").aggregate({"sum": np.sum}), DataFrame)
-    check(S.resample("ME").aggregate({"sum": np.sum, "mean": np.mean}), DataFrame)
-    check(S.resample("ME").aggregate("sum"), Series)
-    check(S.resample("ME").aggregate(s2series), Series)
-    check(S.resample("ME").aggregate(s2scalar), Series)
+    check(assert_type(S.resample("ME").aggregate(np.sum), "Series | DataFrame"), Series)
```
Dr-Irv: All of these from line 373 through 396 should be fixed by fixing the typing

Dr-Irv (changes requested):

tests/arrays/test_datetime_array.py
loicdiridollou: Actually I forgot to remove the changes I had made here, the assert_type are handled in the `if TYPE_CHECKING' block below, will remove.

tests/arrays/test_floating.py
loicdiridollou: Same here, actually I forgot to remove the changes I had made here, the assert_type are handled in the `if TYPE_CHECKING' block below, will remove.

tests/arrays/test_datetime_array.py
Dr-Irv: So be careful. We should only be doing `if TYPE_CHECKING` around code that doesn't work in `pytest`, but does work in normal pandas usage.  I don't think that should be true anyway, except for places  where we do imports that are typing specific.   I think we violate this assumption in various places in the tests.  

So this usage in `tests/frame/test_frame.py` is OK:
```if TYPE_CHECKING:
    from pandas.core.frame import PandasNamedTuple
else:
    PandasNamedTuple: TypeAlias = tuple
```
because `PandasNamedTuple` is something only in our type checking and is not available at runtime.

And this usage might be OK, because we might not be able to do a plot with `pytest` (although we should check on that...)
```python
    if TYPE_CHECKING:  # skip pytest
        df = pd.DataFrame(data={"col1": [1, 1, 2], "col2": [3, 4, 5]})
        df.plot.hist()
        df.plot.scatter(x="col2", y="col1")
```

And this usage is NOT OK, because the code should execute:
```python
    if TYPE_CHECKING:
        assert_type(df.where(where_cond1), pd.DataFrame)
        assert_type(df.where(where_cond2), pd.DataFrame)
        assert_type(df.where(where_cond3), pd.DataFrame)
        assert_type(df.where(where_cond3, None), pd.DataFrame)
        assert_type(df.where(where_cond1, inplace=True), pd.DataFrame)
        assert_type(df.where(where_cond2, inplace=True), pd.DataFrame)
        assert_type(df.where(where_cond3, inplace=True), pd.DataFrame)
        assert_type(df.where(where_cond3, None, inplace=True), pd.DataFrame)
```

Might need to create a separate issue on this to fix these usages.

We also have `if TYPE_CHECKING_INVALID_USAGE:` where we know the code will fail, and we want the type checker to pick up those failures ahead of time.

loicdiridollou: So this approach of doing the `check` in the function and the `assert_type` for all the versions of the overloads is taken care of in the `if TYPE_CHECKING` block.
I think @cmp0xff introduced this parametrized approach when we have a lot of overloads.
I have no preference, I understand it is more compact and less readable but not disturbing to me either.

tests/test_groupby.py
loicdiridollou: Good catch, will update the types to support that!

tests/test_groupby.py
loicdiridollou: Good catch, will fix those!

tests/test_resampler.py
loicdiridollou: Correct, the fix for the the groupby should over those, let me see.

tests/arrays/test_floating.py
loicdiridollou: Will add a comment, mypy sees it as Any, pd.array(Sequence[Any]) is matched by pyright to `Sequence[Just[float]]` for some reason, will add a comment

tests/test_resampler.py
loicdiridollou: Those are fixed now

tests/arrays/test_datetime_array.py
Dr-Irv: I don't see a reason to put an `if TYPE_CHECKING:` around the `assert_type()` calls.  Should include a comment as to what we are doing here - doing the `check` in a parameterized way, but still checking explicit types.

loicdiridollou: So the `check` part is done at runtime with a regular run of `pytest`, the `assert_type` step is done in the block for TYPE_CHECKING. The reason I believe they are in the block is so that the code does not run for every parameter combination since `assert_type` checks are for mypy/pyright.
Let's discuss in a side issue since this was not the core of this PR https://github.com/pandas-dev/pandas-stubs/issues/1750

tests/arrays/test_floating.py
```diff
@@ -66,7 +66,8 @@ def test_construction_sequence(
 def test_construction_sequence_nan(
     data: tuple[Any, ...], typ: Callable[[Sequence[Any]], Sequence[Any]]
 ) -> None:
-    check(assert_type(pd.array(typ(data)), FloatingArray), FloatingArray)  # type: ignore[assert-type]
+    # can't use assert_type as mypy sees the result as Any while pyright matches to FloatingArray
+    check(pd.array(typ(data)), FloatingArray)
```
Dr-Irv: You can keep the `assert_type` here, and just have the `# type: ignore` for `mypy`, with the comment that says that `mypy` is not inferring this correctly.

tests/test_groupby.py
```diff
@@ -327,43 +329,43 @@ def test_series_groupby_resample() -> None:
...
         Series,
     )
     check(
-        assert_type(GB_S.resample("ME").agg(np.sum), DataFrame | Series),
+        assert_type(GB_S.resample("ME").agg(np.sum), "Series[Any]"),
         Series,
     )
     check(
-        assert_type(GB_S.resample("ME").apply(np.sum), DataFrame | Series),
+        assert_type(GB_S.resample("ME").apply(np.sum), "Series[Any]"),
```
Dr-Irv: I don't think you need the `Any` here.  You can just use `Series` (and in other places as well)

tests/test_groupby.py
```diff
@@ -894,7 +887,7 @@ def test_engine() -> None:
             engine_kwargs="not valid",  # pyright: ignore
             other_kwarg="",
         )
-    GB_DF.aggregate("size", engine="cython", engine_kwargs={})
+    assert_type(GB_DF.aggregate("size", engine="cython", engine_kwargs={}), DataFrame)
```
Dr-Irv: why not include `check` here?

Dr-Irv (changes requested):

tests/arrays/test_floating.py
loicdiridollou: Fixed

tests/test_groupby.py
loicdiridollou: Fixed on all

tests/test_groupby.py
loicdiridollou: Not sure what happened here with the script here. Also the type hint was wrong so fixed that too

tests/test_resampler.py
```diff
@@ -24,15 +25,15 @@
     check,
 )
 
+if TYPE_CHECKING:
+    from typing import Any  # noqa: F401
+
```
Dr-Irv: need to remove this here as well

Dr-Irv (changes requested):
just fix `test_resampler.py` and we should be good

tests/test_resampler.py
loicdiridollou: Thanks, that should be good now!

Dr-Irv (approved):
thanks @loicdiridollou

PR #1753, opened 2026-05-29
Make mypy reject `float(Series)` calls

sterliakov:
Similar to the SupportsIndex problem solved by explicit `__index__: ClassVar[None]` addition, there are other magic methods worth explicitly removing. I propose four more: `__float__`, `__int__`, `__trunc__` (pre-3.14 alternative integer conversion), and `__buffer__`. This will ban doing `int(series)` and `float(series)` for `Series[Any]`.

Those already raise an error with `pyright`, and `mypy` does not pick them up due to `__getattr__`. Neither of typechecker implementations is obviously correct (because direct access to protocol member - `series.__float__` - will indeed return a member named `__float__` if it exists), so I suggest to pick an approach that helps all popular checkers.

I'm not a bot, you may have met me in mypy before:)

### DIFF
--- a/pandas-stubs/core/series.pyi
+++ b/pandas-stubs/core/series.pyi
@@ -21,6 +21,7 @@ from datetime import (
     timedelta,
 )
 from pathlib import Path
+import sys
 from typing import (
     Any,
     ClassVar,
@@ -357,6 +358,8 @@ class Series(IndexOpsMixin[S1], ElementOpsMixin[S1], NDFrame):
     # Same as above but for int() and float()
     __float__: ClassVar[None]
     __int__: ClassVar[None]
+    if sys.version_info < (3, 14):
+        __trunc__: ClassVar[None]
     __buffer__: ClassVar[None]
     __hash__: ClassVar[None]  # pyright: ignore[reportIncompatibleMethodOverride]
 
--- a/pandas-stubs/core/series.pyi
+++ b/pandas-stubs/core/series.pyi
@@ -354,6 +354,10 @@ _DataLike: TypeAlias = ArrayLike | dict[str, np_ndarray] | SequenceNotStr[S1]
 class Series(IndexOpsMixin[S1], ElementOpsMixin[S1], NDFrame):
     # Define __index__ because mypy thinks Series follows protocol `SupportsIndex` https://github.com/pandas-dev/pandas-stubs/pull/1332#discussion_r2285648790
     __index__: ClassVar[None]
+    # Same as above but for int() and float()
+    __float__: ClassVar[None]
+    __int__: ClassVar[None]
+    __buffer__: ClassVar[None]
     __hash__: ClassVar[None]  # pyright: ignore[reportIncompatibleMethodOverride]
 
     @overload
--- a/pandas-stubs/core/series.pyi
+++ b/pandas-stubs/core/series.pyi
@@ -21,7 +21,6 @@ from datetime import (
     timedelta,
 )
 from pathlib import Path
-import sys
 from typing import (
     Any,
     ClassVar,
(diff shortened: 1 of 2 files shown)


Dr-Irv:
@sterliakov tell me when you're ready for review, and then I can trigger the CI

sterliakov:
@Dr-Irv thanks, that was quicker than I expected! I think this is good to go now. All tests except for 6 missing TZ names (probably something is misconfigured on my system, will check) passed locally, including the new one

pandas-stubs/core/series.pyi
```diff
@@ -354,6 +355,12 @@ _DataLike: TypeAlias = ArrayLike | dict[str, np_ndarray] | SequenceNotStr[S1]
 class Series(IndexOpsMixin[S1], ElementOpsMixin[S1], NDFrame):
     # Define __index__ because mypy thinks Series follows protocol `SupportsIndex` https://github.com/pandas-dev/pandas-stubs/pull/1332#discussion_r2285648790
     __index__: ClassVar[None]
+    # Same as above but for int() and float()
+    __float__: ClassVar[None]
+    __int__: ClassVar[None]
+    if sys.version_info < (3, 14):
+        __trunc__: ClassVar[None]
+    __buffer__: ClassVar[None]
```
Dr-Irv: can you add tests for these?

tests/series/test_series.py
```diff
@@ -3809,3 +3809,15 @@ def test_series_copy_deprecated() -> None:
         _7 = s.astype(int, copy=True)  # type: ignore[call-overload] # pyright: ignore[reportCallIssue,reportUnknownVariableType]
         # swaplevel
         _8 = s.swaplevel(copy=True)  # type: ignore[call-arg] # pyright: ignore[reportCallIssue,reportUnknownVariableType]
+
+
+def test_series_float_protocols() -> None:
```
Dr-Irv: since you're testing `float` and `int`, either separate the tests, or you could rename this to `test_series_invalid_dunder_methods` (or something like that), and then add tests for `trunc` and `buffer`

Dr-Irv (changes requested):

pandas-stubs/core/series.pyi
sterliakov: They're already there - CI test matrix covers both 3.14 and pre-3.14, and the signatures of `int` and `float` in typeshed are as follows:

```python
# _typeshed.pyi, https://github.com/python/typeshed/blob/1d3abc42077c2ea47985363a4fab387357f3aa77/stdlib/_typeshed/__init__.pyi#L368-L373
if sys.version_info >= (3, 14):
    ConvertibleToInt: TypeAlias = str | ReadableBuffer | SupportsInt | SupportsIndex
else:
    ConvertibleToInt: TypeAlias = str | ReadableBuffer | SupportsInt | SupportsIndex | SupportsTrunc
ConvertibleToFloat: TypeAlias = str | ReadableBuffer | SupportsFloat | SupportsIndex

# builtins.pyi, https://github.com/python/typeshed/blob/1d3abc42077c2ea47985363a4fab387357f3aa77/stdlib/builtins.pyi#L253
class int:
    @overload
    def __new__(cls, x: ConvertibleToInt = 0, /) -> Self: ...
    @overload
    def __new__(cls, x: str | bytes | bytearray, /, base: SupportsIndex) -> Self: ...
    

class float:
    def __new__(cls, x: ConvertibleToFloat = 0, /) -> Self: ...
```

`ReadableBuffer` is an alias to `collections.abc.Buffer` on 3.12+ and an empty ABC before that. `Supports*` protocols are exactly what you think they are, single-member protocol with the magic method implied by their name.

So `__trunc__` is needed to reject `int()`, and `__buffer__` is needed for both.

Dr-Irv: I'm missing something here.  You added tests for `int(Series)` and `float(Series)`, so shouldn't there be tests for `math.trunc(Series)` and `memoryview(Series)` ?

tests/series/test_series.py
sterliakov: I renamed the test and added other operations resembling primitive conversions. This is not about magic methods in general, but only about a subset of them. 

I also removed the `sys.version` guard - we don't want `math.trunc(series)` on 3.14+ too anyway, so let's disable it unconditionally. And I added the missing methods for ceil, floor and complex to be consistent.

pandas-stubs/core/series.pyi
sterliakov: Added memoryview, thanks! My point was that `int(s2)` will be allowed with any of these magic methods overrides removed: we need `Series[Any]` to not conform to *any* of those protocols, so `__trunc__` alone is enough to make `int(s2)` allowed, and `__buffer__` too. 

<s>`memoryview` isn't Buffer-only, it's just another type that accepts `Buffer | something_else` in their constructor, as `int` does. </s> I stand corrected, there are no other overloads or union members, it has only `def __new__(cls, obj: ReadableBuffer) -> Self: ...`

sterliakov:
Added tests for other primitive ops, including allowed ones - this should be ready for another round!

Dr-Irv (approved):
thanks @sterliakov

PR #1754, opened 2026-05-30
GHA workflow updates

TomAugspurger:
This updates a few things in our GitHub Actions. See the individual commits for details.

I've tried to preserve the behavior of things, but we might need to go through a couple iterations to make sure everything still works properly.

### DIFF
--- a/.github/workflows/comment_commands.yml
+++ b/.github/workflows/comment_commands.yml
@@ -25,7 +25,7 @@ jobs:
         # get the sha of the last commit to attach the results of the tests
         if: always()
         id: get-branch-info
-        uses: actions/github-script@v9
+        uses: actions/github-script@3a2844b7e9c422d3c10d287c895573f7108da1b3 # v9
         with:
           github-token: ${{ secrets.GITHUB_TOKEN }}
           script: |
@@ -39,7 +39,7 @@ jobs:
             core.setOutput('repository', pr.data.head.repo.full_name)
 
       - name: Checkout code on the correct branch
-        uses: actions/checkout@v6
+        uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6
         with:
           # context is not aware which branch to checkout so it would otherwise
           # default to main (we also need repo name to source from the right user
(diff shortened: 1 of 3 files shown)


Dr-Irv:
@TomAugspurger I tested the comment command, and it worked.  Output here:  https://github.com/pandas-dev/pandas-stubs/actions/runs/26689612269/job/78663597345

I have no way to judge whether what you've done addresses the security concerns.

I can approve and merge this as is - should I do that, or do you want to add more things?

TomAugspurger:
Just to be clear, membership of the pandas org (or maybe this repo? I'm not sure) lets that comment command work. Non-members can't trigger the action.

It should be good to go.

loicdiridollou:
A little bit unfamiliar with all those settings, one small question, can the user trigger the commands for nightly? That is quite useful so the reviewer does not have to do it.

TomAugspurger:
No, I don't think so, because of the change to this line: https://github.com/pandas-dev/pandas-stubs/pull/1754/changes#diff-7d4a1b384c8f4e89dda2472782e10e36e026ae89e6bc0cd5df7c774d1fd07c32L21. A maintainer will need to do that.

Dr-Irv:
> A little bit unfamiliar with all those settings, one small question, can the user trigger the commands for nightly? That is quite useful so the reviewer does not have to do it.

I think this will be OK.  Our only comment commands are `/pandas_nightly` and `/mypy_nightly` and I think the members should be only doing this in the course of a review.

Dr-Irv (approved):
thanks @TomAugspurger

TomAugspurger:
Thanks!

It looks like the mypy-nightly job failed on main, but seemingly an issue with the type stubs rather than the workflows: https://github.com/pandas-dev/pandas-stubs/actions/runs/26718305826/job/78740713249

```

tests/indexes/test_indexes.py:1638: error: List item 0 has incompatible type "int"; expected "Never"  [list-item]
tests/indexes/test_indexes.py:1638: note: Error code "list-item" not covered by "type: ignore[arg-type]" comment
tests/series/test_series.py:2108: error: Expression is of type "ndarray[tuple[int], dtype[numpy.bool[Any]]]", not "ndarray[tuple[int], dtype[numpy.bool[builtins.bool]]]"  [assert-type]
Found 2 errors in 2 files (checked 345 source files)
```

LMK if you do spot any issues with the workflows and I'd be happy to take a look.

loicdiridollou:
The mypy nightly is not related to the changes no worries you are good, thanks for the PR!

PR #1757, opened 2026-06-03
GH1347 Fix Series/Index arithmetic overloads and ty issue

### LINKED ISSUE #1347: Index arithmetic overloads show wrong types

loicdiridollou:
- [x] Closes #1347

A new version of `ty` does not raise the same overload issues than previous.


tests/test_natype.py
```diff
@@ -84,22 +84,20 @@ def test_arithmetic() -> None:
...
-    #     ),
-    #     tuple,
-    # )
-    # https://github.com/microsoft/pyright/issues/10899.
+    check(
+        assert_type(
+            divmod(na, s_int),
+            tuple[pd.Series, pd.Series],
+        ),
+        tuple,
```
cmp0xff, line 92:
```suggestion
        assert_type(divmod(na, s_int), tuple[pd.Series, pd.Series]),
        tuple,
        pd.Series,
```
Also for the lines below

tests/test_natype.py
```diff
@@ -111,16 +109,14 @@ def test_arithmetic() -> None:
...
-    #     tuple,
-    # )
-    # https://github.com/pandas-dev/pandas-stubs/issues/1347
-    # check(
-    #     assert_type(divmod(idx_int, na), tuple[pd.Index, pd.Index]),
-    #     tuple,
-    # )
+    check(
+        assert_type(divmod(s_int, na), tuple["pd.Series[int]", "pd.Series[int]"]),
+        tuple,
```
cmp0xff:
```suggestion
        assert_type(divmod(s_int, na), "tuple[pd.Series[int], pd.Series[int]]"),
        tuple,
        pd.Series,
```
also for the lines below

cmp0xff (changes requested):

tests/_typing.py
cmp0xff, line 1: Oops, this is to be reverted.

tests/test_natype.py
```diff
@@ -111,16 +111,16 @@ def test_arithmetic() -> None:
...
-    #     assert_type(divmod(s_int, na), tuple[pd.Series, pd.Series]),
-    #     tuple,
-    # )
-    # https://github.com/pandas-dev/pandas-stubs/issues/1347
-    # check(
-    #     assert_type(divmod(idx_int, na), tuple[pd.Index, pd.Index]),
-    #     tuple,
-    # )
+    check(
+        assert_type(divmod(s_int, na), tuple["pd.Series[int]", "pd.Series[int]"]),
```
cmp0xff:
```suggestion
        assert_type(divmod(s_int, na), "tuple[pd.Series[int], pd.Series[int]]"),
```

tests/test_natype.py
```diff
@@ -111,16 +111,16 @@ def test_arithmetic() -> None:
...
+        tuple,
+        pd.Series,
+    )
+    check(
+        assert_type(divmod(idx_int, na), tuple["pd.Index[int]", "pd.Index[int]"]),
```
cmp0xff:
```suggestion
        assert_type(divmod(idx_int, na), "tuple[pd.Index[int], pd.Index[int]]"),
```

cmp0xff (changes requested):

tests/_typing.py
loicdiridollou: Gosh that thing keeps happening when there is a test failure with `poetry run poe pytest`, low key wants to add a post step if `pytest` fails to revert the changes from that step, may look into it since I have been guilty of modifying that file too many times.

cmp0xff (approved):
Thank you @loicdiridollou !

tests/_typing.py
cmp0xff: Yep it's a bit annoying and one often forgets to keep it unchanged. However this was the best design I could work out.

PR #1763, opened 2026-06-06
Allow SQLAlchemy types in DataFrame.to_sql dtype

### LINKED ISSUE #1762: BUG: dtype parameter of DataFrame.to_sql() rejects SQLAlchemy type instances
**Describe the bug**

The `dtype` parameter of `DataFrame.to_sql()` raises a type error when passing SQLAlchemy type *instances* (e.g. `NVARCHAR(length=1024)`). Pandas itself accepts instances — this was explicitly made possible in [pandas#9138](https://github.com/pandas-dev/pandas/issues/9138) — but the stubs do not reflect this.

The error from Pyright/Pylance is:

```
Argument of type "dict[str, type[UNIQUEIDENTIFIER[_UUID_RETURN@UNIQUEIDENTIFIER]] | NVARCHAR]" cannot be assigned to parameter "dtype" of type "DtypeArg | None" in function "to_sql".
```

**To Reproduce**

1. Minimal example:

```python
import pandas as pd
from sqlalchemy import create_engine
from sqlalchemy.types import NVARCHAR
from sqlalchemy.dialects.mssql import UNIQUEIDENTIFIER
import uuid

engine = create_engine("mssql+pyodbc://...")

df = pd.DataFrame(
    columns=['id', 'name'],
    data=[[str(uuid.uuid4()), 'hello']]
)

df.to_sql(
    'test',
    con=engine,
    dtype={
        'id': UNIQUEIDENTIFIER,
        'name': NVARCHAR(length=1024),  # instance, not class
    },
    index=False,
)
```

2. Type checker: **Pyright/Pylance**

3. Error message:
```
Argument of type "dict[str, type[UNIQUEIDENTIFIER[_UUID_RETURN@UNIQUEIDENTIFIER]] | NVARCHAR]" cannot be assigned to parameter "dtype" of type "DtypeArg | None" in function "to_sql".
```

- OS: Windows 11
- OS Version: 10.0.26100
- Python version: 3.13.0
- Type checker version: 2026.2.1
- pandas-stubs version: 3.0.3.260530

**Additional context**

Originally reported at [pandas-dev/pandas#61660](https://github.com/pandas-dev/pandas/issues/61660), where maintainers confirmed this should be filed against pandas-stubs.

The `dtype` argument to `to_sql()` accepts a dict of column names to SQLAlchemy types, and SQLAlchemy types are commonly passed as instances in order to configure parameters like `length`. The `DtypeArg` type does not appear to accommodate SQLAlchemy type instances (only classes), which means valid pandas usage cannot be correctly typed.

SAY-5:
- [x] Closes #1762

`DataFrame.to_sql` passes its `dtype` mapping straight through to SQLAlchemy, so the values can be SQLAlchemy types (either the class or an instance like `NVARCHAR(length=1024)`), but the stub only accepted `DtypeArg`. Widen the `dtype` parameter to also allow a column mapping of `TypeEngine` classes/instances.

### DIFF
--- a/pandas-stubs/core/generic.pyi
+++ b/pandas-stubs/core/generic.pyi
@@ -27,14 +27,14 @@ from pandas import Index
 from pandas.core.resample import DatetimeIndexResampler
 from pandas.core.series import Series
 from sqlalchemy.engine import Connectable
-from sqlalchemy.types import TypeEngine
+from sqlalchemy.sql.type_api import TypeEngineMixin
 
 from pandas._libs.lib import NoDefault
 from pandas._typing import (
     Axis,
     CompressionOptions,
     CSVQuoting,
-    DtypeArg,
+    Dtype,
     DtypeBackend,
     ExcelWriterMergeCells,
     FilePath,
@@ -177,7 +177,11 @@ class NDFrame:
         index_label: IndexLabel = None,
         chunksize: int | None = None,
         dtype: (
-            DtypeArg | Mapping[Hashable, type[TypeEngine[Any]] | TypeEngine[Any]] | None
+            Dtype
+            | type[TypeEngineMixin]
+            | TypeEngineMixin
+            | Mapping[Hashable, Dtype | type[TypeEngineMixin] | TypeEngineMixin]
+            | None
         ) = None,
         method: (
             Literal["multi"]
--- a/pandas-stubs/core/generic.pyi
+++ b/pandas-stubs/core/generic.pyi
@@ -27,6 +27,7 @@ from pandas import Index
 from pandas.core.resample import DatetimeIndexResampler
 from pandas.core.series import Series
 from sqlalchemy.engine import Connectable
+from sqlalchemy.types import TypeEngine
 
 from pandas._libs.lib import NoDefault
 from pandas._typing import (
@@ -175,7 +176,9 @@ class NDFrame:
         index: _bool = True,
         index_label: IndexLabel = None,
         chunksize: int | None = None,
-        dtype: DtypeArg | None = None,
+        dtype: (
+            DtypeArg | Mapping[Hashable, type[TypeEngine[Any]] | TypeEngine[Any]] | None
+        ) = None,
         method: (
             Literal["multi"]
             | Callable[
--- a/tests/test_io.py
+++ b/tests/test_io.py
@@ -1246,6 +1246,28 @@ def test_read_sql_via_sqlalchemy_engine_with_params(tmp_path: Path) -> None:
     engine.dispose()
 
 
+def test_to_sql_dtype_sqlalchemy_type(tmp_path: Path) -> None:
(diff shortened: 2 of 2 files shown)


pandas-stubs/core/generic.pyi
```diff
@@ -175,7 +176,9 @@ class NDFrame:
         index: _bool = True,
         index_label: IndexLabel = None,
         chunksize: int | None = None,
-        dtype: DtypeArg | None = None,
+        dtype: (
+            DtypeArg | Mapping[Hashable, type[TypeEngine[Any]] | TypeEngine[Any]] | None
```
loicdiridollou: I think you may need `TypeEngineMixin` instead.
Please try running tests locally with `poetry run poe test_all` for the full CI cycle or `poetry run poe mypy` for just `mypy`.

tests/test_io.py
```diff
@@ -1246,6 +1246,28 @@ def test_read_sql_via_sqlalchemy_engine_with_params(tmp_path: Path) -> None:
...
+    db_uri = "sqlite:///" + path_str
+    engine = sqlalchemy.create_engine(db_uri)
+
+    check(
+        assert_type(
+            DF.to_sql(
+                "test",
+                con=engine,
+                dtype={
+                    "a": sqlalchemy.types.VARCHAR(16),
```
loicdiridollou, line 1260: You added the case for `dict[str, TypeEngineMixin]`, to be complete you should also add for `TypeEngineMixin` so you test all the cases possible.

loicdiridollou (changes requested):
See the CI failures, please run `poetry run poe test_all` for checking the CI locally.
Added some comments in the code to guide you.

pandas-stubs/core/generic.pyi
```diff
@@ -175,7 +176,13 @@ class NDFrame:
         index: _bool = True,
         index_label: IndexLabel = None,
         chunksize: int | None = None,
-        dtype: DtypeArg | None = None,
+        dtype: (
+            Dtype
+            | type[TypeEngineMixin]
```
loicdiridollou, line 181: I am not too familiar with the use of `INTEGER` vs `INTEGER()` because ideally we would avoid adding `type[TypeEngineMixin]` so that one does not try to pass just `VARCHAR` alone.
Let me know if you have insights

loicdiridollou (commented):
One minor question, the rest looks good.

pandas-stubs/core/generic.pyi
```diff
@@ -175,7 +176,13 @@ class NDFrame:
...
+            | TypeEngineMixin
+            | Mapping[Hashable, Dtype | type[TypeEngineMixin] | TypeEngineMixin]
```
cmp0xff, line 183: I guess the key of the `Mapping` can only be `str` for SQL.

loicdiridollou (approved):
Thanks for the changes @SAY-5

PR #1765, opened 2026-06-10
Remove `pyrefly: ignore-errors` in `test_io.py`

MarcoGorelli:
In most of these cases, they're in `TYPE_CHECKING_INVALID_USAGE`, so the same ignores for mypy/pyright need applying

There's one case where, because `JsonReader` inherits from `Iterator[Any]`, pyrefly was reporting `__iter__` return type as `Any`. `ty` also flags an error there - the fix is just to make it `Iterator[NDFrameT]` 


Dr-Irv (approved):
thanks @MarcoGorelli

PR #1766, opened 2026-06-10
remove unnecessary pyrefly: ignore-errors in tests

MarcoGorelli:
These are all cases where the only type-checking errors are in TYPE_CHECKING_INVALID_USAGE, so the same (or similar) ignores that are already there for mypy/pyright need applying


Dr-Irv (approved):
thanks @MarcoGorelli

Issue #1767, opened 2026-06-10
tracker: resolving all the `pyrefly: ignore-errors`


MarcoGorelli:
There are currently (10th June 2026) 63 `pyrefly: ignore-errors` in the test suite

This is a tracker issue dedicated to resolving them

## PRs
- https://github.com/pandas-dev/pandas-stubs/pull/1756
- https://github.com/pandas-dev/pandas-stubs/pull/1765
- https://github.com/pandas-dev/pandas-stubs/pull/1766
- https://github.com/pandas-dev/pandas-stubs/pull/1768
- https://github.com/pandas-dev/pandas-stubs/pull/1769
- https://github.com/pandas-dev/pandas-stubs/pull/1770

## upstream issues

- https://github.com/facebook/pyrefly/issues/3742
- https://github.com/facebook/pyrefly/issues/3779
- https://github.com/facebook/pyrefly/issues/3819 (this is at the heart of hundreds of errors)

cmp0xff:
After #1756, #1765 and #1766, still 52 `pyrefly: ignore-errors`

MarcoGorelli:
~~16 files left 🥳 ~~

8 now!

MarcoGorelli:
for the remaining issues, i've come across https://github.com/facebook/pyrefly/issues/3891 🤷

PR #1772, opened 2026-06-12
TST: #1767 enable `pyrefly` in `tests\test_natype.py` and `tests\arrays\test_floating.py`

cmp0xff:
Not sure if we want to report the potential issues with `divmod` and in `tests\arrays\test_floating.py`.

### DIFF
--- a/tests/test_natype.py
+++ b/tests/test_natype.py
@@ -84,6 +84,7 @@ def test_arithmetic() -> None:
 
     # __divmod__
     check(assert_type(divmod(na, s_int), tuple[pd.Series, pd.Series]), tuple, pd.Series)
+    # TODO: facebook/pyrefly#3822
     check(
         assert_type(  # pyrefly: ignore[assert-type]
             divmod(na, idx_int), tuple[pd.Index, pd.Index]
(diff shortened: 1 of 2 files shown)


tests/test_natype.py
```diff
@@ -84,25 +83,17 @@ def test_arithmetic() -> None:
...
-            tuple[pd.Series, pd.Series],
-        ),
-        tuple,
-        pd.Series,
-    )
-    check(
-        assert_type(
-            divmod(na, idx_int),
-            tuple[pd.Index, pd.Index],
+        assert_type(  # pyrefly: ignore[assert-type]
```
Dr-Irv, line 89: add a comment here as to why you have a `#pyrefly: ignore`

Dr-Irv (changes requested):

tests/test_natype.py
cmp0xff: 4a54495b7a96f832a5bfa533e368d6ce85dc7a67

tests/test_natype.py
```diff
@@ -84,13 +84,15 @@ def test_arithmetic() -> None:
 
     # __divmod__
     check(assert_type(divmod(na, s_int), tuple[pd.Series, pd.Series]), tuple, pd.Series)
+    # TODO: facebook/pyrefly#3822
```
Dr-Irv, line 87: FYI on that report, you reversed the names of the PYI and PY files.

Dr-Irv (approved):

Dr-Irv:
Thanks @cmp0xff

tests/test_natype.py
cmp0xff: Thank you, they may have bad naming but the names are as intended.

PR #1779, opened 2026-06-19
TYP: #1159 invalidate `Series[int]().str` methods

cmp0xff:
- [x] Closes #1159 (Replace xxxx with the Github issue number)
- [x] Tests added (Please use `assert_type()` to assert the type of any return value)
- [ ] If I used AI to develop this pull request, I prompted it to follow `AGENTS.md`.

It will take time to add tests.

PR #1783, opened 2026-06-22
TYP: refactor string methods

cmp0xff:
The class of `StringMethods` in `pandas-stubs`-`main` contains quite a few type variables. When we want to fix some of them in `self`, as in pandas-dev/pandas-stubs#1779, they break into one `TypeVar` per line, and the code becomes much longer.

Meanwhile, these type variables serve just to distinguish `Index` and `Series`, the only reason we have so many of them is that type variables cannot be generic.

I propose the current PR to reduce the number of `TypeVar`s to one, by creating `IndexStringMethods` and `SeriesStringMethods` separately. The number of code lines also reduced by several hundred.

Caveats:
1. ~~It seems some irrelevant `mypy` checks are failing~~ edit: caused by incompatible versions of `numpy` and `numpy-typing-compact`
2. `StringMethods`, which is a class in `pandas` and the runtime type of `idx.str` and `sr.str`, now does not exist in the stubs any more. This might be acceptable, because it is [not quite a public class by documentation](https://pandas.pydata.org/pandas-docs/stable/search.html?q=StringMethods); there is no stand-alone page for it.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -21,7 +21,7 @@ classifiers = [
   "Topic :: Scientific/Engineering",
   "Typing :: Stubs Only",
 ]
-dependencies = ["numpy>=1.23.5"]
+dependencies = ["numpy>=1.23.5,<2.5"]
 
 [project.urls]
 Homepage = "https://pandas.pydata.org"
--- a/pandas-stubs/core/indexes/base.pyi
+++ b/pandas-stubs/core/indexes/base.pyi
@@ -60,11 +60,10 @@ from pandas.core.indexes.category import CategoricalIndex
 from pandas.core.indexes.datetimes import DatetimeIndex
 from pandas.core.indexes.frozen import FrozenList
 from pandas.core.indexes.interval import IntervalIndex
-from pandas.core.indexes.multi import MultiIndex
 from pandas.core.indexes.period import PeriodIndex
 from pandas.core.indexes.timedeltas import TimedeltaIndex
 from pandas.core.series import Series
-from pandas.core.strings.accessor import StringMethods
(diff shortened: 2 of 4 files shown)


pyproject.toml
```diff
@@ -21,7 +21,7 @@ classifiers = [
   "Topic :: Scientific/Engineering",
   "Typing :: Stubs Only",
 ]
-dependencies = ["numpy>=1.23.5"]
+dependencies = ["numpy>=1.23.5,<2.5"]
```
loicdiridollou, line 24: Do we have a TODO on this one? Or at least an issue for it

loicdiridollou (commented):
Minor question, rest looks good.

pyproject.toml
cmp0xff: It is discussed in #1785 and tracked in #1786

loicdiridollou (approved):
Thanks @cmp0xff

cmp0xff:
Thanks, we'll not be supporting checking `type(series.str) is StringMethods` any more. Hopefully this will not become an issue.

Dr-Irv:
> Thanks, we'll not be supporting checking `type(series.str) is StringMethods` any more. Hopefully this will not become an issue.

This is a nice change.  The complexity of the string accessors had grown over the years.

Would it be possible to have an empty `StringMethods` class that is a superclass of `IndexStringMethods` and `SeriesStringMethods` so then at least `isinstance(Series.str, StringMethods)` and `isinstance(Index.str, StringMethods)` would be accepted?

Having said that, we don't document `StringMethods`, so maybe it's not necessary.

PR #1785, opened 2026-06-22
BUG: update `numpy-typing-compat` and cap `numpy`

cmp0xff:
Pipelines like https://github.com/pandas-dev/pandas-stubs/actions/runs/27935121522/job/82654970732 and https://github.com/pandas-dev/pandas-stubs/actions/runs/27947029379/job/82694274269?pr=1784 because https://github.com/numpy/numpy/releases/tag/v2.5.0 has just been released.
- https://github.com/jorenham/numpy-typing-compat has changed its version format, we need to update
- new `numpy` introduces breaking changes https://github.com/pandas-dev/pandas-stubs/actions/runs/27948511803/job/82699244836

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -69,7 +69,7 @@ dev = [
   "pytz>=2025.2",
   "types-pytz>=2022.1.1",
   "pyiceberg>=0.10.0",
-  "numpy-typing-compat>=20260602.2.4",
+  "numpy-typing-compat>=20260602",
 ]
 
 [build-system]
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -21,7 +21,7 @@ classifiers = [
   "Topic :: Scientific/Engineering",
   "Typing :: Stubs Only",
 ]
-dependencies = ["numpy>=1.23.5"]
+dependencies = ["numpy>=1.23.5,<2.5"]
 
 [project.urls]
 Homepage = "https://pandas.pydata.org"
@@ -69,6 +69,7 @@ dev = [
   "pytz>=2025.2",
   "types-pytz>=2022.1.1",
   "pyiceberg>=0.10.0",
+  "numpy-typing-compat>=20260602.2.4",
 ]
 
 [build-system]
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -21,7 +21,7 @@ classifiers = [
   "Topic :: Scientific/Engineering",
   "Typing :: Stubs Only",
 ]
-dependencies = ["numpy>=1.23.5"]
+dependencies = ["numpy>=1.23.5,<2.5"]  # TODO: pandas-dev/pandas-stubs#1786 remove <2.5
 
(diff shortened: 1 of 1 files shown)


cmp0xff:
https://github.com/pandas-dev/pandas-stubs/actions/runs/27949330176/job/82701989558?pr=1785 always fails upon the first run

pyproject.toml
```diff
@@ -21,7 +21,7 @@ classifiers = [
   "Topic :: Scientific/Engineering",
   "Typing :: Stubs Only",
 ]
-dependencies = ["numpy>=1.23.5"]
+dependencies = ["numpy>=1.23.5,<2.5"]
```
MarcoGorelli: i'm not sure about adding an upper-bound to a runtime dependency

is it possible to pin it just for internal usage so CI becomes green again?

cmp0xff: 39746b41672c9ce7b0518785207a8e27884b7c20

cmp0xff: `numpy` 2.5 behaves very differently. I'd cap `numpy` version for now and investigate separately in #1786.

MarcoGorelli:
i'd still push back on capping numpy in the required dependencies tbh, none of the failures in https://github.com/pandas-dev/pandas-stubs/actions/runs/27948511803/job/82699244836 looks serious enough to warrant that. i think this risks be ing disruptive to users

is it possible to only pin it here in CI?

cmp0xff:
> i'd still push back on capping numpy in the required dependencies tbh, none of the failures in https://github.com/pandas-dev/pandas-stubs/actions/runs/27948511803/job/82699244836 looks serious enough to warrant that. i think this risks be ing disruptive to users
> 
> is it possible to only pin it here in CI?

I understand the concern, and I tried to cap the `numpy` dependency in `dev` only. The problem is, we have a typing test in CI with `panda-stubs` _installed_, and there `numpy` version is updated to 2.5, see https://github.com/pandas-dev/pandas-stubs/actions/runs/27953122507/job/82714675960 from 39746b41672c9ce7b0518785207a8e27884b7c20.

My idea for now is that we do a temporary fix to unblock other PRs, and do not release a new version of `pandas-stubs` until we remove the `numpy` version cap in `dependencies`. What do you think?

MarcoGorelli (approved):
if there's agreement that the pin will be removed before a release, then sure, no objections 👍

cmp0xff:
Hi @Dr-Irv , would you be okay with the change and the release freeze?

cmp0xff:
I'll merge for now to unblock other PRs. We can revert or fix the numpy issues.

Dr-Irv:
> Hi @Dr-Irv , would you be okay with the change and the release freeze?

Yes.

One note - in reading the pypi docs on `numpy-typing-compat`, the author (part of the numpy team) points to `optype`, which we might consider using.  See https://jorenham.github.io/optype/

PR #1790, opened 2026-06-25
Support passing DataFrames to df.clip()

### LINKED ISSUE #1789: DataFrame.clip() should support DataFrame types for upper/lower
Currently mypy complains about `df.clip(upper=df_upper, lower=df_lower)`, but documentation says it's supported. I believe this use case has worked fine since this commit: https://github.com/pandas-dev/pandas/commit/1528859f44608fc1c6d02bc5fa51272e9e3cf69a and has a test covering it: https://github.com/pandas-dev/pandas/blob/6fba097a91566e1e1ee7308f0df1291d9b099669/pandas/tests/frame/methods/test_clip.py#L113

geoffrey-eisenbarth:
- [x] Closes #1789

Extended existing tests and added an overload.

Glad to make any changes if needed!

### DIFF
--- a/pandas-stubs/core/frame.pyi
+++ b/pandas-stubs/core/frame.pyi
@@ -1982,30 +1982,10 @@ class DataFrame(NDFrame, OpsMixin, _GetItemHack):
     @overload
     def clip(
         self,
-        lower: DataFrame | None = ...,
-        upper: DataFrame | None = ...,
+        lower: DataFrame | None = None,
+        upper: DataFrame | None = None,
         *,
-        axis: Axis | None = ...,
-        inplace: bool = False,
-        **kwargs: Any,
-    ) -> Self: ...
-    @overload
-    def clip(
-        self,
-        lower: DataFrame = ...,
-        upper: DataFrame | None = ...,
-        *,
-        axis: Axis | None = ...,
-        inplace: bool = False,
-        **kwargs: Any,
-    ) -> Self: ...
-    @overload
-    def clip(
-        self,
-        lower: DataFrame | None = ...,
-        upper: DataFrame = ...,
-        *,
-        axis: Axis | None = ...,
+        axis: None = None,
         inplace: bool = False,
         **kwargs: Any,
     ) -> Self: ...
--- a/pandas-stubs/core/frame.pyi
+++ b/pandas-stubs/core/frame.pyi
@@ -1989,6 +1989,26 @@ class DataFrame(NDFrame, OpsMixin, _GetItemHack):
         inplace: bool = False,
         **kwargs: Any,
     ) -> Self: ...
+    @overload
+    def clip(
+        self,
+        lower: DataFrame = ...,
+        upper: DataFrame | None = ...,
+        *,
+        axis: Axis | None = ...,
+        inplace: bool = False,
+        **kwargs: Any,
+    ) -> Self: ...
+    @overload
+    def clip(
+        self,
+        lower: DataFrame | None = ...,
+        upper: DataFrame = ...,
+        *,
+        axis: Axis | None = ...,
+        inplace: bool = False,
+        **kwargs: Any,
+    ) -> Self: ...
     @final
     def copy(self, deep: _bool = True) -> Self: ...
     def cummax(
--- a/pandas-stubs/core/frame.pyi
+++ b/pandas-stubs/core/frame.pyi
@@ -1965,7 +1965,7 @@ class DataFrame(NDFrame, OpsMixin, _GetItemHack):
         lower: AnyArrayLike = ...,
         upper: AnyArrayLike | None = ...,
         *,
-        axis: Axis = ...,
+        axis: Axis,
         inplace: bool = False,
         **kwargs: Any,
     ) -> Self: ...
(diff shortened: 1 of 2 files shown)


pandas-stubs/core/frame.pyi
```diff
@@ -1979,6 +1979,16 @@ class DataFrame(NDFrame, OpsMixin, _GetItemHack):
         inplace: bool = False,
         **kwargs: Any,
     ) -> Self: ...
+    @overload
+    def clip(
+        self,
+        lower: DataFrame | None = ...,
+        upper: DataFrame | None = ...,
```
Dr-Irv: Just like in the above 2 overloads for `AnyArrayLike`, you will need one overload where the `lower` is `DataFrame` and `upper` is `DataFrame | None`, and vice-versa.

tests/frame/test_frame.py
```diff
@@ -903,6 +903,10 @@ def test_dataframe_clip() -> None:
         assert_type(df.clip(lower=None, upper=15, axis=0, inplace=True), pd.DataFrame),
         pd.DataFrame,
     )
+    check(
+        assert_type(df.clip(lower=df, upper=df, inplace=True), pd.DataFrame),
+        pd.DataFrame,
+    )
```
Dr-Irv, line 914: can you also add tests without `inplace=True` above where the `inplace=True` tests start (i.e., above line 892).

So you add these 3 similar tests, but without `inplace=True`

It might also be good to create a test that uses different `DataFrame` instances for `lower` and `upper` instead of the `df` itself.

Dr-Irv (changes requested):

tests/frame/test_frame.py
geoffrey-eisenbarth: I've added the tests without `inplace=True`, thanks!

re: using a different `DataFrame`: are you suggesting that we validate the output values of the `clip()` operation? I figured that since we're only testing types here, it would be sufficient to just re-use `df` instead of creating a `df2` inside of `test_dataframe_clip()`. I'm down for whatever you suggest, however; just wanted clarification.

pandas-stubs/core/frame.pyi
geoffrey-eisenbarth: Based on the `mypy` run, I think those overloads might be redundant since `axis: Axis | None`. I suppose there's a question of whether this should just be `axis: None` when `upper` or `lower` are `DataFrame`, though?

cmp0xff: @Dr-Irv , to be honest, I don't see why we need one overload with `lower: AnyArrayLike = ...`, `upper: AnyArrayLike | None = ...` and another overload with reversed annotations. I think it should be sufficient to have one overload `lower: AnyArrayLike | None = None`, `upper: AnyArrayLike | None = None` (note the default value is actually `None`, which we should add). See https://github.com/pandas-dev/pandas/blob/v3.0.3/pandas/core/generic.py#L8605-L8783.

@geoffrey-eisenbarth just in case you did not know: you don't need to rely on the CI to run `mypy` and other type checkers. You could use `poe mypy` and eventually `poe test_all`.

geoffrey-eisenbarth: @cmp0xff thanks for the heads up :) I just forgot to run them locally before the second commit.

Once we hear back from Dr-Irv and decide if a `df2` is necessary (see following comment), I can wrap this up pretty quickly.

Dr-Irv: @cmp0xff you're right.  @loicdiridollou worked on `DataFrame.clip()` last, so I'd like his opinion on this as he wrote the overloads.

tests/frame/test_frame.py
Dr-Irv: My concern is that when we run `pytest`, using the same `DataFrame` for doing the `clip` (especially in the `inplace=True` case) might give some funky results.  So we'd be better off passing data that corresponds to likely pandas usage.

You could just use code similar to what is in the pandas test.

pandas-stubs/core/frame.pyi
loicdiridollou: So the reason we have the overloads is that if you pass `df.clip(None, None, inplace=True)` the return type is `Self` but if you pass anything else in `lower` or `upper` then the type is None. That worked for 2.3.3 but not in 3.0, now it only returns `Self`.

loicdiridollou: There is also an issue if you pass an array-like without specifying the axis that is why the overloads are there (not letting you pass an array without axis so it knows which orientation is correct).
So to be fully correct we should remove the `axis: Axis = ...,` and replace it with `axis: Axis,` but that could be a separate one.

geoffrey-eisenbarth: @loicdiridollou @Dr-Irv If I understand all this correctly:

1. `axis: Axis` should be required when using one of the `AnyArrayLike` overloads,
2. but not when both `upper` and `lower` are None, so

we need two overloads for `AnyArrayLike`, both of which force `axis: Axis` to be required.

I made that change and all prior tests are passing. If I should do something else, please let me know!

Dr-Irv: > @loicdiridollou @Dr-Irv If I understand all this correctly:
> 
> 1. `axis: Axis` should be required when using one of the `AnyArrayLike` overloads,
> 2. but not when both `upper` and `lower` are None, so
> 
> we need two overloads for `AnyArrayLike`, both of which force `axis: Axis` to be required.
> 
> I made that change and all prior tests are passing. If I should do something else, please let me know!

I'm going to defer to @loicdiridollou on this PR.  He's more familiar with that section of the code and why we have the multiple overloads.

loicdiridollou (approved):
Thanks @geoffrey-eisenbarth, nice one!

Issue #1799, opened 2026-06-29
Resolve all `tool.ty.src.exclude`


cmp0xff:
Following #1767, we can also run `ty` on `tests`.

We can just tag this issue in the PR, using the hashtag properly.

Upstream issues:
- [x] astral-sh/ty#3898 -> astral-sh/ty#3323
- [x] astral-sh/ty#3951
- [x] astral-sh/ty#3956
- [x] ~~astral-sh/ty#4005~~ (won't do)
- [x] astral-sh/ty#4022
- [ ] astral-sh/ty#4052 -> astral-sh/ty#2681
- [x] astral-sh/ty#4053
- [ ] astral-sh/ty#4055
- [x] ~~astral-sh/ty#4102~~ -> astral-sh/ty#4111
- [ ] astral-sh/ty#4135
- [ ] astral-sh/ty#4150

PR #1800, opened 2026-06-29
TST: #1799 enable `ty` in `tests/test_windowing.py`

cmp0xff:
- [x] Addresses #1799
- [x] Enable `ty` in `tests/test_windowing.py`
- [x] Add `ty_dist` and `pyrefly_dist`
- [x] Rename `ty_src` and `pyrefly_src`
- [x] Change the order of type checkers so that the fastest `ty` runs first
- [x] Update descriptions

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -39,17 +39,17 @@ jobs:
           os: ${{ matrix.os }}
           python-version: ${{ matrix.python-version }}
 
-      - name: Run mypy on 'tests' (using the local stubs) and on the local stubs
-        run: poetry run poe mypy
-
-      - name: Run ty on 'pandas-stubs' (using the local stubs) and on the local stubs
+      - name: Run ty on 'tests' (using the local stubs) and on the local stubs
         run: poetry run poe ty
 
-      - name: Run pyrefly on the local stubs
+      - name: Run pyrefly on 'tests' (using the local stubs) and on the local stubs
         run: poetry run poe pyrefly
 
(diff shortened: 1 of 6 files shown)


Dr-Irv (commented):
Only thing I'm not sure of is why you changed the order of the tests, i.e., it used to be `mypy`, then `ty`, then `pyrefly`, then `pyright`.  

I actually think it should be `mypy`, 'pyright`, 'pyrefly'`, `ty`, because that probably represents the popularity of the individual type checkers, and we want to catch errors in that order.

But I'm not fixated on any particular order.

cmp0xff:
My argument are the followings:
- originally the order was `mypy`, `ty`, `pyrefly` and `pyright`, which did not make much sense.
- the new order is that the faster type checker runs earlier, so we see the failure faster.
- in the end, `panda-stubs` will support all type checkers, and we would report all incompatibilities, so the popularity does not work well as a sorting key.

Dr-Irv:
> My argument are the followings:
> 
> * originally the order was `mypy`, `ty`, `pyrefly` and `pyright`, which did not make much sense.
> * the new order is that the faster type checker runs earlier, so we see the failure faster.
> * in the end, `panda-stubs` will support all type checkers, and we would report all incompatibilities, so the popularity does not work well as a sorting key.

OK, I'm fine with that.  Only issue is that the failures in the newer type checkers have always been more likely, but I guess they are getting more mature.

Dr-Irv (approved):
thanks @cmp0xff

PR #1802, opened 2026-06-29
BUG: #1786 resume supporting `numpy` 2.5

### LINKED ISSUE #1786: `mypy` raises errors after `numpy` 2.5.0
https://github.com/pandas-dev/pandas-stubs/actions/runs/27948511803/job/82699244836#step:4:18
```text
tests/scalars/test_scalars.py:571: error: Unused "type: ignore" comment  [unused-ignore]
tests/scalars/test_scalars.py:619: error: Unused "type: ignore" comment  [unused-ignore]
tests/scalars/test_scalars.py:907: error: Expression is of type "Any", not "ndarray[tuple[Any, ...], dtype[numpy.bool[builtins.bool]]]"  [assert-type]
tests/scalars/test_scalars.py:908: error: Expression is of type "Any", not "ndarray[tuple[Any, ...], dtype[numpy.bool[builtins.bool]]]"  [assert-type]
tests/scalars/test_scalars.py:910: error: Expression is of type "Any", not "ndarray[tuple[int, int], dtype[numpy.bool[builtins.bool]]]"  [assert-type]
tests/scalars/test_scalars.py:911: error: Expression is of type "Any", not "ndarray[tuple[int, int], dtype[numpy.bool[builtins.bool]]]"  [assert-type]
tests/scalars/test_scalars.py:913: error: Expression is of type "Any", not "ndarray[tuple[int], dtype[numpy.bool[builtins.bool]]]"  [assert-type]
tests/scalars/test_scalars.py:914: error: Expression is of type "Any", not "ndarray[tuple[int], dtype[numpy.bool[builtins.bool]]]"  [assert-type]
tests/scalars/test_scalars.py:1317: error: Expression is of type "Any", not "ndarray[tuple[Any, ...], dtype[numpy.bool[builtins.bool]]]"  [assert-type]
tests/scalars/test_scalars.py:1318: error: Expression is of type "Any", not "ndarray[tuple[Any, ...], dtype[numpy.bool[builtins.bool]]]"  [assert-type]
tests/scalars/test_scalars.py:1320: error: Expression is of type "Any", not "ndarray[tuple[int, int], dtype[numpy.bool[builtins.bool]]]"  [assert-type]
tests/scalars/test_scalars.py:1321: error: Expression is of type "Any", not "ndarray[tuple[int, int], dtype[numpy.bool[builtins.bool]]]"  [assert-type]
tests/scalars/test_scalars.py:1323: error: Expression is of type "Any", not "ndarray[tuple[int], dtype[numpy.bool[builtins.bool]]]"  [assert-type]
tests/scalars/test_scalars.py:1324: error: Expression is of type "Any", not "ndarray[tuple[int], dtype[numpy.bool[builtins.bool]]]"  [assert-type]
tests/arrays/test_datetime_array.py:81: error: Expression is of type "Any", not "DatetimeArray"  [assert-type]
tests/arrays/test_datetime_array.py:92: error: Expression is of type "Any", not "DatetimeArray"  [assert-type]
tests/arrays/test_datetime_array.py:107: error: Expression is of type "Any", not "DatetimeArray"  [assert-type]
tests/arrays/test_datetime_array.py:117: error: Expression is of type "Any", not "DatetimeArray"  [assert-type]
tests/arrays/test_datetime_array.py:123: error: Expression is of type "Any", not "DatetimeArray"  [assert-type]
tests/arrays/test_datetime_array.py:124: error: Expression is of type "Any", not "DatetimeArray"  [assert-type]
tests/arrays/test_datetime_array.py:313: error: Expression is of type "Any", not "DatetimeArray"  [assert-type]
tests/series/test_series.py:2644: error: Expression is of type "ndarray[tuple[int], dtype[datetime64[int]]]", not "ndarray[tuple[int], dtype[datetime64[Any]]]"  [assert-type]
tests/series/test_series.py:2650: error: Expression is of type "ndarray[tuple[int], dtype[timedelta64[int]]]", not "ndarray[tuple[int], dtype[timedelta64[Any]]]"  [assert-type]
tests/series/test_series.py:2657: error: Expression is of type "ndarray[tuple[int], dtype[timedelta64[int]]]", not "ndarray[tuple[int], dtype[timedelta64[Any]]]"  [assert-type]
Found 24 errors in 3 files (checked 345 source files)
```

cmp0xff:
- [x] Partially addresses #1786

### DIFF
--- a/tests/arrays/test_datetime_array.py
+++ b/tests/arrays/test_datetime_array.py
@@ -104,6 +104,12 @@ def test_construction_sequence_pandas(
                 ),
                 DatetimeArray,
             )
+            assert_type(  # type: ignore[assert-type]
+                pd.array(
+                    [np.datetime64("2131-01-05 01:25"), np.datetime64("1748-01-06")]
+                ),
+                DatetimeArray,
+            )
         else:
             assert_type(
                 pd.array(
@@ -111,11 +117,12 @@ def test_construction_sequence_pandas(
                 ),
                 DatetimeArray,
             )
-        # TODO: pandas-dev/pandas-stubs#1786
-        assert_type(  # type: ignore[assert-type]
-            pd.array([np.datetime64("2131-01-05 01:25"), np.datetime64("1748-01-06")]),
-            DatetimeArray,
-        )
+            assert_type(
+                pd.array(
+                    [np.datetime64("2131-01-05 01:25"), np.datetime64("1748-01-06")]
+                ),
+                DatetimeArray,
+            )
         assert_type(
             pd.array([np.datetime64("2130-01-01 01:25"), datetime(1749, 1, 6)]),
             DatetimeArray,
--- a/tests/arrays/test_datetime_array.py
+++ b/tests/arrays/test_datetime_array.py
@@ -79,11 +79,14 @@ def test_construction_sequence_pandas(
             pd.array([pd.Timestamp(2026, 1, 5, tzinfo=ZoneInfo("Africa/Ouagadougou"))]),
             DatetimeArray,
         )
-        # TODO: pandas-dev/pandas-stubs#1786
         if sys.version_info >= (3, 14):
+            # TODO: pandas-dev/pandas-stubs#1786
             assert_type(  # type: ignore[assert-type]
                 pd.array([np.datetime64("2026-01-05 23:27:59")]), DatetimeArray
             )
+        elif sys.version_info >= (3, 12):
+            # TODO: pandas-dev/pandas-stubs#1786 inconsistecy between the stubs repo and the installed stubs
+            pass
         else:
             assert_type(pd.array([np.datetime64("2026-01-05 23:27:59")]), DatetimeArray)
 
@@ -96,8 +99,8 @@ def test_construction_sequence_pandas(
         assert_type(
             pd.array([datetime(2052, 1, 5), pd.Timestamp(2, 1, 6)]), DatetimeArray
         )
-        # TODO: pandas-dev/pandas-stubs#1786
         if sys.version_info >= (3, 14):
+            # TODO: pandas-dev/pandas-stubs#1786
             assert_type(  # type: ignore[assert-type]
                 pd.array(
                     [np.datetime64("2026-01-05 23:27:59"), np.datetime64("1748-01-06")]
@@ -110,6 +113,9 @@ def test_construction_sequence_pandas(
                 ),
                 DatetimeArray,
             )
+        elif sys.version_info >= (3, 12):
+            # TODO: pandas-dev/pandas-stubs#1786 inconsistecy between the stubs repo and the installed stubs
+            pass
         else:
             assert_type(
                 pd.array(
@@ -134,11 +140,14 @@ def test_construction_sequence_pandas(
 
         assert_type(pd.array([datetime(2061, 1, 5, 1), None]), DatetimeArray)
         assert_type(pd.array([pd.Timestamp(1902, 1, 5, 3), None]), DatetimeArray)
-        # TODO: pandas-dev/pandas-stubs#1786
         if sys.version_info >= (3, 14):
+            # TODO: pandas-dev/pandas-stubs#1786
             assert_type(  # type: ignore[assert-type]
                 pd.array([np.datetime64("2111-01-05"), None]), DatetimeArray
             )
+        elif sys.version_info >= (3, 12):
+            # TODO: pandas-dev/pandas-stubs#1786 inconsistecy between the stubs repo and the installed stubs
+            pass
         else:
             assert_type(pd.array([np.datetime64("2111-01-05"), None]), DatetimeArray)
 
@@ -150,11 +159,14 @@ def test_construction_sequence_pandas(
         assert_type(
             pd.array([pd.Timestamp(2102, 1, 5, 3), None, pd.NaT]), DatetimeArray
         )
-        # TODO: pandas-dev/pandas-stubs#1786
         if sys.version_info >= (3, 14):
+            # TODO: pandas-dev/pandas-stubs#1786
             assert_type(  # type: ignore[assert-type]
                 pd.array([np.datetime64("2114-01-05"), None, pd.NaT]), DatetimeArray
             )
+        elif sys.version_info >= (3, 12):
+            # TODO: pandas-dev/pandas-stubs#1786 inconsistecy between the stubs repo and the installed stubs
+            pass
         else:
             assert_type(
                 pd.array([np.datetime64("2114-01-05"), None, pd.NaT]), DatetimeArray
@@ -162,10 +174,13 @@ def test_construction_sequence_pandas(
 
         assert_type(pd.array((datetime(2026, 1, 5),)), DatetimeArray)
         assert_type(pd.array(UserList([pd.Timestamp(2026, 1, 5)])), DatetimeArray)
-        # TODO: pandas-dev/pandas-stubs#1786
         if sys.version_info >= (3, 14):
+            # TODO: pandas-dev/pandas-stubs#1786
             assert_type(pd.array((np.datetime64("1959-01-05"),)), DatetimeArray)  # type: ignore[assert-type]
             assert_type(pd.array(UserList([np.datetime64("1701-01-05")])), DatetimeArray)  # type: ignore[assert-type]
+        elif sys.version_info >= (3, 12):
+            # TODO: pandas-dev/pandas-stubs#1786 inconsistecy between the stubs repo and the installed stubs
+            pass
         else:
             assert_type(pd.array((np.datetime64("1959-01-05"),)), DatetimeArray)
             assert_type(
@@ -359,9 +374,12 @@ def test_constructor() -> None:
     check(assert_type(pd.array([np_dt]), DatetimeArray), DatetimeArray)
     check(assert_type(pd.array([np_dt, None]), DatetimeArray), DatetimeArray)
     dt_nat = cast(list[np.datetime64 | NaTType], [np_dt, pd.NaT])
-    # TODO: pandas-dev/pandas-stubs#1786
     if sys.version_info >= (3, 14):
+        # TODO: pandas-dev/pandas-stubs#1786
         check(assert_type(pd.array(dt_nat), DatetimeArray), DatetimeArray)  # type: ignore[assert-type]
+    elif sys.version_info >= (3, 12):
+        # TODO: pandas-dev/pandas-stubs#1786 inconsistecy between the stubs repo and the installed stubs
+        pass
     else:
         check(assert_type(pd.array(dt_nat), DatetimeArray), DatetimeArray)
 
--- a/tests/arrays/test_datetime_array.py
+++ b/tests/arrays/test_datetime_array.py
@@ -8,6 +8,7 @@ from datetime import (
     datetime,
     time,
 )
+import sys
 from typing import (
     TYPE_CHECKING,
     Any,
@@ -78,7 +79,13 @@ def test_construction_sequence_pandas(
             pd.array([pd.Timestamp(2026, 1, 5, tzinfo=ZoneInfo("Africa/Ouagadougou"))]),
             DatetimeArray,
         )
-        assert_type(pd.array([np.datetime64("2026-01-05 23:27:59")]), DatetimeArray)
+        # TODO: pandas-dev/pandas-stubs#1786
+        if sys.version_info >= (3, 12):
+            assert_type(  # type: ignore[assert-type]
+                pd.array([np.datetime64("2026-01-05 23:27:59")]), DatetimeArray
+            )
+        else:
+            assert_type(pd.array([np.datetime64("2026-01-05 23:27:59")]), DatetimeArray)
 
         assert_type(
             pd.array([datetime(2100, 1, 5, 1), datetime(1, 1, 6, 2)]), DatetimeArray
@@ -89,7 +96,23 @@ def test_construction_sequence_pandas(
         assert_type(
             pd.array([datetime(2052, 1, 5), pd.Timestamp(2, 1, 6)]), DatetimeArray
         )
-        assert_type(
+        # TODO: pandas-dev/pandas-stubs#1786
+        if sys.version_info >= (3, 12):
+            assert_type(  # type: ignore[assert-type]
+                pd.array(
+                    [np.datetime64("2026-01-05 23:27:59"), np.datetime64("1748-01-06")]
+                ),
+                DatetimeArray,
+            )
+        else:
+            assert_type(
+                pd.array(
+                    [np.datetime64("2026-01-05 23:27:59"), np.datetime64("1748-01-06")]
+                ),
+                DatetimeArray,
+            )
+        # TODO: pandas-dev/pandas-stubs#1786
+        assert_type(  # type: ignore[assert-type]
             pd.array([np.datetime64("2131-01-05 01:25"), np.datetime64("1748-01-06")]),
             DatetimeArray,
         )
@@ -104,7 +127,13 @@ def test_construction_sequence_pandas(
 
         assert_type(pd.array([datetime(2061, 1, 5, 1), None]), DatetimeArray)
         assert_type(pd.array([pd.Timestamp(1902, 1, 5, 3), None]), DatetimeArray)
-        assert_type(pd.array([np.datetime64("2111-01-05"), None]), DatetimeArray)
+        # TODO: pandas-dev/pandas-stubs#1786
+        if sys.version_info >= (3, 12):
+            assert_type(  # type: ignore[assert-type]
+                pd.array([np.datetime64("2111-01-05"), None]), DatetimeArray
+            )
+        else:
+            assert_type(pd.array([np.datetime64("2111-01-05"), None]), DatetimeArray)
 
         assert_type(pd.array([datetime(1921, 1, 5, 1), pd.NaT]), DatetimeArray)
         assert_type(pd.array([pd.Timestamp(1872, 1, 5, 3), pd.NaT]), DatetimeArray)
@@ -114,14 +143,27 @@ def test_construction_sequence_pandas(
         assert_type(
             pd.array([pd.Timestamp(2102, 1, 5, 3), None, pd.NaT]), DatetimeArray
         )
-        assert_type(
-            pd.array([np.datetime64("2114-01-05"), None, pd.NaT]), DatetimeArray
-        )
+        # TODO: pandas-dev/pandas-stubs#1786
+        if sys.version_info >= (3, 12):
+            assert_type(  # type: ignore[assert-type]
+                pd.array([np.datetime64("2114-01-05"), None, pd.NaT]), DatetimeArray
+            )
+        else:
+            assert_type(
+                pd.array([np.datetime64("2114-01-05"), None, pd.NaT]), DatetimeArray
+            )
 
         assert_type(pd.array((datetime(2026, 1, 5),)), DatetimeArray)
         assert_type(pd.array(UserList([pd.Timestamp(2026, 1, 5)])), DatetimeArray)
-        assert_type(pd.array((np.datetime64("1959-01-05"),)), DatetimeArray)
-        assert_type(pd.array(UserList([np.datetime64("1701-01-05")])), DatetimeArray)
+        # TODO: pandas-dev/pandas-stubs#1786
+        if sys.version_info >= (3, 12):
+            assert_type(pd.array((np.datetime64("1959-01-05"),)), DatetimeArray)  # type: ignore[assert-type]
+            assert_type(pd.array(UserList([np.datetime64("1701-01-05")])), DatetimeArray)  # type: ignore[assert-type]
+        else:
+            assert_type(pd.array((np.datetime64("1959-01-05"),)), DatetimeArray)
+            assert_type(
+                pd.array(UserList([np.datetime64("1701-01-05")])), DatetimeArray
+            )
 
 
 @pytest.mark.parametrize(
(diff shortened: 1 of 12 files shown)


tests/arrays/test_datetime_array.py
```diff
@@ -78,7 +79,16 @@ def test_construction_sequence_pandas(
...
             DatetimeArray,
         )
-        assert_type(pd.array([np.datetime64("2026-01-05 23:27:59")]), DatetimeArray)
+        if sys.version_info >= (3, 14):
+            # TODO: pandas-dev/pandas-stubs#1786
+            assert_type(  # type: ignore[assert-type]
+                pd.array([np.datetime64("2026-01-05 23:27:59")]), DatetimeArray
+            )
+        elif sys.version_info >= (3, 12):
+            # TODO: pandas-dev/pandas-stubs#1786 inconsistecy between the stubs repo and the installed stubs
```
loicdiridollou: What is the issue here in details? For python 3.12 and 3.13 what is the resulting type?

loicdiridollou (commented):
That sounds like a lot of work and I am a bit worried of loosing track about all the changes and exceptions that were carved out here to make it compliant with numpy 2.5.
Is it something that is bound to stay with `numpy`? Or later get fixed in one way or another?

tests/arrays/test_datetime_array.py
cmp0xff: If you run `poe mypy_dist`, you get `tests\arrays\test_datetime_array.py:88: error: Expression is of type "Any", not "DatetimeArray"  [assert-type]`. This error does not appear without `_dist`.

cmp0xff: What do you think, @loicdiridollou ? Maybe @MarcoGorelli @Dr-Irv as well, since I've never met the case before where `poe mypy` passes but `poe mypy_dist` produces errors. I don't really understand. Things added here are workarounds.

Dr-Irv: I have seen this before in the early days of `pandas-stubs` development, which is why we added tests for local stubs and installed stubs.

My guess is that there is some difference in the environments created for the tests when doing the local tests vs. the `_dist` tests.

I would prefer that we don't have code like this in the tests.  So we should try to figure out what is going one.

cmp0xff: After digging into the GH workflows, I found the following issue:

https://github.com/pandas-dev/pandas-stubs/actions/runs/28395697394/job/84133759130?pr=1802#step:9:43

In words, in the `_dist` step, `numpy` is upgraded from 2.4.6 to 2.5.0, but not its compact typing stubs. Do you know how this can be fixed?

```text
Processing ./dist/pandas_stubs-3.0.3.260530-py3-none-any.whl
Collecting numpy>=1.23.5 (from pandas-stubs==3.0.3.260530)
  Downloading numpy-2.5.0-cp312-cp312-macosx_14_0_arm64.whl.metadata (6.6 kB)
Downloading numpy-2.5.0-cp312-cp312-macosx_14_0_arm64.whl (5.3 MB)
   ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ 5.3/5.3 MB 144.1 MB/s  0:00:00
Installing collected packages: numpy, pandas-stubs
  Attempting uninstall: numpy
    Found existing installation: numpy 2.4.6
    Uninstalling numpy-2.4.6:
      Successfully uninstalled numpy-2.4.6

ERROR: pip's dependency resolver does not currently take into account all the packages that are installed. This behaviour is the source of the following dependency conflicts.
numpy-typing-compat 20260602.2.4 requires numpy<2.5,>=2.4, but you have numpy 2.5.0 which is incompatible.
Successfully installed numpy-2.5.0 pandas-stubs-3.0.3.260530
```

Dr-Irv: So the issue here is that  you have a `dev` dependency on `numpy-typing-compat` but it seems that we need to include that dependency in the installed version of the stubs.  But we should only do that dependency if numpy 2.5 is installed.  I don't know how to specify that in `pyproject.toml` file, but I'm sure there is a way.

We want to say some thing in `dependencies` that means:
- `numpy`>=1.23.5
-  if `numpy` >= 2.5.0, then include `numpy-typing-compat` in the dependencies

Having said that, the question is whether we can remove the `numpy-typing-compat` dependency completely. Reading the docs about `numpy-typing-compat`, they recommend using `optype` instead.

Also, since `pandas` has a minimum dependency on `numpy` 1.26.0, we should upgrade our `numpy` dependency as well.

cmp0xff: Now I've managed to reduce conditional branches.

loicdiridollou (approved):
Overall I think this is good to go, it is a ton of splits but there does not seem to be a better setup than this so thanks @cmp0xff for doing all the heavy lifting!
@Dr-Irv please merge if you are onboard with the changes

tests/scalars/test_scalars.py
```diff
@@ -508,8 +509,12 @@ def test_timedelta_properties_methods() -> None:
     check(assert_type(td.to_pytimedelta(), dt.timedelta), dt.timedelta)
     check(assert_type(td.to_timedelta64(), np.timedelta64), np.timedelta64)
     check(assert_type(td.total_seconds(), float), float)
-    check(assert_type(td.view(np.int64), object), np.int64)
-    check(assert_type(td.view("i8"), object), np.int64)
+    # TODO: pandas-dev/pandas-stubs#1786
+    if sys.version_info >= (3, 12):
+        pass
```
Dr-Irv: can we leave in the tests and just include the ignores here?

tests/arrays/test_datetime_array.py
```diff
@@ -78,7 +79,13 @@ def test_construction_sequence_pandas(
             pd.array([pd.Timestamp(2026, 1, 5, tzinfo=ZoneInfo("Africa/Ouagadougou"))]),
             DatetimeArray,
         )
-        assert_type(pd.array([np.datetime64("2026-01-05 23:27:59")]), DatetimeArray)
+        if sys.version_info >= (3, 12):
+            # TODO: pandas-dev/pandas-stubs#1786
```
Dr-Irv: Change reference to https://github.com/python/mypy/issues/21733

tests/arrays/test_datetime_array.py
```diff
@@ -89,10 +96,33 @@ def test_construction_sequence_pandas(
...
+                    [np.datetime64("2026-01-05 23:27:59"), np.datetime64("1748-01-06")]
+                ),
+                DatetimeArray,
+            )
+            assert_type(
+                pd.array(
+                    [np.datetime64("2131-01-05 01:25"), np.datetime64("1748-01-06")]
+                ),
+                DatetimeArray,
+            )
```
Dr-Irv, line 106: these tests seem to be testing the same thing.  Why do we need both?

tests/arrays/test_datetime_array.py
```diff
@@ -89,10 +96,33 @@ def test_construction_sequence_pandas(
         assert_type(
             pd.array([datetime(2052, 1, 5), pd.Timestamp(2, 1, 6)]), DatetimeArray
         )
-        assert_type(
-            pd.array([np.datetime64("2131-01-05 01:25"), np.datetime64("1748-01-06")]),
-            DatetimeArray,
-        )
+        if sys.version_info >= (3, 12):
+            # TODO: pandas-dev/pandas-stubs#1786
```
Dr-Irv: Change reference to https://github.com/python/mypy/issues/21733

tests/arrays/test_datetime_array.py
```diff
@@ -104,7 +134,13 @@ def test_construction_sequence_pandas(
 
         assert_type(pd.array([datetime(2061, 1, 5, 1), None]), DatetimeArray)
         assert_type(pd.array([pd.Timestamp(1902, 1, 5, 3), None]), DatetimeArray)
-        assert_type(pd.array([np.datetime64("2111-01-05"), None]), DatetimeArray)
+        if sys.version_info >= (3, 12):
+            # TODO: pandas-dev/pandas-stubs#1786
```
Dr-Irv: Change reference to https://github.com/python/mypy/issues/21733

tests/arrays/test_datetime_array.py
```diff
@@ -114,14 +150,27 @@ def test_construction_sequence_pandas(
         assert_type(
             pd.array([pd.Timestamp(2102, 1, 5, 3), None, pd.NaT]), DatetimeArray
         )
-        assert_type(
-            pd.array([np.datetime64("2114-01-05"), None, pd.NaT]), DatetimeArray
-        )
+        if sys.version_info >= (3, 12):
+            # TODO: pandas-dev/pandas-stubs#1786
```
Dr-Irv: Change reference to https://github.com/python/mypy/issues/21733

tests/arrays/test_datetime_array.py
```diff
@@ -114,14 +150,27 @@ def test_construction_sequence_pandas(
...
+            assert_type(
+                pd.array([np.datetime64("2114-01-05"), None, pd.NaT]), DatetimeArray
+            )
 
         assert_type(pd.array((datetime(2026, 1, 5),)), DatetimeArray)
         assert_type(pd.array(UserList([pd.Timestamp(2026, 1, 5)])), DatetimeArray)
-        assert_type(pd.array((np.datetime64("1959-01-05"),)), DatetimeArray)
-        assert_type(pd.array(UserList([np.datetime64("1701-01-05")])), DatetimeArray)
+        if sys.version_info >= (3, 12):
+            # TODO: pandas-dev/pandas-stubs#1786
```
Dr-Irv: Change reference to https://github.com/python/mypy/issues/21733

tests/arrays/test_datetime_array.py
```diff
@@ -310,7 +359,11 @@ def test_constructor() -> None:
     check(assert_type(pd.array([np_dt]), DatetimeArray), DatetimeArray)
     check(assert_type(pd.array([np_dt, None]), DatetimeArray), DatetimeArray)
     dt_nat = cast(list[np.datetime64 | NaTType], [np_dt, pd.NaT])
-    check(assert_type(pd.array(dt_nat), DatetimeArray), DatetimeArray)
+    if sys.version_info >= (3, 12):
+        # TODO: pandas-dev/pandas-stubs#1786
```
Dr-Irv: Change reference to https://github.com/python/mypy/issues/21733

tests/arrays/test_integer.py
```diff
@@ -46,7 +47,13 @@ def test_construction_sequence(
 
     if TYPE_CHECKING:
         assert_type(pd.array([-2, 3]), IntegerArray)
-        assert_type(pd.array([1 << 32, np.int8(1) << 6]), IntegerArray)
+        # TODO: pandas-dev/pandas-stubs#1786
```
Dr-Irv: Change reference to https://github.com/python/mypy/issues/21733

tests/scalars/test_scalars.py
```diff
@@ -903,24 +914,39 @@ def test_timedelta_cmp_array() -> None:
...
-        np.bool,
-    )
-    assert (eq_nd1 != ne_nd1).all()
-    eq_2d1 = check(assert_type(td == arr_2d, np_2darray[np.bool]), np_2darray[np.bool])
-    ne_2d1 = check(assert_type(td != arr_2d, np_2darray[np.bool]), np_2darray[np.bool])
-    assert (eq_2d1 != ne_2d1).all()
-    eq_1d1 = check(assert_type(td == arr_1d, np_1darray_bool), np_1darray_bool)
-    ne_1d1 = check(assert_type(td != arr_1d, np_1darray_bool), np_1darray_bool)
-    assert (eq_1d1 != ne_1d1).all()
+    # TODO: pandas-dev/pandas-stubs#1786 facebook/pyrefly#3977
```
Dr-Irv: Change pandas-stubs reference to [python/mypy#21733](https://github.com/python/mypy/issues/21733), leave in pyrefly reference

tests/scalars/test_scalars.py
```diff
@@ -1322,24 +1348,39 @@ def test_timestamp_cmp_array() -> None:
...
-        np.bool,
-    )
-    assert (eq_nd1 != ne_nd1).all()
-    eq_2d1 = check(assert_type(ts == arr_2d, np_2darray[np.bool]), np_2darray[np.bool])
-    ne_2d1 = check(assert_type(ts != arr_2d, np_2darray[np.bool]), np_2darray[np.bool])
-    assert (eq_2d1 != ne_2d1).all()
-    eq_1d1 = check(assert_type(ts == arr_1d, np_1darray_bool), np_1darray_bool)
-    ne_1d1 = check(assert_type(ts != arr_1d, np_1darray_bool), np_1darray_bool)
-    assert (eq_1d1 != ne_1d1).all()
+    # TODO: pandas-dev/pandas-stubs#1786 facebook/pyrefly#3977
```
Dr-Irv: Change pandas-stubs reference to [python/mypy#21733](https://github.com/python/mypy/issues/21733),

tests/scalars/test_scalars.py
```diff
@@ -1722,7 +1763,10 @@ def test_period_add_subtract() -> None:
     check(assert_type(as_np_td + p, pd.Period), pd.Period)
     check(assert_type(p.__radd__(as_np_td), pd.Period), pd.Period)
 
-    check(assert_type(as_np_i64 + p, pd.Period), pd.Period)
+    # TODO: pandas-dev/pandas-stubs#1786
```
Dr-Irv: Is that the right TODO?  That points to a mypy issue, not a pyrefly one

tests/series/test_series.py
```diff
@@ -2643,26 +2644,60 @@ def test_types_to_numpy() -> None:
...
+        check(
+            assert_type(
+                s_date.to_numpy(dtype=np.dtype("datetime64[ns]")),
+                "np_1darray[np.datetime64[int]]",
+            ),
+            np_1darray,
+            np.datetime64,
+        )
+    else:
+        # TODO: pandas-dev/pandas-stubs#1786
```
Dr-Irv: Isn't this a pyrefly issue, not a mypy one?

tests/series/test_series.py
```diff
@@ -2643,26 +2644,60 @@ def test_types_to_numpy() -> None:
...
+        check(
+            assert_type(
+                s_td_small.to_numpy(dtype=np.dtype("timedelta64[ns]")),
+                "np_1darray[np.timedelta64[int]]",
+            ),
+            np_1darray,
+            np.timedelta64,
+        )
+    else:
+        # TODO: pandas-dev/pandas-stubs#1786
```
Dr-Irv: Isn't this a pyrefly issue, not a pandas one?

tests/test_dtypes.py
```diff
@@ -138,9 +139,17 @@ def test_sparse_dtype() -> None:
     check(assert_type(pd.SparseDtype(np.int64), pd.SparseDtype), pd.SparseDtype)
     check(assert_type(pd.SparseDtype(str), pd.SparseDtype), pd.SparseDtype)
     check(assert_type(pd.SparseDtype(float), pd.SparseDtype), pd.SparseDtype)
-    check(assert_type(pd.SparseDtype(np.datetime64), pd.SparseDtype), pd.SparseDtype)
-    check(assert_type(pd.SparseDtype(np.timedelta64), pd.SparseDtype), pd.SparseDtype)
-    check(assert_type(pd.SparseDtype("datetime64"), pd.SparseDtype), pd.SparseDtype)
+    # TODO: pandas-dev/pandas-stubs#1786, also disable "datetime64"
+    if sys.version_info >= (3, 12):
+        pass
```
Dr-Irv: keep the tests with the appropriate ignores

tests/test_dtypes.py
```diff
@@ -162,9 +171,13 @@ def test_sparse_dtype_fill_value_subtype_compatibility() -> None:
     check(assert_type(s_dt_bool.fill_value, Scalar | None), bool)
 
     # datetime64 subtype: default fill_value is NaT
-    s_dt_dt = pd.SparseDtype(np.datetime64)
-    check(assert_type(s_dt_dt.subtype, np.dtype), np.dtypes.DateTime64DType)
-    check(assert_type(s_dt_dt.fill_value, Scalar | None), np.datetime64)
+    # TODO: pandas-dev/pandas-stubs#1786, also disable "datetime64"
+    if sys.version_info >= (3, 12):
+        pass
```
Dr-Irv: keep the tests with the appropriate ignores

Dr-Irv (changes requested):
So I figured out most of the problems and created a `mypy` issue that explains why we have to do the tests for python 3.12 in various places.

for `pd.array`, if you comment out:
```python
@overload
def array(
    data: Sequence[Any], dtype: BuiltinObjectDtypeArg | None = None, copy: bool = True
) -> NumpyExtensionArray: ...
```
then we get the correct results for all of the `pd.array()` calls where you needed that test.

And there are other places too (e.g., `Timedelta.__eq__()`, that have the same issue.  So if they fix that in `mypy`, we won't need any of the funky tests.

I think it is worth changing the tests that say `if sys.version_info >= (3, 12):` to use a constant `NUMPY_GE_251` that is defined in `tests/__init__.py` to be ` sys.version_info >= (3, 12)`, then it is clear that what we are testing for is the newer version of numpy, not for the difference in the python version.

tests/scalars/test_scalars.py
cmp0xff: 9d9c9a42e54c48084f130038de20cbcdb57b0a10

tests/test_dtypes.py
cmp0xff: a5bb5400d1ea556780f96234c3a47f938fd74258

tests/test_dtypes.py
cmp0xff: 9d9c9a42e54c48084f130038de20cbcdb57b0a10

tests/arrays/test_integer.py
cmp0xff: cc378489c496a4f667f1a5353921d8ec9b9968dc

tests/scalars/test_scalars.py
cmp0xff: cc378489c496a4f667f1a5353921d8ec9b9968dc

tests/series/test_series.py
cmp0xff: cc378489c496a4f667f1a5353921d8ec9b9968dc

tests/series/test_series.py
cmp0xff: cc378489c496a4f667f1a5353921d8ec9b9968dc

tests/arrays/test_datetime_array.py
cmp0xff: f62f79cebac133e227d1c8234918243924b3f943

tests/arrays/test_datetime_array.py
cmp0xff: f62f79cebac133e227d1c8234918243924b3f943

tests/arrays/test_datetime_array.py
cmp0xff: f62f79cebac133e227d1c8234918243924b3f943

tests/arrays/test_datetime_array.py
cmp0xff: f62f79cebac133e227d1c8234918243924b3f943

tests/arrays/test_datetime_array.py
cmp0xff: f62f79cebac133e227d1c8234918243924b3f943

tests/arrays/test_datetime_array.py
cmp0xff: f62f79cebac133e227d1c8234918243924b3f943

tests/scalars/test_scalars.py
cmp0xff: f62f79cebac133e227d1c8234918243924b3f943

tests/scalars/test_scalars.py
cmp0xff: f62f79cebac133e227d1c8234918243924b3f943

tests/arrays/test_datetime_array.py
cmp0xff: e4a0736d9f43e30f64ed1c52c0df84ba710dd2af

cmp0xff (commented):
Thank you @Dr-Irv , I planned to postpone further tunings to the next PRs, but you have already done much of the research.

> I think it is worth changing the tests that say `if sys.version_info >= (3, 12):` to use a constant `NUMPY_GE_251` that is defined in `tests/__init__.py` to be ` sys.version_info >= (3, 12)`, then it is clear that what we are testing for is the newer version of numpy, not for the difference in the python version.

Unfortunately type checkers do not use such a variable to skip conditional branches, they skip only when seeing `if sys.version_info >= (3, 12)` explicitly, not when seeing `if NUMPY_GE_251` where `NUMPY_GE_251 = sys.version_info >= (3, 12)`.

Dr-Irv (changes requested):
> Unfortunately type checkers do not use such a variable to skip conditional branches, they skip only when seeing if sys.version_info >= (3, 12) explicitly, not when seeing if NUMPY_GE_251 where NUMPY_GE_251 = sys.version_info >= (3, 12).

OK, but I think we need better comments here as opposed to just cross-referencing the mypy issue or a pyrefly issue.

Every time that you have 
```python
if sys.version_info >= (3, 12):
```
there are one of 2 cases:

1. We need the test because `numpy >= 2.5` has broken behavior with `mypy` as referred to in the issue, and with python 3.12, we are using `numpy >= 2.5`
2. We need the test because `numpy >= 2.5` gives us the right behavior, and we are using `numpy=2.4.6` with python 3.11 so we have to add ignores to deal with that.

So I'd prefer a comment in each case (in addition to the cross referencing comments) that says which of those 2 cases is applicable.

Could be as simple as doing
```python
if sys.version_info >= (3, 12):   # numpy  >= 2.5
```
so that we remember the reason is because of the numpy versioning, NOT because of the python version.

tests/arrays/test_datetime_array.py
cmp0xff: 367f5672a282abfb2649b3f344876bafcfca2b3f

tests/arrays/test_datetime_array.py
cmp0xff: 367f5672a282abfb2649b3f344876bafcfca2b3f

tests/scalars/test_scalars.py
cmp0xff: 367f5672a282abfb2649b3f344876bafcfca2b3f

cmp0xff (commented):
Explained in more detail in 367f5672a282abfb2649b3f344876bafcfca2b3f

Dr-Irv (approved):
thanks @cmp0xff

PR #1809, opened 2026-07-02
TST: #1799 enable `ty` in `tests/test_resampler.py`

cmp0xff:


pandas-stubs/core/resample.pyi
```diff
@@ -42,7 +42,7 @@ _FrameGroupByFuncTypes: TypeAlias = (
     _FrameGroupByFunc | str | list[_FrameGroupByFunc | str]
 )
 _FrameGroupByFuncArgs: TypeAlias = (
-    _FrameGroupByFuncTypes | Mapping[Hashable, _FrameGroupByFuncTypes]
+    _FrameGroupByFuncTypes | Mapping[Any, _FrameGroupByFuncTypes]
```
loicdiridollou: Is this to please `ty`? I don't see an added test for a non-hashable element

loicdiridollou (changes requested):
Minor question on the change from `Hashable` to `Any`

pandas-stubs/core/resample.pyi
cmp0xff: Yes this is to make ty happy for some reason. I did not manage to separate a minimal example. I don't think python allows non-Hashable objects as Mapping keys, so the change should not matter. But feel free to challenge me.

loicdiridollou: Yes indeed it will fail at runtime, @Dr-Irv are you okay with this change?

Dr-Irv: > Yes indeed it will fail at runtime, @Dr-Irv are you okay with this change?

No, because I think there are types that match `Mapping` that allow keys that are hashable at runtime, but not `Hashable` in terms of static types.

loicdiridollou: @cmp0xff what would be the errors if you revert the change? If it’s small enough maybe worth opening an issue with ty

cmp0xff: Will investigate later, I am having very little capacity at the moment.

pandas-stubs/core/resample.pyi
```diff
@@ -32,30 +31,22 @@ from pandas._typing import (
     np_ndarray_float,
 )
 
-_FrameGroupByFunc: TypeAlias = (
-    Callable[[DataFrame], Scalar]
-    | Callable[[DataFrame], Series]
-    | Callable[[DataFrame], DataFrame]
-    | np.ufunc
-)
+_FrameGroupByFunc: TypeAlias = Callable[[DataFrame], Scalar | Series | DataFrame]
```
cmp0xff, line 34: It is weird. `np.mean` does not belong to either `Callable[[DataFrame], Scalar | Series | DataFrame]` or `np.ufunc`, yet some of the tests pass, like `assert_type(DF.resample("ME").aggregate(np.sum), DataFrame)`. I can't understand how it can pass.

cmp0xff: astral-sh/ty#3956

pandas-stubs/core/resample.pyi
cmp0xff: Raised astral-sh/ty#3956

loicdiridollou (approved):
Good for the TODO, let’s keep an eye on this as hopefully this gets solved, thanks @cmp0xff

cmp0xff:
Thanks @loicdiridollou , it surprised me that `np.ufunc` was actually not functional; `assert_type(np.mean, np.ufunc)` is `False`.

PR #1810, opened 2026-07-03
GH1796 Add super class StringMethod with __iter__ banned

### LINKED ISSUE #1796: Enable `isinstance(Index.str, StringMethods)` and `isinstance(Series.str, StringMethods)`
Would it be possible to have an empty `StringMethods` class that is a superclass of `IndexStringMethods` and `SeriesStringMethods` so then at least `isinstance(Series.str, StringMethods)` and `isinstance(Index.str, StringMethods)` would be accepted?

_Originally posted by @Dr-Irv in https://github.com/pandas-dev/pandas-stubs/issues/1783#issuecomment-4802995241_
            
Note that `__iter__(self) -> Never` could be moved to the common base class.

loicdiridollou:
- [x] Closes #1796

### DIFF
--- a/tests/test_string_accessors.py
+++ b/tests/test_string_accessors.py
@@ -4,6 +4,7 @@ from typing import assert_type
 
 import numpy as np
 import pandas as pd
+from pandas.core.strings.accessor import StringMethods
 import pytest
 
 from tests import (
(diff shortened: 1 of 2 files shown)


tests/test_string_accessors.py
```diff
@@ -16,6 +17,14 @@
 DATA_BYTES = [b"applep", b"bananap"]
 
 
+def test_accessors_isinstance() -> None:
+    """Test that Series.str and Index.str supertype is `StringMethods`."""
+    s_str = pd.Series(DATA)
+    i_str = pd.Series(DATA)
+    isinstance(s_str, StringMethods)
+    isinstance(i_str, StringMethods)
```
Dr-Irv: I don't think this test is correct.  Don't you want to test `isinstance(s.str, StringMethods)` ?  And shouldn't the test be `assert isinstance(s.str, StringMethods))` ?

Dr-Irv (changes requested):

tests/test_string_accessors.py
loicdiridollou: Yes indeed, wasn't sure how to test it, let me fix it.

Dr-Irv (approved):
thanks @loicdiridollou

PR #1811, opened 2026-07-07
ENH: Change signature of `is_number` and `is_hashable` (EDIT: originally included also "is_array_like")

janpipek:
Modified to type guards:
- is_number
- is_hashable (+fixed the extra arg)

No known issue-

Note: TypeGuard checks just do not work for me.

No agents.

Originally, I added also the following four functions they are not part of pandas public API hence not relevant for the stubs(?).

- is_array_like
- is_nested_list_like
- is_sequence
- is_dataclass


pandas-stubs/core/dtypes/api.pyi
```diff
@@ -1,5 +1,6 @@
 from pandas.core.dtypes.common import (
     is_any_real_numeric_dtype as is_any_real_numeric_dtype,
+    is_array_like as is_array_like,
```
loicdiridollou: Any reason to expose those? I am not sure this is documented with `pandas` so in general we don't expose those.

pandas-stubs/core/dtypes/inference.pyi
```diff
@@ -8,10 +14,11 @@ is_scalar = lib.is_scalar
 is_list_like = lib.is_list_like
 is_iterator = lib.is_iterator
 
-def is_number(obj: object) -> bool: ...
+def is_number(obj: object) -> TypeGuard[Number | np.number]: ...
```
loicdiridollou, line 17: Have you tested those changes, I only see tests for is_array_like? Same for is_hashable

loicdiridollou (changes requested):
@Dr-Irv do you have background on `is_array_like`? I don't believe it is documented.
`is_number` and `is_hashable` are documented so fine with those (although needs tests).

Dr-Irv:
> @Dr-Irv do you have background on `is_array_like`? I don't believe it is documented. `is_number` and `is_hashable` are documented so fine with those (although needs tests).

Seems like `is_array_like` is not documented, but it is the only function in `pandas.api.types` that is not.  I created an issue for this:  https://github.com/pandas-dev/pandas/issues/66211

Let's see what the rest of the pandas dev team says about it.

Dr-Irv:
> > @Dr-Irv do you have background on `is_array_like`? I don't believe it is documented. `is_number` and `is_hashable` are documented so fine with those (although needs tests).
> 
> Seems like `is_array_like` is not documented, but it is the only function in `pandas.api.types` that is not. I created an issue for this: [pandas-dev/pandas#66211](https://github.com/pandas-dev/pandas/issues/66211)
> 
> Let's see what the rest of the pandas dev team says about it.

Discussed in dev meeting 7/8/26, goal is to deprecate `is_array_like()`, so it is not public, so it should not be in the stubs.

loicdiridollou:
Thanks @Dr-Irv!
@janpipek happy to keep the rest of the PR since it is document though it’ll need some testing.

pandas-stubs/core/dtypes/inference.pyi
janpipek: I am afraid `assert_type` does not work nicely with the type-guards but correct me if I am wrong, I am not an expert here. When I change `check(assert_type(api.is_number(obj), bool), bool)` to `check(assert_type(api.is_number(obj), TypeGuard[Number | np.number]), bool)`, pyright and pyrefly fail even with the correct type while mypy or ty succeed even with a wrong type (like TypeGuard[list]).

janpipek:
> > > @Dr-Irv do you have background on `is_array_like`? I don't believe it is documented. `is_number` and `is_hashable` are documented so fine with those (although needs tests).
> > 
> > 
> > Seems like `is_array_like` is not documented, but it is the only function in `pandas.api.types` that is not. I created an issue for this: [pandas-dev/pandas#66211](https://github.com/pandas-dev/pandas/issues/66211)
> > Let's see what the rest of the pandas dev team says about it.
> 
> Discussed in dev meeting 7/8/26, goal is to deprecate `is_array_like()`, so it is not public, so it should not be in the stubs.

That is sad as it was my motivation for this PR in the first place :-(

pandas-stubs/core/dtypes/inference.pyi
Dr-Irv: You should use this form:  `check(assert_type(api.is_number(obj), bool), bool)` because the type of the call is `bool`.  That should work across type checkers.  If not, let us know.

janpipek: Which is already there and the return type update does not require any change... 🙂

tests/test_api_types.py
```diff
@@ -171,6 +171,10 @@ def test_is_hashable() -> None:
     )
     check(assert_type(api.is_hashable(ind), bool), bool)
 
+    for allow_slice in [False, True]:
+        # Check the extra argument
+        check(assert_type(api.is_hashable(obj, allow_slice=allow_slice), bool), bool)
```
loicdiridollou:
```suggestion
    check(assert_type(api.is_hashable(obj, allow_slice=True), bool), bool)
    check(assert_type(api.is_hashable(obj, allow_slice=False), bool), bool)
```
Considering there are only two choices it makes it more readable and from a static checking it is recommended not to use `for` loops

loicdiridollou (changes requested):
Minor request and then it will be good.

tests/test_api_types.py
janpipek: Makes sense.

loicdiridollou (approved):
Thanks @janpipek

PR #1816, opened 2026-07-11
GH1676 Restrict str to Literal for aggregate/transform Groupby

loicdiridollou:


tests/frame/test_groupby.py
```diff
@@ -313,16 +221,19 @@ def wrapped_min(x: pd.Series) -> Scalar:
...
     # Here, MyPy infers dict[object, object], so it must be explicitly annotated
-    agg_dict3: dict[str | int, str | Callable[..., Any]] = {
+    agg_dict3: dict[str | int, Literal["max"] | Callable[..., Any]] = {
         "col2": min,
         "col3": "max",
         0: wrapped_min,
     }
     check(assert_type(df.groupby("col1").agg(agg_dict3), pd.DataFrame), pd.DataFrame)
-    agg_dict4 = {"col2": "sum"}
+    agg_dict4: dict[str, Literal["sum"]] = {"col2": "sum"}
```
loicdiridollou: @Dr-Irv This is where i think we will lose the user, I don't feel that overtyping would make it easy for the user so my suggestion was to leave the `str` there, it is broader and not perfect but it feels more user friendly, what do you think?

.github/setup/action.yaml
loicdiridollou, line 26: Ignore this, I haven't pulled the latest main locally

tests/frame/test_groupby.py
Dr-Irv: For the test, could you try:
```python
check(assert_type(df.groupby("col1").agg({"col2": "sum"}), pd.DataFrame), pd.DataFrame)
```
It might be that by putting it in a variable, you are not letting the bidirectional mapping that type checkers do to occur.

tests/frame/test_frame.py
```diff
@@ -2223,6 +2112,9 @@ def test_types_aggregate() -> None:
     check(assert_type(df.aggregate("mean", axis=1), pd.Series), pd.Series)
     check(assert_type(df.aggregate({"A": "mean"}), pd.Series), pd.Series)
     check(assert_type(df.aggregate({"A": "mean", "B": "sum"}), pd.Series), pd.Series)
+    if TYPE_CHECKING_INVALID_USAGE:
+        df.aggregate("randoom_fuc")  # type: ignore[call-overload] # pyright: ignore[reportCallIssue, reportArgumentType]
```
Dr-Irv: Not that it matters, but maybe `"random_func"` instead of `"randoom_fuc"`

tests/frame/test_groupby.py
```diff
@@ -290,19 +200,17 @@ def test_types_groupby_agg() -> None:
     check(
         assert_type(df.groupby("col1").agg(["min", "max"]), pd.DataFrame), pd.DataFrame
     )
-    agg_dict1 = {"col2": "min", "col3": "max", 0: "sum"}
+    agg_dict1: dict[str | int, Literal["min", "max", "sum"]] = {
+        "col2": "min",
+        "col3": "max",
+        0: "sum",
+    }
     check(assert_type(df.groupby("col1").agg(agg_dict1), pd.DataFrame), pd.DataFrame)
```
Dr-Irv: Or maybe:
```python
check(assert_type(df.groupby("col1").agg({
        "col2": "min",
        "col3": "max",
        0: "sum",
    }), pd.DataFrame), pd.DataFrame)
```
IMHO, it's more likely that people would put the actual `dict` in the arguments as opposed to storing the dict in a variable.

tests/frame/test_groupby.py
```diff
@@ -313,16 +221,19 @@ def wrapped_min(x: pd.Series) -> Scalar:
...
     }
     check(assert_type(df.groupby("col1").agg(agg_dict3), pd.DataFrame), pd.DataFrame)
-    agg_dict4 = {"col2": "sum"}
+    agg_dict4: dict[str, Literal["sum"]] = {"col2": "sum"}
     check(assert_type(df.groupby("col1").agg(agg_dict4), pd.DataFrame), pd.DataFrame)
-    agg_dict5 = {0: "sum"}
+    agg_dict5: dict[int, Literal["sum"]] = {0: "sum"}
     check(assert_type(df.groupby("col1").agg(agg_dict5), pd.DataFrame), pd.DataFrame)
+    if TYPE_CHECKING_INVALID_USAGE:
+        df.groupby("col1").agg("randoom_fuc")  # type: ignore[arg-type] # pyright: ignore[reportCallIssue, reportArgumentType]
```
Dr-Irv: same comment about `randoom_func`

tests/series/test_series.py
```diff
@@ -1373,25 +1320,16 @@ def test_types_transform() -> None:
     )
     check(assert_type(s.transform({"index": "abs"}), pd.DataFrame), pd.DataFrame)
     check(assert_type(s.transform({"index": abs}), pd.DataFrame), pd.DataFrame)
+    if TYPE_CHECKING_INVALID_USAGE:
+        s.transform("randoom_fuc")  # type: ignore[arg-type] # pyright: ignore[reportCallIssue, reportArgumentType]
```
Dr-Irv: here too

tests/test_windowing.py
```diff
@@ -149,6 +149,10 @@ def _mean(df: DataFrame) -> Series:
     # func: np.ufunc | Callable[..., Any] | str | list[Callable[..., Any] | str, np.ufunc] | dict[Hashable, Callable[..., Any] | str | np.ufunc | list[Callable[..., Any] | str]]
     check(assert_type(DF.rolling(10).agg("sum"), DataFrame), DataFrame)
 
+    if TYPE_CHECKING_INVALID_USAGE:
+        DF.rolling(10).aggregate("randoom_fuc")  # type: ignore[arg-type] # pyright: ignore[reportArgumentType]
```
Dr-Irv: here too

Dr-Irv (changes requested):
aside from fixing the issue with `action.yaml`, I think part of the problem with the tests is that we're storing the `dict` arguments as a variable, instead of just passing the value inline.

loicdiridollou:
I think this would be the ideal setup but pyrefly complains a lot while none of mypy, pyright and ty complains when we have a list of functions, it is inferred as `list[str]` and not `list[Literal[...]]` so I am a bit confused which is the best route moving forward, @Dr-Irv let me know if you have some thoughts.

Dr-Irv:
> I think this would be the ideal setup but pyrefly complains a lot while none of mypy, pyright and ty complains when we have a list of functions, it is inferred as `list[str]` and not `list[Literal[...]]` so I am a bit confused which is the best route moving forward, @Dr-Irv let me know if you have some thoughts.

So I think we may be doing something wrong here.  We have:
```python
AggFuncTypeBase: TypeAlias = Callable[P, Any] | Literal[list, of, strings] | np.ufunc
AggFuncTypeDictSeries: TypeAlias = Mapping[HashableT, AggFuncTypeBase[P]]
```
When we refer to `AggFuncTypeBase` in the stubs, we have `AggFuncTypeBase[...]`.  But that only applies to the `Callable` using the `ParamSpec`

So maybe we should split `AggFunctTypeBase` into 2 parts - one part that uses the `ParamSpec` and the `Callable` and the other that uses the `Literal[list, of, strings] | np.ufunc`.  Then wherever we use `AggFuncTypeBase[...]`, we split that into the 2 parts.

That may be the issue with `pyrefly`, although I am not sure.

cmp0xff:
Possibly relevant: microsoft/pyright#11152

cmp0xff:
@loicdiridollou would you continue working on this?

loicdiridollou:
I am stretched a bit at the moment, feel free to take over if you have some time, otherwise will finish it eventually

PR #1817, opened 2026-07-11
GH1815 Fix nightly warnings and rollback function for nightly checks

loicdiridollou:

### DIFF
--- a/scripts/test/run.py
+++ b/scripts/test/run.py
@@ -1,6 +1,7 @@
 from __future__ import annotations
 
 from pathlib import Path
+import re
 import subprocess
 import sys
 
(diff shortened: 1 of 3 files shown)


scripts/test/run.py
```diff
@@ -108,11 +109,13 @@ def nightly_pandas() -> None:
 
 
 def _get_version_from_pyproject(program: str) -> str:
+    """Find version of a package from the pyproject.toml file."""
```
loicdiridollou, line 112: Fixing this function as the current one does not find the version of the package in the environment so after running `poetry run poe pytest --nightly` it does not roll back to pandas 3.0.3 so it left the environment in the wrong config.

Dr-Irv (approved):
thanks @loicdiridollou

scripts/test/run.py
```diff
@@ -108,11 +109,13 @@ def nightly_pandas() -> None:
...
     text = Path("pyproject.toml").read_text()
-    version_line = next(
-        line for line in text.splitlines() if line.startswith(f"{program} = ")
-    )
-    return version_line.split('"')[1]
+    # handle <, >, ==, <=, >= cases
+    match = re.search(rf'"{re.escape(program)}[=<>~!]+([^"]+)"', text)
+    if match is None:
+        raise AssertionError(f"Could not find {program} in pyproject.toml")
```
cmp0xff: I would do instead
```suggestion
    assert match, f"Could not find {program} in pyproject.toml"
```
but I won't block the merging

loicdiridollou: Usually this raises S101 (https://docs.astral.sh/ruff/rules/assert/#assert-s101) does not really matter here but got a lint on my end.
I changed it to `KeyError` which is a bit more accuratea anyway, please merge when you are good.

cmp0xff (approved):
Looks good also to me. Thank you @loicdiridollou !

PR #1820, opened 2026-07-14
Fix/rollforward rollback return type

### LINKED ISSUE #1818: BUG: rollforward and rollback return type incorrectly typed as datetime instead of Timestamp
Originally noted by @sam-s in pandas-dev/pandas#66195.

## Reproducible Example

import pandas as pd
from pandas.tseries.holiday import USFederalHolidayCalendar as usfed

bmb = pd.tseries.offsets.CustomBusinessMonthBegin(calendar=usfed())
bme = pd.tseries.offsets.CustomBusinessMonthEnd(calendar=usfed())

reveal_type(bmb.rollforward)
reveal_type(bme.rollback)

## mypy output

Revealed type is "def (dt: datetime.datetime) -> datetime.datetime"
Revealed type is "def (dt: datetime.datetime) -> datetime.datetime"

## Expected behavior

Per the docs, rollforward and rollback should:
- accept `datetime | Timestamp` as parameter
- return `Timestamp`, not `datetime`

## Environment
- pandas 3.0.3
- pandas-stubs 3.0.3.260530
- mypy 2.2.0

krishgarg344:
- [x] closes #1818 
- [x] Related to pandas-dev/pandas#66195
- [x] Fix rollforward and rollback to accept datetime | Timestamp and return Timestamp as in docs
- [x] Tests added, used `assert_type()` checks for `rollback` and `rollforward` to return type  as `Timestamp`


pandas-stubs/_libs/tslibs/offsets.pyi
```diff
@@ -96,8 +96,8 @@ class BaseOffset:
     def rule_code(self) -> str: ...
     @property
     def freqstr(self) -> str: ...
-    def rollback(self, dt: datetime) -> datetime: ...
-    def rollforward(self, dt: datetime) -> datetime: ...
+    def rollback(self, dt: datetime | Timestamp) -> Timestamp: ...
+    def rollforward(self, dt: datetime | Timestamp) -> Timestamp: ...
```
cmp0xff, line 100: `Timestamp` is a subclass of `datetime`, so I think the following is sufficient

```suggestion
    def rollback(self, dt: datetime) -> Timestamp: ...
    def rollforward(self, dt: datetime) -> Timestamp: ...
```

tests/test_holidays.py
```diff
@@ -97,3 +97,22 @@ def test_custom_business_month() -> None:
...
     )
+
+
+def test_rollforward_rollback_return_type() -> None:
+
+    bmb = pd.offsets.CustomBusinessMonthBegin()
+    bme = pd.offsets.CustomBusinessMonthEnd()
+
+    check(
+        assert_type(bmb.rollforward(datetime(2024, 1, 1)), pd.Timestamp), pd.Timestamp
```
cmp0xff, line 108: Would it work if you insert `datetime` and `Timestamp` with non-zero seconds as well?

cmp0xff (changes requested):

pandas-stubs/_libs/tslibs/offsets.pyi
krishgarg344: Thanks for the suggestion! Indeed `Timestamp` is a subclass of `datetime` but see the [docs](https://pandas.pydata.org/docs/reference/api/pandas.tseries.offsets.DateOffset.rollback.html), they explicitly ask for datetime | Timestamp as input parameters. So maybe keeping this makes it more accurate. Otherwise would like maintainer review if there's a strong preference either way.

tests/test_holidays.py
krishgarg344: Good point, added tests with non-zero seconds for both datetime and Timestamp inputs.

pandas-stubs/_libs/tslibs/offsets.pyi
cmp0xff: OK.

I realised that you are changing the base class `BaseOffset` here, and you are referring to the docs of `DateOffset`. However in the tests you added, only `CustomBusinessMonthBegin` and `CustomBusinessMonthEnd` are tested.

Please add at least tests for `DateOffset` as well.

cmp0xff (changes requested):

krishgarg344:
Added tests for DateOffset as well

cmp0xff (approved):
Thank you @krishgarg344 !

PR #1821, opened 2026-07-14
TST: #1799 enable `ty` in `tests/test_pandas.py`

cmp0xff:
It is very hard to isolate minimal script to report in this one.

- [x] Addresses #1799

### DIFF
--- a/tests/test_pandas.py
+++ b/tests/test_pandas.py
@@ -715,30 +715,19 @@ def test_isna() -> None:
     if pd.notna(nullable1):
         check(assert_type(nullable1, str), str)
     if not pd.isna(nullable1):
-        check(
-            assert_type(nullable1, str),  # ty: ignore[assert-type-unspellable-subtype]
-            str,
-        )
+        check(assert_type(nullable1, str), str)
     if pd.isna(nullable1):
         assert_type(nullable1, NaTType | NAType | None)
     if not pd.notna(nullable1):
-        assert_type(  # ty: ignore[assert-type-unspellable-subtype]
-            nullable1, NaTType | NAType | None
-        )
+        assert_type(nullable1, NaTType | NAType | None)
 
     nullable2 = random.choice([2, None])
     if pd.notna(nullable2):
         check(assert_type(nullable2, int), int)
     if not pd.isna(nullable2):
-        check(
-            assert_type(nullable2, int),  # ty: ignore[assert-type-unspellable-subtype]
-            int,
-        )
+        check(assert_type(nullable2, int), int)
     if pd.isna(nullable2):
-        check(
-            assert_type(nullable2, None),  # ty: ignore[type-assertion-failure]
-            type(None),
-        )
+        check(assert_type(nullable2, None), type(None))
     if not pd.notna(nullable2):
         check(assert_type(nullable2, None), type(None))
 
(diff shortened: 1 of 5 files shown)


tests/test_pandas.py
```diff
@@ -118,22 +118,41 @@ def test_types_concat_none() -> None:
     )
     check(assert_type(pd.concat([None, df]), pd.DataFrame), pd.DataFrame)
     check(
-        assert_type(pd.concat([None, series, df], axis=1), pd.DataFrame), pd.DataFrame
+        assert_type(  # ty: ignore[type-assertion-failure]
```
cmp0xff, line 124: I cannot work out a minimal example to reproduce the error family around `concat`. Can we create an internal `ty` issue master tracker and postpone the reporting to `ty`?

cmp0xff:
And there is a feedback suggesting using `@final` above `class NAType` https://github.com/astral-sh/ty/issues/4005#issuecomment-4978470163. What do you think @Dr-Irv @MarcoGorelli since you are the main authors of this class in the stubs?

MarcoGorelli:
> ty is correct here. These errors only happen because NAType is not final, so there could be common subclasses of NAType and int

ooh, interesting!

seems fine to use final anyway, i don't think `NaType` in pandas is meant to be subclassed

Dr-Irv:
> And there is a feedback suggesting using `@final` above `class NAType` [astral-sh/ty#4005 (comment)](https://github.com/astral-sh/ty/issues/4005#issuecomment-4978470163). What do you think @Dr-Irv @MarcoGorelli since you are the main authors of this class in the stubs?

If it works, I don't have a problem with it.

tests/test_pandas.py
```diff
@@ -529,38 +590,123 @@ def test_isna() -> None:
...
 
-    assert check(assert_type(pd.isna(pd.NA), bool), bool)
-    assert not check(assert_type(pd.notna(pd.NA), bool), bool)
+    # assert-type-unspellable-subtype is ty-specific astral-sh/ty#4005
+    assert check(
+        assert_type(  # ty: ignore[assert-type-unspellable-subtype]
+            pd.isna(pd.NA), bool
+        ),
+        bool,
+    )
```
cmp0xff: In https://github.com/astral-sh/ty/issues/4005#issuecomment-4983779126 it was also suggested to ignore `assert-type-unspellable-subtype` globally to retain the compatibility across type checkers, since `assert-type-unspellable-subtype` seems to be `ty`-specific. Currently I am ignoring them in-line. Which way do way prefer?

Dr-Irv: I don't have a problem with disabling that via `pyproject.toml`.  We used to do some similar things for `pyright` , and do similar things for `mypy`

When we "officially" support `ty`, we should document that we support `ty` with that feature turned off.

tests/test_pandas.py
cmp0xff: Maybe just use #1799.

cmp0xff: f7e1dc7b5a411980dc4d6fadb9f20c456e91a846 astral-sh/ty#4022

loicdiridollou (approved):
Nice one here! Thanks @cmp0xff

PR #1834, opened 2026-07-22
TYP: propagate astral-sh/ruff#26964 and upgrade to `ty` 0.0.64

cmp0xff:


pyproject.toml
```diff
@@ -37,7 +37,7 @@ dev = [
   "pyarrow>=10.0.1",
   "pytest>=8.4.2",
   "pyright>=1.1.410",
-  "ty>=0.0.61",
+  "ty>=0.0.62",
```
loicdiridollou: What if we lock this to 0.0.62 for the moment?

loicdiridollou:
Maybe good to check with 0.0.63 and if good then happy to merge this.

pyproject.toml
cmp0xff: It also works for 0.0.63.

cmp0xff: Now 0.0.64, version flies.

loicdiridollou (approved):
Great thanks @cmp0xff

PR #1835, opened 2026-07-22
CLN: python/typeshed#16072 replace `SupportsGetItem`

cmp0xff:
`SupportsGetItem` has two slots, but in our application we never made use of the first slot. I figured out we could use ~~`Viewable`~~`Iterable` instead.

The initial idea was to use `SupportsIter`, but it turns out that it would require the annotation `SupportsIter[Iterator[S1]]` which is cumbersome, see python/typeshed#16072.


loicdiridollou (approved):
Looks good, great cleanup!

PR #1837, opened 2026-07-23
CLN: update `ruff` and enable `A`, `ERA` except `ERA001`, and `RUF`

cmp0xff:
We may not want [`A001`](https://docs.astral.sh/ruff/rules/builtin-variable-shadowing/) and [`A004`](https://docs.astral.sh/ruff/rules/builtin-import-shadowing/) since `pandas` does shadow the built-in `eval`. But this is the only case.


loicdiridollou (approved):
Nice clean up thanks @cmp0xff

PR #1839, opened 2026-07-24
TYP: stricter `pyright`

cmp0xff:
With #1171 and #1639 we have made `pandas-stubs` passing `pyright` in `strict` mode.

However there are still disabled `pyright` rules even in `strict` mode, see https://github.com/microsoft/pyright/blob/main/docs/configuration.md#diagnostic-settings-defaults.

In this PR I have included all of them in `pyproject.toml` and enabled some of them.

Enabling `reportMissingSuperCall` has led to two (non-functional) changes in `tests/extension/decimal/array.py`.

Note that `reportMissingModuleSource` always gives warning. We can make `pyright` return `1` upon having warnings by adding `--warnings` flag. I have added the flag and removed `reportMissingModuleSource = true`.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -303,9 +303,21 @@ typeCheckingMode = "strict"
 stubPath = "."
 include = ["tests", "pandas-stubs"]
 enableTypeIgnoreComments = false    # use pyright-specific ignores
-reportUnnecessaryTypeIgnoreComment = true
 # enable optional checks
-reportMissingModuleSource = true
+# https://github.com/microsoft/pyright/blob/main/docs/configuration.md#diagnostic-settings-defaults
+deprecateTypingAliases = true
+enableExperimentalFeatures = false
+reportCallInDefaultInitializer = true
+reportImplicitOverride = "none"
+reportImplicitStringConcatenation = true
+reportImportCycles = "none"
+reportMissingSuperCall = true
+reportPropertyTypeMismatch = "none"
+reportUninitializedInstanceVariable = "none"
+reportUnnecessaryTypeIgnoreComment = true
+reportUnreachable = "none"
+reportUnusedCallResult = "none"
+
 
 [tool.codespell]
 ignore-words-list = "indext, mose, sav, ser, ttest"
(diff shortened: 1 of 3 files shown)


pyproject.toml
```diff
@@ -307,9 +307,21 @@ typeCheckingMode = "strict"
...
+reportCallInDefaultInitializer = true
+reportImplicitOverride = "none"
+reportImplicitStringConcatenation = true
+reportImportCycles = "none"
+reportMissingSuperCall = true
+reportPropertyTypeMismatch = "none"
+reportUninitializedInstanceVariable = "none"
+reportUnnecessaryTypeIgnoreComment = true
+reportUnreachable = "none"
+reportUnusedCallResult = "none"
```
Dr-Irv, line 319: If these are the default values, why do we need to include these in our `pyproject.toml` ?  I'm missing why you're doing this?

cmp0xff: My idea is we can either enable them later (I can add `# TODO` if that is clearer), or comment with a reason why we cannot enable them.

Dr-Irv: But by saying `strict`, they are enabled, so our only issue is that we have to use `#pyright: ignore` in the stubs?

I don't think that's a big deal.  What's more important is that a user with a codebase can use `strict` and still use our stubs.  I think we are testing that, right?

cmp0xff: > No these rules are the exceptions, that even if `strict` is enabled, these rules are not turned on. See the table https://github.com/microsoft/pyright/blob/main/docs/configuration.md#diagnostic-settings-defaults. In the last column `Strict`, there are not only `true` and `"error"` and `"warning"`, but also `"none"` and `false`. These are pasted in this PR.

So this would be some kind of "super-strict" support?

cmp0xff: > So this would be some kind of "super-strict" support?

Yes (accidentally you edited my reply instead of adding a new one 😄)

loicdiridollou (approved):
Thanks @cmp0xff

PR #1840, opened 2026-07-27
allow pd.col().str

Dr-Irv:
Using `df.assign(x=pd.col["column"].str.title())` didn't work, so this fixes it.

Had to pin `ty` to 0.0.61 because 0.0.63 has failures for @cmp0xff to investigate


loicdiridollou (approved):
Looks good to me, okay for the `ty` pinning for now. Thanks @Dr-Irv

PR #1841, opened 2026-07-27
MNT: Upgrade ty and fix nightly warnings with changes in read_excel default engine

loicdiridollou:


tests/indexes/test_indexes.py
```diff
@@ -76,7 +76,11 @@ def test_index_isin() -> None:
 
     mi = pd.MultiIndex.from_arrays([[1, 2, 3]])
     check(assert_type(mi.isin([[3]]), np_1darray_bool), np_1darray_bool)
-    check(assert_type(mi.isin({iter([3])}), np_1darray_bool), np_1darray_bool)
+
+    if PD_LTE_31:
+        # TODO: pandas-dev/pandas#0000 pandas bug
```
Dr-Irv: can we get a better reference here?

tests/test_io.py
```diff
@@ -1034,107 +1034,235 @@ def test_read_excel(tmp_path: Path) -> None:
...
-    )
-    check(
-        assert_type(
-            pd.read_excel(path_str, names=(("higher", "lower"),), header=0),
+
+    with pytest_warns_bounded(
+        errors.Pandas4Warning,
+        "The default engine for reading 'xlsx' files will ch",
+        lower="3.0.99",
+    ):
```
Dr-Irv: Given how often we have to do this, can we just create something in `tests/__init__.py` that returns the `pytest_warns_bounded(...)` here and then if it turns out that we need to change the bounds or the message, we have one line to change?

This will also shorten the number of lines changed in the PR, and make it easier to review.

Dr-Irv (changes requested):

tests/indexes/test_indexes.py
loicdiridollou: Indeed, I had not opened the issue yet, wanted to push it to CI first, done

tests/test_io.py
loicdiridollou: I have created a wrapper for pytest_warns_pandas4 for those and we could clean up the rest of the warnings with that wrapper.

tests/test_pandas.py
```diff
@@ -123,6 +123,7 @@ def test_types_concat_none() -> None:
         assert_type(pd.concat([None, series, df], axis=1), pd.DataFrame), pd.DataFrame
     )
     check(assert_type(pd.concat([None, series, df]), pd.DataFrame), pd.DataFrame)
+    check(assert_type(pd.concat([None, series, df]), pd.DataFrame), pd.DataFrame)
```
Dr-Irv: duplicate line

tests/__init__.py
```diff
@@ -206,6 +208,14 @@ def pytest_warns_conditioned(
     return suppress(upper_exception)
 
 
+def pytest_warns_pandas4(
```
Dr-Irv: I'd rather not introduce this.  What I would like to do is introduce this:

```python
def pytest_warns_excel_pandas4():
    return  pytest_warns_bounded(
        Pandas4Warning,
        "The default engine for reading 'xlsx' files will ch",
        lower="3.0.99",
    )
```
Then all the excel messages are in one place

Dr-Irv (changes requested):

tests/__init__.py
loicdiridollou: Agree, adjusted

Dr-Irv (approved):
Thanks @loicdiridollou

PR #1845, opened 2026-07-28
TST: #1799 enable `ty` in some of `tests/arrays/**/*`

cmp0xff:
Due to the following two `ty` issues, the progress is limited.

- astral-sh/ty#3199: issues around nested `numpy` `dtype`
- ~~astral-sh/ty#4102: issues around `list[str | numpy.str_]`~~ astral-sh/ty#4111: issues around `Sequence[str | Just[float]]`


tests/arrays/test_base_string_.py
```diff
@@ -1,3 +1,5 @@
+# ty: ignore[no-matching-overload,type-assertion-failure]
+# TODO: remove the above ignores astral-sh/ty#3199, astral-sh/ty#4111
```
loicdiridollou, line 2: 4111 has been marked as unplanned by the ty team, does this impact the work here?

cmp0xff: astral-sh/ty#4102 is marked as unplanned by I myself, it wasn't the real cause of problems.

astral-sh/ty#4111 is the real problem and has been resolved. Unfortunately it does not seem to be contained in ty 0.0.65.

loicdiridollou: Okay sounds good i will ship it and we will see in the next versions of ty

loicdiridollou (approved):

PR #1849, opened 2026-07-30
TST: #1799 enable `ty` in `tests/test_string_accessors.py`

cmp0xff:
In `tests/test_string_accessors.py`, there are 368 `: ignore[`'s, which is a multiple of 4. In other words, we should have the same number of `ignore`s for all 4 type checkers in that script.


loicdiridollou (approved):
Thanks @cmp0xff

PR #1850, opened 2026-07-30
TST: #1799 enable `ty` in `tests/test_groupby.py`

cmp0xff:
In `tests/test_groupby.py`, there are 88+2 `: ignore[`'s. Two are `# pyright: ignore[reportUnusedFunction]`. The rest is a multiple of 4. In other words, we should have the same number of `ignore`s for all 4 type checkers in that script.

### DIFF
--- a/tests/test_groupby.py
+++ b/tests/test_groupby.py
@@ -111,7 +111,7 @@ def test_frame_groupby_resample() -> None:
 
     # fillna (deprecated)
     if TYPE_CHECKING_INVALID_USAGE:
-        GB_DF.resample("ME").fillna("ffill")  # type: ignore[operator] # pyright: ignore  # pyrefly: ignore[not-callable]
+        GB_DF.resample("ME").fillna("ffill")  # type: ignore[operator] # pyright: ignore[reportCallIssue] # pyrefly: ignore[not-callable] # ty: ignore[call-non-callable]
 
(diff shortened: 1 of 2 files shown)


tests/test_groupby.py
```diff
@@ -111,7 +111,7 @@ def test_frame_groupby_resample() -> None:
 
     # fillna (deprecated)
     if TYPE_CHECKING_INVALID_USAGE:
-        GB_DF.resample("ME").fillna("ffill")  # type: ignore[operator] # pyright: ignore  # pyrefly: ignore[not-callable]
```
loicdiridollou, line 114: Did not know we were even allowed to pass a broad ignore for pyright in this codebase.

loicdiridollou (approved):
Thanks @cmp0xff

tests/test_groupby.py
cmp0xff: I found `black` allows a long line if the first comment is a `type: ignore`.

Dr-Irv: I'd prefer being more specific about which kind of pyright error that we are ignoring here.

cmp0xff: Hi @Dr-Irv, yes it was an issue in the old code. I've fixed it in this MR.

PR #1858, opened 2026-07-31
GH1848 Drop str | np.str_ split as np.str_ is a subclass of str

### LINKED ISSUE #1848: Simplify `str | np.str_` (and add a test)
At a few places we have `str | np.str_`. The latter turns out to be a subclass of the built-in `str`. We could add a test to nail it down and simplify `str | np.str_` to `str`.

> [`np._str` is defined as a subclass of `str`](https://github.com/numpy/numpy/blob/5e1d03ffac5f2c0a9c39bfcaa9fc853b2b83151e/numpy/__init__.pyi#L7443), so `str | np._str` can be simplified to `str`.

 _Originally posted by @sharkdp in [#4102](https://github.com/astral-sh/ty/issues/4102#issuecomment-5102515226)_

loicdiridollou:
- [x] Closes #1848 

### DIFF
--- a/pandas-stubs/_typing.pyi
+++ b/pandas-stubs/_typing.pyi
@@ -932,8 +932,6 @@ PyArrowNotStrDtypeArg: TypeAlias = (
     | PyArrowBytesDtypeArg
 )
 
-StrLike: TypeAlias = str | np.str_
-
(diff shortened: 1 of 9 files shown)


cmp0xff (approved):
Thank you @loicdiridollou !

PR #1843, opened 2026-07-27
change Mapping arg to use a TypeVar in read_csv

Dr-Irv:
- [X] Closes #1842(Replace xxxx with the Github issue number)
- [X] Tests added (Please use `assert_type()` to assert the type of any return value)
- [ ] If I used AI to develop this pull request, I prompted it to follow `AGENTS.md`.

PR #1859, opened 2026-08-01
TYP: use HashableT TypeVars for Mapping parameters

### LINKED ISSUE #1844: Change `Mapping[Hashable,` to `Mapping[HashableTx,` in stubs
Following the issue reported in #1842 that I fixed in #1843 , the solution is to not use `Mapping[Hashable` in various places, but to use `Mapping[HashableT` or `Mapping{HashbleTx,` where x=1,2,3,4,5,6.  the `TypeVar` chosen needs to be unique within the parameter types of the function.

There are 28 places where we use this.

dgvj-work:
- [x] Closes #1844

## Summary

Replace remaining `Mapping[Hashable, ...]` parameter annotations with `Mapping[HashableTx, ...]` so typed `dict[str, ...]` arguments are accepted. Same approach as #1843 for `read_csv`.

Touched:
- `DtypeArg` / `ConvertersArg` / `ColspaceArgType` aliases
- `read_table` / `read_clipboard` / `read_json` dtype mappings
- `DataFrame.isin`, `to_html` / `to_string` formatters
- `NDFrame.to_sql` dtype mapping
- `Resampler.aggregate` mappings

Left unchanged on purpose:
- return types like `MutableMapping[Hashable, ...]`
- `JsonReader.dtype` attribute (unbound TypeVar is invalid there)

## Testing

- `poetry run poe pyright` - 0 errors
- `poetry run poe mypy` - success (349 files)
- `poetry run poe ty` - all checks passed
- `poetry run pytest` - passed
- pre-commit - passed


tests/test_io.py
```diff
@@ -900,6 +900,19 @@ def test_read_csv_dtype_dict() -> None:
     check(assert_type(pd.read_csv(csv, dtype=my_col_types), pd.DataFrame), pd.DataFrame)
 
 
+def test_read_table_dtype_dict() -> None:
```
loicdiridollou: This seems a little small coverage for the amount of change you made.
Could you add more tests to show that if you revert the changes in the pyi files then it will fail the CI?

loicdiridollou (changes requested):

loicdiridollou:
You probably want to change in `_typing.pyi` where it has `DtypeArg: TypeAlias = Dtype | Mapping[HashableT0, Dtype]` (I already made the change in the snippet).

tests/test_io.py
cmp0xff: I would like to add that there is already a `test_read_table` below. I think it would be better to add the new test there in the existing function.

dgvj-work: Good call, I expanded the coverage so the typed `dict[str, ...]` pattern is exercised for the other Mapping sites in this PR (`read_table` / clipboard / json dtype, `isin`, `to_html`/`to_string` formatters, `to_sql` dtype, and resample `aggregate`). Those should fail type-checking if the stub changes are reverted.

dgvj-work: Agreed - folded it into `test_read_table` instead of a separate function.

dgvj-work:
@loicdiridollou updated `DtypeArg` (and the related `ConvertersArg` / `ColspaceArgType` aliases) in `_typing.pyi` to use `HashableT0` as you suggested. Also added broader typed-dict tests per the review thread.

tests/frame/test_indexing.py
```diff
@@ -229,6 +229,9 @@ def test_frame_isin() -> None:
     check(assert_type(df.isin(pd.Index([1, 3, 5])), pd.DataFrame), pd.DataFrame)
     check(assert_type(df.isin(df), pd.DataFrame), pd.DataFrame)
     check(assert_type(df.isin({"x": [1, 2]}), pd.DataFrame), pd.DataFrame)
+    # GH 1844
+    isin_map: dict[str, list[int]] = {"x": [1, 2]}
```
loicdiridollou: @Dr-Irv @cmp0xff what is your view on adding more types here, I think the typecheckers are good enough so we don't need to add the types when declaring the variable

Dr-Irv: In the tests, we don't want to declare the types of variables.  We want to see what the type checkers will infer.

Sometimes, if you do `foo = {"x": [1,2]}` followed by `df.some_method(foo)` you get different results than doing `df.some_method({"x": [1,2]})`.  We prefer to test the latter.  That's not to say that we couldn't test the former, but it should be done without declaring a specific type for the variable, or doing it by writing:
```python
def func(goo: dict[str, list[int]]):
     check(df.some_method(goo), pd.DataFrame), pd.DataFrame)

foo = {"x": [1,2]}
func(foo)
```

tests/test_io.py
```diff
@@ -223,6 +226,12 @@ def test_clipboard() -> None:
         assert_type(read_clipboard(dtype={"first": "f8"}), DataFrame),
         DataFrame,
     )
+    # GH 1844
+    clipboard_dtypes: dict[str, str] = {"first": "f8"}
```
loicdiridollou: Please remove the extra types here, this is not how we would use it in general context, the stubs want to adapt the same usage as a regular user. Same for all the other places

tests/test_io.py
```diff
@@ -914,6 +923,15 @@ def test_read_table(tmp_path: Path) -> None:
         assert_type(read_table(path_str, dtype={"first": "f8"}), DataFrame),
         DataFrame,
     )
+    # GH 1844 — typed dict[str, type] must be accepted for dtype=
+    table_col_types: dict[str, type] = {
```
loicdiridollou: Remove extra type here, the type checkers already do that

tests/test_io.py
```diff
@@ -1486,6 +1507,17 @@ def test_to_sql_dtype_sqlalchemy_type(tmp_path: Path) -> None:
         ),
         int,
     )
+    # GH 1844
+    sql_dtypes: dict[str, Any] = {
```
loicdiridollou: Same here remove

tests/test_io.py
```diff
@@ -1624,6 +1656,9 @@ def test_read_sql_query_via_sqlalchemy_engine_with_tuple_valued_params() -> None
 
 def test_read_html(tmp_path: Path) -> None:
     check(assert_type(DF.to_html(), str), str)
+    # GH 1844
+    html_formatters: dict[str, Callable[[object], str]] = {"a": str}
```
loicdiridollou: Same here

tests/test_io.py
```diff
@@ -1752,6 +1787,9 @@ def test_all_read_without_lxml_dtype_backend(tmp_path: Path) -> None:
     check(
         assert_type(read_json(path_str, dtype={"MatchID": str}), DataFrame), DataFrame
     )
+    # GH 1844
+    json_dtypes: dict[str, type] = {"MatchID": str}
```
loicdiridollou: Same here

tests/test_resampler.py
```diff
@@ -106,6 +106,12 @@ def test_aggregate() -> None:
         ),
         DataFrame,
     )
+    # GH 1844
+    resample_agg: dict[str, str] = {"col1": "sum", "col2": "mean"}
```
loicdiridollou: Same here

loicdiridollou (changes requested):
Please remove the extra typehints, the typecheckers already take care of that so you don't need to add extra information for them

tests/frame/test_indexing.py
dgvj-work: Agreed and removed the explicit annotations. Left the unannotated dict variables so we still cover the pass-via-variable case, and let the type checkers infer the types.

dgvj-work: Thanks! updated accordingly. Dropped the variable annotations so inference is what gets tested, consistent with the existing `test_read_csv_dtype_dict` pattern.

tests/test_io.py
dgvj-work: Removed the extra type annotation here (and the same pattern elsewhere).

tests/test_io.py
dgvj-work: Done - annotation removed.

tests/test_io.py
dgvj-work: Done - annotation removed.

tests/test_io.py
dgvj-work: Done - annotation removed.

tests/test_io.py
dgvj-work: Done - annotation removed.

tests/test_resampler.py
dgvj-work: Done - annotation removed.

dgvj-work (commented):
Addressed the requested changes: removed the extra type hints from the test variables so inference is what gets checked.

dgvj-work:
@loicdiridollou @Dr-Irv thanks for the feedback - removed the explicit dict type annotations from the GH 1844 tests so the type checkers infer them, matching the preferred style. Left the unannotated variables in place to keep the pass-via-variable coverage.

tests/test_io.py
```diff
@@ -914,6 +920,15 @@ def test_read_table(tmp_path: Path) -> None:
         assert_type(read_table(path_str, dtype={"first": "f8"}), DataFrame),
         DataFrame,
     )
+    # GH 1844
+    table_col_types = {
+        "a": int,
+        "b": float,
+    }
```
loicdiridollou:
```suggestion
    table_col_types = {"a": int, "b": float}
```

tests/test_io.py
```diff
@@ -1486,6 +1504,17 @@ def test_to_sql_dtype_sqlalchemy_type(tmp_path: Path) -> None:
         ),
         int,
     )
+    # GH 1844
+    sql_dtypes = {
+        "a": sqlalchemy.types.INTEGER,
+        "b": sqlalchemy.types.FLOAT,
+    }
```
loicdiridollou:
```suggestion
    sql_dtypes = {"a": sqlalchemy.types.INTEGER, "b": sqlalchemy.types.FLOAT}
```

loicdiridollou (changes requested):
Minor formatting for the tests but the rest is looking good.

tests/test_io.py
dgvj-work: Done - compacted to a single-line dict.

tests/test_io.py
dgvj-work: Done - compacted to a single-line dict.

dgvj-work (commented):
Thanks - applied the formatting suggestions for the two dict literals. Ready for another look.

loicdiridollou (approved):
Great job thanks @dgvj-work

cmp0xff:
@dgvj-work please rebase onto main / merge main into your branch to see if the CI can be fixed

dgvj-work:
@cmp0xff  Done with the rebase onto latest main (includes #1884).

The new Test workflow is waiting on maintainer approval for the fork PR run:
https://github.com/pandas-dev/pandas-stubs/actions/runs/31410722464

Could someone please approve that workflow when convenient?

PR #1864, opened 2026-08-02
TST: #1799 enable `ty` in `tests/frame/test_frame.py`

cmp0xff:
- [x] Addresses #1799


pandas-stubs/core/series.pyi
```diff
@@ -1906,6 +1906,9 @@ class Series(IndexOpsMixin[S1], ElementOpsMixin[S1], NDFrame):
     def __add__(
         self: Supports_ProtoAdd[S2_contra, S2], other: S2_contra | Sequence[S2_contra]
     ) -> Series[S2]: ...
+    # TODO: pandas-dev/pandas-stubs#1799 the following overload causes ty
+    # crashing with test_compute_values in tests/frame/test_frame.py.
+    # Investigate and report to ty.
```
cmp0xff: This is hard to figure out

tests/frame/test_frame.py
```diff
@@ -2950,7 +2944,12 @@ def test_dataframe_pct_change() -> None:
 def test_compute_values() -> None:
     df = pd.DataFrame({"x": [1, 2, 3, 4]})
     s: pd.Series = pd.Series([10, 20, 30, 40])
-    check(assert_type(df["x"] + s.values, pd.Series), pd.Series, np.int64)
+    # TODO: pandas-dev/pandas-stubs#1799 the original code causes ty
+    # crashing with a marked overload of Series.__add__ in
+    # pandas-stubs/core/series.pyi. Investigate and report to ty. When fixed,
+    # use the following line and remove the line with __add__.
+    # check(assert_type(df["x"] + s.values, pd.Series), pd.Series, np.int64)
+    check(assert_type(df["x"].__add__(s.values), pd.Series), pd.Series, np.int64)
```
cmp0xff: This is hard to figure out

loicdiridollou: Do we have a way to have a timeout for the ty run, it is pretty transient so a little hard to nail down but would prefer that CI does not end up deadlocked because of this.

loicdiridollou (approved):
Thanks @cmp0xff

tests/frame/test_frame.py
cmp0xff: Thank you for the approval! The CI timeout is 10 minutes. But on local machine it is annoying. Also it does not freeze anymore because of my workaround.

cmp0xff: Oh it does not freeze anymore, let me fix it

cmp0xff: It still freezes. Rolling back. https://github.com/pandas-dev/pandas-stubs/actions/runs/31049878204

cmp0xff: It may have to do with astral-sh/ty#4195

PR #1836, opened 2026-07-23
TST: "strict" modes for `mypy`, `pyrefly` and `ty`

cmp0xff:
We used to have a `pyright_strict` which was integrated into `pyright` with #1171 and #1639.

We can apply the same idea to `mypy`, `pyrefly` and `ty`:
- https://mypy.readthedocs.io/en/stable/config_file.html#confval-strict
- https://pyrefly.org/en/docs/configuration/#preset
- https://docs.astral.sh/ty/rules/#rule-levels

Just a note, for now I see a value in making [`missing-source`](https://pyrefly.org/en/docs/error-kinds/#missing-source) work for `pyrefly_strict`.

PR #1507, opened 2025-11-26
type `isin`, remove redundant `Iterable | Series`

MarcoGorelli:
`Series` is already `Iterable`, so `Iterable | Series` was redundant here

`Series.isin` doesn't accept `dict` in the same way that `DataFrame.isin` does:
- in the dataframe case, you can pass a per-column mapping
- in the series one, it accidentally works but just because it's an iterable (i.e., only the keys are checked anyway)

- [ ] Closes #xxxx (Replace xxxx with the Github issue number)
- [ ] Tests added (Please use `assert_type()` to assert the type of any return value)
- [ ] If I used AI to develop this pull request, I prompted it to follow `AGENTS.md`.

PR #1486, opened 2025-11-13
Type Index methods: is_, ravel, take, repeat, to_flat_index,   `to_frame`, `set_names`, `rename`, remove undocumented `__copy__`, `__deepcopy__`, `__array_wrap__`

MarcoGorelli:
towards #1485 

- [ ] Closes #xxxx (Replace xxxx with the Github issue number)
- [ ] Tests added: Please use `assert_type()` to assert the type of any return value

PR #1867, opened 2026-08-02
CLN: #1836 merge `pyrefly_strict` into `pyrefly`, make `ty_all` and `pyrefly_all` passing

cmp0xff:
- [x] Addresses #1836 (Replace xxxx with the Github issue number)


.github/workflows/optional.yml
```diff
@@ -97,4 +81,4 @@ jobs:
         python-version: '3.14'
 
     - name: Run ty tests with all rules raising errors
-      run: poetry run poe ty_all
+      run: poetry run poe ty_all
```
cmp0xff:
```suggestion
      run: poetry run poe ty_all

```

cmp0xff:
Basically I'm adding `@override`s. Now `ty_all` and `pyrefly_all` pass. I think `ty_all` is not very stable when `ty` updates, so I decide not to merge it to `ty` for now. Maybe in the future when `ty` adds a strict mode astral-sh/ty#527.

pandas-stubs/_libs/missing.pyi
```diff
@@ -11,14 +11,18 @@ import numpy as np
...
 
 @final
 class NAType:
     def __new__(cls, *args: Any, **kwargs: Any) -> Self: ...
+    @override
     def __format__(self, format_spec: str) -> str: ...
+    @override
     def __hash__(self) -> int: ...
+    @override
     def __reduce__(self) -> str: ...
```
Dr-Irv: If we are doing `@override` on methods that are part of the `object` definition in the typeshed fallback, maybe we should delete these methods in the stubs, since they are defined there.  As long as our override is consistent with the `object` declaration, then we don't need to include the `override` in the stubs. 

would have to change this here and other places where we do this.

pandas-stubs/core/arrays/categorical.pyi
```diff
@@ -190,7 +193,9 @@ class Categorical(NDArrayBackedExtensionArray, Generic[CategoricalValueT]):
         self, removals: Hashable | SequenceNotStr[Hashable] | AnyArrayLike
     ) -> Self: ...
     def remove_unused_categories(self) -> Self: ...
+    @override
     def __eq__(self, other: object) -> bool: ...
+    @override
     def __ne__(self, other: object) -> bool: ...
```
Dr-Irv: I think we can delete these, right?

pandas-stubs/core/arrays/masked.pyi
```diff
@@ -24,11 +25,14 @@ from pandas._typing import (
 
 class BaseMaskedArray(ExtensionArray):
     @overload
+    @override
     def __getitem__(self, item: ScalarIndexer) -> Any: ...
     @overload
     def __getitem__(self, item: SequenceIndexer) -> Self: ...
+    @override
     def __iter__(self) -> Iterator[Any]: ...
```
Dr-Irv: Since this is the same declaration as the superclass, we can delete it.

pandas-stubs/core/arrays/masked.pyi
```diff
@@ -40,10 +44,13 @@ class BaseMaskedArray(ExtensionArray):
         self, dtype: NpDtype | None = None, copy: bool | None = None
     ) -> np_1darray: ...
     def __arrow_array__(self, type: DtypeArg | None = None) -> Any: ...
+    @override
     def copy(self) -> Self: ...
```
Dr-Irv: remove since inherits from superclass

pandas-stubs/core/arrays/masked.pyi
```diff
@@ -40,10 +44,13 @@ class BaseMaskedArray(ExtensionArray):
...
     def value_counts(self, dropna: bool = True) -> Series[int]: ...
+    @override
     def isna(self) -> np_1darray_bool: ...
```
Dr-Irv: remove since inherits from superclass

pandas-stubs/core/arrays/masked.pyi
```diff
@@ -40,10 +44,13 @@ class BaseMaskedArray(ExtensionArray):
...
     @property
+    @override
     def nbytes(self) -> int: ...
```
Dr-Irv: remove since inherits from superclass

pandas-stubs/core/arrays/string_.pyi
```diff
@@ -48,15 +51,18 @@ class StringDtype(ExtensionDtype, Generic[_StorageT]):
...
+    @override
     def dtype(self) -> StringDtype[_StorageT]: ...
 
 class StringArray(BaseStringArray[Literal["python"]], NumpyExtensionArray):
     def __init__(
         self, values: np_ndarray_object | Self, copy: bool = False
     ) -> None: ...
     def __arrow_array__(self, type: DtypeArg | None = None) -> pa.StringArray: ...
+    @override
     def __setitem__(self, key: Any, value: Any) -> None: ...
```
Dr-Irv: I think we can remove this.  Can inherit from `ExtensionArray` that has better typing anyway for the `key` argument.

pandas-stubs/core/arrays/timedeltas.pyi
```diff
@@ -108,14 +110,14 @@ class TimedeltaArray(TimelikeOps):
     ) -> np_1darray_int64: ...
     @overload
     def __rfloordiv__(self, other: Any) -> Self: ...
+    @override
     def __mod__(self, other: Any) -> Self: ...
+    @override
     def __rmod__(self, other: Any) -> Self: ...
```
Dr-Irv: remove since inherits from superclass

pandas-stubs/core/indexes/base.pyi
```diff
@@ -440,16 +441,20 @@ class Index(IndexOpsMixin[S1], ElementOpsMixin[S1]):
...
     def get_level_values(self, level: int | _str) -> Index: ...
     @property
+    @override
     def is_monotonic_increasing(self) -> bool: ...
     @property
+    @override
     def is_monotonic_decreasing(self) -> bool: ...
     @property
+    @override
     def is_unique(self) -> bool: ...
```
Dr-Irv: remove since inherits from superclass

pandas-stubs/core/indexes/base.pyi
```diff
@@ -440,16 +441,20 @@ class Index(IndexOpsMixin[S1], ElementOpsMixin[S1]):
...
     @property
     def has_duplicates(self) -> bool: ...
     @property
     def inferred_type(self) -> _str: ...
     @property
+    @override
     def hasnans(self) -> bool: ...
```
Dr-Irv: remove since inherits from superclass

pandas-stubs/core/indexes/category.pyi
```diff
@@ -31,13 +33,18 @@ class CategoricalIndex(ExtensionIndex[S1], PandasDelegate):
...
     def inferred_type(self) -> str: ...
     @property
+    @override
     def is_unique(self) -> bool: ...
     @property
+    @override
     def is_monotonic_increasing(self) -> bool: ...
     @property
+    @override
     def is_monotonic_decreasing(self) -> bool: ...
```
Dr-Irv: remove since inherits from superclass

pandas-stubs/core/indexes/datetimes.pyi
```diff
@@ -62,28 +63,31 @@ class DatetimeIndex(
...
     def __rtruediv__(self, other: np_ndarray) -> Never: ...  # type: ignore[override]
     @final
+    @override
     def to_series(
         self, index: Index | None = None, name: Hashable | None = None
     ) -> Series[Timestamp]: ...
     def snap(self, freq: Frequency = "S") -> Self: ...
     @property
+    @override
     def inferred_type(self) -> str: ...
```
Dr-Irv: remove since inherits from superclass

pandas-stubs/core/indexes/datetimes.pyi
```diff
@@ -98,12 +102,15 @@ class DatetimeIndex(
     def to_julian_date(self) -> Index[float]: ...
     def isocalendar(self) -> DataFrame: ...
     @property
+    @override
     def tzinfo(self) -> _tzinfo | None: ...
```
Dr-Irv: remove since inherits from superclass

pandas-stubs/core/indexes/interval.pyi
```diff
@@ -243,15 +244,22 @@ class IntervalIndex(ExtensionIndex[IntervalT, np.object_], IntervalMixin):
...
+    def __contains__(  # pyright: ignore[reportOverlappingOverload]
+        self, key: object
+    ) -> Literal[False]: ...
+    @override
+    def astype(self, dtype: DtypeArg, copy: bool = True) -> IntervalIndex: ...  # type: ignore[override] # pyright: ignore[reportIncompatibleMethodOverride] # pyrefly: ignore[bad-override] # ty: ignore[invalid-method-override]
     @property
+    @override
     def inferred_type(self) -> str: ...
+    @override
     def memory_usage(self, deep: bool = False) -> int: ...
```
Dr-Irv: remove since inherits from superclass

pandas-stubs/core/indexes/interval.pyi
```diff
@@ -243,15 +244,22 @@ class IntervalIndex(ExtensionIndex[IntervalT, np.object_], IntervalMixin):
...
     @property
     def is_overlapping(self) -> bool: ...
+    @override
     def get_loc(self, key: Label) -> int | slice | np_1darray_bool: ...
```
Dr-Irv: remove since inherits from superclass

pandas-stubs/core/indexes/multi.pyi
```diff
@@ -123,33 +124,45 @@ class MultiIndex(Index):
...
-    def view(self, cls: NumpyNotTimeDtypeArg | NumpyTimedeltaDtypeArg | NumpyTimestampDtypeArg | type[np_ndarray] | None = None) -> MultiIndex: ...  # type: ignore[override] # pyrefly: ignore[bad-override] # pyright: ignore[reportIncompatibleMethodOverride] # ty: ignore[invalid-method-override]
+    @override
+    def view(self, cls: NumpyNotTimeDtypeArg | NumpyTimedeltaDtypeArg | NumpyTimestampDtypeArg | type[np_ndarray] | None = None) -> MultiIndex: ...  # type: ignore[override] # pyright: ignore[reportIncompatibleMethodOverride] # pyrefly: ignore[bad-override] # ty: ignore[invalid-method-override]
     @property
+    @override
     def dtype(self) -> np.dtype: ...
     @property
     def dtypes(self) -> pd.Series[Dtype]: ...
+    @override
     def memory_usage(self, deep: bool = False) -> int: ...
```
Dr-Irv: remove since inherits from superclass

pandas-stubs/core/indexes/multi.pyi
```diff
@@ -123,33 +124,45 @@ class MultiIndex(Index):
...
     @property
+    @override
     def nbytes(self) -> int: ...
```
Dr-Irv: remove since inherits from superclass

pandas-stubs/core/indexes/multi.pyi
```diff
@@ -123,33 +124,45 @@ class MultiIndex(Index):
...
+    @override
     def __len__(self) -> int: ...
     @property
+    @override
     def is_monotonic_increasing(self) -> bool: ...
     @property
+    @override
     def is_monotonic_decreasing(self) -> bool: ...
```
Dr-Irv: remove since inherits from superclass

pandas-stubs/core/indexes/multi.pyi
```diff
@@ -123,33 +124,45 @@ class MultiIndex(Index):
...
+    @override
     def dropna(self, how: AnyAll = "any") -> Self: ...
```
Dr-Irv: remove since inherits from superclass

pandas-stubs/core/indexes/multi.pyi
```diff
@@ -159,10 +172,12 @@ class MultiIndex(Index):
     def to_flat_index(self) -> Index: ...
     def remove_unused_levels(self) -> MultiIndex: ...
     @property
+    @override
     def nlevels(self) -> int: ...
```
Dr-Irv: remove since inherits from superclass

pandas-stubs/core/indexes/multi.pyi
```diff
@@ -198,13 +215,14 @@ class MultiIndex(Index):
         self, before: IndexLabel | None = None, after: IndexLabel | None = None
     ) -> MultiIndex: ...
     @overload  # type: ignore[override]
-    def isin(  # pyrefly: ignore[bad-override]
-        self, values: Iterable[Any], level: Level
-    ) -> np_1darray_bool: ...
+    @override
+    # pyrefly: ignore[bad-override]
+    def isin(self, values: Iterable[Any], level: Level) -> np_1darray_bool: ...
```
Dr-Irv: the argument for `level` in the superclass allows `level` to be `None`, so that probably means this one is wrong and can be deleted.

pandas-stubs/core/indexes/multi.pyi
```diff
@@ -198,13 +215,14 @@ class MultiIndex(Index):
...
     @overload
     def isin(  # ty: ignore[invalid-method-override] # pyright: ignore[reportIncompatibleMethodOverride]
         self, values: Collection[Iterable[Any]], level: None = None
     ) -> np_1darray_bool: ...
+    @override
     def set_names(
```
Dr-Irv, line 207: remove since inherits from superclass

pandas-stubs/core/indexes/period.pyi
```diff
@@ -56,23 +61,28 @@ class PeriodIndex(DatetimeIndexOpsMixin[Period, np.object_], PeriodIndexFieldOps
...
     ) -> NaTType: ...
+    @override
     def asof_locs(
         self, where: pd.DatetimeIndex | Self, mask: np_ndarray_bool
     ) -> np_1darray_intp: ...
     @property
     def is_full(self) -> bool: ...
     @property
+    @override
     def inferred_type(self) -> str: ...
```
Dr-Irv: remove since inherits from superclass

pandas-stubs/core/indexes/range.pyi
```diff
@@ -46,26 +47,36 @@ class RangeIndex(IndexSubclassBase[int, np.int64]):
     @property
     def step(self) -> int: ...
     @property
+    @override
     def nbytes(self) -> int: ...
+    @override
     def memory_usage(self, deep: bool = ...) -> int: ...
```
Dr-Irv: remove since inherits from superclass

pandas-stubs/core/indexes/range.pyi
```diff
@@ -46,26 +47,36 @@ class RangeIndex(IndexSubclassBase[int, np.int64]):
...
     def is_unique(self) -> bool: ...
     @property
+    @override
     def is_monotonic_increasing(self) -> bool: ...
     @property
+    @override
     def is_monotonic_decreasing(self) -> bool: ...
     @property
+    @override
     def has_duplicates(self) -> bool: ...
```
Dr-Irv: remove since inherits from superclass

pandas-stubs/core/indexes/range.pyi
```diff
@@ -46,26 +47,36 @@ class RangeIndex(IndexSubclassBase[int, np.int64]):
...
+    @override
     def factorize(
         self, sort: bool = False, use_na_sentinel: bool = True
     ) -> tuple[np_1darray_intp, RangeIndex]: ...
     @property
+    @override
     def size(self) -> int: ...
```
Dr-Irv: remove since inherits from superclass

pandas-stubs/core/indexes/timedeltas.pyi
```diff
@@ -136,20 +146,24 @@ class TimedeltaIndex(
...
     @overload
+    @override
     def __rfloordiv__(self, other: np_ndarray_num | np_ndarray_dt) -> Never: ...
     @overload
     def __rfloordiv__(  # pyright: ignore[reportIncompatibleMethodOverride]
         self, other: _DT_FACTOR_SEQ | Self
     ) -> Index[int]: ...
     @property
+    @override
     def inferred_type(self) -> str: ...
```
Dr-Irv: remove since inherits from superclass

pandas-stubs/core/base.pyi
```diff
@@ -56,6 +57,7 @@ from pandas._typing import (
 T_INTERVAL_NP = TypeVar("T_INTERVAL_NP", bound=np.bytes_ | np.str_)
 
 class NoNewAttributesMixin:
+    @override
     def __setattr__(self, key: str, value: Any) -> None: ...
```
Dr-Irv: remove since inherits from superclass.  Might be able to just delete this class

pandas-stubs/core/frame.pyi
```diff
@@ -1824,6 +1830,7 @@ class DataFrame(NDFrame, OpsMixin, _GetItemHack):
     @property
     def dtypes(self) -> Series: ...
     @property
+    @override
     def empty(self) -> _bool: ...
```
Dr-Irv: remove since inherits from superclass

pandas-stubs/core/frame.pyi
```diff
@@ -1838,8 +1845,10 @@ class DataFrame(NDFrame, OpsMixin, _GetItemHack):
     @property
     def loc(self) -> _LocIndexerFrame[Self]: ...
     @property
+    @override
     def ndim(self) -> int: ...
```
Dr-Irv: remove since inherits from superclass

pandas-stubs/core/frame.pyi
```diff
@@ -1838,8 +1845,10 @@ class DataFrame(NDFrame, OpsMixin, _GetItemHack):
...
     @property
+    @override
     def size(self) -> int: ...
```
Dr-Irv: remove since inherits from superclass

pandas-stubs/core/series.pyi
```diff
@@ -662,6 +662,7 @@ class Series(IndexOpsMixin[S1], ElementOpsMixin[S1], NDFrame):
         copy: bool | None = None,
     ) -> Series: ...
     @property
+    @override
     def hasnans(self) -> bool: ...
```
Dr-Irv: remove since inherits from superclass

pandas-stubs/core/series.pyi
```diff
@@ -1115,6 +1118,7 @@ class Series(IndexOpsMixin[S1], ElementOpsMixin[S1], NDFrame):
     @overload
     def __rmatmul__(self, other: np_ndarray_num) -> np_ndarray_num: ...
     @overload
+    @override
     def searchsorted(
```
Dr-Irv: remove since inherits from superclass

pandas-stubs/core/series.pyi
```diff
@@ -4455,6 +4488,7 @@ class Series(IndexOpsMixin[S1], ElementOpsMixin[S1], NDFrame):
         axis: AxisIndex = ...,
     ) -> Series[_bool]: ...
     @final
+    @override
     def nunique(self, dropna: _bool = True) -> int: ...
```
Dr-Irv: remove since inherits from superclass

pandas-stubs/core/series.pyi
```diff
@@ -4595,8 +4629,10 @@ class Series(IndexOpsMixin[S1], ElementOpsMixin[S1], NDFrame):
         min_count: int = ...,
         **kwargs: Any,
     ) -> _T: ...
+    @override
     def to_list(self) -> list[S1]: ...
```
Dr-Irv: remove since inherits from superclass

pandas-stubs/core/series.pyi
```diff
@@ -4820,6 +4856,7 @@ class Series(IndexOpsMixin[S1], ElementOpsMixin[S1], NDFrame):
         na_value: Scalar = ...,
         **kwargs: Any,
     ) -> np_1darray: ...
+    @override
     def tolist(self) -> list[S1]: ...
```
Dr-Irv: remove since inherits from superclass

pandas-stubs/io/sas/sas7bdat.pyi
```diff
@@ -1,8 +1,12 @@
-from pandas import DataFrame
+from pandas.core.frame import DataFrame
+from typing_extensions import override
 
 from pandas.io.sas.sasreader import SASReader
 
 class SAS7BDATReader(SASReader):
+    @override
     def close(self) -> None: ...
```
Dr-Irv, line 8: remove since inherits from superclass

pandas-stubs/io/sas/sas7bdat.pyi
```diff
@@ -1,8 +1,12 @@
...
+    @override
     def __next__(self) -> DataFrame: ...
+    @override
     def read(self, nrows: int | None = None) -> DataFrame: ...
```
Dr-Irv, line 12: remove since inherits from superclasses

pandas-stubs/io/sas/sas_xport.pyi
```diff
@@ -1,8 +1,12 @@
...
 from pandas.io.sas.sasreader import SASReader
 
 class XportReader(SASReader):
+    @override
     def close(self) -> None: ...
-    def __next__(self) -> pd.DataFrame: ...
-    def read(self, nrows: int | None = None) -> pd.DataFrame: ...
+    @override
+    def __next__(self) -> DataFrame: ...
+    @override
```
Dr-Irv, line 11: remove since inherits from superclasses

pandas-stubs/io/stata.pyi
```diff
@@ -103,6 +104,7 @@ class StataReader(StataParser, Iterator[Any]):
         exc_value: BaseException | None,
         traceback: TracebackType | None,
     ) -> None: ...
+    @override
     def __next__(self) -> DataFrame: ...
```
Dr-Irv, line 107: I think if the class is `Iterator[DataFrame]` rather than `Iterator[Any]`, you can then remove this

pandas-stubs/util/version/__init__.pyi
```diff
@@ -1,11 +1,15 @@
+from typing_extensions import override
+
 def parse(version: str) -> LegacyVersion | Version: ...
 
 class _BaseVersion:
     def __lt__(self, other: _BaseVersion) -> bool: ...
     def __le__(self, other: _BaseVersion) -> bool: ...
+    @override
     def __eq__(self, other: object) -> bool: ...
```
Dr-Irv: remove since inherits from superclass

pandas-stubs/util/version/__init__.pyi
```diff
@@ -1,11 +1,15 @@
...
     def __ge__(self, other: _BaseVersion) -> bool: ...
     def __gt__(self, other: _BaseVersion) -> bool: ...
+    @override
     def __ne__(self, other: object) -> bool: ...
```
Dr-Irv: remove since inherits from superclass

Dr-Irv (changes requested):
I have reviewed all the PYI files, github workflows, docs, but not what is in `tests`.  

Main thing that I see is that we have overrides/method declarations that are unnecessary because the superclass defines the method exactly the same way, so we should remove the declaration from the subclass.

I don't know if I caught all of them.

I'll review the test functions later.

pandas-stubs/_libs/missing.pyi
cmp0xff: Thank you for the review! Please allow me to address the comments one by one. 1ec5b3d310c0971246456fec55cef38f28b2465d

- `__format__`: Seems the default one is `__format__(self, format_spec: str, /) -> str`, only supports one positional argument `format_spec`, no keyword argument is allowed. Override is necessary. Tests added.
- `__hash__`: Override is necessary, https://github.com/pandas-dev/pandas-stubs/pull/746#discussion_r1266975292
- `__reduce__`: I can tighten it.

Tests added.

cmp0xff:
Well, pyright complains. But when I use pyright to check only the files with errors, it does not complain any more 😕 

> poetry run pyright tests/series/timedelta/test_mul.py
> 0 errors, 0 warnings, 0 informations

pandas-stubs/_libs/missing.pyi
Dr-Irv: With respect to `__format__()`, I think this is an issue of the implementation in `pandas` being in CPython which does allow a keyword argument, which one could argue is incorrect.  For the stubs, we should just be supporting the default `object` implementation that requires a positional argument.

cmp0xff: d03e3c6f75411a91aa3543f8111cb08868128bac

pandas-stubs/core/arrays/categorical.pyi
cmp0xff: Not really, if deleted, `__eq__` would give a `bool`, which is not true. 81fe691673ce8bd0c6a7683cae88f604b0862a65 pandas-dev/pandas-stubs#1901

cmp0xff (commented):
https://github.com/pandas-dev/pandas-stubs/pull/1867#issuecomment-5300925096: microsoft/pyright#11644

pandas-stubs/core/indexes/multi.pyi
cmp0xff: This is from pandas-dev/pandas-stubs#1507 with tests.

pandas-stubs/core/indexes/multi.pyi
cmp0xff: This is from #1486 with tests, different from the superclass

pandas-stubs/io/sas/sas_xport.pyi
cmp0xff: - `read`, `close`: the methods in superclass are abstract, I'd keep them. Also reported astral-sh/ty#4307
- `__next__`: it is abstract in the base class, still needs an override

pandas-stubs/io/stata.pyi
cmp0xff: It is abstract in the base class `Iterator`, still needs an override

pandas-stubs/io/sas/sas7bdat.pyi
cmp0xff: base class is abstract, we need it

pandas-stubs/io/sas/sas7bdat.pyi
cmp0xff: - `read`, `close`: the methods in superclass are abstract, I'd keep them. Also reported astral-sh/ty#4307
- `__next__`: it is abstract in the base class, still needs an override

pandas-stubs/core/arrays/interval.pyi
```diff
@@ -104,9 +104,6 @@ class IntervalArray(IntervalMixin, ExtensionArray):
     @override
     def dtype(self) -> IntervalDtype: ...
     @property
-    @override
-    def nbytes(self) -> int: ...
-    @property
     def size(self) -> int: ...
```
Dr-Irv: I think you can remove `size` here

pandas-stubs/core/indexes/multi.pyi
Dr-Irv: Yes, but the code defines `isin` allowing `None` as the argument for `level` here:
https://github.com/pandas-dev/pandas/blob/4e1b1f77dbc700e505c808e8b6bc164abce87955/pandas/core/indexes/multi.py#L4884

So I think we can delete this override and the tests should still work.

pandas-stubs/core/indexes/multi.pyi
Dr-Irv: OK - the implementation is just in `Index`, but tests for `MultiIndex` being required if `Mapping` is the argument

tests/frame/test_frame.py
```diff
@@ -301,7 +302,7 @@ def test_assign() -> None:
     )
 
     def my_named_func_1(df: pd.DataFrame) -> pd.Series[str]:
-        return df["a"]
+        return cast("pd.Series[str]", df["a"])
```
Dr-Irv: can we add a comment here why we're OK with the cast in the test?

tests/frame/test_groupby.py
```diff
@@ -628,7 +629,7 @@ def test_groupby_apply() -> None:
     df_gb = df.groupby("col1")
 
     def sum_mean(x: pd.DataFrame) -> float:
-        return x.sum().mean()
+        return cast("pd.Series[float] | pd.Series[int]", x.sum()).mean()
```
Dr-Irv: Not a fan of the `cast` here.  Why should a user be required to do the cast to get this to work?

tests/test_groupby.py
```diff
@@ -350,7 +350,7 @@ def test_series_groupby_resample() -> None:
         DataFrame,
     )
 
-    def f(val: Series) -> float:
+    def f(val: Series[int] | Series[float]) -> float:
```
Dr-Irv: the point of this test was to correspond to code a user might write, which would not specify the generic type of the `Series`.  Can we get this to work without this change?

tests/test_groupby.py
```diff
@@ -384,7 +384,7 @@ def h(val: Series) -> Series:
     def s2series(val: Series) -> Series:
         return Series(val)
 
-    def s2scalar(val: Series) -> float:
+    def s2scalar(val: Series[int] | Series[float]) -> float:
```
Dr-Irv: same comment as above

tests/test_groupby.py
Dr-Irv, line 1: Multiple places here (I didn't comment on all of them), but we want functions declared as:
```python
def foo(val: Series) -> float:
```
to be acceptable usage.  We can't force people to use the generic declarations like `Series[int]` and `Series[float]`.

Is there a way to keep the tests as they were?

tests/test_pandas.py
Dr-Irv, line 1: same comment here as in `test_groupby.py` about the function declarations

tests/test_resampler.py
Dr-Irv, line 1: `DatetimeIndexResampler` is not documented, so I don't think we should use it in tests.  Also, have functions with `Series[int]` in declaration, but we can't expect users to do that.

tests/test_windowing.py
Dr-Irv, line 1: same comment about having functions with `Series[int] | Series[float]` as argument types

Dr-Irv (changes requested):
did go through the tests, so those are reviewed.

There is one open comment from previous review - others were all resolved.

pandas-stubs/core/indexes/multi.pyi
cmp0xff: https://github.com/pandas-dev/pandas-stubs/blob/b2a9b49b7afe142bb6aaf2cf71d1d759be6054fe/tests/indexes/test_indexes.py#L1634-L1637 I think this should be sufficient?

tests/frame/test_groupby.py
cmp0xff: `ty` reveals `x.sum()` as `Series[Any]`, and `Series[Any]().mean()` as `Unknown`. What do you think? We can make `Series[Any]().mean()` giving `Any`.

tests/test_groupby.py
cmp0xff: `ty` reveals `Series[Any]().mean()` as `Unknown`. What do you think? We can make `Series[Any]().mean()` giving `Any`.

tests/test_resampler.py
cmp0xff: Based on the `assert`s, I think `DatetimeIndexResampler` is indeed needed for typing purposes 😕  pandas-dev/pandas-stubs#848

tests/frame/test_groupby.py
Dr-Irv: At least for `pyright`, `x.sum().mean()` is revealed as `float` there.  So check the other type checkers, as this might be a bug in `ty`

tests/test_groupby.py
Dr-Irv: and `pyright` reveals `Series[Any].mean()` as `float` .  So see what the other type checkers do.  `ty` may be incorrect here.

tests/test_resampler.py
Dr-Irv: I think those `assert` should be removed.

But I also think that in `pandas`, `DatetimeIndexResampler`, `PeriodIndexResampler` and `TimedeltaIndexResampler` should be added to the `pandas.api.typing` module since `pandas.api.typing.DataFrameGroupBy.resample` can return those types.

Dr-Irv: Created https://github.com/pandas-dev/pandas/issues/66850

tests/frame/test_frame.py
cmp0xff: `ty` thinks `Series[str]` is not a subtype of `Series[Any]`. Do you think this is true? https://play.ty.dev/cf48fa3d-d784-4eed-adf1-458fc64865a6

tests/frame/test_groupby.py
cmp0xff: For all type checkers, `x.sum()` is revealed as `Series[Any]`. The next step might be a `ty` issue. astral-sh/ty#4360 astral-sh/ty#4135

I added an inline comment.

tests/test_groupby.py
cmp0xff: I think this is the same as in another comment. astral-sh/ty#4360 astral-sh/ty#4135. I added an inline comment.

tests/test_groupby.py
cmp0xff: I think this is the same as in another comment. astral-sh/ty#4360 astral-sh/ty#4135. I added an inline comment.

tests/test_groupby.py
cmp0xff: Reported in astral-sh/ty#4360 (might be the same as astral-sh/ty#4135).

tests/test_groupby.py
```diff
@@ -868,8 +868,8 @@ def test_series_groupby_ewm() -> None:
     if TYPE_CHECKING_INVALID_USAGE:
         GB_DF.ewm(1).agg(np.mean)  # type: ignore[arg-type] # pyright: ignore[reportArgumentType] # pyrefly: ignore[no-matching-overload] # ty: ignore[no-matching-overload]
 
-        def _func(x: Series) -> float:
-            return sum(x)  # type: ignore[no-any-return]
+        def _func(x: Series[int] | Series[float]) -> float:
+            return sum(x)
```
cmp0xff: @Dr-Irv this one is special. When `x` is merely `Series`, even `mypy` will complain `error: Returning Any from function declared to return "float"  [no-any-return]` (now I've turned on `no-any-return` in `mypy`). `sum(Iterable[Any])` is indeed `Any`, and if we do stricter type checking, this is not allowed (`no-any-return` by `mypy`, `unsound-return-statement` by `ty`).

tests/test_pandas.py
cmp0xff: I added inline comments.

tests/test_windowing.py
cmp0xff: I added inline comments.

tests/test_resampler.py
cmp0xff: `assert`s removed.

pandas-stubs/core/indexes/multi.pyi
```diff
@@ -195,14 +195,8 @@ class MultiIndex(Index):
...
     ) -> MultiIndex: ...
-    @overload  # type: ignore[override]
     @override
-    # pyrefly: ignore[bad-override]
-    def isin(self, values: Iterable[Any], level: Level) -> np_1darray_bool: ...
-    @overload
-    def isin(  # ty: ignore[invalid-method-override] # pyright: ignore[reportIncompatibleMethodOverride]
-        self, values: Collection[Iterable[Any]], level: None = None
-    ) -> np_1darray_bool: ...
+    def isin(self, values: Collection[Iterable[Any]], level: None = None) -> np_1darray_bool: ...  # type: ignore[override] # pyright: ignore[reportIncompatibleMethodOverride] # pyrefly: ignore[bad-override] # ty: ignore[invalid-method-override]
```
Dr-Irv: All 4 type checkers are requiring some form of `ignore` here.   So if you remove the `@override` designation, does that reduce the number of type checkers needing an `ignore`?  Or does it just change the specific `ignore` that is needed?

Would prefer less `ignores` if possible.

pandas-stubs/core/indexes/multi.pyi
Dr-Irv: yes

tests/frame/test_frame.py
Dr-Irv: If the other type checkers are OK without the `cast`, then use a `#ty: ignore` and document that.

tests/frame/test_groupby.py
Dr-Irv: If it is a `ty` issue, then don't change the test, but put in an `ignore` in the test for `ty` with the comment referring to the `ty` issue.

tests/test_groupby.py
Dr-Irv: Again, I'd rather leave the test to be what typical code would look like (not using the generic declarations) and if `ty` can't handle it, then put in an `ignore` for `ty` in the tests and an appropriate comment.

tests/test_groupby.py
Dr-Irv: Again, I'd rather leave the test to be what typical code would look like (not using the generic declarations) and if ty can't handle it, then put in an ignore for ty in the tests and an appropriate comment.

tests/test_groupby.py
Dr-Irv: I'd rather leave the test to be what typical code would look like (not using the generic declarations) and if ty can't handle it, then put in an ignore for ty in the tests and an appropriate comment.

tests/test_pandas.py
Dr-Irv: Again, I'd rather leave the test to be what typical code would look like (not using the generic declarations) and if ty can't handle it, then put in an ignore for ty in the tests and an appropriate comment.

tests/test_resampler.py
Dr-Irv: Keeping conversation open as I'd rather leave the test to be what typical code would look like (not using the generic declarations) and if ty can't handle it, then put in an ignore for ty in the tests and an appropriate comment.

tests/test_windowing.py
Dr-Irv: Again, I'd rather leave the test to be what typical code would look like (not using the generic declarations) and if ty can't handle it, then put in an ignore for ty in the tests and an appropriate comment.

Dr-Irv (changes requested):
So the main open issue is using things like `x: pd.Series[int] | pd.Series[float]` in function declarations in the tests.  We want to have the tests where we avoid using the generic declarations.  It seems it's only needed for `ty`, so use `ty: ignore` instead, and keep the comment about the relevant `ty` issue.

Basic idea here is that if only one type checker can't handle a test, then we do an `ignore` for that type checker.  We've had to do that for `mypy` in a few places.  Don't change the test to appease the type checker.

I've kept those conversations open so I can remember to look at them again.

tests/frame/test_frame.py
cmp0xff: c50866d9f3b694a11e049c1c866a98bb54790977

tests/frame/test_groupby.py
cmp0xff: c50866d9f3b694a11e049c1c866a98bb54790977

tests/test_groupby.py
cmp0xff: 055152f913316d4a194e4c8ab7e9db75577e66e7

tests/test_groupby.py
cmp0xff: 055152f913316d4a194e4c8ab7e9db75577e66e7

tests/test_groupby.py
cmp0xff: 055152f913316d4a194e4c8ab7e9db75577e66e7

tests/test_pandas.py
cmp0xff: 055152f913316d4a194e4c8ab7e9db75577e66e7

tests/test_resampler.py
cmp0xff: Sorry, I'm not sure what are the proposals here?

tests/test_windowing.py
cmp0xff: 055152f913316d4a194e4c8ab7e9db75577e66e7

pandas-stubs/core/indexes/multi.pyi
cmp0xff: This `isin` is different from the one in `Index`, "in a bad way", as the type checkers say, even thought they know it is an `override`. Even before this change, there were three typing ignores. I guess removing the first overload is the last straw on mypy's back.

tests/test_groupby.py
cmp0xff: @Dr-Irv I've added an additional overload for `Series[Never].sum()` giving `float`.

tests/test_groupby.py
```diff
@@ -1152,3 +1128,8 @@ def test_dataframe_groupby_dtypes() -> None:
 
         def _0() -> None:  # pyright: ignore[reportUnusedFunction]
             assert_type(GB_DF.dtypes, Never)
+
+
+def s2scalar(val: Series) -> float:
+    # TODO: remove ty ignore astral-sh/ty#4360 astral-sh/ty#4135
+    return val.mean()  # ty: ignore[unsound-return-statement]
```
Dr-Irv: can we move this function to the top of the file - makes it easier to see the reference

tests/test_groupby.py
```diff
@@ -868,8 +844,8 @@ def test_series_groupby_ewm() -> None:
     if TYPE_CHECKING_INVALID_USAGE:
         GB_DF.ewm(1).agg(np.mean)  # type: ignore[arg-type] # pyright: ignore[reportArgumentType] # pyrefly: ignore[no-matching-overload] # ty: ignore[no-matching-overload]
 
-        def _func(x: Series) -> float:
-            return sum(x)  # type: ignore[no-any-return]
+        def _func(x: Iterable[float]) -> float:
+            return sum(x)
```
Dr-Irv: concerned that we've changed the meaning of this test by changing the arg from `Series` to `Iterable[float]`.  So can  you put back the original function definition?  And if you want to test `Iterable[float]`, that's OK to do too

tests/test_resampler.py
```diff
@@ -420,3 +408,8 @@ def test_getitem() -> None:
         assert_type(DF.resample("ME")[["col1", "col2"]], DataFrameGroupBy),
         DataFrameGroupBy,
     )
+
+
+def s2scalar(val: Series) -> float:
+    # TODO: remove ty ignore astral-sh/ty#4360 astral-sh/ty#4135
+    return val.mean()  # ty: ignore[unsound-return-statement]
```
Dr-Irv: move to top of file

tests/test_resampler.py
Dr-Irv: it's ok - you addressed it

tests/test_windowing.py
```diff
@@ -484,3 +476,8 @@ def test_groupby_kurt() -> None:
         assert_type(S.groupby(S).agg("kurt"), Series),
         Series,
     )
+
+
+def _mean(s: Series) -> float:
+    # TODO: remove ty ignore astral-sh/ty#4360 astral-sh/ty#4135
+    return s.mean()  # ty: ignore[unsound-return-statement]
```
Dr-Irv: move to top of file

pandas-stubs/core/indexes/multi.pyi
Dr-Irv: Ugh.  So looking at the code, I think  your declaration is incorrect.  

Things like this should work:
```python
mi = pd.MultiIndex.from_arrays( [[1, 2, 3], ["red", "blue", "green"]], names=["number", "color"])
mi.isin(["red", "orange", "yellow"], level="color")
mi.isin([(1, "red"), (3, "red")])
```
In addition, the `values` argument could be a `MultiIndex`, or a `Generator`

So can you fix this and add appropriate tests.

Dr-Irv (changes requested):

tests/test_groupby.py
cmp0xff: So what we want to do is test typing. The pain point here is really `sum(x)` when x is merely `Series[Any]`. Would it be fine to use `s2scalar` instead of `_func`?

cmp0xff:
OK, now I see there is one more comment, please hold on

cmp0xff:
OK, now all should have been addressed.

tests/indexes/test_indexes.py
```diff
@@ -77,17 +77,30 @@ def test_index_isin() -> None:
...
+    check(assert_type(mi.isin([[1, "red"]]), np_1darray_bool), np_1darray_bool)
+    check(
+        assert_type(
+            mi.isin(["red", "orange", "yellow"], level="color"), np_1darray_bool
+        ),
+        np_1darray_bool,
+    )
+    check(
+        assert_type(mi.isin([(1, "red"), (3, "red")]), np_1darray_bool), np_1darray_bool
+    )
```
Dr-Irv, line 105: Need to add a test for `mi.isin(mi)`, i.e. a `MultiIndex` can be passed as a parameter.

tests/indexes/test_indexes.py
```diff
@@ -77,17 +77,30 @@ def test_index_isin() -> None:
...
-    mi = pd.MultiIndex.from_arrays([[1, 2, 3]])
-    check(assert_type(mi.isin([[3]]), np_1darray_bool), np_1darray_bool)
+    mi = pd.MultiIndex.from_arrays(
+        [[1, 2, 3], ["red", "blue", "green"]], names=["number", "color"]
+    )
+    check(assert_type(mi.isin([[1, "red"]]), np_1darray_bool), np_1darray_bool)
+    check(
+        assert_type(
+            mi.isin(["red", "orange", "yellow"], level="color"), np_1darray_bool
+        ),
```
Dr-Irv, line 95: The only reason this works is because `Collection[Iterable[Hashable]]` matches a list of strings, because the strings are `Iterable[Hashable]`.  But if you did `mi.isin([1, 3,5], level="number")`, I think it wouldn't work.

So you have to add tests and update `isin`.  

One option here is to do a separate PR for the `isin` changes, I'll review/approve that, and then get it working here.  But I'm fine either way.

tests/test_groupby.py
Dr-Irv: OK

Dr-Irv (commented):
On `MultiIndex.isin()` is left to get right

tests/indexes/test_indexes.py
cmp0xff: From my perspective, `isin` is good now.

pandas-stubs/core/indexes/multi.pyi
```diff
@@ -195,8 +195,14 @@ class MultiIndex(Index):
     def truncate(
         self, before: IndexLabel | None = None, after: IndexLabel | None = None
     ) -> MultiIndex: ...
-    @override
-    def isin(self, values: Collection[Iterable[Hashable]], level: Level | None = None) -> np_1darray_bool: ...  # type: ignore[override] # pyright: ignore[reportIncompatibleMethodOverride] # pyrefly: ignore[bad-override] # ty: ignore[invalid-method-override]
+    @override  # type: ignore[override]
+    @overload
+    # pyrefly: ignore[bad-override]
+    def isin(self, values: Iterable[Any], level: Level) -> np_1darray_bool: ...
```
Dr-Irv, line 201: Is `Iterable[Any]` too wide?  Maybe `Iterable[Hashable]` ?

Dr-Irv (changes requested):
only one question

tests/indexes/test_indexes.py
```diff
@@ -77,17 +77,62 @@ def test_index_isin() -> None:
...
+    )
+    check(
+        assert_type(
+            mi.isin({iter((1, "red")): 0, (3, "red"): 0}.keys()), np_1darray_bool
+        ),
+        np_1darray_bool,
+    )
+    check(assert_type(mi.isin(mi), np_1darray_bool), np_1darray_bool)
+    check(assert_type(mi.isin(iter(["1"]), 1), np_1darray_bool), np_1darray_bool)
+    check(assert_type(mi.isin(iter([["1"]]), 1), np_1darray_bool), np_1darray_bool)
```
cmp0xff, line 121: https://github.com/pandas-dev/pandas-stubs/pull/1867/changes#r3848365216

> Is `Iterable[Any]` too wide? Maybe `Iterable[Hashable]` ?

@Dr-Irv it actually works at runtime for Iterable[non-hashable] for some reason. I don't want to forbid users from doing it without a good reason.

pandas-stubs/core/indexes/multi.pyi
cmp0xff: https://github.com/pandas-dev/pandas-stubs/pull/1867/changes#r3852620715

tests/indexes/test_indexes.py
cmp0xff: This is also added.

tests/indexes/test_indexes.py
Dr-Irv: OK, probably an improvement to make in pandas, but not something to worry about.

Dr-Irv (approved):
thanks @cmp0xff

PR #1869, opened 2026-08-03
TYP: #1867 `Categorical`

cmp0xff:
- [x] Towards #1867 which is gigantic, reducing `pyrefly` ignores around `Categorical`
- [x] Tests added

### DIFF
--- a/pandas-stubs/core/arrays/categorical.pyi
+++ b/pandas-stubs/core/arrays/categorical.pyi
@@ -36,9 +36,9 @@ from pandas._typing import (
     SequenceNotStr,
     np_1darray,
     np_1darray_bool,
-    np_1darray_str,
     np_ndarray_anyint,
     np_ndarray_float,
+    np_ndarray_str,
 )
 
 from pandas.core.dtypes.dtypes import (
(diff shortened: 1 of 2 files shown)


pandas-stubs/core/arrays/categorical.pyi
```diff
@@ -62,10 +62,10 @@ class Categorical(NDArrayBackedExtensionArray, Generic[CategoricalValueT]):
     def __new__(  # pyright: ignore[reportOverlappingOverload]
         cls,
         values: (
-            list[str] | np_1darray_str | SequenceNotStr[str] | Series[str] | Index[str]
+            list[str] | np_ndarray_str | SequenceNotStr[str] | Series[str] | Index[str]
```
loicdiridollou, line 65: I am not sure that categorical accepts a 2-day array, is there a test for this one?

loicdiridollou (changes requested):
One minor question, the rest looks good.

pandas-stubs/core/arrays/categorical.pyi
cmp0xff: It does not. The problem is np.array(["1"], dtype=str) is typed as an array with indefinite dimensions. The principle here should be to be more tolerant on input, and be strict for our own output.

loicdiridollou: Okay that makes sense!

loicdiridollou (approved):

PR #1870, opened 2026-08-03
CLN: #1867 `Styler`

cmp0xff:
- [x] Towards #1867 which is gigantic. The original `Protocol`s confuse `pyrefly` in its `strict` mode. These could be replaced with `Callable`s with modern Python.


loicdiridollou (commented):
Nice clean up, thanks @cmp0xff

PR #1878, opened 2026-08-06
BUG: relocate `to_offset` and other `pyrefly`-inspired changes

cmp0xff:
`to_offset` has been moved from `pandas-stubs/tseries/frequencies.pyi` to `pandas-stubs/_libs/tslibs/offsets.pyi`.

- [x] Towards #1867 which is gigantic

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -8,7 +8,7 @@ repos:
     -   id: ruff-check
         args: [--exit-non-zero-on-fix]
 -   repo: https://github.com/codespell-project/codespell
-    rev: v2.4.2
+    rev: v2.4.3
     hooks:
     - id: codespell
       args: [-L, "THIRDPARTY"]
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -3,7 +3,7 @@ ci:
     autofix_prs: false
 repos:
 -   repo: https://github.com/astral-sh/ruff-pre-commit
-    rev: v0.15.22
+    rev: v0.16.1
     hooks:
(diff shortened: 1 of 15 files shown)


.pre-commit-config.yaml
loicdiridollou: While we are here maybe 0.16.1

loicdiridollou (approved):
Thanks @cmp0xff

tests/frame/test_frame.py
```diff
@@ -4572,12 +4575,14 @@ def func_e(x: pd.DataFrame, k: int) -> pd.Series:
         return x.max() - k * x.min()
 
     check(
+        # pyrefly: ignore[implicity-any-lambda]
```
Dr-Irv, line 4578: IMHO, I think we should always ignore this from `pyrefly`, because the pattern of doing `lambda`s this way is common.

cmp0xff: Will do in #1867 or so

cmp0xff: What about pyright [reportUnknownLambdaType](https://github.com/microsoft/pylance-release/blob/main/docs/diagnostics/reportUnknownLambdaType.md)?

Dr-Irv: same - I don't think we should be expecting our tests to handle that, because it is pretty natural to NOT add types to lambda functions.

My philosophy on the tests is that we want to have in our tests code that people are likely to write.  Trying to make that fully compliant with any/all type checkers with some of the more strict errors puts a burden on people using the code.   We can make our own choices as to which error messages we are willing to support.

PR #1879, opened 2026-08-06
CLN: #1878 disable `pyright` `reportUnknownLambdaType` and `pyrefly` `implicity-any-lambda`

cmp0xff:
The idea comes from https://github.com/pandas-dev/pandas-stubs/pull/1878#discussion_r3729661783.

- [x] Addresses #1878, towards #1867 which is gigantic


pandas-stubs/core/dtypes/missing.pyi
```diff
@@ -23,8 +24,8 @@ from pandas._typing import (
     np_ndarray_bool,
 )
 
-isposinf_scalar = ...
-isneginf_scalar = ...
+isposinf_scalar: Callable[[], Any] = ...
+isneginf_scalar: Callable[[], Any] = ...
```
Dr-Irv: This should be declared as a `TypeAlias`

tests/series/test_series.py
```diff
@@ -3579,7 +3569,7 @@ def test_apply_dateoffset() -> None:
     check(
         assert_type(
             s.apply(
-                lambda x: pd.DateOffset(  # pyright: ignore[reportUnknownArgumentType,reportUnknownLambdaType]
+                lambda x: pd.DateOffset(  # pyright: ignore[reportUnknownArgumentType]
                     months=x  # pyright: ignore[reportUnknownArgumentType]
```
Dr-Irv, line 3573: If the `reportUnknownArgumentType` is only occurring when we are doing stuff with `lambda`, we should not worry about that one as well.

If it is occurring in other places, it may be worth a look to see if we should do something different.

Dr-Irv (changes requested):

Dr-Irv:
Not sure why it didn't run CI??

cmp0xff:
Thank you for the review @Dr-Irv ! GitHub has issues so CI does not run at the moment, see https://www.githubstatus.com/.

loicdiridollou:
Prob you can repush now to trigger the jobs on your PR

pandas-stubs/core/dtypes/missing.pyi
cmp0xff: https://github.com/pandas-dev/pandas/blob/d0d07d18f9fe855529997e3fe16cf1d0c8ce5eb8/pandas/_libs/missing.pyi#L17-L18

But I don't find them in the pandas docs, they should be private. Removing in 04c8c5bb796dcc8a6a4833f1fd4805abfb38692b.

tests/series/test_series.py
cmp0xff: They are not _all_ related to `lambda`. Creating #1880.

Dr-Irv:
Coincidentally, the typing people published a guide about typing with `lambda`:   https://typing.python.org/en/latest/guides/lambda.html

You might find that interesting

Dr-Irv (approved):
thanks @cmp0xff

PR #1819, opened 2026-07-13
type `JsonReader.__init__`, `AbstractHolidayCalendar` attributes, `Holiday` attributes

MarcoGorelli:
closes #1814 

I've:
- Typed `JsonReader.__init__`, and kept it aligned with `read_json`
  - while there, I noticed that `dtype` was using `HashableT`, even though it only appears once in the signature, so I just switched that to `Hashable`
- Missing attributes in `Holiday`
- Missing attribute `name` in `AbstractHolidayCalendar`

Related pandas PR to make sure these appear in the api docs: https://github.com/pandas-dev/pandas/pull/66267

- [ ] Closes #xxxx (Replace xxxx with the Github issue number)
- [ ] Tests added (Please use `assert_type()` to assert the type of any return value)
- [ ] If I used AI to develop this pull request, I prompted it to follow `AGENTS.md`.

PR #1883, opened 2026-08-08
GH1795 Add pyrefly coverage step

### LINKED ISSUE #1795: test_completeness checks are timing out
Not totally sure what to do on this one, it seemed like it was working fine a few days ago but in the interim we increase the timeout currently set at 5 to 10 and see what is going on. I was able to run it locally and it took around 3 min.
@MarcoGorelli I believe you worked on the type checking action so if that is the case do you have any insights on it?

loicdiridollou:
- [x] Closes #1795

Adding pyrefly coverage, needed a few tweaks for the data at the top of test files but this is now looking quite good.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -45,6 +45,9 @@ jobs:
       - name: Run pyrefly on 'tests' (using the local stubs) and on the local stubs
         run: poetry run poe pyrefly
 
+      - name: Run pyrefly's type coverage check on 'tests'
+        run: poetry run poe pyrefly_coverage
+
       - name: Run pyright on 'tests' (using the local stubs) and on the local stubs
         run: poetry run poe pyright
 
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -85,7 +85,7 @@ help = "Run all tests"
 script = "scripts.test:run_tests(src=True, dist=True)"
 
 [tool.poe.tasks.test]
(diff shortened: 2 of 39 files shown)


loicdiridollou:
@MarcoGorelli 

Tried to add pyrefly coverage to the stubs, interesting failure on the 3.11 versions, let me know if you have some background, yes it is much faster but is it worth the hassle to maintain, not totally convinced.

cmp0xff:
- It seems worthy reporting to ty
- You can try `if sys.python_version: ...`
- Alternatively you can use `# ty: ignore[invalid-assignment,unused-ignore-comment,unused-ignore-comment]` to bypass

MarcoGorelli:
thanks for the ping! i was looking at this recently, i think there may a simpler way to do it, will update you

MarcoGorelli:
I've pushed to https://github.com/pandas-dev/pandas-stubs/compare/main...MarcoGorelli:pyrefly-coverage-check?expand=1

It bumps pyrefly to 1.2.0, so that the `tool.pyrefly.coverage` section in `pyproject.toml` can be used. I've excluded tests and am using the `--public-only` flag, so that the rest can all stay the same

Note that #1819 added missing annotations brought up by pyrefly coverage check, and that's already  been merged, so that's why nothing else needs changing now

If you agree, do you want to cherry-pick from my branch and update this PR?

MarcoGorelli:
should we close this now, as it's been incorporated into #1910 ?

loicdiridollou:
Absolutely we don't need this one anymore

PR #1885, opened 2026-08-11
TYP: Allow list as other in DataFrame.dot

### LINKED ISSUE #1862: Argument of type "list[int]" cannot be assigned to parameter "other" of type "Series[Any]" in function "dot"
```python
import pandas as pd

time = pd.DataFrame({"HH": [1, 2, 3], "MM": [4, 5, 6]})
seconds = time.dot([3600, 60])
```

Khan3K:
## Summary

Allow a plain `list` to be passed as `other` to `DataFrame.dot`, fixing the type error reported in #1862.

`DataFrame.dot` accepts a Series, DataFrame or array-like per the pandas docs, but the stub only accepted `ArrayLike` (ndarray/ExtensionArray), `DataFrame`, and `Series`. `list` is now accepted too, consistent with the docs and with `Series.dot`, which already accepts `Sequence`.

```python
time = pd.DataFrame({"HH": [1, 2, 3], "MM": [4, 5, 6]})
seconds = time.dot([3600, 60])  # previously: Argument of type "list[int]" cannot be assigned
```

- [x] closes #1862
- [x] tests added / updated
- [x] passes `poe test` (mypy, pyright, pyrefly, ty, pytest, style) locally for the touched files' scope

## Notes
The `other: DataFrame | ArrayLike` overload returns `Self`; a `list` is treated the same way (a 2-D list returns a `DataFrame` at runtime). This matches the existing ndarray convention.


pandas-stubs/core/frame.pyi
```diff
@@ -416,7 +416,7 @@ class DataFrame(NDFrame, OpsMixin, _GetItemHack):
     ) -> Iterator[tuple[Any, ...]]: ...
     def __len__(self) -> int: ...
     @overload
-    def dot(self, other: DataFrame | ArrayLike) -> Self: ...
+    def dot(self, other: DataFrame | ArrayLike | list[Any]) -> Self: ...
```
cmp0xff: Likely better:
```suggestion
    def dot(self, other: DataFrame | ArrayLike | Sequence[_SupportsMul[Any]] | Sequence[Sequence[_SupportsMul[Any]]]) -> Self: ...
```
`_SupportsMul` comes from `pandas-stubs/core/series.pyi`

Probably rename it so it does not start with `_`

cmp0xff (changes requested):
What about ```If I used AI to develop this pull request, I prompted it to follow `AGENTS.md`.```? This is in the default PR description but is missing here.

tests/frame/test_frame.py
```diff
@@ -2802,6 +2802,8 @@ def test_types_dot() -> None:
     check(assert_type(df1.dot(np_array), pd.DataFrame), pd.DataFrame)
     check(assert_type(df1 @ s1, pd.Series), pd.Series)
     check(assert_type(df1.dot(s1), pd.Series), pd.Series)
+    list_like = [[0, 1], [1, 2], [-1, -1], [2, 0]]
+    check(assert_type(df1.dot(list_like), pd.DataFrame), pd.DataFrame)
```
cmp0xff, line 2806: The original example is missing from the test with `[3600, 60]`

Khan3K:
Thanks for the review, @cmp0xff! I've addressed both points:

1. **Precise typing instead of `list[Any]`**: I adopted the `_SupportsMul`-based approach. One note: I named the protocol `SupportsSelfMul` rather than `SupportsMul` because `SupportsMul` is already imported from `_typeshed` in `series.pyi` (the two-argument `__mul__` protocol used by `__rmul__`/`rmul`). The new name also matches the existing `SupportsSelfSub` convention in the same file (both are `@type_check_only`).

2. **The original `[3600, 60]` example**: added to `test_types_dot`. Since a 1-D list returns a `Series` at runtime (per the pandas docs), I mapped the single-level `Sequence[SupportsSelfMul[Any]]` to the `Series` overload and kept `Sequence[Sequence[SupportsSelfMul[Any]]]` on the `Self` (DataFrame) overload:
   - `df1.dot([[0, 1], [1, 2], ...])` → `DataFrame`
   - `time.dot([3600, 60])` → `Series`

Verified locally with mypy (`pandas-stubs/core/frame.pyi`, `pandas-stubs/core/series.pyi`, `tests/frame/test_frame.py` — no errors) and pyright (no new errors vs. base).

pandas-stubs/core/frame.pyi
cmp0xff: c716eae9dbb129995cad145e4d9e08e13f9d2798

cmp0xff (approved):
Thank you @Khan3K @Faiaz-To-Faiaz

PR #1895, opened 2026-08-14
CLN: #1880 disable `pyright` `reportUnknownArgumentType`

### LINKED ISSUE #1880: Investigate if `pyright` `reportUnknownArgumentType` should be disabled
> If the `reportUnknownArgumentType` is only occurring when we are doing stuff with `lambda`, we should not worry about that one as well.
>
> If it is occurring in other places, it may be worth a look to see if we should do something different.

_Originally posted by @Dr-Irv in https://github.com/pandas-dev/pandas-stubs/pull/1879#discussion_r3732034653_

comment: there are `24` occurrences, a few do not seem to have to do with `lambda`.

cmp0xff:
- [x] Closes #1880


Dr-Irv (approved):
thanks @cmp0xff

PR #1898, opened 2026-08-14
BUG: #1896 `DataFrame.apply`

### LINKED ISSUE #1896: DataFrame.apply / GroupBy.apply: first parameter named `f` in stubs but `func` at runtime
## Describe the bug

The first parameter of `DataFrame.apply` (and `DataFrameGroupBy.apply`) is named **`f`** in pandas-stubs, but the runtime pandas signature names it **`func`**. This makes the keyword-argument form inconsistent between the stubs and runtime:

- `df.apply(func=...)` — **valid at runtime**, but **rejected by the stubs** (no overload variant matches).
- `df.apply(f=...)` — **accepted by the stubs**, but **raises `TypeError` at runtime**.

Only the positional form (`df.apply(callable, ...)`) works with both. So code that correctly calls `apply(func=...)` fails type checking, and code written to satisfy the stubs (`apply(f=...)`) crashes at runtime.

The stubs declare `f` (`pandas-stubs/core/frame.pyi`):

```python
def apply(self, f: Callable[..., ListLikeExceptSeriesAndStr | Series], axis: AxisIndex = ...): ...
```

while runtime pandas declares `func` (`pandas/core/frame.py`):

```python
def apply(self, func: AggFuncType, axis: Axis = 0, ...): ...
```

## To Reproduce

1. Minimal example — correct at runtime, but rejected by the stubs:

```python
import pandas as pd

result = df.apply(func=lambda row: row["a"] * 2, axis="columns")  # runs fine at runtime
```

2. Type checker: **mypy** (2.1.0)

3. Error message from mypy:

```
error: No overload variant of "apply" of "DataFrame" matches argument types "Callable[[Any], Any]", "str"  [call-overload]
note:     def apply(self, f: Callable[..., ...], axis: Literal['index', 0] = ..., ...) -> ...
```

Conversely, the form the stubs accept **crashes at runtime**:

```python
df.apply(f=lambda row: row["a"] * 2, axis="columns")   # mypy: Success: no issues found
# runtime: TypeError: DataFrame.apply() missing 1 required positional argument: 'func'
```

## System specifications

- OS and its version: macOS 26.5.2 (Darwin 25.5.0)
- python version: 3.14.6
- type checker and its version: mypy 2.1.0
- version of installed `pandas-stubs`: 3.0.5.260730 (with `pandas` 3.0.5)

## Additional context

Renaming the stub parameter from `f` to `func` (matching the runtime signature) would make the keyword form consistent across all `DataFrame.apply` overloads. The same mismatch affects `DataFrameGroupBy.apply`, where the parameter is additionally declared positional-only (`/`), so no keyword form works there at all.

DSeaStar:
- [x] Closes #1896

`DataFrame.apply` overloads named the first parameter `f`, but pandas names it `func`. `df.apply(func=...)` is valid at runtime and was rejected by the stubs; `df.apply(f=...)` type-checked and then raised `TypeError`.

Rename the parameter to `func` on every overload. `DataFrameGroupBy.apply` / `Series.apply` already used `func`.

### DIFF
--- a/pandas-stubs/core/frame.pyi
+++ b/pandas-stubs/core/frame.pyi
@@ -1449,7 +1449,7 @@ class DataFrame(NDFrame, OpsMixin, _GetItemHack):
     @overload
     def apply(
         self,
-        f: Callable[..., ListLikeExceptSeriesAndStr | Series],
+        func: Callable[..., ListLikeExceptSeriesAndStr | Series],
         axis: AxisIndex = ...,
         raw: _bool = ...,
         result_type: None = None,
@@ -1460,7 +1460,7 @@ class DataFrame(NDFrame, OpsMixin, _GetItemHack):
     def apply(
         self,
         # Use S2 (TypeVar without `default=Any`) instead of S1 due to https://github.com/python/mypy/issues/19182.
-        f: Callable[..., S2 | NAType],
+        func: Callable[..., S2 | NAType],
         axis: AxisIndex = ...,
         raw: _bool = ...,
         result_type: None = None,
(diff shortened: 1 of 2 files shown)


loicdiridollou (approved):
Thanks @DSeaStar

Issue #1609, opened 2025-12-29
`mypy` recognises `[1]` also as `SequenceNotStr[str]`.


cmp0xff:
## Describe the bug

`mypy` thinks `[1]` also fits `SequenceNotStr[str]`.

Since `SequenceNotStr` is our own definition, I think it worths to discuss here first, before sending the report to `mypy`.

### Minimal example

```py
from collections.abc import Iterator, Sequence
from typing import Any, Protocol, SupportsIndex, overload
from typing_extensions import reveal_type, TypeVar


_T_co = TypeVar("_T_co", covariant=True)


class SequenceNotStr(Protocol[_T_co]):
    @overload
    def __getitem__(self, index: SupportsIndex, /) -> _T_co: ...
    @overload
    def __getitem__(self, index: slice, /) -> Sequence[_T_co]: ...
    def __contains__(self, value: object, /) -> bool: ...
    def __len__(self) -> int: ...
    def __iter__(self) -> Iterator[_T_co]: ...
    def index(self, value: Any, start: int = ..., stop: int = ..., /) -> int: ...
    def count(self, value: Any, /) -> int: ...
    def __reversed__(self) -> Iterator[_T_co]: ...


def foo(a: SequenceNotStr[str]) -> None:
    reveal_type(a[0])  # mypy and pyright gives str


reveal_type(foo([1]))  # mypy gives None, pyright gives error
```

### Indicate which type checker you are using

both

## Please complete the following information:
 - OS: Debian
 - OS Version: 12
 - python version: 3.12
 - version of type checker: `mypy` 1.19.1
 - version of installed `pandas-stubs`: None

**Additional context**

randolf-scholz:
`MutableSequence` is invariant, not covariant, so `foo(seq)` fails.

`foo(["1 day"])` succeeds because they convert the list literal into a function call `[T](arg: T) -> list[T]` and then `T=str | float` is a valid solution for the argument `"1 day"`.

randolf-scholz:
Not really, that's the price for mutability. Otherwise, stuff like this would be problematic:

```python
def add_zero(arg: MutableSequence[int]) -> None:
    arg.append(0)

class MyInteger(int):
    def special_method(self): ...

l: list[MyInteger] = []
add_zero(l)
l[-1].special_method()  # kaboom
```

cmp0xff:
[...]

PR #1899, opened 2026-08-14
BUG: #1890 widen `read_sql`/`read_sql_query` `params`

### LINKED ISSUE #1890: `read_sql` and `read_sql_query` reject a `params` dict that mixes a scalar and a sequence
## Describe the bug

`params` accepts two kinds of dict, `Mapping[str, Scalar]` and `Mapping[str, tuple[Scalar, ...]]`. A dict with a date under one key and a tuple under another matches neither. So filtering by a date and an `IN` list in one query does not type check.

Positional params have the same gap: `[date, (1, 2)]` matches neither `list[Scalar]` nor `tuple[tuple[Scalar, ...], ...]`.

pandas itself annotates `params` as `list[Any] | Mapping[str, Any] | None`. #997 added tuple support for #996 as a second option instead of widening the value type, which is why no type fits the mixed dict.

## To Reproduce

Runs on sqlite and prints `[1, 2]`. The `read_sql_query` call is rejected.

```python
import datetime as dt

import pandas as pd
import sqlalchemy as sa

engine = sa.create_engine("sqlite://")
with engine.begin() as con:
    con.execute(sa.text("CREATE TABLE t (id INTEGER, d DATE)"))
    con.execute(sa.text("INSERT INTO t VALUES (1, '2026-08-11'), (2, '2026-08-11')"))

statement = sa.text("SELECT id FROM t WHERE d = :d AND id IN :ids").bindparams(
    sa.bindparam("ids", expanding=True)
)
with engine.connect() as con:
    df = pd.read_sql_query(statement, con, params={"d": dt.date(2026, 8, 11), "ids": (1, 2)})

print(df["id"].tolist())
```

pyright 1.1.406:

```
error: Argument of type "dict[str, date | tuple[int, int]]" cannot be assigned to parameter
"params" of type "list[Scalar] | ... | Mapping[str, tuple[Scalar, ...]] | None"
in function "read_sql_query"
    Type parameter "_VT_co@Mapping" is covariant, but "date | tuple[int, int]" is not a
    subtype of "Scalar"
```

mypy 1.19.1, pyrefly 1.2.0 and ty 0.0.69 reject it the same way.

## System specifications

- OS and its version: Rocky Linux 9 / Linux 6.1.29
- python version: 3.12.13
- type checker and its version: pyright 1.1.406, mypy 1.19.1, pyrefly 1.2.0, ty 0.0.69
- version of installed `pandas-stubs`: 3.0.5.260730 (pandas 3.0.5, SQLAlchemy 2.0.51)

## Additional context

One type for the bind value fixes both cases:

```python
_SQLBindValue: TypeAlias = Scalar | None | Sequence[Scalar | None]

params: Sequence[_SQLBindValue] | Mapping[str, _SQLBindValue] | None = None,
```

This covers all five current options, so nothing that passes today starts failing. `None` is included because binding NULL is common and `Scalar` omits it. Sets and nested sequences stay rejected.

 #997's tuple test is skipped because it needed Postgres; the example above runs on sqlite, so it can be unskipped.

notatallshaw:
- [x] Closes #1890
- [x] Tests added

This fixes `params` in `read_sql` and `read_sql_query` being able to take a mixed type mapping.

This introduces one regression in `read_sql_query` that already exists on main for `read_sql`, `params` will now accept a `str` (and `bytes` / `bytearray` / `memoryview`), which is incorrect as it mismatches documentation and functionality. I tried to fix this for both  `read_sql_query` and `read_sql` but I found both short comings in `SequenceNotStr` and using a protocol (same issue as #1609) in mypy, happy to take guidance on this.

### DIFF
--- a/tests/test_io.py
+++ b/tests/test_io.py
@@ -1617,32 +1617,125 @@ def test_read_sql_query_via_sqlalchemy_engine_with_params(tmp_path: Path) -> Non
     engine.dispose()
 
 
-@pytest.mark.skip(
-    reason="Only works in Postgres (and MySQL, but with different query syntax)"
-)
-def test_read_sql_query_via_sqlalchemy_engine_with_tuple_valued_params() -> None:
-    db_uri = "postgresql+psycopg2://postgres@localhost:5432/postgres"
+def test_read_sql_query_via_sqlalchemy_engine_with_tuple_valued_params(
+    tmp_path: Path,
+) -> None:
+    path_str = str(tmp_path / str(uuid.uuid4()))
+    db_uri = "sqlite:///" + path_str
     engine = sqlalchemy.create_engine(db_uri)
 
-    check(
-        assert_type(
-            read_sql_query(
-                "select * from test where a in %(a)s", con=engine, params={"a": (1, 2)}
+    check(assert_type(DF.to_sql("test", con=engine), int | None), int)
+    statement = sqlalchemy.text("select * from test where a in :a").bindparams(
+        sqlalchemy.bindparam("a", expanding=True)
+    )
+    with engine.connect() as conn:
+        check(
+            assert_type(
+                read_sql_query(statement, con=conn, params={"a": (1, 2)}), DataFrame
             ),
             DataFrame,
-        ),
-        DataFrame,
-    )
-    check(
+        )
+    if TYPE_CHECKING:
+        # %s paramstyle needs psycopg2 or MySQLdb, so this is checked but not run
         assert_type(
             read_sql_query(
                 "select * from test where a in %s", con=engine, params=((1, 2),)
             ),
             DataFrame,
+        )
+    engine.dispose()
+
+
+def test_read_sql_query_with_mixed_params(tmp_path: Path) -> None:
+    # GH 1890
+    path_str = str(tmp_path / str(uuid.uuid4()))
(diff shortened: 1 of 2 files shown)


tests/test_io.py
```diff
@@ -1617,32 +1617,125 @@ def test_read_sql_query_via_sqlalchemy_engine_with_params(tmp_path: Path) -> Non
...
+    engine.dispose()
+
+
+def test_read_sql_query_with_list_params(tmp_path: Path) -> None:
+    # GH 1890
+    path_str = str(tmp_path / str(uuid.uuid4()))
+    con = sqlite3.connect(path_str)
+    frame = DataFrame({"a": ["x", "y"], "b": [1, 2]})
+    check(assert_type(frame.to_sql("test", con=con), int | None), int)
+    names: list[str] = ["x"]
```
loicdiridollou: No need for the extra typehint here, the goal is to let the typechecker do the work.

loicdiridollou (approved):
One minor thing but looking clean otherwise

loicdiridollou (changes requested):
One minor thing but looking clean otherwise

tests/test_io.py
notatallshaw: Fixed

loicdiridollou (approved):
Looks good, thanks @notatallshaw

PR #1900, opened 2026-08-15
Update fill value series dataframe

### LINKED ISSUE #1797: Specifying `fill_value` in `pd.Series.eq()` with `'string'` dtype
Right now I see the [pandas docs](https://pandas.pydata.org/docs/reference/api/pandas.Series.eq.html) show that `fill_value` should be `None or float value`, but if the `Series` has dtype `'string'`, then using a string value works as expected (although `mypy` obviously balks).

I know this is kind of a pandas issue too; I tried navigating the pandas issue tracker for something mentioning it and found this: https://github.com/pandas-dev/pandas/issues/15533. I got a little lost trying to parse everything in the issue, so I'm a little unsure of where this stands; would a pandas doc PR be sufficient for now? Should pandas-stubs support passing a string to `fill_value`?

I'm happy to put together any necessary PRs here or for docs in the pandas repo if it's agreed upon that this is the correct approach.

krishgarg344:
Closes #1797

`fill_value` on the comparison and arithmetic flex methods was typed as
`float | None` regardless of the Series/DataFrame's actual dtype

- `Series`: `fill_value` is now typed as `S1 | None` on `eq`, `ne`, `ge`,
  `gt`, `le`, `lt`, and all arithmetic methods (`add`, `radd`, `sub`,
  `rsub`, `mul`, `rmul`, `truediv`, `rtruediv`, `floordiv`, `rfloordiv`,
  `mod`, `rmod`, `pow`, `rpow`, `divmod`, `rdivmod`).
- `DataFrame`: since DataFrame isn't dtype-generic (columns can differ),
  there's no `S1` equivalent. Went with `fill_value: Scalar | None` on
  the 14 arithmetic methods instead, matching the general `scalar`
  description in the pandas docs

Open question for review: is `Scalar | None` the right call for
`DataFrame`, or should this be scoped differently?

- [x] Tests added


tests/frame/test_frame.py
```diff
@@ -4551,20 +4511,34 @@ def test_frame_pipe() -> None:
...
-        assert_type(df.expanding().pipe(lambda x: x.min() - x.max()), pd.DataFrame),
+        assert_type(  # pyrefly: ignore[assert-type]
+            df.expanding().pipe(lambda x: x.min() - x.max()), pd.DataFrame
+        ),
         pd.DataFrame,
     )
     check(assert_type(df.expanding().pipe(func_e, k=2), pd.DataFrame), pd.DataFrame)
+
+
+def test_fill_value_scalar_dataframe() -> None:
```
cmp0xff, line 4539: We already have a `test_types_scalar_arithmetic`. Please try to reuse the existing test function and integrate the new tests there.

Furthermore, you changed more than `add`. Please include tests for all functional changes.

pandas-stubs/core/series.pyi
```diff
@@ -1998,23 +1959,23 @@ class Series(IndexOpsMixin[S1], ElementOpsMixin[S1], NDFrame):
         self: Series[Never],
         other: _str,
         level: Level | None = None,
-        fill_value: float | None = None,
+        fill_value: S1 | None = None,
```
cmp0xff, line 1962: I think this may not work. `S1` is the `TypeVar` in `self` when it is unannotated. However this is not the case here; `self` is annotated as `Series[Never]`, so I think `S1` is meaningless.

In the meantime, you are making a lot of changes here. Each requires a test. You may look for `test_add.py` and there are many of them, not to mention you are also changing `sub`, `mul`, etc. Making comprehensive changes + tests is not easy.

Looking back at the original issue, only `Series.eq` is mentioned, not the rich arithmetic operations. I would suggest focusing on the original issue first, and if you are interested in the rich arithmetic functions, you may create a master tracker and fix them one by one.

cmp0xff (changes requested):
Please follow the PR template and also claim your AI usage, if there is any.

cmp0xff:
@krishgarg344 would you continue working on this?

PR #1910, opened 2026-08-20
Use Pyrefly for type-coverage instead of pyright

### LINKED ISSUE #1795: test_completeness checks are timing out

MarcoGorelli:
This builds on https://github.com/pandas-dev/pandas-stubs/pull/1883 and keeps @loicdiridollou as co-author

closes https://github.com/pandas-dev/pandas-stubs/issues/1795


.github/workflows/test.yml
```diff
@@ -51,6 +51,9 @@ jobs:
       - name: Run mypy on 'tests' (using the local stubs) and on the local stubs
         run: poetry run poe mypy
 
+      - name: Verify type completeness using pyrefly
+        run: poetry run poe type_completeness
```
MarcoGorelli, line 55: this now takes 2 seconds in CI, so I think it's fine to run it alongside the other checks rather than in a separate job

pyproject.toml
```diff
@@ -38,7 +38,7 @@ dev = [
   "pytest>=8.4.2",
   "pyright>=1.1.410",
   "ty>=0.0.67",
-  "pyrefly>=1.1.0",
+  "pyrefly (>=1.2.0)",
```
loicdiridollou: Shouldn’t we keep the same format as the other specifications for this section? Like `“pyrefly>=1.2.0”`

loicdiridollou (commented):
One minor question, the rest looks great!

pyproject.toml
MarcoGorelli: yup thanks!

loicdiridollou (approved):
Great step forward thanks @MarcoGorelli

PR #1911, opened 2026-08-21
TYP: accept datetime.timedelta as freq for round, floor, and ceil

ghackebeil:
Extends `datetime.timedelta` support added to `to_offset` in https://github.com/pandas-dev/pandas-stubs/pull/1707 to additional functions:
```
Timedelta.floor
Timedelta.ceil
Timedelta.round
Timestamp.floor
Timestamp.ceil
Timestamp.round
```


loicdiridollou:
@Dr-Irv @cmp0xff this opens another question, it seems like most of the instance where Frequency (`str | BaseOffset`) is used has been touched, except `Groupby.shift` (for `freq`) and pandas-stubs/core/indexes/timedeltas
In many of the other places it was already something like `Frequency | dt.Timedelta | ...`
Should it be something we could add to Frequency?

Dr-Irv:
> @Dr-Irv @cmp0xff this opens another question, it seems like most of the instance where Frequency (`str | BaseOffset`) is used has been touched, except `Groupby.shift` (for `freq`) and pandas-stubs/core/indexes/timedeltas In many of the other places it was already something like `Frequency | dt.Timedelta | ...` Should it be something we could add to Frequency?

No.  I want to be consistent here with what is in the pandas source in terms of the `Frequency` definition

loicdiridollou:
> > @Dr-Irv @cmp0xff this opens another question, it seems like most of the instance where Frequency (`str | BaseOffset`) is used has been touched, except `Groupby.shift` (for `freq`) and pandas-stubs/core/indexes/timedeltas In many of the other places it was already something like `Frequency | dt.Timedelta | ...` Should it be something we could add to Frequency?
> 
> No. I want to be consistent here with what is in the pandas source in terms of the `Frequency` definition

Oh okay I thought this was an internal TypeAlias. 100% we need to stay aligned

loicdiridollou (approved):
Looks good, thanks @ghackebeil !

PR #1930, opened 2026-08-30
CLN: load tests._typing from pandas-stubs/_typing.pyi without file swap

cmp0xff:
## Summary

Replace the tests/conftest.py file-swap mechanism with a dual-mode tests/_typing.py shim, so importing or running conftest.py (for example, from VSCode test discovery) no longer mutates tracked files.

- tests/_typing.py re-exports the stub aliases from pandas._typing under `if TYPE_CHECKING:` for static type checkers and, at runtime only, executes pandas-stubs/_typing.pyi in its own namespace.
- tests/conftest.py is deleted; tests/_typing.pyi is no longer created.
- The now-obsolete tests/_typing.pyi entry is removed from .gitignore.

No existing issue to close. No new assert_type tests were needed: existing tests already import tests._typing, and the full validation suite passes.

- [ ] Tests added (Please use `assert_type()` to assert the type of any return value) — no new type assertions; covered by the existing suite.

<details><summary>AI Implementation Plan</summary>

## Motivation

pandas-stubs tests/conftest.py replaced tests/_typing.py with pandas-stubs/_typing.pyi at import time and restored it in a session-scoped fixture. Tools that import conftest.py without running a full pytest session (for example VSCode test discovery) triggered the replacement without the rollback, leaving tracked files modified and a stray tests/_typing.pyi behind.

## Approach

Keep the single source of truth (pandas-stubs/_typing.pyi) and remove all disk mutation:

- tests/_typing.py keeps its explicit 79-name import list and `__all__`, but the import is guarded by `if TYPE_CHECKING:` so the four type checkers resolve the aliases from the stub package.
- A `if not TYPE_CHECKING:` branch executes pandas-stubs/_typing.pyi in the module namespace at runtime. The .pyi content is valid Python and currently already ran as tests/_typing.py after the old copy step, so runtime behavior is unchanged.
- tests/conftest.py is deleted.

## Experiments

- Verified the dual-mode pattern in a scratch project against pyright (strict), mypy (--strict), pyrefly (strict preset), and ty; all four resolved the re-exported names with no errors.
- Verified runtime import against pandas 3.0.5: tests._typing.np_ndarray_bool resolves to the expected numpy alias.

## Attempts and failures

- First full validation run failed at pre-commit ruff-check with S102 (use of `exec`) and PLE0605 (`__all__ = _all` is not a list/tuple literal).
- Fix: added `# noqa: S102` to the intentional exec line and dropped the runtime `__all__` restore. Nothing imports `*` from tests._typing (verified by search), so leaving the runtime `__all__` as the stub file's own `["npt", "type_t"]` has no effect.
- Windows CI then failed at pytest: tests/test_typing.py::test_dtype_arg_aliases compared `'datetime64[\xce\xbcs]'` (mojibake) against `'datetime64[μs]'`. Cause: `Path.read_text()` uses the locale encoding, which is cp1252 on Windows, so the UTF-8 `μ` characters in pandas-stubs/_typing.pyi string literals were decoded incorrectly at runtime.
- Fix: read the source with `encoding="utf-8"`. The `# noqa: S102` comment now sits on the `exec(` line so black keeps it stable.
- Re-ran the full suite locally: all steps passed (src and dist mode).

## Caveats

- After import at runtime, tests._typing.__all__ is the stub file's `["npt", "type_t"]`, not the shim's explicit list; no code depends on it.
- The explicit import list in tests/_typing.py must stay in sync with the names used by test modules; the four type checkers fail the build if a name is missing.

## Outcome

Full validation `run_tests(src=True, dist=True)` passes locally: ty, pyrefly, mypy, pyright, pytest, pre-commit, plus dist-mode checks. Both the worktree and the main checkout are clean after the run.

## Out of scope

- No new alias-consistency test; type checkers already enforce it.
- No changes to pandas-stubs/_typing.pyi itself.

## To-do

- None.

</details>


Dr-Irv (commented):
I wonder if we should put any definitions of aliases that are not in `pandas/_typing.py` in a separate PYI file (e.g., `pandas-stubs/_typing_aliases.pyi`), and have `pandas-stubs/_typing.pyi` import `_typing_aliases`, so then we know that `_typing_aliases` are the "extra" types we have added to the stubs that we use for testing the stubs.

If we did that, wouldn't this make things easier to maintain?

cmp0xff:
> I wonder if we should put any definitions of aliases that are not in `pandas/_typing.py` in a separate PYI file (e.g., `pandas-stubs/_typing_aliases.pyi`), and have `pandas-stubs/_typing.pyi` import `_typing_aliases`, so then we know that `_typing_aliases` are the "extra" types we have added to the stubs that we use for testing the stubs.

That would be nice. Unfortunately we already make extensive use of definitions in our tests, which are not in `pandas/_typing.py`. In particular, `np_ndarray_something` and `np_1darray_something`. Such things have to live in a `.py` file when tests run.

> If we did that, wouldn't this make things easier to maintain?

Maybe we can think about such a refactor, but from my argument above, I'd say it's not easy.

Dr-Irv:
The other option is that we could start adding these types into `pandas` at `pandas.api.typing.aliases`, then import them from there in the tests.

loicdiridollou (approved):
Clean! Thanks @cmp0xff

PR #1941, opened 2026-09-09
TYP: `first_valid_index`/`last_valid_index` return `Scalar | None`

samskiter:

`Series.first_valid_index`, `Series.last_valid_index`, `DataFrame.first_valid_index` and `DataFrame.last_valid_index` are annotated as returning `Scalar`, but pandas documents them as "Return index for first/last non-NA value or None, if no non-NA value is found", and an all-NA (or empty) Series/DataFrame returns `None` at runtime. This changes the four return annotations to `Scalar | None` and adds runtime-checked `assert_type` tests for both the non-NA and the all-NA case.

<details><summary>AI Implementation Plan</summary>

**Motivation.** Downstream code running mypy with `warn_unreachable` gets a false "unreachable" report on `if s.first_valid_index() is None:` guards, because the stubs say the call can never return `None`. Removing the guard would crash on all-NaN input, so the stub is wrong, not the caller.

**Changes.**
- `pandas-stubs/core/series.pyi`: `first_valid_index` / `last_valid_index` → `Scalar | None`.
- `pandas-stubs/core/frame.pyi`: same two methods → `Scalar | None`.
- `tests/series/test_series.py`: `test_series_first_last_valid_index` checks a partially-NA Series returns an `int` label and an all-NA Series returns `None`.
- `tests/frame/test_frame.py`: `test_dataframe_first_last_valid_index`, same shape for DataFrame.

**Validation.** `poe test` locally: ty, pyrefly, mypy and pyright all pass with no diagnostics; pre-commit style hooks pass; pytest passes (the only failures in the environment were `ZoneInfoNotFoundError` from a missing `tzdata` package, unrelated, and they pass once `tzdata` is installed).

**Caveats / out of scope.** Callers that previously relied on the non-`None` annotation will now need an `is None` check or an `assert`; that is the correct runtime behaviour. `NDFrame` in `generic.pyi` does not declare these methods, so nothing else needed updating.

</details>

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01HXUrhGQv3vdDJe2VAMRoR3

---
_Generated by [Claude Code](https://claude.ai/code/session_01HXUrhGQv3vdDJe2VAMRoR3)_


pandas-stubs/core/frame.pyi
```diff
@@ -2112,7 +2112,7 @@ class DataFrame(NDFrame, OpsMixin, _GetItemHack):
         axis: Axis | None = None,
     ) -> Self: ...
     @final
-    def first_valid_index(self) -> Scalar: ...
+    def first_valid_index(self) -> Scalar | None: ...
```
cmp0xff, line 2115:
```suggestion
    def first_valid_index(self) -> Hashable | None: ...
```

Consider the following example:

```py
pd.Series([1], index=[()]).first_valid_index()
```

pandas-stubs/core/frame.pyi
```diff
@@ -2175,7 +2175,7 @@ class DataFrame(NDFrame, OpsMixin, _GetItemHack):
         **kwargs: Any,
     ) -> Series: ...
     @final
-    def last_valid_index(self) -> Scalar: ...
+    def last_valid_index(self) -> Scalar | None: ...
```
cmp0xff, line 2178:
```suggestion
    def last_valid_index(self) -> Hashable | None: ...
```

```py
pd.Series([1], index=[()]).last_valid_index()
```

pandas-stubs/core/series.pyi
```diff
@@ -1917,9 +1917,9 @@ class Series(IndexOpsMixin[S1], ElementOpsMixin[S1], NDFrame):
         freq: Frequency | timedelta | None = None,
     ) -> Series[float]: ...
     @final
-    def first_valid_index(self) -> Scalar: ...
+    def first_valid_index(self) -> Scalar | None: ...
```
cmp0xff, line 1920:
```suggestion
    def first_valid_index(self) -> Hashable | None: ...
```

pandas-stubs/core/series.pyi
```diff
@@ -1917,9 +1917,9 @@ class Series(IndexOpsMixin[S1], ElementOpsMixin[S1], NDFrame):
...
     @final
-    def last_valid_index(self) -> Scalar: ...
+    def last_valid_index(self) -> Scalar | None: ...
```
cmp0xff, line 1922:
```suggestion
    def last_valid_index(self) -> Hashable | None: ...
```

tests/frame/test_frame.py
```diff
@@ -2985,6 +2985,16 @@ def test_dataframe_pct_change() -> None:
     check(assert_type(df.pct_change(fill_method=None), pd.DataFrame), pd.DataFrame)
 
 
+def test_dataframe_first_last_valid_index() -> None:
```
cmp0xff, line 2988: Use existing `test_types_idxmin` / `test_types_idxmax` instead; also add tests for the new cases of non-`Scalar` results.

tests/series/test_series.py
```diff
@@ -498,6 +498,16 @@ def test_series_pct_change() -> None:
     )
 
 
+def test_series_first_last_valid_index() -> None:
```
cmp0xff, line 501: Use existing `test_types_idxmin` / `test_types_idxmax` instead; also add tests for the new cases of non-`Scalar` results.

cmp0xff (changes requested):

cmp0xff:
Please rebase onto `main` or merge `main` into your branch, so that CI can be fixed.

PR #1945, opened 2026-09-12
Bump Pyrefly version

MarcoGorelli:
Quite a few `# pyrefly: ignore`s get to be removed

Unfortunately, some new ones have crept in since https://github.com/facebook/pyrefly/issues/4913


tests/series/test_series.py
```diff
@@ -3685,9 +3691,11 @@ def to_str(x: int) -> str:
 
     unknown_series = pd.Series([1, 0, None])
     check(
-        # TODO: facebook/pyrefly#3268
-        # pyrefly: ignore[assert-type]
-        assert_type(unknown_series.map({1: True, 0: False, None: None}), pd.Series),
+        # pyrefly now infers Series[bool | None] here (mypy/pyright still infer
```
cmp0xff, line 3694: If we don't want to allow Series[bool | None], maybe we should not do a `bound` for `S1`, but exact types? This was discussed before, but here is another hint. cc @Dr-Irv @loicdiridollou p.s.  I will have no access to GitHub until next Sunday.

tests/test_groupby.py
```diff
@@ -251,7 +250,6 @@ def h(val: Resampler[DataFrame]) -> Series:
 
     def i(val: Resampler[DataFrame]) -> float:
         assert isinstance(val, Resampler)
```
cmp0xff, line 252: I think we can clean up and remove these `assert`s, but not necessarily in this PR.

cmp0xff (approved):
Thank you @MarcoGorelli !

tests/series/test_series.py
Dr-Irv: I think `pyrefly` is incorrect here.  The docs say: 

> Bounded type variables and constrained type variables have different semantics in several important ways. Using a bounded type variable means that the TypeVar will be solved using the most specific type possible:

Now maybe `pyrefly` is saying that `bool | None` is the "most specific type", but my interpretation of that phrase means that a union type cannot be specific.

Also at https://typing.python.org/en/latest/spec/generics.html, it says "type constraints cause the inferred type to be _exactly_ one of the constraint types"

So based on that, I consider this a pyrefly bug

MarcoGorelli: thanks - I think there is a Pyrefly issue here, reported at https://github.com/facebook/pyrefly/issues/4937

PR #1946, opened 2026-09-13
TYP: Include missing values in scalar indexer returns

### LINKED ISSUE #1937: `Scalar` is missing types that `.loc` and `.at` can return
## Describe the bug

[`Scalar`](https://github.com/pandas-dev/pandas-stubs/blob/a3824373b6f3c8ce5a3f2185e64f6eea98fb03ae/pandas-stubs/_typing.pyi#L140-L142) is the return type of [`DataFrame.loc[r, c]`](https://github.com/pandas-dev/pandas-stubs/blob/a3824373b6f3c8ce5a3f2185e64f6eea98fb03ae/pandas-stubs/core/frame.pyi#L258-L265), `.iloc[i, j]`, [`.at`](https://github.com/pandas-dev/pandas-stubs/blob/a3824373b6f3c8ce5a3f2185e64f6eea98fb03ae/pandas-stubs/core/frame.pyi#L345-L347), `.iat`, `Series.at`, `Series.iat` and `itertuples()` row attributes. pandas [documents](https://github.com/pandas-dev/pandas/blob/80d2663b37792511e1f9a5531065d723efb79a3f/doc/source/reference/aliases.rst#L137-L138) the alias as "Types that can be stored in Series with non-object dtype", but these non-object values are not in it:

| dtype | `df.at[0, c]` at runtime |
|---|---|
| bool | `np.bool_` |
| Int64 or string with a missing value | `NAType` |
| datetime64 with a missing value | `NaTType` |
| period | `Period` |
| interval | `Interval` |

The setters already accept all of these through [`ScalarOrNA`](https://github.com/pandas-dev/pandas-stubs/blob/a3824373b6f3c8ce5a3f2185e64f6eea98fb03ae/pandas-stubs/core/frame.pyi#L349-L350).

## To Reproduce

1. 

```python
import pandas as pd

df = pd.DataFrame({
    "flag": [True],
    "n": pd.array([None], dtype="Int64"),
    "ts": [pd.NaT],
    "p": [pd.Period("2020-01")],
    "iv": [pd.Interval(0, 1)],
})

n = df.at[0, "n"]
if n is pd.NA:
    print("missing")  # printed

ts = df.at[0, "ts"]
if ts is pd.NaT:
    print("missing")  # printed

print(df.at[0, "p"].year)  # 2020
print(df.at[0, "iv"].left)  # 0
print(type(df.at[0, "flag"]))  # <class 'numpy.bool'>
```
2. mypy (pyright reports the attribute errors but not the identity checks):

```
error: Non-overlapping identity check (left operand type: "str | bytes | date | datetime | timedelta | <7 more items> | complex | integer[Any] | floating[Any] | complexfloating[Any, Any]", right operand type: "NAType")  [comparison-overlap]
error: Non-overlapping identity check (left operand type: "str | bytes | date | datetime | timedelta | <7 more items> | complex | integer[Any] | floating[Any] | complexfloating[Any, Any]", right operand type: "NaTType")  [comparison-overlap]
error: Item "str" of "str | bytes | date | datetime | timedelta | <7 more items> | complex | integer[Any] | floating[Any] | complexfloating[Any, Any]" has no attribute "year"  [union-attr]
error: Item "str" of "str | bytes | date | datetime | timedelta | <7 more items> | complex | integer[Any] | floating[Any] | complexfloating[Any, Any]" has no attribute "left"  [union-attr]
```

## System specifications

- OS and its version: Linux 6.18.33 (WSL2)
- python version: 3.12.13
- type checker and its version: mypy 2.3.0, pyright 1.1.411
- version of installed `pandas-stubs`: main at a382437 (2026-08-31), pandas 3.0.5

## Additional context

I found three times `Scalar` was widened (#501, #550, #1182), each because a value passed to a function was rejected. Its use as the return type of `.loc` and `.at` seems to date from the original stubs, and I could not find any time it was revisited.

Since pandas-dev/pandas#61735, pandas' own [alias](https://github.com/pandas-dev/pandas/blob/80d2663b37792511e1f9a5531065d723efb79a3f/pandas/_typing.py#L139-L157) is this union plus `Interval`, so the "This is wider than what is in pandas" [comment](https://github.com/pandas-dev/pandas-stubs/blob/a3824373b6f3c8ce5a3f2185e64f6eea98fb03ae/pandas-stubs/_typing.pyi#L139) is out of date.

One possible fix is to add `np.bool_`, `Period`, `Interval`, `NAType` and `NaTType` to `Scalar`. Another is to add `np.bool_`, `Period` and `Interval` to `Scalar` and have the methods listed above return `ScalarOrNA` instead of `Scalar`.

 Related: #1506 on `bool | np.bool`, #835 on `NAType`, #1892 on whether those methods should return `Scalar` or `Any` (which is where I started investigating and trying to understand the `Scalar` type).

naseweisssss:
- [x] Closes #1937

`Scalar` now includes `np.bool_`, `Period`, and `Interval`. Scalar-returning DataFrame indexers, Series `.at`/`.iat`, and `itertuples()` attributes now use `ScalarOrNA` so their annotations include `NAType`, `NaTType`, and `None`.

The regression tests cover `.loc`, `.iloc`, `.at`, `.iat`, Series scalar access, and `itertuples()` attributes.

Validation: `poetry run poe test_all`


pandas-stubs/_typing.pyi
```diff
@@ -138,7 +138,14 @@ _IndexIterScalar: TypeAlias = (
...
+    _IndexIterScalar
+    | complex
+    | np.bool_
+    | np.integer
+    | np.floating
+    | np.complexfloating
+    | Period
+    | Interval
 )
 ScalarOrNA: TypeAlias = Scalar | NAType | NaTType | None
```
cmp0xff, line 150: Now that you use `ScalarOrNA` in tests, it should be in `tests/_typing.py`

tests/frame/test_frame.py
```diff
@@ -55,6 +55,10 @@
     pytest_warns_bounded,
     pytest_warns_excel_pandas4,
 )
+
+if TYPE_CHECKING:
+    from pandas._typing import ScalarOrNA  # noqa: F401
```
cmp0xff, line 60: Re-export `ScalarOrNA` in `tests/_typing.py` and import from `tests._typing`, so you can also drop all the forwarded `"ScalarOrNA"` in tests, just `ScalarOrNA`

tests/frame/test_frame.py
```diff
@@ -738,6 +742,7 @@ def test_types_itertuples() -> None:
 
     for t1 in df.itertuples():
         check(assert_type(t1, PandasNamedTuple), tuple)
+        assert_type(t1.col1, "ScalarOrNA")
```
cmp0xff, line 745: You also need a `check` in almost all cases when you have `assert_type`

tests/frame/test_indexing.py
```diff
@@ -119,9 +121,27 @@ def test_types_iloc_iat() -> None:
 def test_types_loc_at() -> None:
     df = pd.DataFrame(data={"col1": [1, 2], "col2": [3, 4]})
     check(assert_type(df.loc[[0], "col1"], pd.Series), pd.Series)
-    check(assert_type(df.loc[0, "col1"], Scalar), np.integer)
+    check(assert_type(df.loc[0, "col1"], "ScalarOrNA"), np.integer)
+
+    check(assert_type(df.at[0, "col1"], "ScalarOrNA"), np.integer)
+
+
+def test_scalar_indexers_include_extension_scalars_and_missing_values() -> None:
```
cmp0xff, line 129: Why new function instead of appending to the existing function

cmp0xff (changes requested):
Your agent did not follow `AGENTS.md`. Take a look at `PR body — concise for humans, comprehensive for agents` there.

